PR #171, opened 2017-09-27
Add Waitrose

zerebubuth:
[Waitrose](http://www.waitrose.com/) is a chain of supermarkets (mostly) in the UK. Their store finder interface (usually) contains data about location, address, contact details and opening hours. It appears that some or all of these are manually entered, and so occasionally are missing or have typos.

This is the first one of these that I've attempted, so please let me know if it's outputting an acceptable set of properties in the GeoJSON, and if the `scrapy` use is idiomatic.


iandees:
Thanks for doing this, @zerebubuth! It looks great.

PR #214, opened 2017-11-29
Blue Bottle Cafe

### LINKED ISSUE #157: Blue Bottle Coffee
https://bluebottlecoffee.com/cafes

colanzio5:
Fixes #157.

Is there a defined set of properties that the GeojsonPointItems need to contain, or should they just contain everything that can be pulled off the page that isn't redundant.

Otherwise, I think this is all set.

### DIFF
--- a/locations/spiders/blue_bottle_cafe.py
+++ b/locations/spiders/blue_bottle_cafe.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+import scrapy
+import json
+from locations.items import GeojsonPointItem
+
+class BlueBottleCafeSpider(scrapy.Spider):
+    name = "bluebottlecafe"
+    allowed_domains = ["www.bluebottlecoffee.com"]
+    start_urls = (
+        'https://bluebottlecoffee.com/api/cafe_search/fetch.json?coordinates=false&query=true&search_value=all',
+    )
+    def parse(self, response):
+        results = json.loads(response.body_as_unicode())
+        for e in results["cafes"]:
+            for store_data in results["cafes"][e]:
+
+                add = store_data['address'].replace('\n', ' ').replace('\r', '').replace('<br>', ',')
+                properties = {
+                    'name': store_data['name'],
+                    'addr:full': add,
+                    'addr:city': store_data['region'],
+                    'website': store_data['url'],
+                    'ref': store_data['id'],
+                }
+                lon_lat = [
+                    float(store_data['longitude']),
+                    float(store_data['latitude']),
+                ]
+
+                yield GeojsonPointItem(
+                    properties=properties,
+                    lon_lat=lon_lat,
+                )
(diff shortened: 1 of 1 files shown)


locations/spiders/blue_bottle_cafe.py
```diff
@@ -0,0 +1,33 @@
...
+
+class BlueBottleCafeSpider(scrapy.Spider):
+    name = "bluebottlecafe"
+    allowed_domains = ["www.bluebottlecoffee.com"]
+    start_urls = (
+        'https://bluebottlecoffee.com/api/cafe_search/fetch.json?coordinates=false&query=true&search_value=all',
+    )
+    def parse(self, response):
+        results = json.loads(response.body_as_unicode())
+        for e in results["cafes"]:
```
iandees: Please use a descriptive name for this variable instead of just `e`. Something like `region_name`.

locations/spiders/blue_bottle_cafe.py
```diff
@@ -0,0 +1,33 @@
...
+            for store_data in results["cafes"][e]:
+
+                add = store_data['address'].replace('\n', ' ').replace('\r', '').replace('<br>', ',')
+                properties = {
+                    'name': store_data['name'],
+                    'addr:full': add,
+                    'addr:city': store_data['region'],
```
iandees: The `region` in the JSON object from the store doesn't seem to be a city, just a region code of some kind. It looks like you might be able to parse out the city from the address field.

locations/spiders/blue_bottle_cafe.py
```diff
@@ -0,0 +1,33 @@
...
+                    'website': store_data['url'],
+                    'ref': store_data['id'],
+                }
+                lon_lat = [
+                    float(store_data['longitude']),
```
iandees: These are already floats in the JSON so you don't need to wrap them in `float`.

locations/spiders/blue_bottle_cafe.py
```diff
@@ -0,0 +1,38 @@
...
+            for store_data in results["cafes"][region_name]:
+
+                address_string = store_data['address'].replace('\n', ' ').replace('\r', '').replace('<br>', ', ')
+                address_data = address_string.split(", ")
+
+                properties = {
+                    'name': store_data['name'],
+                    'addr:full': address_string,
+                    'addr:city': address_string.split(", ")[1].replace(" ", ""),
+                    'region': store_data['region'].title(),
```
iandees: We don't need to include this region in the output. It's a store region for their use, not something useful for our output.

locations/spiders/blue_bottle_cafe.py
```diff
@@ -0,0 +1,38 @@
...
+        for region_name in results["cafes"]:
+            for store_data in results["cafes"][region_name]:
+
+                address_string = store_data['address'].replace('\n', ' ').replace('\r', '').replace('<br>', ', ')
+                address_data = address_string.split(", ")
+
+                properties = {
+                    'name': store_data['name'],
+                    'addr:full': address_string,
+                    'addr:city': address_string.split(", ")[1].replace(" ", ""),
```
iandees: The last `replace()` makes it so two-word cities like "New York" end up squished together like "NewYork".

PR #219, opened 2017-11-30
Initial commit of Marshally Spider, hours need to be parsed to OSM.

### LINKED ISSUE #207: Marshall's

kristinepetrosyan:
Closes #207

### DIFF
--- a/locations/spiders/marshalls.py
+++ b/locations/spiders/marshalls.py
@@ -1,5 +1,6 @@
-import scrapy
 import json
+import re
+import scrapy
 from locations.items import GeojsonPointItem
 
 STATES = ["AL", "AK", "AZ", "AR", "CA", "CO", "CT", "DC", "DE", "FL", "GA",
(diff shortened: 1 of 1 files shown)


mendoncaangelo:
@iandees https://s3.amazonaws.com/placescraper-results/runs/1512015627.html why does it say 0 results? I see the log showing more than 1 value.

iandees:
I'm not sure why the count is off, but you're only getting one result because the `ref` property is the same for every result. It dedupes on the value of the `ref` property.

kristinepetrosyan:
@iandees I have added OSM hours format for Marshalls. Please review the code and let me know.

iandees:
Ack, I didn't mean to merge this one.

@kristinepetrosyan, can you please take a look at the other spiders to see what the output looks like? The output on this spider doesn't match the desired output.

kristinepetrosyan:
@iandees this spider has this output 
image

Applebees has this format. Do you want me to remove some of the additional fields? I thought I would leave them in there so that we have more data.

iandees:
Yes, please standardize the attributes and remove the extras. Things like Region, RegionID, Facilities, Directions, Comment, Departments, NewStore, etc. aren't really useful to us.

kristinepetrosyan:
@iandees will do. Will keep this in mind for the other spiders. Thanks once again for reviewing. I will push to the same branch.

iandees:
Thanks for pointing it out – I am making a change in #222 so that the expected properties are more obvious.

PR #224, opened 2017-12-02
Initial commit of Panera Bread, OSM hours remaining

### LINKED ISSUE #107: Panera Bread
https://www.panerabread.com

kristinepetrosyan:
Closes #107


iandees:
Looks like this is missing quite a few locations yet.

kristinepetrosyan:
@iandees The following states have no Panera Bread.
AZ,AR,DE,HI,ID,AK,IN,KY,ME,GA,MN,MT,NM,KS,NV,NE,OR,TX,SD,VA,WY,WA

iandees:
Yep, I'm thinking of the states that do have Panera bread that are missing. For example, there's one in Glendale, WI and several in central Virginia that I've been to that aren't in the results.

iandees:
For example, these in Virginia: https://www.panerabread.com/en-us/mypanera/find-a-panera.html#24450_10

kristinepetrosyan:
@iandees  for wisconsin I have 
image

kristinepetrosyan:
@iandees Wisconsin seems right. But I need to probably enter full state name for the ones that are missing. 

image

kristinepetrosyan:
@iandees  seems like the build is not running for all states. I have a build running at the moment on my local which has 1680 items at the moment.

kristinepetrosyan:
@iandees  @nvkelso [mapanera.txt](https://github.com/alltheplaces/alltheplaces/files/1524496/mapanera.txt)

Complete listing of Panera locations.

iandees:
While looking around for a better URL to use for each store, I found this site: http://locations.panerabread.com/index.html

This would be way easier and more efficient to scrape than the search-based thing we're using now. Take a look at the other directory-based spiders like [the one for Home Depot](https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/homedepot.py). You should be able to keep most of what you already have here, just a different way of getting to the individual store pages.

iandees:
Also, I merged in #222, so the item now has a bunch of explicit columns it expects rather than just a latlon and a properties dictionary.

kristinepetrosyan:
@iandees so you dont want me to use the json search based thing. Just parse the text from the locations? Can I make a new PR for that? OR just push to this one? Something like CVS, right?

iandees:
Yea, just reuse this branch. Pulling it out of the HTML shouldn't be too bad. Let me know if you run into problems.

kristinepetrosyan:
@iandees Sure no problem. I can handle the HTML parsing. I can handle it. No worries. I did the same for CVS. So similar style. Thanks for the suggestion Ian.

kristinepetrosyan:
@iandees I have changed the implementation to parsing HTML instead of the JSON.

kristinepetrosyan:
@iandees this is much efficient, thanks.

image

locations/states_counties.py
```diff
@@ -0,0 +1,613 @@
+state_county = {
```
iandees: I don't think you need this file anymore.

locations/spiders/panera_bread.py
```diff
@@ -0,0 +1,107 @@
...
+            '//div[@class="nap-info-left"]//span[@class="c-address-postal-code"]/text()').extract_first()
+        props['phone'] = loc.xpath(
+            '//div[@class="nap-info-left"]//a[@class="c-phone-number-link c-phone-main-number-link"]/text()').extract_first()
+        props['website'] = loc.url
+        props['ref'] = loc.url
+        hours = loc.xpath(
+            '//div[@class="c-location-hours-details-wrapper js-location-hours"]/@data-days').extract_first()
+        props['opening_hours'] = self.store_hours(json.loads(hours))
+
+        return GeojsonPointItem(**props)
```
iandees: Need to extract `lat` and `lon`.

locations/spiders/panera_bread.py
```diff
@@ -0,0 +1,107 @@
...
+    )
+
+    def store_hours(self, day_hours):
+        day_groups = []
+        this_day_group = {}
+
+        for day_hour in day_hours:
+            hours = ''
+            day, intervals = day_hour['day'], day_hour['intervals']
+            short_day = DAYS[day]
```
iandees: Instead of doing this lookup you can do `day.title()[:2]`.

locations/states_counties.py
kristinepetrosyan: Do u think we might need this for future spiders?

locations/spiders/panera_bread.py
kristinepetrosyan: Will add that.

locations/spiders/panera_bread.py
kristinepetrosyan: Thank you @iandees  :)

iandees:
Looks good, thanks!

PR #225, opened 2017-12-03
Added spider for bathandbodyworks

### LINKED ISSUE #186: Beauty Stores
- [ ] [Sephora](https://www.sephora.com/storelist)
- [ ] [Ulta](http://www.ulta.com/ulta/stores/storelocator.jsp)
- [ ] Bath and Body Works [US](http://www.bathandbodyworks.com/store-locator), [world](http://www.bathandbodyworks.com/global-locations/global-locations.html)

AustenHolmberg:
Fixes #186. Scrapes US and global locations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,13 +17,13 @@ To get started, you'll want to install the dependencies for this project.
 1. With `pipenv` installed, make sure you have the `all-the-places` repository checked out
 
    ```
-   git clone git@github.com:iandees/all-the-places.git
+   git clone git@github.com:alltheplaces/alltheplaces.git
    ```
 
 1. Then you can install the dependencies for the project
 
    ```
-   cd all-the-places
+   cd alltheplaces
    pipenv install
    ```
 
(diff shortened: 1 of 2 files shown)


AustenHolmberg:
Oops, I think it failed because I didn't include a lon_lat item at all for locations missing that data. Fixing.

AustenHolmberg:
Hmm, not sure what's tripping it up. It's complaining about "properties" being an unsupported GeojsonPointItem field.

nvkelso:
@iandees any tips?

iandees:
I merged in #222 so the [item](https://github.com/alltheplaces/alltheplaces/blob/master/locations/items.py#L11) has explicit columns its looking for rather than a single `properties` dictionary. Take a look at what I did in the other scrapers for an example:

I still create a `properties` dict, but I include `lat` and `lon`:

https://github.com/alltheplaces/alltheplaces/blob/a9b3ee9aecd35e5be503e969b55f34f60befb709/locations/spiders/perkins.py#L77-L86

...and then I pass that dict into the item's constructor with a kwarg expansion (the `**` prefix):

https://github.com/alltheplaces/alltheplaces/blob/a9b3ee9aecd35e5be503e969b55f34f60befb709/locations/spiders/perkins.py#L98

You'll need to do something similar for this spider.

PR #232, opened 2017-12-03
Chipotle

sledjama:
Chipotle has 5 sites that this crawls.

- [ ] chipotle.com

- [ ] chipotle.ca

- [ ] chipotle.de

- [ ] chipotle.fr

- [ ] chipotle.co.uk

We can only get locations through searches which returns JSON,
address is needed and radius can be specified to return paginated results of locations within the radius of the address given.
There is a need to go from page to page of each address search until it we get nothing.
we give it a generous radius of 500miles and gave states of respective countries as address.
it returned 2392 total locations which hopefully covers all their locations in these 5 countries.


iandees:
Note that after #222 the item to yield has specific properties to pass in rather than a `properties` and a `lon_lat` property. See https://github.com/alltheplaces/alltheplaces/pull/225#issuecomment-348793352 for more info.

iandees:
Also, I appreciate the documentation you're adding. Thanks!

sledjama:
@iandees i think i should just add fax as extras

iandees:
Yea, let's put fax in extras for now.

locations/spiders/chipotle.py
```diff
@@ -0,0 +1,267 @@
...
+                    properties = {"addr_full": full_address,
+                                  "city": city,
+                                  "state": state,
+                                  "postcode": zip_code,
+                                  "country": country,
+                                  "phone": self.process_phone(phone),
+                                  "website": website,
+                                  "ref": link_id,
+                                  "opening_hours": self.process_hours(open_hours),
+                                  "extras": "fax: "+self.process_phone(fax),
```
iandees: We're expecting `extras` to be a `dict`. Can you make this look like:

```
"extras": {
    "fax": self.process_phone(fax),
}
```

iandees:
Looks good. Should I merge now or did you want to keep going with other domains?

sledjama:
i actually added other domains, you can merge.
Thanks.

iandees:
Great!

MiniCodeMonkey:
I just discovered by accident that it looks like all US locations can be retrieved with a single request:
https://order.chipotle.com/api/restaurant?latitude=38.45&longitude=-100.59&radius=99999999&timeZoneOffset=-240&pageIndex=0&pageSize=100000

iandees:
@MiniCodeMonkey if you're up for it you can rewrite the spider to use this new endpoint you found.

PR #240, opened 2017-12-05
Quiznos

MagnetonIO:
Added a quiznos spider. Please check.

### DIFF
--- a/locations/spiders/quiznos.py
+++ b/locations/spiders/quiznos.py
@@ -6,10 +6,10 @@ import traceback
 
 from locations.items import GeojsonPointItem
 
-
+URL = "http://restaurants.quiznos.com"
 class QuiznosSpider(scrapy.Spider):
     name = "quiznos"
-    allowed_domains = ["restaurants.quiznos.com"]
+    allowed_domains = [URL]
     start_urls = (
         'http://restaurants.quiznos.com/data/stores.json?callback=storeList',
     )
@@ -77,5 +77,5 @@ class QuiznosSpider(scrapy.Spider):
                 city=store.get('city'),
                 state=store.get('statecode'),
                 postcode=store.get('zipcode'),
-                website=store.get('url'),
+                website=URL + store.get('url'),
             )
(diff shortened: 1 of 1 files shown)


locations/spiders/quiznos.py
```diff
@@ -0,0 +1,40 @@
...
+    start_urls = (
+        'http://restaurants.quiznos.com/data/stores.json?callback=storeList',
+    )
+
+    def parse(self, response):
+
+        try:
+            data = response.body_as_unicode()
+            stores = json.loads(re.search('storeList\((.*)\)', data).group(1))
+        except:
```
iandees: Please don't catch exceptions like this. I'd prefer to have them bubble up and cause the spider to stop.

MagnetonIO: Don't you need statements for validation?

iandees: Not in this case. I'd rather the spider throw an exception and have it get logged through the normal scrapy logging system than have it get caught and logged separately.

MagnetonIO:
I updated quiznos for exception. issue. Can you check again?

locations/spiders/quiznos.py
```diff
@@ -0,0 +1,35 @@
...
+                lon=store.get('longitude'),
+                ref=str(store.get('storeid')),
+                phone=store.get('phone'),
+                name=store.get('restaurantname'),
+                opening_hours=store.get('businesshours'),
+                addr_full=store.get('address1'),
+                city=store.get('city'),
+                state=store.get('statecode'),
+                postcode=store.get('zipcode'),
+                website=store.get('url'),
```
iandees: The website is not a fully formed URL. Looks like you need to prepend something to the front of the value from `url`.

locations/spiders/quiznos.py
```diff
@@ -0,0 +1,35 @@
...
+
+        for store in stores:
+
+            yield GeojsonPointItem(
+                lat=store.get('latitude'),
+                lon=store.get('longitude'),
+                ref=str(store.get('storeid')),
+                phone=store.get('phone'),
+                name=store.get('restaurantname'),
+                opening_hours=store.get('businesshours'),
```
iandees: The hours format from the website doesn't match our desired hours format. Take a look at other scrapers to see how they parse the various hours formats into something consistent.

PR #250, opened 2017-12-06
Kohl's

### LINKED ISSUE #120: Kohl's
http://www.kohls.com/stores.shtml

kristinepetrosyan:
Fixes #120.


iandees:
This seems to periodically fail because they block the spider. We can slow it down later if that continues to be a problem.

PR #252, opened 2017-12-06
TractorSupply

### LINKED ISSUE #153: Tractor Supply
http://www.tractorsupply.com/tsc/store-locator

whoissteven:
Fixes #153.

Having issues sending properties to GeojsonPointItem.
"KeyError: 'GeojsonPointItem does not support field: fieldname"

Any ideas why I run into this..? Thanks!

Also, my spider runs so fast that eventually their servers block my IP.
Should I use a simple sleep to avoid any issues?

### DIFF
--- a/locations/spiders/tractorsupply.py
+++ b/locations/spiders/tractorsupply.py
@@ -3,6 +3,7 @@ import scrapy
 from scrapy.linkextractors import LinkExtractor
 import json
 import re
+import time
 
 from locations.items import GeojsonPointItem
 
@@ -110,5 +111,6 @@ class TractorSupplySpider(scrapy.Spider):
         address = self.address(data['address'])
         if address:
             properties.update(address)
+        time.sleep(2)
 
         yield GeojsonPointItem(**properties)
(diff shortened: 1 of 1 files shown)


whoissteven:
Added a 2 second sleep after the properties dict is built, but it seems it didn't benefit anything. 9 less results that run.

Open to suggestions on this one. Thanks.

iandees:
Scrapy "items" only allow you to set certain properties  (see the list of properties on the [GeojsonPointItem](https://github.com/alltheplaces/alltheplaces/blob/master/locations/items.py) class). Read [the Scrapy documentation about items](https://docs.scrapy.org/en/latest/topics/items.html), too.

To slow the spider down, use the `download_delay` property on your spider. See [this other spider](https://github.com/alltheplaces/alltheplaces/blob/0f6fa8463b38493f031c252fee43d5553cc781a1/locations/spiders/sephora.py#L11) as an example. Also, there is [documentation on `download_delay` here](https://doc.scrapy.org/en/latest/topics/settings.html?highlight=download_delay#download-delay).

whoissteven:
What is the max download we should be setting to avoid the script timing out?

iandees:
There shouldn't be a max download set: the timeout is for testing only and won't be set when we run to get all the data.

PR #259, opened 2017-12-06
potbelly sandwich

### LINKED ISSUE #124: Potbelly Sandwich Shop
http://www.potbelly.com/Shops/ShopLocator.aspx

sweetdream0729:
Fixes #124.

@iandees , I have taken "pickup_menu_url" for "website" field.
But there is also "facebook_url".
I am not sure if it is correct to pull "pickup_menu_url" for "website".
Thanks.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -13,7 +13,7 @@ name = "pypi"
 
 scrapy = "*"
 botocore = "*"
-
+demjson = "*"
 
 [requires]
 
--- a/Pipfile
+++ b/Pipfile
@@ -13,7 +13,6 @@ name = "pypi"
 
 scrapy = "*"
 botocore = "*"
-demjson = "*"
 
 [requires]
 
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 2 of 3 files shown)


iandees:
Let's use the Facebook URL for now instead of the menu URL.

It also looks like the server is rejecting the spider's request for some reason. Does it work for you when you run it locally?

sweetdream0729:
yes, but not now.
let me check more.

iandees:
There are many more stores missing from the output. It looks like the the full list of locations can be found in `https://www.potbelly.com/dist/main.js` if you search for `staticLocations=[`. The trick is to find the end of the array.

sweetdream0729:
ok, let me check more.
Thanks.

sweetdream0729:
wow~, contents of staticLocations=[ is not JSON format. 
Properties and values have not double quotes:(

sweetdream0729:
@iandees, added demjson package in order to convert Javascript object to python dictionary.
Please have a look it and let me know any issues.
Thanks.

iandees:
@sweetdream0729 demjson worked but it was taking a very long time, so I added a couple regex runs to make the JSON correct and parse it with the normal JSON parser.

PR #267, opened 2017-12-07
LA Breweries

### LINKED ISSUE #262: LA Breweries
http://labeerhop.com/breweries/

whoissteven:
Fixes #262

There is no geo data on these pages. How should that be handled?

### DIFF
--- a/locations/spiders/labreweries.py
+++ b/locations/spiders/labreweries.py
@@ -81,10 +81,13 @@ class LaBreweriesSpider(scrapy.Spider):
         response.selector.remove_namespaces()
         city_urls = response.xpath('//url/loc/text()').extract()
         for path in city_urls:
-            yield scrapy.Request(
-                path.strip(),
-                callback=self.parse_store,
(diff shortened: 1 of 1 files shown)


whoissteven:
Still need to clean up hours. Just wondering how to handle geo data missing.

iandees:
Pull the address out and we can geocode it later.

whoissteven:
For clarification, is the only needed data: lat, lon, and full address(addr, city, state, and post)?

iandees:
@whoissteven please pull in as much data as you can find on the site. If you can't find a lat/lon we can do without it as long as there's an address we can pull out.

whoissteven:
The error is on this page:
http://labeerhop.com/breweries/1056/

Which is a list of ALL breweries in LA Area. I am already pulling all the breweries from the sitemap XML, I'll need to disallow this URL. Will look into this. Otherwise data pulled looks good. Open to critique otherwise.

PR #280, opened 2017-12-08
La Salsa

### LINKED ISSUE #84: La Salsa
http://lasalsa.com/locations/

AnthoneyKalasho:
Fixes #84


AnthoneyKalasho:
@iandees Sorry for the trouble, the Anthony's Restaurants spider uses Regex in a similar fashion, I'll go back and update it. Should regex be avoided altogether or just whenever possible?

iandees:
Yea, if you could use XPath that'd be helpful because it's more reliable when parsing XML.

PR #287, opened 2017-12-08
Olive Garden

### LINKED ISSUE #106: Olive Garden
http://www.olivegarden.com/locations/location-search

whoissteven:
Fixes #106.


nvkelso:
Why did the results go from 314 to 47?

iandees:
The server was blocking us so I added a download delay between requests. This slows down the spider and results in a timeout here, but without any errors from the server. When it runs in a global build context (without a timeout) it will hopefully give more complete results.

PR #299, opened 2017-12-08
whidbeycoffee

### LINKED ISSUE #101: Whidbey Coffee
http://www.whidbeycoffee.com/pages/locations

josh0203e:
Fixes #101.


iandees:
@josh0203e thanks for removing the print. Can you also make it so the `ref` field points to a unique value per store?

PR #300, opened 2017-12-08
McDonalds Puerto Rico

SelfnessAid:
added the Spider to McDonalds - Puerto Rico (pr)

### DIFF
--- a/locations/spiders/mcdonald_pr.py
+++ b/locations/spiders/mcdonald_pr.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+import scrapy
+import json
+from locations.items import GeojsonPointItem
+
+
+class McDonaldsPR(scrapy.Spider):
+
+    name = "mcdonalds_pr"
+    allowed_domains = ["www.mcdonalds.com.pr", "www.mcdonalds.com"]
+    start_urls = (
+        'http://www.mcdonalds.com.pr/api/restaurantsByCountry?country=PR',
+    )
+
+    def parse(self, response):
+        results = json.loads(response.body_as_unicode())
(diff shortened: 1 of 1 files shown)


locations/spiders/mcdonald_pr.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def parse(self, response):
+        results = json.loads(response.body_as_unicode())
+        results = results["content"]["restaurants"]
+        for data in results:
+            properties = {
+                'ref': data['id'],
+                'lon': data['longitude'],
+                'lat': data['latitude'],
+                'name': data['name'][:data['name'].find("<br")]
```
iandees: It looks like there's an address after the name. Can you put that in `addr_full`?

locations/spiders/mcdonald_pr.py
```diff
@@ -0,0 +1,26 @@
...
+        'http://www.mcdonalds.com.pr/api/restaurantsByCountry?country=PR',
+    )
+
+    def parse(self, response):
+        results = json.loads(response.body_as_unicode())
+        results = results["content"]["restaurants"]
+        for data in results:
+            properties = {
+                'ref': data['id'],
+                'lon': data['longitude'],
```
iandees: Can you cast the `lat` and `lon` to `float()` please.

locations/spiders/mcdonald_pr.py
SelfnessAid: okay, I'll do.
but we don't need this Spider anymore.
I merged this Spider to mclocalizer Spider.
so I'll implement your requirement on mclocalizer Spider.
okay?

Issue #132, opened 2017-01-17
Aldi


maxerickson:
https://storelocator.aldi.us/Presentation/AldiSued/en-us/Start

iandees:
There's Aldi's all over the world (See https://aldi.de/). We should add spiders for the Aldi's all over the world.

Eagle19243:
https://www.aldi.ie/store-finder
@iandees, As you can see, they want to type place name.
At this point, how can we scrap stores from the above website?
Exactly, we need to know all place names in Ireland?

iandees:
If you type in something like "Dublin" and look at the browser's developer tools when you do the search, you can see an XHR request being made like this:

https://www.aldi.ie/api/store-finder/search?q=dublin&address=Dublin&latitude=53.3498053&longitude=-6.260309699999993

Try adapting that and see what you can get.

Eagle19243:
@iandees, `Aldi LU` and `Aldi BE` has no (lat, lon) info on their website.
So I have ignored lat and lon in spider.

iandees:
This issue hasn't been worked on in a while. I'm removing the assignment so others can work on it. If you're still working on it, please reassign yourself and reference a pull request or branch where you're doing your work.

Cj-Malone:
- [ ] Germany - Aldi Nord
- [ ] Germany - Aldi Süd
- [ ] Australia - Aldi Süd
- [ ] Austria - Hofer Austria
- [ ] Belgium - Aldi Nord
- [ ] China - Aldi China Süd
- [ ] Denmark - Aldi Marked Nord
- [ ] France - Aldi Marché Nord
- [ ] Hungary - Aldi Hungary 
- [x] Ireland - Aldi Ireland Süd
- [ ] Italy - Aldi Italy 
- [ ] Luxembourg - Aldi Nord
- [ ] Netherlands - Aldi Nord
- [ ] Poland - Aldi Nord
- [ ] Portugal - Aldi Nord
- [ ] Slovenia - Hofer Slovenia
- [ ] Spain - Aldi Nord
- [ ] Switzerland - Aldi Suisse
- [x] United Kingdom - Aldi UK Süd
- [ ] United States - Aldi US Süd

List from Wikipedia, some of these are already done, some are broken.

PR #302, opened 2017-12-08
Aldi US

Eagle19243:
For #132.

### DIFF
--- a/locations/spiders/aldi.py
+++ b/locations/spiders/aldi.py
@@ -1,0 +1,115 @@
+# -*- coding: utf-8 -*-
+import scrapy
+import json
+import re
+from uszipcode import ZipcodeSearchEngine
+from locations.items import GeojsonPointItem
+
+STATES = ['AL', 'AK', 'AZ', 'AR', 'CA', 'CO', 'CT', 'DC', 'DE', 'FL', 'GA',
+          'HI', 'ID', 'IL', 'IN', 'IA', 'KS', 'KY', 'LA', 'ME', 'MD',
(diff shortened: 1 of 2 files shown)


Eagle19243:
@iandees, would you please check this commit?

iandees:
Can you rename this spider to `aldi_us.py` so we can have spiders for other countries too?

Also, instead of searching by zipcode you can use their store search API to be more efficient.

To get an idea for where all the stores are, start at 
https://www.yellowmap.de/Presentation/AldiSued/en-US/ResultList?LocX=&LocY=&Lux=-115.224609375&Luy=58.17070248348609&Rlx=-62.666015625&Rly=17.14079039331665&ZoomLevel=4

Inside the `container` element there's a JSON `addressResult` blob. Pass those X and Y's to the API like https://www.yellowmap.de/Presentation/AldiSued/en-US/ResultList?LocX=-93.15121&LocY=43.9212&ZoomLevel=6 (with `ZoomLevel` increased by 2).

Then do it again and you'll get to https://www.yellowmap.de/Presentation/AldiSued/en-US/ResultList?LocX=-93.2905&LocY=44.73248&ZoomLevel=8 that gives you results and more `addressResult`s. If you keep recursing down into these address results and pulling out the stores I think you can make fewer queries.

This might work for other countries, too.

locations/spiders/aldi.py
```diff
@@ -0,0 +1,115 @@
+# -*- coding: utf-8 -*-
+import scrapy
+import json
+import re
+from uszipcode import ZipcodeSearchEngine
```
iandees: Please see my comment in the PR to more efficiently query this site without including this library.

PR #329, opened 2017-12-10
Rubios

### LINKED ISSUE #56: Rubio's
https://www.rubios.com/store-locator

whoissteven:
Fixes #56

### DIFF
--- a/locations/spiders/rubios.py
+++ b/locations/spiders/rubios.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+import scrapy
+import re
+
+from locations.items import GeojsonPointItem
+
+
+class RubiosSpider(scrapy.Spider):
+    name = "rubios"
+    allowed_domains = ['rubios.com']
+    start_urls = (
+        'https://www.rubios.com/sitemap.xml',
+    )
+
+    def parse(self, response):
+        response.selector.remove_namespaces()
+        city_urls = response.xpath('//url/loc/text()').extract()
+        for path in city_urls:
+            regex = re.compile(r'http\S+rubios.com/store-locations/\S+/\S+/\S+')
+            if not re.search(regex,path):
+                pass
+            else:
+                yield scrapy.Request(
+                    path.strip(),
+                    callback=self.parse_store,
+                )
+
+    def parse_store(self, response):
+
+        properties = {
+            'name': response.css('.store-info').extract_first().replace('\t','').split('<span itemprop="name">')[1].split('</span>')[0],
+            'ref': response.css('.store-info').extract_first().replace('\t','').split('"addressLocality">')[1].split('</span>')[0],
+            'addr_full': response.css('.store-info').extract_first().replace('\t','').split('itemprop="streetAddress">')[1].split('</span>')[0],
+            'city': response.css('.store-info').extract_first().replace('\t','').split('"addressLocality">')[1].split('</span>')[0],
+            'state': response.css('.store-info').extract_first().replace('\t','').split('itemprop="addressRegion">')[1].split('</span>')[0],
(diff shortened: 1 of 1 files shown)


AnthoneyKalasho:
Might be too late but I found this earlier when snooping around, could be useful if you find a way to make it readable

https://www.rubios.com/store-location-data

whoissteven:
Will fix tomorrow.

locations/spiders/rubios.py
```diff
@@ -0,0 +1,58 @@
...
+    allowed_domains = ['rubios.com']
+    start_urls = (
+        'https://www.rubios.com/sitemap.xml',
+    )
+
+    def parse(self, response):
+        response.selector.remove_namespaces()
+        city_urls = response.xpath('//url/loc/text()').extract()
+        for path in city_urls:
+            regex = re.compile(r'http\S+rubios.com/store-locations/\S+/\S+/\S+')
```
iandees: You don't need to compile the regex every time through the for loop. Please move the `re.compile()` before the for loop.

locations/spiders/rubios.py
```diff
@@ -0,0 +1,58 @@
...
+            if not re.search(regex,path):
```
iandees: Instead of checking the negative condition here, only `yield scrapy.Request()` if there's a match.

locations/spiders/rubios.py
```diff
@@ -0,0 +1,58 @@
...
+            'postcode': response.css('.store-info').extract_first().replace('\t','').split('itemprop="postalCode">')[1].split('</span>')[0],
+            'phone': response.css('.store-info').extract_first().replace('\t','').split('"tel:')[1].split('"')[0],
+            'website': response.request.url,
+            'opening_hours': response.css('.store-info').extract_first().replace('\t', '').split('<span class="oh-display-label" style="width: 6.6em;">')[1].split('<br></span>')[0].replace('</span><span class="oh-display-times oh-display-hours">', '').strip(),
+            'lon': float(response.xpath('//head/script[9]').extract_first().split('"lon":')[1].split('}')[0]),
+            'lat': float(response.xpath('//head/script[9]').extract_first().split('"lat":')[1].split(',')[0]),
+        }
+        else:
+            properties = {
+            'name': response.css('.store-info').extract_first().replace('\t','').split('<span itemprop="name">')[1].split('</span>')[0],
```
iandees: This looks like a repeat of the code above. Is there anything different about it?

locations/spiders/rubios.py
```diff
@@ -0,0 +1,58 @@
...
+                yield scrapy.Request(
+                    path.strip(),
+                    callback=self.parse_store,
+                )
+
+    def parse_store(self, response):
+
+        if response.css('.store-info').extract_first().replace('\t', '').split('<span class="oh-display-label" style="width: 6.6em;">')[1].split('<br></span>')[0].replace('</span><span class="oh-display-times oh-display-hours">', '').strip() is not None or '':
+            properties = {
+            'name': response.css('.store-info').extract_first().replace('\t','').split('<span itemprop="name">')[1].split('</span>')[0],
```
iandees: Try your hardest to not use splits as they're pretty inefficient and brittle (it's what's causing the exception in the log). It would be better to use `xpath()` here like this:

```
'name': response.xpath('//span[@itemprop="name"]/text()').extract_first(),
...
'telephone': response.xpath('//span[@itemprop="telephone"]/a/text()').extract_first(),
```

locations/spiders/rubios.py
whoissteven: I apologize, I just saw these changes. Will adjust.

locations/spiders/rubios.py
whoissteven: Yes, not all pages have the opening_hours listed as they are "Coming Soon", I was attempting to utilize an if statement to see if the element is present or not.

PR #333, opened 2017-12-10
Serious Coffee

### LINKED ISSUE #61: Serious Coffee (BC)
http://www.seriouscoffee.com/locations/

josh0203e:
Fixes #61.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,20 +1,20 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "3ec89dd612978e5090d9643366227e41c39175833c742f2f9b578993f6b82807"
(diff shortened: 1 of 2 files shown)


Pipfile.lock
```diff
@@ -1,20 +1,20 @@
 {
```
iandees: Please don't include changes unrelated to the spider you're adding.

locations/spiders/serious_coffee.py
```diff
@@ -0,0 +1,128 @@
...
+        'http://www.seriouscoffee.com/locations/',
+    )
+
+    def parse_day(self, day):
+        if re.search('-', day):
+            days = day.split('-')
+            osm_days = []
+            if len(days) == 2:
+                for day in days:
+                    osm_day = DAY_MAPPING[day.strip()]
```
iandees: Instead of using a mapping dictionary like this you can just call `day.strip().title()[:2]`.

PR #337, opened 2017-12-10
Exxonmobil

sledjama:
Exxonmobil has 1,679 locations worldwide
This crawler crawls all earth's geographic locations except Antartica.


sledjama:
I wonder where that much difference is from, I ran it twice  on my local machine and got same 1,679.
CI returned 1,710.

iandees:
There are *way* more than 1700 ExxonMobil stations around the world. Looking at this via the map it looks like it just timed out before finishing. It might have also not found anything outside of North America, too.

iandees:
I'm going to merge this and see what we can get as I run the spider without any time limits.

PR #373, opened 2017-12-11
Disco Argentina

### LINKED ISSUE #90: Disco (Argentina)
Supermarket chain, owned by Cencosud.

The [webpage with the list of branches](https://www.disco.com.ar/Login/Sucursales.aspx) uses some client-side-template Javascript monstrosity. The data can be retrieved with:
```curl -H "Content-Type: application/json" -d "{}" https://www.disco.com.ar/Login/PreHomeService.aspx/TraerLocales```
which will return a JSON object with a "d" key with a string with more JSON (?!)

Has address, phone number, lat, lon (at least one pointing to null island), payment methods. No opening hours.

sweetdream0729:
Fixes #90.

### DIFF
--- a/locations/spiders/disco_argentina.py
+++ b/locations/spiders/disco_argentina.py
@@ -1,0 +1,39 @@
+import json
+import scrapy
+from locations.items import GeojsonPointItem
+
+class DiscoArgentinaSpider(scrapy.Spider):
+    name = "disco_argentina"
+    allowed_domains = ["www.disco.com.ar"]
+
+    def start_requests(self):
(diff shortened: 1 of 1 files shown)


sweetdream0729:
@iandees , please review this spider. 
It gives me a warning for duplicated items.
Thanks.

sweetdream0729:
@iandees , I pulled master and switched new branch.
I am getting a error for import request package which comes from carrefour.py.
So I have reverted to a previous commit.
Please have a look into it.
Thanks.

iandees:
From their webpage it looks like there are at least 30 stores, but the spider is only pulling in 4.

https://www.disco.com.ar/Login/Sucursales.aspx

PR #385, opened 2017-12-12
McDonalds - France

SelfnessAid:
For #1 

Created spider for France of McDonalds.

### DIFF
--- a/locations/spiders/mcdonalds_fr.py
+++ b/locations/spiders/mcdonalds_fr.py
@@ -77,7 +77,7 @@ class McDonalsFRSpider(scrapy.Spider):
                 'city': data['location']['city'],
                 'ref': data['id'],
                 'addr_full': data['location']['streetLabel'],
-                'phone': data['datasheet']['tel'],
+                'phone': data['datasheet']['tel']['number'],
                 'state': data['location']['countryISO'],
                 'postcode': data['location']['postalCode'],
                 'name': data['name'],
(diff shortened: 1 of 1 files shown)


iandees:
If you look at the map it looks like the radius search isn't big enough.

SelfnessAid:
yes, I saw it.
but they return maximum 1000 stores.
we need to fix this?

SelfnessAid:
okay, I think we can fix this problem.
I'll update this spider soon.

iandees:
Yep, we should fix it. One option would be to change the requests: one for the north part of France and one for the southern part.

SelfnessAid:
Yep, it's just my thought. :)

SelfnessAid:
okay, it seems be better now.

locations/spiders/mcdonalds_fr.py
```diff
@@ -0,0 +1,93 @@
...
+
+        results = json.loads(match.groups()[1])
+        results = results["poiList"]
+        for item in results:
+            data = item["poi"]
+            properties = {
+                'city': data['location']['city'],
+                'ref': data['id'],
+                'addr_full': data['location']['streetLabel'],
+                'phone': data['datasheet']['tel'],
```
iandees: If you look at the output, it looks like this is a dictionary that has "number" inside of it rather than the phone number itself.

SelfnessAid: okay, let me fix it now.

PR #386, opened 2017-12-12
Debugging AWS Build

iandees:
Permissions seem wrong, so debugging the upload step.


iandees:
Looks to be working now.

PR #392, opened 2017-12-12
McDonaldsQStore

SelfnessAid:
For #1 

Merged Australia and New Zealand into one new Spider - mcdonalds_qstore.
Removed McDonaldsAUSpider(mcdonalds_au.py).


iandees:
I think this is a bug in the CI spider runner not handling renames very well.

SelfnessAid:
thanks

PR #433, opened 2017-12-14
Publix

### LINKED ISSUE #343: Publix
http://store.publix.com/publix/

whoissteven:
Fixes #343

No GEO Lat/Lon data on page.

### DIFF
--- a/locations/spiders/publix.py
+++ b/locations/spiders/publix.py
@@ -61,6 +61,5 @@ class PublixSpider(scrapy.Spider):
         'lon': response.xpath('///div[@class="store-info-group"][4]/a/@href').extract_first().split('//')[2].split(',')[1],
         }
 
-        print(str(properties))
 
         yield GeojsonPointItem(**properties)
--- a/locations/spiders/publix.py
+++ b/locations/spiders/publix.py
@@ -1,0 +1,53 @@
+# -*- coding: utf-8 -*-
+import scrapy
+import re
+
+from locations.items import GeojsonPointItem
+
+
+class PublixSpider(scrapy.Spider):
+    name = "publix"
+    allowed_domains = ['publix.com']
+    start_urls = (
+        'http://www.publix.com/sitemap.xml',
+    )
+
+    def parse(self, response):
+        response.selector.remove_namespaces()
+        city_urls = response.xpath('//url/loc/text()').extract()
+        regex = re.compile(r'http\S+publix.com/locations/\d+\S+')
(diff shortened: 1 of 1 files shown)


whoissteven:
This one requires splits to clean up the data scraped for certain parameters.
IF ELSE yielded more results, but still resulted in an exception..
Will see if I can think of something more creative..
Also, looks like there are google maps url's in the page that I can pull the LAT/LON from.

AnthoneyKalasho:
Not 100% sure if its required but I've noticed a lot of other spiders format their hours like such. 
`Mo-Th: 7:00-23:00; Fr-Sa: 7:00-12:00; Su: 7:00-23:00;`
Maybe consider doing that as it's nice to keep things consistent

iandees:
Looks like something is limiting you here. This should find at least 900 stores.

locations/spiders/publix.py
```diff
@@ -0,0 +1,55 @@
...
+            storeHoursHTML = response.xpath('//div[@class="store-info-group"]').extract()[4]
+            p = re.compile(r'<.*?>')
+            storeHours = p.sub('', storeHoursHTML)
+            storeHours = storeHours.replace('\t', '').replace('\r', '').replace('\n', '').replace('       ', ' ')
+            storeHours = "".join(storeHours.strip())
+
+
+        properties = {
+        'name': " ".join(response.xpath('//h1[@id="content_2_TitleTag"]/text()').extract_first().split()),
+        'ref': "".join(response.xpath('//div[@class="store-info-group"]/text()').extract_first().strip()),
```
iandees: You shouldn't need any of these `"".join()`'s in here. `extract_first()` already returns a string.

whoissteven:
Will be using http://weeklyad.publix.com/Publix/Entry/Locations/ to pull locations as sitemap isn't full. May be only a part.

whoissteven:
Please check the JSON.
I had the .join to help with formatting of the strings.
If that is cleaned up post and this is fine I will leave as is.

iandees:
Please don't delete or dismiss my review comments. Once you delete them I can't see them :).

You can remove the extra characters around the fields by using the `strip()` function. For example:

```python
>>> "\r\n        Crow's Corner\r\n        \r\n    ".strip()
"Crow's Corner"
```

whoissteven:
@iandees I apologize, I would dismiss and leave notes.. I'll make sure not to dismiss them again. 

Cleaned up the json

PR #440, opened 2017-12-14
McDonalds - Singapore

SelfnessAid:
For #1 

Created the Spider for Singapore of McDonalds.
I put raw opening_hours because of its response.
Please see comments #1


iandees:
Yea, this'll work for now. Thanks!

PR #443, opened 2017-12-14
McDonalds - Norway

SelfnessAid:
For #1 

Added the Norway to McdonaldsSpider.


SelfnessAid:
I am not sure why here is two Spiders.

SelfnessAid:
@iandees 
we're getting timeout at mcdonalds.py.
we need to create another Spider to avoid timeout?

iandees:
Don't worry about the timeout – it indicates the spider takes longer than a minute to run.

SelfnessAid:
okay, got you.

Issue #346, opened 2017-12-11
SPAR


iandees:
Multiple countries, e.g.:

- https://www.spar.co.uk/store-locator
- https://www.spar.co.za/Store-Finder?showall=1&province=-1

http://spar-international.com/
Total:
SPAR UK - Done
SPAR Norway - Done
SPAR Portugal - no shops position  info

vlad555:
So it this avaliable?

iandees:
Yes I think there are other countries left.

vlad555:
If site doesn't have lan/lon information of their stores does  we need to scrape it?

iandees:
Skip it for now, but note in here which countries you're working on and which you're skipping and why. That way we won't duplicate work in the future.

CloCkWeRX:
This turned into SPAR.pl

### Brand name

Piotr i Paweł

Polish supermarket chain

### Wikidata ID

Q2096356
https://www.wikidata.org/wiki/Q2096356
https://www.wikidata.org/wiki/Special:EntityData/Q2096356.json

### Store finder url(s)

Official Url(s): http://www.piotripawel.pl
- [x] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q2096356 http://www.piotripawel.pl`
```
```

CloCkWeRX:
#9379 adds
            "country_code": "SI"
            "country_code": "HR"
            "country_code": "AT",
           
            "country_code": "HU",

CloCkWeRX:
"There are now over 13,900 SPAR branded stores in over 48 countries on four continents, meeting the needs of over 14.7 million consumers every day."

PR #448, opened 2017-12-15
Spar UK

AustenHolmberg:
For #346 
This spider scrapes the U.K. locations for Spar (most of them are located in the U.K.)
It yields 2201 locations, although Wikipedia says that there are 2352 in the U.K. I'm thinking this might not be due to a problem with the scraper, because this spider sends a request for nearly every postcode in the country.


AustenHolmberg:
Hmm, it's understandable that it would timeout because of the gargantuan number of requests and the download delay (~3000 requests, download_delay set to 0.5). Let me know if you'd like me to make any modifications to the spider.

vlad555:
Is it necessary to collect data from site without shop positions?
For example SPAR in Portugal
http://www.spar.pt/rede-de-lojas/

PR #453, opened 2017-12-15
update lovinghut

MagnetonIO:


MagnetonIO:
@iandees , Can you check please?

PR #469, opened 2017-12-15
make coopfood

### LINKED ISSUE #412: Co-op food
Co-op is a brand which, amongst other things, runs a supermarket chain in the UK.

https://finder.coop.co.uk/food

josh0203e:
Fixes #412

### DIFF
--- a/locations/spiders/coopfood.py
+++ b/locations/spiders/coopfood.py
@@ -1,41 +1,46 @@
 import json
-import re
 import scrapy
+
 from locations.items import GeojsonPointItem
+
+
 class CoopFoodSpider(scrapy.Spider):
     name = "coopfood"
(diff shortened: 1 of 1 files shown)


locations/spiders/coopfood.py
```diff
@@ -0,0 +1,49 @@
...
+                yield scrapy.Request(
+                    url=url,
+                    callback=self.parse
+                )
+
+    def parse(self, response):
+      try:
+        data = json.loads(response.body_as_unicode())
+        if(data['next']==None):
+            self.state = False
```
iandees: Rather than using this `state` variable, you can start with the first page in the `start_requests` list and then in `parse` yield another `scrapy.Request()` for the next page after yielding the items. Then when there's no more results you can stop yielding requests. That would clean this up a bit.

PR #478, opened 2017-12-16
Boston market

### LINKED ISSUE #81: Boston Market
https://www.bostonmarket.com/locations/

vikranthkr:
Fixes #81


iandees:
Thanks for adding the extras, but I took them out because they aren't generally useful and are specific to Boston Market.

Issue #88, opened 2017-01-15
Lowe's


ianmcorvidae:
https://www.lowes.com/store/

nicolas17:
GET https://www.lowes.com/wcs/resources/store/10151/storelocation/v1_0?maxResults=30&query=<ZIPCODE>.

If I set maxResults high enough, I get 221 stores, which might be all of them (in which case the specific zipcode used doesn't matter as long as it's valid). Needs more testing to ensure that's the case, or maybe finding out elsewhere if they do have 221 stores and not more.

ianmcorvidae:
https://en.wikipedia.org/wiki/Lowe%27s says 1.840 stores, so probably not.

iandees:
No progress for 2 days, re-opening this for others.

ethan626:
This link is broken. 
https://www.lowes.com/wcs/resources/store/10151/storelocation/v1_0?maxResults=30&query=

Attempted to use https://lowes.com/store, this will yeild at most 10 items. So it appears that one is required to enter new search items in order to recieved new locations. 

Looking for a workaround. 
A potential solution is brute force. Generate all possible zip codes and search using those. (This does lead to a very large number of requests) 

Ideas?

ethan626:
Using a state as a search query or setting the max results to a very high number does not seem to work. The response will only return a max of ten locations. The proposed solution is to generate zip codes and use those as search queries.

PR #513, opened 2017-12-17
Lowes Spider

ethan626:
Lowes spider scapes via zipcode. 

Website does not allow for more than 10 results at a time, making using max results in the url or searching by State not feasible. Proposed solution is to generate zip codes as search queries. This does create a lot of requests though.

#88

### DIFF
--- a/lowes.py
+++ b/lowes.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+import scrapy
+from locations.items import GeojsonPointItem
+import json 
+from urllib.parse import urlencode
+
+class LowesSpider(scrapy.Spider):
+    """" This spider scrapes Lowes store locations """
+    name = "lowes"
+    allowed_domains = ["https://www.lowes.com"]
+    start_urls = ('https://www.lowes.com/store/')
+
+    def zip_codes(self):
+        """ Generates all possible zip codes as strings """
+        import itertools as it
+
+        for i in it.product('0123456789', repeat=5): # Create all possible 5 digit zip codes as tuples
+            yield ''.join(i)                         # Join together into a zip code 
+            
+    def start_requests(self):
+        base_url = 'https://www.lowes.com/wcs/resources/store/10151/storelocation/v1_0?%s'
+        
+        headers = {'Accept-Encoding': 'gzip, deflate, br', 'Accept-Language': 'en-US,en;q=0.5',
+                   'Authorization': 'Basic QWRvYmU6ZW9pdWV3ZjA5ZmV3bw==',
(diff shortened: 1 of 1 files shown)


iandees:
I would prefer that we not make this many requests yet.

iandees:
Found this: http://lowes.know-where.com/lowes/cgi/region_list

ethan626:
@iandees I'll have a look at that page and work on building a spider for it. That page seems like a much better target. 
Thanks!

ethan626:
@iandees http://lowes.know-where.com does not appear to provide latitude and longitude data on the page nor in the responses. I created a spider which grabs the address listed on the page and committed to the Lowes branch.

lowes.py
```diff
@@ -0,0 +1,52 @@
+# -*- coding: utf-8 -*-
+import scrapy
+from locations.items import GeojsonPointItem
+import json 
+
+states = ["AL", "AK", "AZ", "AR", "CA", "CO", "CT", "DC", "DE", "FL", "GA",
```
iandees: This looks like an extra file. This should probably be in `locations/spiders/`?

locations/spiders/lowes.py
```diff
@@ -0,0 +1,54 @@
...
+
+class LowesSpider(scrapy.Spider):
+    """" This spider scrapes Lowes store locations """
+    name = 'lowes'
+    allowed_domains = ['https://know-where.com']
+    start_url = 'http://lowes.know-where.com/lowes/cgi/region_list'
+    
+    def start_requests(self):
+        for state in states:
+            url = ''.join(['http://lowes.know-where.com/lowes/cgi/region?all=true&country=US&region=',
```
iandees: Please use string formatting instead of `join()`.

locations/spiders/lowes.py
```diff
@@ -0,0 +1,54 @@
...
+                     state, '&design=default&lang=en&option=&mapid=NorthAmerica'])
+            request = scrapy.Request(url, callback=self.parse, headers=headers)
+            request.meta['state'] = state
+
+            yield request
+            
+        for province in provinces:
+            url = ''.join(['http://lowes.know-where.com/lowes/cgi/region?country=CA&region=',
```
iandees: Please use string formatting instead of `join()`.

lowes.py
ethan626: Deleted the extra file.

locations/spiders/lowes.py
ethan626: Will do.

locations/spiders/lowes.py
ethan626: Will do.

Issue #183, opened 2017-11-07
Childcare chains


dianashk:
- [ ] [Goddard School](https://www.goddardschool.com/locations)
- [ ] [Montessori School](https://www.montessori.com/montessori-schools/find-a-school/)
- [ ] [Children of America](http://www.childrenofamerica.com/locations.cfm)
- [ ] [The Learning Experience](https://thelearningexperience.com/our-centers/directory)
- [ ] [Kiddie Academy](https://kiddieacademy.com/academies/)
- [ ] [Boys and Girls Club](https://www.bgca.org/get-involved/find-a-club)
- [ ] [Big Brothers Big Sisters of America](http://www.bbbs.org/)
- [ ] YWCA [local](http://www.ywca.org/site/c.cuIRJ7NTKrLaG/b.7527667/k.C931/Local_Associations/apps/kb/cs/contactsearch.asp), [world](http://www.worldywca.org/ywca-member-associations-list/)

meghanhade:
- [ ] [Bright Horizons](https://www.brighthorizons.com/child-care-locator)

whoissteven:
I can tackle Bright Horizons

bealbrown:
I'll take YWCA local and world

iandees:
This issue hasn't been worked on in a while. I'm removing the assignment so others can work on it. If you're still working on it, please reassign yourself and reference a pull request or branch where you're doing your work.

PR #518, opened 2017-12-18
Bright Horizons

whoissteven:
for #183


whoissteven:
Hmm.. Seems like the regex URL validation takes a while so it times out. If this is an issue I will create a simpler regex validation string.

PR #537, opened 2017-12-18
Harris Teeter

### LINKED ISSUE #148: Harris Teeter
Mostly southeastern https://www.harristeeter.com/store/#/app/store-locator

vlad555:
Fixes #148


vlad555:
It's sad to admit,
But it was necessary to integrate cookies into spider :-(
They are dynamically generated and site don't give answer without them.
So they could expire in future

iandees:
I think the session cookies time out after a while. It might be a good idea to start at a normal page, [let it set cookies](https://blog.scrapinghub.com/2016/03/23/scrapy-tips-from-the-pros-march-2016-edition/) and then spider into the other pages?

vlad555:
Sad to  admit, but it's necessary javascript to generate cookies.
I don't understand how to move further without javascript rendering here

evanmosseri:
I'm on it

vlad555:
cool!!
do we have permission to use javascript rendering libs?
it's not first one I need these libs.

evanmosseri:
You don't need javascript for this, you just need to get the cookies from a response object and pass it down to further requests

vlad555:
:-) I'll be glad if it will be enough
Let's wait your solution .

vlad555:
I don't know if you need it, but cookies you need to send request:
_gag
HTAPIADMSESID
_ga
_gid
First pair you really got from first request
Look where to take second.
Good luck

PR #561, opened 2017-12-19
The Library

### LINKED ISSUE #38: The Library
https://www.worldcat.org/libraries

whoissteven:
Fixes #38

### DIFF
--- a/locations/spiders/thelibrary.py
+++ b/locations/spiders/thelibrary.py
@@ -27,7 +27,6 @@ class TheLibrarySpider(scrapy.Spider):
         response.selector.remove_namespaces()
         xml_urls = response.xpath('//url/loc/text()').extract()
         for path in xml_urls:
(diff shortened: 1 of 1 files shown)


whoissteven:
I'm having some issues as each country has their own formats. Some have an additional line prior to the address, and the address is found in a single <p>, so I am needing to split the address to make it work. Is it fine to just have addr_all, and combine it in a single line or keep looking into better options?

whoissteven:
Idea:
I can easily scrape US addresses.
I see that there is a google maps URL for each library, will try to see if I can do something with that. At the very least I can pull library names and lat/lon. I will see if I can think of a cleaner method to pull the international addresses.

iandees:
I think this should be closed as a dupe of #590. Feel free to reopen if you're interested in tackling the international stuff.

PR #565, opened 2017-12-19
McDonalds -  United Arab Emirates

SelfnessAid:
For #1 

Added  United Arab Emirates to McDonaldsSpider


SelfnessAid:
@iandees 
this branch is related with #562 
we need to keep both of this and #562 .
thanks.

Issue #40, opened 2017-01-15
A&W restaurants


kgjenkins:
http://www.awrestaurants.com/locations

carhutt:
Got it. @kgjenkins do I need to self-assign?

iandees:
Yep, if you're working on one of these please do self-assign.

carhutt:
Do I need to be added as a collaborator?  Sorry, this is my first one.

iandees:
Invited!

iandees:
Removing the assignee on this so others can give it a shot.

iandees:
This issue hasn't been worked on in a while. I'm removing the assignment so others can work on it. If you're still working on it, please reassign yourself and reference a pull request or branch where you're doing your work.

PR #578, opened 2017-12-19
A&W Restaurants

MagnetonIO:
For #40


iandees:
I'm going to merge this, but we need to spider across zipcodes so we can pull in more stores.

PR #604, opened 2017-12-20
Menards

### LINKED ISSUE #116: Menards
http://www.menards.com/main/storeLocator.html

DerekMcDaniel:
Added spider for Menards. 

Bypassed Incapsula but did not have access to javascript for lat/lon or store list. I was able to get full address however. 

Fixes #116 

@iandees  --side note: Is it ok to create new issues? Running out of stuff to work on. 

Thanks.

### DIFF
--- a/locations/spiders/menards.py
+++ b/locations/spiders/menards.py
@@ -1,0 +1,87 @@
+import scrapy
+from locations.items import GeojsonPointItem
+
+base_url = 'https://216.222.184.133/main/storeDetails.html?store='
+store_list = [str(i) for i in range(3012, 3345)]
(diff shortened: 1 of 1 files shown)


iandees:
Yep, it's ok to create new issues. Please make sure you aren't duplicating a ticket that already exists. Also, it'd be preferable to include stores that are local to you so we can get a diverse array of places.

maxerickson:
The addresses and coordinates are present in the html of https://www.menards.com/main/storeLocator.html with `storeList` on each line that has store data. Bypassing using the IP address works.

    wget --no-check-certificate https://216.222.184.133/main/storeLocator.html
    grep storeList storeLocator.html | wc -l
    307

PR #621, opened 2017-12-20
McDonalds - Portugal

SelfnessAid:
For #1 

Created the Spider for Portugal of McDonalds.

### DIFF
--- a/locations/spiders/mcdonalds_pt.py
+++ b/locations/spiders/mcdonalds_pt.py
@@ -103,8 +103,8 @@ class McDonaldsPTSpider(scrapy.Spider):
             properties = {
                 'ref': store['id'],
                 'phone': store['phone'],
-                'lon': store['coordinates']['latitude'],
-                'lat': store['coordinates']['longitude'],
+                'lat': store['coordinates']['latitude'],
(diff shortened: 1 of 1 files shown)


SelfnessAid:
this website need to input all postal code of Portugal to get all mcdonald's restaurants in Portugal.
I solved this by crawling [wiki page](https://en.wikipedia.org/wiki/List_of_postal_codes_in_Portugal) first.
I think we can crawl websites which require to enter all zipcode with this way, without usage of any other package.

locations/spiders/mcdonalds_pt.py
```diff
@@ -0,0 +1,130 @@
...
+        match = re.sub('<[^<]+?>', '', address)
+        return ' '.join(match.split())
+
+    def parse_store(self, response):
+        stores = json.loads(response.body_as_unicode())
+        for store in stores:
+            properties = {
+                'ref': store['id'],
+                'phone': store['phone'],
+                'lon': store['coordinates']['latitude'],
```
iandees: You have latitude and longitude flipped here.

SelfnessAid: haha, sorry for it.
I'll fix it now.

PR #622, opened 2017-12-20
Hy-Vee

### LINKED ISSUE #347: Hy-Vee
https://www.hy-vee.com/Stores/default.aspx

sledjama:
Fixes #347 

HyVee has 155 locations, this scraps all.
HyVee implemented a server-side javascript detector that returns the HTML head with an empty body asking that we enable javascript to browse the site.
The only shortcut around this seems to be to use the PUT method which then required re-implementing the start_request method.


iandees:
Thanks for the great documentation here. Very helpful!

PR #715, opened 2017-12-21
McDonalds - Israel

SelfnessAid:
For #1 

Created the Spider for Israel of McDonalds.


SelfnessAid:
@iandees 
I noticed there is one additional field - _extra_ after pull last master branch and crawling speed is a little slow.
it's okay?

iandees:
The `extras` field was added a few weeks ago so people could add extra information that might be useful. It's not required and don't worry about filling it in.

Issue #655, opened 2017-12-21
Regis Salons


iandees:
https://www.regissalons.com/salon-locator.html

DerekMcDaniel:
working on this

PR #728, opened 2017-12-21
Regis Salons

DerekMcDaniel:
For #655 

Added spider for Regis Salons

### DIFF
--- a/locations/spiders/regis_salon.py
+++ b/locations/spiders/regis_salon.py
@@ -1,0 +1,101 @@
+import scrapy
+from locations.items import GeojsonPointItem
+import json
+import re
+
+regex_times = r"\d{1,2}:\d{1,2}\s?[Aa]?[Mm]?[Pp]?[Mm]?\s?-\s?\d{1,2}:\d{1,2}" \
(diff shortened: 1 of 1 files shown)


DerekMcDaniel:
ah somethings wrong, I think ill have to search by full state rather than abbreviation. Will fix !

DerekMcDaniel:
Actually turns out that even though it includes the search state in the url it defaults back to lat/lon search. I think I can use the form to populate the lat/lon however

DerekMcDaniel:
Still need to check smaller radius for more stores. Results are capped at 100 per. Also, Regis owns a lot of brands such as supercuts, its unclear at the moment whether all of those locations will be served

DerekMcDaniel:
@iandees Searching by 200 mile radius returned ~30 less results. Also I checked store locator on supercuts vs Regis salons and they do not match up. So while Regis Salons is reported to have more than 9,000 locations it looks like they will be have to done individually.  With the large radius I can include Canada and the UK without too many requests.

DerekMcDaniel:
Picking up Canada but not UK. Looks like UK has its own website

iandees:
Please make another spider for the UK.

PR #829, opened 2017-12-27
Harris Teeter fixes PR #537

IAlwaysBeCoding:
It is important to notice, that there was no need to generate any dynamic cookies as the cookies were given on the first request(empty your cookies and refresh the page and see).

Secondly, since there was no need to generate cookies, there was also not a need to extract cookies from the Response and pass them to the next request. Scrapy does that for you automatically, and unless there are cookies generated dynamically via Javascript, you shouldn't need to insert any cookies at all.

On top of that the cookies `_ga` , `_gid` and `_gag` are google cookies that are set by their analytics and ads scripts. There is no way the sites could check for those cookies and know what they mean. In other words, those cookies can be discarded and ignore because those cookies will be shown on like 90% of the sites but the sites themselves do not use them, google does though.

Now, let's get on the fix.  In order to know what really needed to be changed, one must understand the http status code that was returned. Due to the fact that this is a simple Ajax driven API, there are going to be some check up to see if the session is still active. When the site is first requested, the url `https://www.harristeeter.com/store#/app/store-locator` will first be loaded. After that, the browser does a simple POST request to `https://www.harristeeter.com/api/checkLogin' to check if the session is still active.

This request will return a 401 http status code on the first load, since cookies are new. Therefor, Scrapy will not processed any Requests that are non 200'ish, and 301-303. This is the default behaviour. So we need to allow Scrapy on a per-spider basis(such as this case) to allow it to process 401 requests.

We do this by setting this property as a class attribute on the Spider it self, `handle_httpstatus_list = [401]` . The `handle_httpstatus_list` needs to be a list, so in this case a list with 1 value.

I also removed the manual hard-coded cookies that were there,and added the check_login request.


iandees:
Please make sure you're adding this file alongside the other spiders in `locations/spiders/`.

vlad555:
Great work!!!

IAlwaysBeCoding:
@iandees Sorry about that, small mistake. @vlad555 thanks man I appreciate your kind words!

Issue #822, opened 2017-12-24
Cheesecake Factory


iandees:
http://locations.thecheesecakefactory.com/

PR #845, opened 2018-01-05
Add The Cheesecake Factory spider

SrihariThalla:
For #822


SrihariThalla:
@iandees I'm not sure how to handle the `addr_full` param. Any ideas?

PR #901, opened 2018-10-05
Shake Shack spider

LaurenSpiegel:
Retrieves 178 out of 189 locations

### DIFF
--- a/locations/spiders/corepower_yoga.py
+++ b/locations/spiders/corepower_yoga.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+import scrapy
+import json
+from locations.items import GeojsonPointItem
+
+
+class CorePowerYogaSpider(scrapy.Spider):
+    name = 'corepower_yoga'
+    allowed_domains = ['www.corepoweryoga.com']
+    start_urls = (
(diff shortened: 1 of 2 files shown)


iandees:
Hi Lauren! Thanks for this pull request! Do you know why it's only pulling in 178 instead of all 189?

LaurenSpiegel:
Hi @iandees ,

The format of the location pages are not completely consistent.  For instance, the one I started with was Birmingham, AL and you can see that the phone number is just in a <p> along with the address:'


screen shot 2018-10-05 at 9 51 21 am

This works to pull the phone number in my spider. However, the Detroit location nests the phone number in another span tag:

screen shot 2018-10-05 at 9 52 34 am

I'm also getting a few errors on pulling the city name since I'm assuming there is a comma in the sr-only tag and sometimes there isn't like with the Dubai International airport location -- https://www.shakeshack.com/location/dubai-dxb-airport

I'm pretty new to this. Any tips on dealing with inconsistent pages like this? Someone recommended this: https://github.com/datamade/usaddress

Meanwhile, I added CorePower Yoga which was pretty simple since they are calling for json.

Thanks

PR #947, opened 2018-12-05
Add spider for Texas Roadhouse

### LINKED ISSUE #803: Texas Roadhouse
https://www.texasroadhouse.com/locations/

thismakessand:
Fixes #803

### DIFF
--- a/locations/spiders/texas_roadhouse.py
+++ b/locations/spiders/texas_roadhouse.py
@@ -1,0 +1,153 @@
+import datetime
+import xml.etree.ElementTree as ET
+
+import scrapy
+
+from locations.items import GeojsonPointItem
+from locations.hours import OpeningHours
+
+MAX_RETRIES = 3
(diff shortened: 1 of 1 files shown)


locations/spiders/texas_roadhouse.py
```diff
@@ -0,0 +1,153 @@
...
+
+        return hours.as_opening_hours()
+
+    def parse_store(self, response):
+        retries = 0
+        ref = response.meta['id']
+        name = response.meta['name']
+
+        try:
+            root = ET.fromstring(response.text)
```
iandees: The response already comes in parsed as XML. You can use `response.xpath()` to query the XML document so you don't have to re-parse it inside the spider.

thismakessand: @iandees  Thanks!  I made the switch to response.xpath and cleaned up some other stuff.

PR #949, opened 2018-12-10
Banfield pet hospital spider

### LINKED ISSUE #872: Banfield Pet Hospital
https://www.banfield.com/our-hospitals/hospital-locations

thismakessand:
Fixes #872 

1037/1037 scraped.

Also included an update to hours.OpeningHours where previously not adding a range for a day (ie, store is closed) would still create an entry for that day.
Eg: ``"Mo-Sa 07:00-14:00; Su "`` instead of ``"Mo-Sa 07:00-14:00"``

### DIFF
--- a/locations/hours.py
+++ b/locations/hours.py
@@ -55,7 +55,9 @@ class OpeningHours(object):
             opening_hours = '24/7'
         else:
             for day_group in day_groups:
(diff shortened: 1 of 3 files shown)


iandees:
Awesome, thanks!

PR #1024, opened 2019-02-07
feat(spider): add Casino spider

EpicKiwi:
This is a spider for French shops Casino.
This spider retrieve 408 shops from https://magasins.supercasino.fr/fr and all sub pages.

I added `addr_2_housenumber` because some supermarkets have multiple numbers.

### DIFF
--- a/locations/spiders/casino.py
+++ b/locations/spiders/casino.py
@@ -43,7 +43,7 @@ class TemplateSpider(scrapy.Spider):
             "postcode": postalCode,
             "housenumber": roadNum,
             "street": street,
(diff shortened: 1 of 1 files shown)


locations/spiders/casino.py
```diff
@@ -0,0 +1,55 @@
+# -*- coding: utf-8 -*-
+import scrapy
+from locations.items import GeojsonPointItem
+import re
+
+class CasinoGeojsonPointItem(GeojsonPointItem):
+    addr_2_housenumber = scrapy.Field()
```
iandees: Instead of creating your own item, please add extra fields to the existing `extras` property.

PR #1076, opened 2019-04-10
Fix dollar general spider

### LINKED ISSUE #1027: Dollar General
The results from the Dollar General spider #344 currently only includes 8,146 locations out of approximately 15,000 locations.

wessport:
Closes #1027

Corrected the issue where the scraper only returned stores 50 miles from the center of each state. 

To avoid having to make thousands of zip code requests, I created a US grid of minimally overlapping circles representing a 50mile search radius. The scraper passes the lat/lon of each centroid point (2,108 in total) in the post request.

I stored the points in a new directory under ‘searchable_points’ as ‘us_centroids_50mile_radius.csv’.

Hopefully these searchable points could be used for other scrapers that allow lat/lon searches. I have a few other versions for 100, 25, and 10miles in addition to the 50mile radius if they might also be helpful. I included a state attribute as well that could be used for instances where it’s known that a store only exists in certain states.


iandees:
Nice! That's a great idea and should help with a bunch of other spiders ([the ones that require latlon](https://github.com/alltheplaces/alltheplaces/labels/requires%20latlon%20search) and maybe [the ones that require zipcode](https://github.com/alltheplaces/alltheplaces/issues?q=is%3Aissue+is%3Aopen+label%3A%22requires+zip+search%22), too).

PR #1079, opened 2019-04-13
Deduped point CSV files

swingley:
The recently added [CSV files with points](https://github.com/alltheplaces/alltheplaces/tree/master/locations/searchable_points) used to hit APIs that require lat long pairs are a big help but they contain a bunch of duplicate points. The files included in this PR remove all the duplicate points from those files. Here's the [code I used to discard duplicates](https://gist.github.com/swingley/9390a3ac50b09b1c68135c7de6082d13).

/cc @wessport since they added the original files


swingley:
Also here's a gif with transparency on points set to 0.2 and toggling between the two versions of the file so you can see what changed:

iandees:
Thanks!

wessport:
@swingley 

Excited to see these points already being used!

A search radius may overlap multiple states especially when it’s centered near a state boundary. This creates a one to many relationship between the search radius point and the states covered in that search zone. To allow for quick filtering in instances where it’s known that a store only appears in certain states, a state attribute was included with the points, resulting in records that share the same latlon associated to differing states. That way a radius that covers both Tennessee and Kentucky can be used for requests for either state. 

When using the points at a national scale, there will be duplicate requests, though the duplicate results should be automatically filtered out by the DuplicatesPipeline. 

I think the proper fix here is to include a separate set of searchable point files intended for use at a national scale that doesn’t include the state attribute and another set that allows for filtering by state, otherwise we lose that functionality. I can add both sets and more documentation when I get back from travel Tuesday.

iandees:
Ah, that makes sense @wessport. Did one of the spiders require a state to be specified?

swingley:
@wessport thanks for clarifying, the duplicates pipeline does prevent duplicate results but it seemed wasteful to repeatedly do the same request only to discard the results. Having files for the state and national levels sounds good.

Issue #1034, opened 2019-02-19
Braum's


wessport:
https://www.braums.com/store-locator/

davidhicks:
WPStoreLocator spider which only returns a maximum of 100 results per 500 mile radius geographic search query e.g. `https://www.braums.com/wp-admin/admin-ajax.php?action=store_search&lat=35.46756&lng=-97.51643&max_results=100&search_radius=500`

The store finder page states:
> "We have over 300 Braum’s locations throughout Oklahoma, Kansas, Texas, Missouri, and Arkansas."

Thus the number of searches required is hopefully not too great a number, possibly just one query per state.

CloCkWeRX:
No wikidata?

PR #1082, opened 2019-04-18
Searchable Points Update

wessport:
Adds a new set of searchable points intended for situations in which it's desirable to filter the points by state first e.g. #1034 .  

I've added some clarifying language to the README as well to help outline the difference between the set of points intended for a national search and the set intended for state searches. 

Let me know if there are any questions!


iandees:
Looks good. Can you resolve the conflict with the ones you added before?

wessport:
@iandees no problem! Should be resolved now.

PR #1098, opened 2019-05-24
Fix Costco spider

thismakessand:
Costco removed its list of warehouses and they are now only available via lat/long searches.

Scrapes 538/538 stores in the US.

### DIFF
--- a/locations/spiders/costco.py
+++ b/locations/spiders/costco.py
@@ -23,13 +23,16 @@ DAYS_NAME = {
 class CostcoSpider(scrapy.Spider):
     name = "costco"
     allowed_domains = ['www.costco.com']
-
-    download_delay = 0.5
+    start_urls = (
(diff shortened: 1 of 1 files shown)


thismakessand:
Hmm... I'm not getting any timeout errors when I `pipenv run scrapy crawl costco` locally with the User-Agent override in the script.  If I don't override the User-Agent and use the default ` 'Mozilla/5.0 (Linux; rv:1.0) https://github.com/alltheplaces/alltheplaces'` the requests are timing out, so I'm guessing we're blocked?
If so, I guess we can decommission this one?

iandees:
Yea, they're probably blocking it from Travis CI IPs. Let's keep it in here and see what happens when the larger run happens. At the very least, it'll be useful to have here so someone can run it locally if they want.

PR #1100, opened 2019-05-28
Fix broken Olive Garden spider

thismakessand:
Olive Garden spider currently yields 0 records.

Updated to match on https urls and switch to better data source within location pages.

862 locations scraped.


iandees:
Did you mean to close this? The "timeout" thing is ok.

thismakessand:
Yeah, I noticed right after opening the PR that some of the addresses are getting messed up and I want to fix it first, I'll reopen probably tomorrow

Issue #242, opened 2017-12-05
Burger King


iandees:
https://locations.bk.com/index.html

josh0203e:
Have done  the burger king

PR #1103, opened 2019-06-03
fix Burger King spider

wessport:
Previously only captured 4,470 stores. Updated to capture all 7,233/7233 stores in the US.


iandees:
Related to #242 .

Issue #152, opened 2017-01-18
Dennys


ianmcorvidae:
https://www.dennys.com/locations/

PR #1116, opened 2019-07-01
Fix address targets

hereticaneue:
Sorry, this one is perfect now. Address targets needed to be updated in previous commit.


iandees:
For Denny's #152

Issue #121, opened 2017-01-16
Dunham's Sports


maxerickson:
http://www.dunhamssports.com/store-map/

AnthoneyKalasho:
http://www.dunhamssports.com/wp-admin/admin-ajax.php?action=simloc_get_locations&lat=32.7258378&lng=-116.95580669999998&radius=200000

PR #1120, opened 2019-07-18
Fix spider for Dunham's Sports

hereticaneue:
Added name field and fixed ref so that blank "duplicates" no longer get dropped. Now properly scrapes all 234 locations.


iandees:
#121

Issue #20, opened 2017-01-15
Subway


kgjenkins:
http://www.subway.com/en-us/findastore

iandees:
No progress for 6 days, re-opening this for others.

vlad555:
same situation as walgreens
we need lat/lon pair to get near subways.
page sends request to google service to get lan/lon from search string (city or zip)
If we can use google service to get position from ZIP and after that scan all ZIPs in USA
I can create file with all ZIPs and lat/lon which correspond them.
And we can use this file in several spiders.
What do U think?

Cj-Malone:
Unless I'm missing something there is a Subway spider, and it currently works.

PR #1129, opened 2019-07-22
Subway Scraper - progress made on Issue #20

clonergan-indeed:
Add spider for Subway (US) locations. Subway limits access, so a default request delay of 2 seconds is required. This makes the data acquisition very slow, but it can be run successfully from a home machine. To get worldwide locations, we will need searchable points for other countries. 

Current set-up and searchable points will pull in non-US subways near US borders.


iandees:
https://github.com/alltheplaces/alltheplaces/issues/20

PR #1134, opened 2019-07-25
Brookdale

### LINKED ISSUE #968: Brookdale
https://www.brookdale.com/en/locations.html

nbeecher:
Closes #968

### DIFF
--- a/locations/spiders/brookdale.py
+++ b/locations/spiders/brookdale.py
@@ -30,7 +30,6 @@ class TemplateSpider(scrapy.Spider):
         i = 0
         while i < len(data):
 
-            print(data[i]['name'])
             properties = {
                 "ref": data[i]['community_id'],
                 "name": data[i]['name'],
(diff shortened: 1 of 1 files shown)


locations/spiders/brookdale.py
```diff
@@ -0,0 +1,48 @@
...
+            url = ''.join([URL, state])
+            yield scrapy.Request(url, callback=self.parse_info)
+
+
+    def parse_info(self, response):
+
+        data = json.loads(response.body_as_unicode())
+
+        i = 0
+        while i < len(data):
```
iandees: It looks like you could just use a for loop here instead of a while loop. For example:

```python
for row in data:
  properties = { ... }
  yield GeojsonPointItem(**properties)
```

locations/spiders/brookdale.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def parse_info(self, response):
+
+        data = json.loads(response.body_as_unicode())
+
+        i = 0
+        for row in data:
+
+            properties = {
+                "ref": data[i]['community_id'],
```
iandees: Now you can remove the `i` variable and use `row` instead of `data[i]` throughout.

iandees:
Awesome, thank you!

PR #1147, opened 2019-08-06
Mcdonalds

nbeecher:
For McDonalds locations in the US there are store numbers we can scrape as well.  I added an extras: number: to the output for USA locations.

### DIFF
--- a/locations/spiders/mcdonalds.py
+++ b/locations/spiders/mcdonalds.py
@@ -83,7 +83,7 @@ class McDonaldsSpider(scrapy.Spider):
                     'lat': store['geometry']['coordinates'][1],
                     'extras':
                     {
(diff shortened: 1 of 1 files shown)


iandees:
The "ref" property is supposed to be the store number … Does that not line up?

nbeecher:
The "ref" property for McDonalds is a global unique identifier.  But it appears the US locations also have a 4 digit identifier that also matches a store number.  ("ref": 195500253343-en-us, "number": 6300).  Not all countries appear to have this same store number format.

iandees:
Gotcha, ok. For future reference, can you add a comment in the spider near line 86 (where you're adding "number" to extras) that mentions what the value is?

PR #1153, opened 2019-08-12
Add spider for Ernst & Young

thismakessand:
Scrapes 816 EY office locations worldwide

### DIFF
--- a/locations/spiders/ernst_young.py
+++ b/locations/spiders/ernst_young.py
@@ -8,8 +8,8 @@ from locations.items import GeojsonPointItem
 from locations.hours import OpeningHours
 
 
-class EySpider(scrapy.Spider):
-    name = "ey"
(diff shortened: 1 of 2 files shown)


thismakessand:
Is this the first place in Greenland? :greenland:

iandees:
I'm feeling like we might want to spell out "EY". I don't think I've ever heard them referred to with that acronym before.

thismakessand:
Yeah, I wasn't sure what to call it, but their website is all branded "EY" and it states at the bottom "EY refers to the global organization, and may refer to one or more, of the member firms of Ernst & Young Global Limited, each of which is a separate legal entity."

If you just mean the spider name/file name, then yeah I can for sure update that - the name that is being scraped is the name provided in the data.

iandees:
Yea can you update the spider name and file name to `ernst_young`?

PR #1182, opened 2019-09-16
Fix spider for Tim Hortons

nbeecher:
Fix for Tim Hortons Spider

### DIFF
--- a/locations/spiders/timhortons.py
+++ b/locations/spiders/timhortons.py
@@ -2,52 +2,67 @@ import scrapy
 import re
 from locations.items import GeojsonPointItem
 
-class TimhortonsSpider(scrapy.Spider):
 
+class TimhortonsSpider(scrapy.Spider):
     name = "timhortons"
     allowed_domains = ["locations.timhortons.com"]
+    custom_settings = {
+        'DUPEFILTER_CLASS': 'scrapy.dupefilters.BaseDupeFilter',
+    }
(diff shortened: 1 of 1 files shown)


locations/spiders/timhortons.py
```diff
@@ -2,52 +2,67 @@
 import re
 from locations.items import GeojsonPointItem
 
-class TimhortonsSpider(scrapy.Spider):
 
+class TimhortonsSpider(scrapy.Spider):
     name = "timhortons"
     allowed_domains = ["locations.timhortons.com"]
+    custom_settings = {
+        'DUPEFILTER_CLASS': 'scrapy.dupefilters.BaseDupeFilter',
```
iandees: Why did you need to add this?

nbeecher: I had to hit parse_city_stores sometimes twice (in def parse_city:) depending if there were multiple stores for one URL so I had to do a try except to figure out if the url contained 1 store or multiple.  Scrapy by default does not allow you to send a request on the same URL twice. The original scraper assumed single store/multiple stores based on regex of the URL but that was incorrect.

iandees:
Thank you for doing this! While I was testing your changes, it looked like they're using the same company as several others, so it was an easy copy/paste from an existing spider. I did that in https://github.com/alltheplaces/alltheplaces/pull/1199.

PR #1184, opened 2019-09-17
Add spider for Victra mobile phone stores

thismakessand:
Victra is an authorized Verizon reseller so it looks like these would also be scraped by the Verizon spider.

Not sure if having multiple sources for the same locations would be useful to the project, so feel free to close this PR if not.


iandees:
Sure I think that's fine.

PR #1193, opened 2019-10-01
Add searchable points for Canada

wessport:
Generated searchable points for Canada to help with Canadian stores that require a latlon search similar to #1076 . Should help with scrapers like #1129 (tagging @clonergan-indeed).  

They're setup the same way as the US points. There's one set that's intended for national searches and another set that includes a territory attribute for instances where searching across territories like Nunavut is unnecessary (discussed in more depth previously here https://github.com/alltheplaces/alltheplaces/pull/1079#issuecomment-482915993).


iandees:
Nice, thanks. Do you think it would be helpful to update the docs related to these?

wessport:
@iandees Definitely! Added some language mentioning CA points are also available :thumbsup:

PR #1196, opened 2019-10-02
Add spider for Haggen

thekuiperbelt:
Add spider for Haggen: http://www.haggen.com/find-our-stores/

Latitude/Longitude not available (they are being geocoded via google on the fly)


thekuiperbelt:
Scrapes 15/15 locations

PR #1203, opened 2019-10-04
Hollister

nbeecher:
Added spider for Hollister.  I had to add the custom setting for user agent because the request will be blocked otherwise.


nbeecher:
@iandees do I need to edit this spider or is it okay to be merged in?

PR #1213, opened 2019-10-15
Add spider for First Cash

nbeecher:
I had to use the searchable points for this scraper which got all the stores in the US.  But 60% of first cash stores are in latin america, so this spider will need to be updated once we have searchable points for that region of the world.


nbeecher:
@iandees just wondering the status of this spider?

iandees:
I read that you wanted to wait for more searchable points. I can merge it in for now.

Issue #204, opened 2017-11-20
Dollar Tree


iandees:

whoissteven:
Task Completed.
Pending addition as Contributor to push.

nvkelso:
Your your Github registered email for the contributor invite.

whoissteven:
Steven@WhoisSteven.com

PR #1224, opened 2019-10-28
Fix Dollar Tree (was scraping 0 locations)

thismakessand:
Also cleaned up opening hours to make use of the OpeningHours class.


iandees:
#204

PR #1225, opened 2019-10-28
Add spider for Simon Malls

thekuiperbelt:


thekuiperbelt:
Scrapes 205/206 US locations, skips 1 location that is not open yet.

Issue #899, opened 2018-08-28
There is issue with the seven eleven spider.


vdkotian:
I was trying to scrape seven eleven website, but to my notice that the spider was not able to fetch and scrape data. I tried digging a bit and found that the API mentioned in the no code was no logger working. I have created spider for seven eleven which is scraping data, it would be wonderful if you could review and let me contribute towards this project.

iandees:
Hello! Please file a pull request so that we can see your updated scraper and include it.

PR #1228, opened 2019-10-28
Fix 7-11 spider, was scraping 0 locations

thismakessand:
They now have a store directory so this spider now uses that instead of the old api


thismakessand:
Closes #899

PR #1238, opened 2019-10-29
Add spider for Sherwin Williams

thismakessand:
Scrapes 246 paint, commercial paint, and other facilities in US and CA.


thismakessand:
Hmm, those results look weird, let me look at it some more

thismakessand:
Switched to 50mile radius search and added extra lat/lon searches for a few markets that had >25 results in less than 50 mile radius.

PR #1243, opened 2019-10-29
Update Advance Auto spider to include store name and number

hereticaneue:


iandees:
Don't forget to update your pull request name to include something about the spider you're modifying for easier book keeping.

locations/spiders/advanceautoparts.py
```diff
@@ -25,6 +29,8 @@ def parse_stores(self, response):
             'website': response.url,
             'lat': response.xpath('normalize-space(//meta[@itemprop="latitude"]/@content)').extract_first(),
             'lon': response.xpath('normalize-space(//meta[@itemprop="longitude"]/@content)').extract_first(),
+            'name': response.xpath('//div[@class="LocationName-geo"]/text()').extract_first(),
+            'extras': {'number':store_num}
```
iandees: `ref` is supposed to be the store number as long as it's unique among all the stores for this spider. Does it make sense to switch?

hereticaneue: Oh right, I meant to change the ref to the store number! I can do that.

PR #1247, opened 2019-10-30
Fix for Gucci

nbeecher:
Gucci spider was only scraping 64 places.  Now it is scraping all 574 places.

### DIFF
--- a/locations/spiders/gucci.py
+++ b/locations/spiders/gucci.py
@@ -1,69 +1,84 @@
+# -*- coding: utf-8 -*-
 import scrapy
 import re
+import json
+
 from locations.items import GeojsonPointItem
+from locations.hours import OpeningHours
+
+DAYS_NAME = {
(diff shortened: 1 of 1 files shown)


locations/spiders/gucci.py
```diff
@@ -1,69 +1,84 @@
...
+                'city': city,
+                'state': state,
+                'postcode': postal,
+                'country': store["properties"]["address"]["country"],
+                'lat': store["properties"]["latitude"],
+                'lon': store["properties"]["longitude"],
+                'phone': store["properties"]["address"]["phone"],
+                'website': response.urljoin(store["properties"]["url"]),
+                'extras': {
+                    'number': store["properties"]["storeCode"]
```
iandees: The store number should be in the `ref` property and not in `extras`.

iandees: Ah, gotcha.

PR #1249, opened 2019-10-30
Fixes AT&T post code

wessport:
Fixes the xpath for capturing post code.


locations/spiders/att.py
```diff
@@ -1,6 +1,8 @@
-import scrapy
 import json
 import re
+
+import scrapy
```
iandees: Can you put this line back at the top of the file?

wessport: Will do!

wessport:
@iandees thanks for catching that! I think the most recent commit cleans everything up. Let me know if I missed anything!

Issue #527, opened 2017-12-18
Family Dollar


iandees:
https://www.familydollar.com/content/familydollar/en/store-results.html

They're a client of Brandify, which means we can re-use the AJAX parsing we do in other Brandify stores.

iandees:
This issue hasn't been worked on in a while. I'm removing the assignment so others can work on it. If you're still working on it, please reassign yourself and reference a pull request or branch where you're doing your work.

skorasaurus:
is the parsing for the brandify stores shared somewhere? a repo search of brandify didn't turn up anything

iandees:
Not yet, but it should be.

PR #1261, opened 2019-10-31
Fix for family dollar

nbeecher:
Around 12 records were missing due to an opening hours error (they did not have opening hours).  I added a try except to account for these


iandees:
#527

Issue #12, opened 2017-01-12
Taco Bell


iandees:
https://www.tacobell.com/locations

iandees:
Requires an annoying signed token generated for every request.

iandees:
No activity in several days, so I'm un-assigning. Please re-assign yourself if you're actively working on it.

PR #1267, opened 2019-11-01
Fix Taco Bell spider to collect postal code

hereticaneue:
postalcode field was collecting the locality instead, so I updated it to grab the postalcode


iandees:
#12

PR #1268, opened 2019-11-02
Qdoba

wessport:
Repairs Qdoba spider which was only returning results for NY. Now scrapes all 732 US and CA locations.

### DIFF
--- a/locations/spiders/qdoba.py
+++ b/locations/spiders/qdoba.py
@@ -1,101 +1,85 @@
+# -*- coding: utf-8 -*-
 import scrapy
 import json
 import re
-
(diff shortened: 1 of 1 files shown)


iandees:
Do you think you could rebase this branch from master so that there's not a ton of other commits on it?

wessport:
Yeah sure thing. Let me see what I can do.

wessport:
@iandees I think I finally figured it out. Heck of a way to kick off a Friday night 😄! Let me know what you think!

iandees:
Woo git party! Looks good, thanks.

PR #1272, opened 2019-11-03
add spider for Vue Cinemas

chris48s:
Vue is a chain of cinemas with branches in the UK and Republic of Ireland. This spider covers both countries.


chris48s:
I'll pick up fixing the failures on this one in the week. I hadn't spotted when running locally that https://www.myvue.com/cinema/doncaster/getting-here and https://www.myvue.com/cinema/darlington/getting-here have a slightly different page format. I need to find a slightly more robust xpath selector for pulling out the address text :man_facepalming:

chris48s:
updated - should be good for review now

PR #1281, opened 2019-11-05
Fedex Office

wessport:
Scrapes 2784 Fedex Office US locations including FedEx Freight and FedEx Ship Centers. Excludes drop box locations. 

I included in the extras the location type and location code in case anyone needs to filter the output to specific locations types. There is a considerable run time ~ 13hrs. Their site doesn’t allow for the searchable points so in order to circumvent the zipcode search I had to visit each state’s list of cities which also returns nearby cities leading to duplicate filtering.

### DIFF
--- a/locations/spiders/fedex_office.py
+++ b/locations/spiders/fedex_office.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+import json
+import re
+
+import scrapy
+
+from locations.items import GeojsonPointItem
+from locations.hours import OpeningHours
+
+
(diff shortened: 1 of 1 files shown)


iandees:
Would it be faster to use the sitemap as a starting point? https://local.fedex.com/sitemap.xml

wessport:
@iandees the state-level-sitemap is a wash, but starting at the individual location types works beautifully, plus we can get CA locations and their store hours to boot.

Now scrapes 3,212 locations in...13min :wink:

Issue #64, opened 2017-01-15
Starbucks 


skorasaurus:
On mobile, so didn't find an endpoint yet

iandees:
This issue hasn't been worked on in a while. I'm removing the assignment so others can work on it. If you're still working on it, please reassign yourself and reference a pull request or branch where you're doing your work.

Cj-Malone:
This can be closed. There is a Starbucks and Starbucks EU scraper.

skorasaurus:
there is! thanks for pointing it out - https://github.com/alltheplaces/alltheplaces/blob/6737cea2da20a2e05a0f3d4619df352d6c7f933b/locations/spiders/starbucks.py

PR #1282, opened 2019-11-05
Add Canadian locations to Starbucks spider and additional attributes

thismakessand:
Gets Canadian Starbucks now

#64

### DIFF
--- a/locations/spiders/starbucks.py
+++ b/locations/spiders/starbucks.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import csv
 import scrapy
 import json
 from locations.items import GeojsonPointItem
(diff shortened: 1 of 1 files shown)


locations/spiders/starbucks.py
```diff
@@ -16,7 +16,20 @@ def start_requests(self):
         with open('./locations/searchable_points/us_centroids_50mile_radius.csv') as points:
             next(points)
             for point in points:
-                row = point.split(',')
+                row = point.strip().split(',')
```
iandees: Instead of splitting on comma, can you use `csv.reader` or `csv.DictReader`? That'd make it a bit cleaner and more robust.

PR #1295, opened 2019-11-15
Add spider for Calvin Klein

nbeecher:


iandees:
It looks like there are some errors handling stores in Europe: https://transfer.sh/1PFco/log.txt

Can you see if those are easily fixed?

nbeecher:
@iandees sorry about that.  It should be fixed now.

PR #1306, opened 2019-11-28
Add spider for Five Below

andrewbrown:


locations/spiders/five_below.py
```diff
@@ -0,0 +1,86 @@
...
+from locations.items import GeojsonPointItem
+
+class FiveBelowSpider(scrapy.Spider):
+    name = "five_below"
+    allowed_domains = [ "locations.fivebelow.com" ]
+    start_urls = (
+        "https://locations.fivebelow.com/sitemap.xml",
+    )
+
+    def start_requests(self):
```
iandees: The default behavior is to send requests for the `start_urls` list. You don't need to define `start_requests()` if you're not overriding the default.

andrewbrown: Thanks.  I submitted an update that removes `start_requests()`

iandees:
Thanks!

PR #1301, opened 2019-11-19
Add spider for Tommy Hilfiger

nbeecher:

PR #1320, opened 2019-12-11
Add spider for Tommy Hilfiger

nbeecher:
Fix for body_as_unicode


iandees:
We can keep using #1301

PR #1325, opened 2019-12-11
fixed name attribute for pizza_hut

Starfish699:
Previous name attribute was grabbing the information from the 'nearby locations' at the bottom of the branch page instead of from the top.


iandees:
#9

Issue #675, opened 2017-12-21
Hilton Hotels


iandees:
http://www3.hilton.com/en/about/locations/index.html

This might be a dupe of #17

iandees:
This issue hasn't been worked on in a while. I'm removing the assignment so others can work on it. If you're still working on it, please reassign yourself and reference a pull request or branch where you're doing your work.

PR #1332, opened 2019-12-13
Fix Hilton spider

thismakessand:
Some site changes were causing ~50% of the hotels to fail.
Also updated subsidiary brand names.
#675


thismakessand:
I think the exception is something ephemeral but I added some handling for it.

Issue #1048, opened 2019-03-12
Dickey's Barbecue Pit


maxerickson:
https://www.dickeys.com/location/search-by-state

davidhicks:
Completed per fbc9e21e5012a8b1b6fefac087b00f1695dfaf0f

PR #1368, opened 2019-12-30
Add spider for Dickey's Barbecue Pit

### LINKED ISSUE #1176: Little Caesars
https://order.littlecaesars.com/en-us/

andrewbrown:
Notes: 
- lat/long are not included, because these do not seem to appear in the network traffic at all
- A few URLs return HTTP 500.  This does not appear to be due to the spider; the same response was observed for these URLs from Chrome and Firefox, as well.


iandees:
#1048

Issue #576, opened 2017-12-19
Sears


evanmosseri:
http://www.sears.com/stores.html

iandees:
This issue hasn't been worked on in a while. I'm removing the assignment so others can work on it. If you're still working on it, please reassign yourself and reference a pull request or branch where you're doing your work.

Cj-Malone:
So a Sears spider was added, but is now broken. I'm closing this in favour of #3100

PR #1372, opened 2019-12-31
Adds spider for Sears

wessport:
Sears has store locations available across three different sites based on the store type. Scrapes 708 stores total, excluding the stores that have recently closed e.g. the sears outlet in Des Moines is listed in the sitemap, but appears to be closed (404). The outlet stores do not include latlons.


iandees:
#576

PR #1376, opened 2020-01-02
Add longitude key for Sears hometown stores

wessport:
Looks like the longitude key for the home town stores is slightly different than the latitude key. Should now include latlons for the home town stores as well.


iandees:
#576

PR #1381, opened 2020-01-06
Adding locations for Delaware North

thekuiperbelt:
This is a bare-bones spider... :skull_and_crossbones: 

The address was inconsistent & included international addresses, so instead of parsing the address, I left all the properties in addr_full.

Hope this is okay! :spider:


iandees:
That's ok for now. Maybe they'll add lat/lons later. Thanks!

PR #1388, opened 2020-01-08
Add spider for Harvey's Canada

nbeecher:


nbeecher:
The session ID is expiring which is causing this spider to fail here.  I will make the fix for this and re-commit.

nbeecher:
hmm still failing the checks even though I added code to grab a new session id for each run.  not sure what the issue is then.

nbeecher:
rerun check

PR #1393, opened 2020-01-09
Adds spider for Greyhound

wessport:
Scrapes 1820 Greyhound locations, including a handful of locations in Mexico - the Canadian locations appear to be hosted on a separate site. 

I included an extra attribute 'location_type' to hopefully help indicate which locations are bus stops only so they can be filtered out depending on the use case. Also encountered a handful of 500 response for locations like Richmond, IN. These appear to be broken pages when viewed through the browser.

### DIFF
--- a/locations/spiders/greyhound.py
+++ b/locations/spiders/greyhound.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+import json
+import re
+from urllib.parse import urlencode
+
+import scrapy
+
+from locations.items import GeojsonPointItem
+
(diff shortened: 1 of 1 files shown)


iandees:
I'm ok with the "location_type" thing, but we might want to think about replacing it with a "category" in the future. This would be a more widely useful property for other spiders, too. I just have to think about which category system to use.

wessport:
I agree, there's been a few others in the past that would benefit from a broader "category" property. Sounds like a plan!

PR #1409, opened 2020-01-24
Add spider for sparkasse.de

kalpeshtawde:
Spider for sparkasse.de

### DIFF
--- a/locations/spiders/sparkasse.py
+++ b/locations/spiders/sparkasse.py
@@ -1,0 +1,84 @@
+import scrapy
+import json
+import re
+
+from locations.items import GeojsonPointItem
+from locations.hours import OpeningHours
+
+
+class SparkasseSpider(scrapy.Spider):
+    name = "sparkasse"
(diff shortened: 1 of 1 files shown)


locations/spiders/sparkasse.py
```diff
@@ -0,0 +1,84 @@
...
+
+from locations.items import GeojsonPointItem
+from locations.hours import OpeningHours
+
+
+class SparkasseSpider(scrapy.Spider):
+    name = "sparkasse"
+    allowed_domains = ["www.sparkasse.de"]
+    start_urls = []
+    for c in "abcdefghijklmnopqrstuvwxyz":
```
iandees: Instead of doing this work at the class level, please implement the `start_requests()` function as seen in here: https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/joann_fabrics.py#L21-L29

kalpeshtawde: Makes sense. Done the needful.

iandees:
Thanks!

PR #1412, opened 2020-01-27
Add spider for Scotiabank

thismakessand:
Scrapes Scotiabank branch locations in CA.

### DIFF
--- a/locations/spiders/scotiabank.py
+++ b/locations/spiders/scotiabank.py
@@ -1,0 +1,99 @@
+# -*- coding: utf-8 -*-
+import csv
+import json
+import math
+import re
+
+import scrapy
+
+from locations.items import GeojsonPointItem
+from locations.hours import OpeningHours
+
+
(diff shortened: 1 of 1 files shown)


locations/spiders/scotiabank.py
```diff
@@ -0,0 +1,99 @@
...
+    def parse(self, response):
+        data = json.loads(response.body_as_unicode())
+        branches = (data["branchInfo"] or {}).get("marker", [])
+        if branches:
+            if data["branchCount"] == 20:
+                refs = {x["@attributes"]["id"] for x in branches}
+                if refs.issubset(self.refs):
+                    return
+                max_distance = branches[-1]["@attributes"]["distance"]
+                for b in [45.0, 135.0, 225.0, 315.0]:
```
iandees: Can you add a comment here about what this is doing? It looks like you're making requests in smaller increments than the lat/lon points file, but let's make it obvious.

PR #1414, opened 2020-01-28
Feature/add spider for vr bank

kalpeshtawde:
Added spider for vr.de bank


locations/spiders/vr_bank.py
```diff
@@ -0,0 +1,188 @@
...
+        for item in list:
+            yield scrapy.Request(
+                url=item,
+                callback=self.parse_details,
+                meta={'url': item}
+            )
+
+    def parse(self, response):
+        index = [
+            "https://www.vr.de/service/filialen-a-z/a.html",
```
iandees: These requests should be created in the `start_requests()` function, not the `parse()` function.

locations/spiders/vr_bank.py
```diff
@@ -0,0 +1,188 @@
...
+        try:
+            website = response.xpath('//li[@itemprop="url"]/a/span/text()').get()
+        except Exception:
+            pass
+
+        try:
+            m = re.search(r'lat&quot;:([-+]?[0-9]*\.?[0-9]*)', response.text)
+            if m:
+                latitude = m.group(1)
+        except Exception:
```
iandees: Please don't wrap all these in exception catches. If something fails, we want to fix it.

kalpeshtawde: Actually these are expected to fail. Like some banks do not provide working hours. In that case the regex will fail.

iandees: Please add a comment and use a more restrictive exception in the `except` block so we don't swallow all exceptions.

kalpeshtawde: Please check now.

locations/spiders/vr_bank.py
```diff
@@ -0,0 +1,138 @@
...
+            print("Error: longitude not found for url: {}".format(response.meta.get('url')))
+
+        try:
+            hours = response.xpath('//p[@itemprop="openingHoursSpecification"]/text()').getall()
+        except Exception:
+            print("Error: working hours information not found for url: {}".format(response.meta.get('url')))
+
+        properties = {
+            'ref': response.meta.get('url'),
```
iandees: There's no need to set a `meta` for `url`, you can use `response.request.url`.
```suggestion
            'ref': response.request.url,
```

locations/spiders/vr_bank.py
iandees: What I meant by "use a more restrictive exception" is to figure out which exception is thrown that you want to catch and catch only that one instead of catching `Exception`.

kalpeshtawde: Done. I have removed unwanted exception handling and elsewhere made it more restrictive to ValueError.

locations/spiders/vr_bank.py
kalpeshtawde: Done.

iandees:
Thanks!

PR #1426, opened 2020-02-22
Update spider for Bank of America

thomas536:
Looks like some of the CSS has changed such that this was returning zero
results. Picked some new arbitrary CSS to key off of. Also uses ld+json in
the leaf to get opening hours. [Nit] Moves parse_center to a 4 space indent.


iandees:
Thanks!

PR #1448, opened 2020-04-08
Update sonic spider to use ld+json

thomas536:
This was returning zero results. Adds wikidata.


iandees:
@thomas536 I added you as a contributor to the primary repo. Can you switch your clone's remote to this repo instead of your fork? I think that might help with the test failures in the future.

thomas536:
Thanks. Will do.

Issue #1449, opened 2020-04-08
WSP


LadyLatitude:
https://www.wsp.com/en-GB/who-we-are/our-offices

PR #1451, opened 2020-04-08
Wsp

LadyLatitude:
Closes issue #1449 

Ian, it appears I can no longer push new branches to the main repo. Could I be addd?


iandees:
You're added!

PR #1457, opened 2020-04-10
Add spider for First Watch

nbeecher:


iandees:
The API that this is using seems to return 404 and 403's now.

PR #1461, opened 2020-04-14
Adds searchable points for Europe

wessport:
Adds searchable points for Europe to make point searches possible for European store locators where appropriate. 

They're setup the same way as the previous point sets, with a minor difference of km as the distance unit. This set includes a 2 digit country code (from the [NUTS boundaries file](https://ec.europa.eu/eurostat/web/gisco/geodata/reference-data/administrative-units-statistical-units/nuts#nuts16) from the EU's eurostat office) to help filter points for specific countries.


iandees:
Thanks!

PR #1476, opened 2020-05-12
Add Phillips 66, Conoco, and 76 Gas Stations

reidab:


reidab:
This one ran in under a minute locally, not sure what's happening in the CI run.

iandees:
I think GitHub Actions is blocking the service I use to temporarily store the output of the spider.

PR #1481, opened 2020-05-13
Fix Pilot Flying J parser: rewrite using CSV export, add fuel tags

reidab:
The existing parser was failing on the individual store pages. This rewrites it using a CSV export available on their site. It takes quite a while to resolve the CSV download, but once it does we get all the records at once.

This also adds differentiation between the Pilot and Flying J brands for individual stations.

Alternatively, I found that appending `.json` to the end of any store page returns a nice JSON payload: https://locations.pilotflyingj.com/us/pa/mill-hall/5868-nittany-valley-drive.json

### DIFF
--- a/locations/spiders/pilot_flyingj.py
+++ b/locations/spiders/pilot_flyingj.py
@@ -1,4 +1,4 @@
-import json
+import csv
 
 import scrapy
 from locations.items import GeojsonPointItem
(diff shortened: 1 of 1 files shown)


locations/spiders/pilot_flyingj.py
```diff
@@ -1,81 +1,56 @@
...
+                name=store['Name'],
+                addr_full=store['Address'],
+                city=store['City'],
+                state=store['State/Province'],
+                postcode=store['Zip'],
+                country=store['Country'],
+                phone=store['Phone'],
+                ref=store['Store#'],
+                extras={
+                    'amenity:fuel': True,
```
iandees, line 33: Should this be `'amenity': 'fuel'` to match OSM-ey tags?

reidab: I was wondering about that myself. I went with this format because we only have one entry per larger place here, instead of individual nodes per feature in OSM. So for something like a Costco, we might want both `amenity=fuel` and `amenity=pharmacy` on the same listing.

I think it would be just as reasonable to go with `amenity=fuel;pharmacy` 🤷‍♂️

iandees: This is OK for now. I'm glad you're adding this, but I think I'd like to use full-fledged categories instead of OSM tags to specify categorization. We can adjust that later after we decide which set of categories we want to use.

reidab: 👍 I'd be happy to switch stuff over to a more robust categorization system once we figure that part out.

PR #1516, opened 2020-05-19
Update Kroger spider to include subbrands and more info

reidab:
Since all of Kroger's sub-brands run the same website software, the Kroger scraper can apply to all of them. 

That said, I'm leaving this as a draft PR because running this last week locally ran into rate limiting restrictions and my IP still can't access any of the Kroger store sites.

A better approach might be to use Kroger's public locations API: https://developer.kroger.com

The locations endpoint requires authentication with client credentials, which we probably wouldn't just want to publish here. Is there a mechanism in place for exposing secrets to the spider runs?

### DIFF
--- a/locations/spiders/kroger.py
+++ b/locations/spiders/kroger.py
@@ -7,24 +7,125 @@ from locations.items import GeojsonPointItem
 from scrapy.selector import Selector
 
 
-default_headers = {
+DEFAULT_HEADERS = {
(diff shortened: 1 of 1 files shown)


iandees:
> Is there a mechanism in place for exposing secrets to the spider runs?

We don't have that, no. I can sign up for a key with a All The Places-specific email address if that'd help.

PR #1531, opened 2020-06-23
Fix Circle-K lat/lon + website format

reidab:
#1521 removed all non-digit characters, but sadly `-` and `.` are considered non-digit characters and we kinda need those


reidab:
Also added a necessary prefix since the `url` param is just a path

PR #1538, opened 2020-07-08
Fixed broken spider for McDonald's NL

thekuiperbelt:


iandees:
The best McDonald's I've ever been to is the one in Schiphol.

PR #1541, opened 2020-07-10
Adding new spider for Superdrug

thekuiperbelt:
Please note that **this spider requires a user agent** setting in oreder to successfully run. **I did not include the user agent in this commit.**


thekuiperbelt:
+ Bump

PR #1542, opened 2020-07-13
Adding spider for JD Wetherspoon

thekuiperbelt:


thekuiperbelt:
@iandees + Bump. This one should be good to go.

PR #1549, opened 2020-08-19
Add Pennsylvania Wine & Spirits

jleedev:


iandees:
Thanks for putting this spider together! It looks like at least one result gets put at null island (0.0, 0.0 lat/lon). Can you add a filter to your code so that if there's a zero lat/lon it'll just leave the lat/lon off the result?

Issue #1566, opened 2020-09-28
Langham Hotels


thekuiperbelt:
https://www.langhamhotels.com/en/locations

iriyagupta:
@thekuiperbelt What exactly are we looking to resolve here?

thekuiperbelt:
@iriyagupta  Hello, thanks for reaching out. I've labeled issues in the repository that are good options for creating your first spider. The ultimate goal is to scrape the URL linked above (or a better one if you find it) for [location information](https://github.com/alltheplaces/alltheplaces/blob/master/DATA_FORMAT.md) regarding the business. The project is built using [scrapy](https://docs.scrapy.org/en/latest/url), so there is more documentation there the can also assist you as well. Feel free to comb through the spiders in the locations folder as well. If you find an issue that you think looks interesting, go ahead and assign it to yourself and feel free to reach out if you have more questions.

moohco:
I'd like to take this

thekuiperbelt:
@moohco Thank you, please go ahead and assign it to yourself! 👍

moohco:
Unless I am missing something I can't assign it to myself?

thekuiperbelt:
@moohco Ohhh thanks for notifying me, I have assigned this to you. Thank you for your collaboration :)

moohco:
@thekuiperbelt thanks, should be fun :) you can assign the other too.

PR #1574, opened 2020-10-01
Added Langham Hotels spider

moohco:
For #1566

Hey guys, 

here is my first PR for your project, I was able to extract most of the locations nicely.
Some of the hotels provide invalid JSON or no JSON at all, so these are not included for now.
The only way to get the locations for those would be to crawl the footers, whereas there is no common pattern to do it the same way for all of them. But still doable. 

I wanted to get feedback from you how you proceed in such a case. 

Thanks
Mo


moohco:
Ignore the ralph_lauren spider, should not be in this PR.. haven't started with that yet.

iandees:
Thanks! Looks like there's several missing but they're harder to get to. This is a good start :+1:

moohco:
Thanks, yep, there is really no common pattern I saw to get them extracted in a generic way.

PR #1580, opened 2020-10-02
added iron mountain spider

RitikMody:


RitikMody:
Can you please review ?

locations/spiders/iron_mountain.py
```diff
@@ -0,0 +1,51 @@
...
+            yield scrapy.Request(url=url, callback=self.parse_location,meta={'country': response.meta.get("country")})
+    
+    def parse_location(self,response):
+        data = re.sub("'","",response.xpath('//*[@id="gmap"]/script/text()').extract()[0])
+        data = data.replace("false","False")
+        data = data.replace("true","True")
+        data = data.replace("null","None")
+        data = data.replace("RLS.defaultData = ","")
+        data = data.replace(";","")
+        address = eval("{"+str.join(",",eval(data)["markerData"][0]["info"].replace("<\\/div>","").split(",")[1:]))
```
iandees, line 39: I'd prefer to not use `eval()` as that's potentially dangerous. Can you parse it into JSON?

locations/spiders/iron_mountain.py
```diff
@@ -0,0 +1,51 @@
...
+        yield GeojsonPointItem(**{
```
iandees, line 40: No need to create a dict and then `**` it. Just pass the args straight in like so:

```python
yield GeojsonPointItem(
  addr_full="..."
```

RitikMody:
It was working properly in my command line ,  I don't know why the test-spider is showing that it failed. Can you please tell me what am I doing wrong?

locations/spiders/iron_mountain.py
RitikMody: Okay I will change that.

locations/spiders/iron_mountain.py
RitikMody: I tried with json.loads( ) first, but that was giving many ValueErrors @iandees . Is there any other method?

iandees: Nope. If it doesn't parse as JSON then you'll have to do more cleanup (like you're doing with `data.replace()` above) or find another way to extract the useful information.

RitikMody:
@iandees I have made the changes , please review them.

RitikMody:
@iandees I have made the changes, kindly review them

locations/spiders/iron_mountain.py
```diff
@@ -0,0 +1,51 @@
...
+        #     state=response.xpath('//*[@id="main-container"]/div[2]/div[1]/ol/li[2]/a/text()').extract()[0].strip(" "),
+        #     postcode=address["post_code"],
+        #     lat=float(address["lat"]),
+        #     lon=float(address["lng"]),
+        #     phone=address["local_phone"],
+        #     website=address["url"].replace("\\",""),
+        #     country=response.meta.get("country"),
+        #     ref=address["url"].replace("\\","")
+        # )
+        yield {"data":data}
```
iandees, line 51: You don't want to yield a `dict` here, you need to yield a `scrapy.GeojsonPointItem`

RitikMody: I'm sorry I had commented it out for debugging.

RitikMody:
I've rectified it @iandees

PR #1581, opened 2020-10-02
Adds spider for Vancouver Coastal Health

### LINKED ISSUE #1427: Vancouver Coastal Health
http://www.vch.ca/locations-services

wessport:
Scrapes 978 locations. Closes #1427


iandees (commented):
It looks like they have a couple bad geocodes on the map. Can you add a check that sets the lat/lon to None if lat is < 45 or something like that?

wessport:
Thanks @iandees !

PR #1592, opened 2020-10-04
added Deichmann locations

### LINKED ISSUE #1590: Deichmann
https://stores.deichmann.com/index.html

Planning to put more Germany/Europe on the map :)

moohco:
closes #1590

added Deichmann (shoe store) location spider.

### DIFF
--- a/locations/spiders/deichmann.py
+++ b/locations/spiders/deichmann.py
@@ -1,0 +1,59 @@
+# -*- coding: utf-8 -*-
+import scrapy
+import json
+import base64
+from locations.items import GeojsonPointItem
+
+class Deichmann(scrapy.Spider):
+    name = "deichmann"
+    allowed_domains = ["stores.deichmann.com","stores.dosenbach.ch"]
+    start_urls = (
+        'https://stores.deichmann.com/index.html',
+    )
+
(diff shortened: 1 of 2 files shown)


moohco:
runs locally, tested again
here it seems that it ran into a JSON format error: 
`json.decoder.JSONDecodeError: Invalid control character at: line 20 column 39 (char 1032)`

so I added JSON format validation

locations/spiders/deichmann.py
```diff
@@ -0,0 +1,67 @@
...
+        stores = response.xpath('//a[@class="button-link"]/@href').extract()
+
+        for store in stores:
+            yield scrapy.Request(response.urljoin(store), callback=self.parse_locations)
+
+    def parse_locations(self, response):
+        # get json which provides all data
+        data = response.xpath('//script[@type="application/ld+json"]/text()').extract_first()
+
+        if data and self.isValidJson(data):
```
iandees, line 39: Instead of parsing and throwing away the result to check for valid JSON, just parse it here, catch the Exception and `return` instead of letting it keep going.

moohco:
Right, think I got it. I will update that and the langham spider (where I did the same check too) in a minute

PR #1594, opened 2020-10-04
add Screwfix spider

### LINKED ISSUE #1557: Screwfix
https://www.screwfix.com/jsp/tradeCounter/tradeCounterAllStoresPage.jsp

rubyruins:
Fixes #1557. 

@iandees, could you please review? I wasn't able to reliably extract the city/state/postcode since the format seems different in the few different locations I checked. I got the full address instead, is it okay?

### DIFF
--- a/locations/spiders/screwfix.py
+++ b/locations/spiders/screwfix.py
@@ -3,32 +3,29 @@ import scrapy
 from locations.items import GeojsonPointItem
 
 class ScrewfixSpider(scrapy.Spider):
-	name = "screwfix"
-	allowed_domains = ["www.screwfix.com"]
-	download_delay = 0.1
-	start_urls = (
-		"https://www.screwfix.com/jsp/tradeCounter/tradeCounterAllStoresPage.jsp",
-	)
+    name = "screwfix"
+    allowed_domains = ["www.screwfix.com"]
+    download_delay = 0.1
+    start_urls = (
+        "https://www.screwfix.com/jsp/tradeCounter/tradeCounterAllStoresPage.jsp",
+    )
 
-	def parse(self, response):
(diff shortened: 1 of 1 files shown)


rubyruins:
Ah, the return of the dreaded test-spider fail :D any idea why this happens?

locations/spiders/screwfix.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+import scrapy
+from locations.items import GeojsonPointItem
+
+class ScrewfixSpider(scrapy.Spider):
+	name = "screwfix"
```
iandees, line 6: Please use four spaces instead of a tab.

locations/spiders/screwfix.py
```diff
@@ -0,0 +1,34 @@
...
+	download_delay = 0.1
+	start_urls = (
+		"https://www.screwfix.com/jsp/tradeCounter/tradeCounterAllStoresPage.jsp",
+	)
+
+	def parse(self, response):
+		storeurls = response.xpath(".//ul[@class='store-name']//a/@href").extract()
+		storeurls = [response.urljoin(i) for i in storeurls]
+		for i in range(len(storeurls)):
+			yield scrapy.Request(storeurls[i], callback=self.parse_location)
```
iandees, line 17: Instead of iterating over the indexes of the list, just index over the list of joined URLs:

```suggestion
		for storeurl in response.xpath(".//ul[@class='store-name']//a/@href").extract():
			yield scrapy.Request(response.urljoin(storeurl), callback=self.parse_location)
```

locations/spiders/screwfix.py
rubyruins: Sorry, I forgot. I'll change to spaces.

rubyruins:
I've updated it now!

Issue #1556, opened 2020-09-22
Aurecon


thekuiperbelt:
https://www.aurecongroup.com/locations

chethanagopinath:
@thekuiperbelt Hi, I was looking for beginner-friendly issues for Hacktoberfest and I was wondering if you could provide more detail to this?

thekuiperbelt:
@chethanagopinath Hello, thanks for reaching out. I've labeled issues in the repository that are good options for creating your first spider. The ultimate goal is to scrape the URL linked above (or a better one if you find it) for [location information](https://github.com/alltheplaces/alltheplaces/blob/master/DATA_FORMAT.md) regarding the business.  The project is built using [scrapy](https://docs.scrapy.org/en/latest/topics/spiders.html), so there is more documentation there the can also assist you as well. Feel free to comn through the spiders in the [locations folder ](https://github.com/alltheplaces/alltheplaces/tree/master/locations/spiders) as well. If you find one that you think looks interesting, go ahead and assign it to youself and feel free to reach out if you have more questions.

rubyruins:
@thekuiperbelt, I would like to work on this. Will you guys be adding the `Hacktoberfest` label to the repository? Without it, contributions will not be counted, unfortunately.

rubyruins:
@iandees, is this issue available? I'd like to take it up!

iandees:
Yep, go ahead!

iandees:
Pick a string that's unique to each location that will likely stay consistent. Maybe the phone number?

rubyruins:
Makes sense. Will do!

thekuiperbelt:
Closes #1556

PR #1596, opened 2020-10-05
add Aurecon Group spider

rubyruins:
Fixes issue #1556 
@iandees could you please review this?

### DIFF
--- a/locations/spiders/aurecongroup.py
+++ b/locations/spiders/aurecongroup.py
@@ -14,18 +14,16 @@ class AureconGroupSpider(scrapy.Spider):
         for location in response.xpath('.//h4'):
             addr = location.xpath('.//following-sibling::div')[0].xpath('.//div/span/following-sibling::div')[0]
             addr = ' '.join([addr.xpath('.//span/text()').extract()[i].replace('\t', '').replace('\n', '').replace('\r', '') for i in range(2)])
-            try: 
-                maps = str(location.xpath('.//following-sibling::div//a[@target="_blank"]/@href').extract_first())
-                maps = (maps.split('=')[1]).split(',')
-                properties = {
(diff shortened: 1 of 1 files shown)


locations/spiders/aurecongroup.py
```diff
@@ -0,0 +1,31 @@
...
+                    'brand': 'Aurecon Group',
+                    'city': location.xpath('.//strong/text()').extract_first().replace('\t', '').replace('\n', '').replace('\r', ''),
+                    'addr_full': addr,
+                    'phone': location.xpath('.//following-sibling::div//span[@itemprop="telephone"]/text()').extract_first().strip(),
+                    'lat': float(maps[0]),
+                    'lon': float(maps[1])
+                }
+                yield GeojsonPointItem(**properties)
+            except: 
+                pass
```
iandees, line 31: Can you remove the try/except/pass here? I'd rather see the exceptions than swallow them and potentially miss errors.

rubyruins: Actually, I put the try/except in there since a couple of the locations seem to not have any coordinates set for the latitude and longitude. So when I try splitting the string to extract the coordinates, I get the list index out of bounds error. Should I still remove the try/except?

iandees: You can still include the locations without lat/lon data.

rubyruins: Oh, I didn't know that! Cool, I'll make the relevant changes. So for those locations with lat/lon, I'll extract the coordinates. And for locations without lat/lon, I should still get the rest of the details, right?

iandees: Yep!

rubyruins: I've removed the try/except now! Could you please take a look?

PR #1634, opened 2020-10-25
Add Taco John's Scraper

KatherineMichel:
Hi, I opted to find a site on my own to scrape. I am counting this toward Hacktoberfest. Thank you!

PR #1639, opened 2020-10-29
Improve regex

KatherineMichel:
My code worked, but I found a couple of bugs in the regex. The `.` should have been escaped, but wasn't and I don't think the parenthesis negation was needed, so I changed the pattern.


iandees:
Related to #1634

iandees:
Thanks for spotting this.

KatherineMichel:
@iandees Thank you! I realized it while I was writing a blog post about the scraper! https://github.com/KatherineMichel/portfolio/blob/master/regular-blog-posts/building-a-taco-johns-scraper.md

PR #1643, opened 2020-10-31
Big W Australia

markman123:
It has some 500 errors for out of date stores from the sitemap... how to ignore these elegantly?


iandees:
The errors should be ok. If you look at the map in the CI build you can see that all the lat/lons are at 0,0. Can you double-check that those are getting parsed right?

markman123:
i've accidently put all of the other commits into this branch... not sure if that muffs the whole lot up? Added dc25c5d

iandees:
Yea, do you think you could re-do this branch and force-push it to clean up?

iandees:
Cleaned up and put Big W spider in #3444.

PR #1648, opened 2020-11-06
Chipotle

nbeecher:
Fixed spider.  The spider wasn't scraping any locations and now it is scraping all the locations worldwide (2,697)


iandees:
Woot, thanks!

PR #1657, opened 2020-11-20
Panera Bread - Fix parsing of cities with multiple locations

ctrlaltf2:
I noticed some Panera Bread locations in my area were missing from the alltheplaces data. I investigated it a bit and it turned out that for cities that had more than one location, it wasn't crawling the right pages (404 error) so it just wasn't getting added to the output geojson. I wrote a small fix and tested it, I was able to get it to produce the locations in Tucson, AZ; Fort Wayne, IN; Evansville, IN, which all have more than one location. This should add around 888 new locations if the total number of locations [here](https://locations.panerabread.com/index.html) is right.


iandees:
Related to #107

PR #1666, opened 2020-12-07
Repairs Nike spider

wessport:
Nike spider was previously returning zero results, now scrapes 1495 Nike locations internationally.


iandees:
Thanks!

PR #1667, opened 2020-12-07
Added spider for Mercy

thekuiperbelt:


iandees:
Is this related to an existing ticket?

thekuiperbelt:
@iandees No, but can be confused with https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/mercyhealth.py. But double checked, they are two separate entities.

iandees:
Sounds good!

PR #1700, opened 2020-12-31
Remove deprecated duplicate spider, fix Mercy Health spider class name

wessport:
Noticed the `kindred_health.py` spider is a duplicate of `kindred_healthcare.py` (I should have noticed sooner!). Removed the old one in favor of the most recent (`kindred_healthcare.py`). 

Also happened to notice a typo in the `mercyhealth.py` - renamed the class to be more appropriate :smile:


iandees:
Looks like the CI bot fails when things are deleted. That's ok, though!

PR #1703, opened 2021-02-18
Fix regal theaters start url

davidchiles:


davidchiles:
🤔 Curious one location is showing up in Antarctica.

iandees:
Huh, weird. That's probably ok for now. The data is allowed to be a little messy.

PR #1704, opened 2021-02-18
Added roundtable pizza

davidchiles:
13 of the locations are missing lat/lon coordinates. What is the correct way to handle these locations?


iandees:
It's ok to skip lat/lon if they aren't there. In the future, we might geocode these if they have address information.

PR #1711, opened 2021-03-09
Adds spider for KBP Foods

wessport:
Scrapes all 760 KBP Foods locations using the 100mi searchable points.


iandees:
Hmm, the difference between brand and store_brands is interesting. I'm ok with how it's in there now, but maybe we need to think about an "operator" or "owner" vs "brand". 🤔

PR #1715, opened 2021-03-11
Fix Whataburger spider

wessport:
Updates the xpath for the store hours, spider once again returns all 848 locations! 🍔


iandees:
Woohoo!

PR #1723, opened 2021-03-31
Adding spider for Lands' End

thekuiperbelt:


iandees:
It looks like this is missing out on a bunch of stores for some reason. Can you try using 100 radius instead of 50 in the request?

thekuiperbelt:
Ahh, looking at it now 👍

thekuiperbelt:
Fixed 🌎

PR #1731, opened 2021-04-07
Fix subway

jleedev:
- Switch from a search to using their store directory
- Remove the crawl delay as it was unnecessary

Now produces 31343 items


iandees:
#1129

iandees:
https://github.com/alltheplaces/alltheplaces/issues/20

PR #1734, opened 2021-04-11
Add toyota

jleedev:


iandees:
Thanks for adding this. Do you think there's any way to get Toyota locations for the whole world? If not, maybe we can rename this spider toyota_us?

jleedev:
toyota_us would make sense; the others seem to live on different domains and are probably structured differently.

PR #1744, opened 2021-04-15
Add coen markets

jleedev:
55 locations in PA, WV, and OH; some former Cogo's.


iandees:
Check the logs from the spider here: https://placescraper-results.s3.amazonaws.com/ci/750193140/coen_markets/log.txt

Looks like one of the fields isn't always there so using `foo['bar']` raises an error.

jleedev:
Ah thanks, I was wondering where the output was going.

PR #1753, opened 2021-04-27
Update spider for Tractor Supply Co

LadyLatitude:
Fixes issue #153


LadyLatitude:
This spider takes hours and hours to run, FYI.

PR #1763, opened 2021-05-07
Added scraper for apotheken.de

kalpeshtawde:


iandees:
This fails on some pages when parsing JSON: https://placescraper-results.s3.amazonaws.com/ci/818881440/apotheken_de/log.txt

Can you skip over those bad entries, please?

Is Apotheken.de a website for a specific Apotheke or does it aggregate data? I'd prefer to only include listings scraped from primary data sources.

PR #1769, opened 2021-05-11
Added scraper for deutschepost.de

kalpeshtawde:


locations/spiders/deutschepost_de.py
```diff
@@ -0,0 +1,72 @@
...
+               5: 'Fr',
+               6: 'Sa',
+               7: 'Su'}
+
+class DeutschepostDeSpider(scrapy.Spider):
+    name = "deutschepost_de"
+    allowed_domains = ["www.deutschepost.de"]
+
+    start_urls = []
+    with open('./locations/searchable_points/germany_centroids_80km_radius_country.csv') as csv_file:
```
iandees, line 23: Can you use `start_requests()` instead of filling the `start_urls` list? For example:

https://github.com/alltheplaces/alltheplaces/blob/073f1338a5b495b2057b0ec19432502975f19758/locations/spiders/anytime_fitness.py#L13-L15

kalpeshtawde: Done.

Issue #523, opened 2017-12-18
Enterprise Rent-a-car


iandees:
Directory of all locations available. For example:

https://www.enterprise.com/en/car-rental/locations.html
https://www.enterprise.com/en/car-rental/locations/us.html
https://www.enterprise.com/en/car-rental/locations/us/pa/philadelphia.html
https://www.enterprise.com/en/car-rental/locations/us/pa/darby-17a5.html

PR #1773, opened 2021-05-11
fix spider for Enterprise

LadyLatitude:
Fixes issue #523

### DIFF
--- a/locations/spiders/enterprise.py
+++ b/locations/spiders/enterprise.py
@@ -29,7 +29,7 @@ class EnterpriseSpider(scrapy.Spider):
                     'name': d.get('name'),
                     'phone': d.get('formatted_phone'),
                     'website': d.get('url'),
-                    'addr_full': d.get('address_lines'),
+                    'addr_full': ' '.join(d.get('address_lines') or []),
                     'city': d.get('city'),
                     'state': d.get('state'),
                     'postcode': d.get('postal_code'),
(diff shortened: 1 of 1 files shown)


locations/spiders/enterprise.py
```diff
@@ -6,7 +6,6 @@
 
 class EnterpriseSpider(scrapy.Spider):
     name = "enterprise"
-    item_attributes = { 'brand': "Enterprise Rent-A-Car", 'brand_wikidata': "Q17085454" }
```
iandees, line 9: Did you mean to remove this?

LadyLatitude: Yes, because the response includes the brand, so I added it to the properties.

iandees: Ah, ok. I don't see `brand` in the list of properties below. Can you please add it there?

LadyLatitude: Woops! I'll it in here shortly.

LadyLatitude: Turns out I was mistaken, I must have gotten the responses from sister companies Alamo and National confused. I added back in the above line.

locations/spiders/enterprise.py
```diff
@@ -26,16 +26,16 @@ def parse_country(self, response):
         if data:
             for d in data:
                 properties = {
-                    'name': d.get('shortTitle'),
-                    'phone': d.get('formattedPhone'),
+                    'name': d.get('name'),
+                    'phone': d.get('formatted_phone'),
                     'website': d.get('url'),
-                    'addr_full': ' '.join(d.get('addressLines')),
+                    'addr_full': d.get('address_lines'),
```
iandees, line 32: I'm gong to flip this back to what it was before, because it puts an array of data in here rather than the expected string.

Issue #1774, opened 2021-05-11
Alamo


LadyLatitude:
https://www.alamo.com/en/car-rental-locations.html

davidhicks:
Implemented and working per https://data.alltheplaces.xyz/runs/2023-05-31-05-12-59/stats/alamo.json

PR #1779, opened 2021-05-12
fix phone attribute and scrape int'l locations

LadyLatitude:
fixes issue #1774


iandees:
See the error here: https://placescraper-results.s3.amazonaws.com/ci/836805739/alamo/log.txt

Looks like not all entries have `address` key.

LadyLatitude:
Alrighty then, I added some handling.

PR #1783, opened 2021-05-14
Added scraper for alnatura

kalpeshtawde:


locations/spiders/alnatura_de.py
```diff
@@ -0,0 +1,90 @@
...
+            'lon': response.meta.get('lng'),
+            'name': store['StoreName'],
+            'street': store['Street'],
+            'city': store['City'],
+            'postcode': store['PostalCode'],
+            'phone': store['Tel'],
+            'country': store['Country'],
+            'ref': response.meta.get('id'),
+            'extras': {
+                'time': store['OpeningTime']
```
iandees, line 66: Let's leave this in `opening_hours` only and not duplicate it here.

locations/spiders/alnatura_de.py
```diff
@@ -0,0 +1,90 @@
...
+            }
+        }
+
+        if store['OpeningTime']:
```
iandees, line 70: You might want to use `store.get('OpeningTime')` so that you get a `None` instead of a `KeyError` here if a store doesn't have this key in it.

```suggestion
        if store.get('OpeningTime'):
```

locations/spiders/alnatura_de.py
kalpeshtawde: Done.

locations/spiders/alnatura_de.py
kalpeshtawde: Done.

PR #1874, opened 2021-05-28
Fix Spider for Kroger

nbeecher:


nbeecher:
Runs locally for me.  But it does run very slow due to 25 mile lat/long requests.

PR #2094, opened 2021-06-29
Add additional countries for Nandos

wessport:
Adds a set of spiders for more of Nandos international locations (US, CA, AU, etc.) Most individual country sites had their own structure and specific quirks so I organized them into their own files. I didn’t include a spider for South Africa or Botswana. These two were not easily scrapable without searchable points - maybe a good excuse to work on some points for Africa in the future 😄


iandees:
Thanks! These look great.

Issue #2690, opened 2021-08-27
Spider providence_health_services is broken


scraperbot:
During the global build at 2021-08-25-14-42-15, spider **providence_health_services** failed with **451 features** and **4 errors**.

Here's [the log](https://data.alltheplaces.xyz/runs/2021-08-25-14-42-15/logs/providence_health_services.txt) and [the output](https://data.alltheplaces.xyz/runs/2021-08-25-14-42-15/output/providence_health_services.geojson) ([on a map](https://data.alltheplaces.xyz/map.html?show=https://data.alltheplaces.xyz/runs/2021-08-25-14-42-15/output/providence_health_services.geojson))

scraperbot:
During the global build at 2021-09-15-14-42-44, spider **providence_health_services** failed with **451 features** and **4 errors**.

Here's [the log](https://data.alltheplaces.xyz/runs/2021-09-15-14-42-44/logs/providence_health_services.txt) and [the output](https://data.alltheplaces.xyz/runs/2021-09-15-14-42-44/output/providence_health_services.geojson) ([on a map](https://data.alltheplaces.xyz/map.html?show=https://data.alltheplaces.xyz/runs/2021-09-15-14-42-44/output/providence_health_services.geojson))

Issue #2439, opened 2021-08-06
Spider providence_health_services is broken


scraperbot:
During the global build at 2021-08-04-14-42-45, spider **providence_health_services** failed with **451 features** and **4 errors**.

Here's [the log](https://data.alltheplaces.xyz/runs/2021-08-04-14-42-45/logs/providence_health_services.txt) and [the output](https://data.alltheplaces.xyz/runs/2021-08-04-14-42-45/output/providence_health_services.geojson) ([on a map](https://data.alltheplaces.xyz/map.html?show=https://data.alltheplaces.xyz/runs/2021-08-04-14-42-45/output/providence_health_services.geojson))

Issue #1837, opened 2021-05-28
Spider providence_health_services is broken


scraperbot:
During the global build at 2021-05-26-14-42-23, spider **providence_health_services** failed with **494 features** and **3 errors**.

Here's [the log](https://data.alltheplaces.xyz/runs/2021-05-26-14-42-23/logs/providence_health_services.log) and [the output](https://data.alltheplaces.xyz/runs/2021-05-26-14-42-23/output/providence_health_services.geojson) ([on a map](https://data.alltheplaces.xyz/map.html?show=https://data.alltheplaces.xyz/runs/2021-05-26-14-42-23/output/providence_health_services.geojson))

PR #2863, opened 2021-09-11
Fix providence_health_services

jleedev:
Rewrite spider using sitemap.xml and ldjson

Seems to be serving 500's quite often, including for the sitemap. Not sure how to handle those.


jleedev:
#2690 
#2439 
#1837

PR #2867, opened 2021-09-13
Fixing bug in Cintas

boomerwv1:
Fixing bug found in Cintas Spider.


boomerwv1:
THis file fixes a bug in the cintas.py file that I pushed earlier.

iandees:
Can you resolve the conflict here? Looks like maybe you modified your existing commit instead of making a new commit for the fix?

PR #2975, opened 2021-09-23
Spider for State of South Carolina

boomerwv1:
Adding spider for the State of South Carolina. State uses an API for locations, which requires use of postal code. I added a .csv with only SC zips to avoid using nationwide file. Let me know what you think about doing this.

API example https://applications.sc.gov/PortalMapApi/api/Map/GetMapItemsByZipcodeCategoryId/29414/1,2,3,4,5,6,7

Thanks,


jleedev:
Seems this URL will fetch the whole file: https://applications.sc.gov/PortalMapApi/api/Map/GetMapItemsByCategoryId/1,2,3,4,5,6,7

boomerwv1:
Yeah that'll work even better.  I'll update it using that. Thanks.

PR #2976, opened 2021-09-23
Spider for South Carolina

boomerwv1:
Adding spider for South Carolina.

### DIFF
--- a/locations/spiders/south_carolina.py
+++ b/locations/spiders/south_carolina.py
@@ -9,7 +9,6 @@ from locations.items import GeojsonPointItem
 class SCSpider(scrapy.Spider):
     download_delay = 0.2
     name = "south_carolina"
-    item_attributes = {'brand': "State of South Carolina", 'brand_wikidata': "Q53709977"}
     allowed_domains = ["sc.gov"]
     start_urls = (
         'https://applications.sc.gov/PortalMapApi/api/Map/GetMapItemsByCategoryId/1,2,3,4,5,6,7',
--- a/locations/spiders/south_carolina.py
+++ b/locations/spiders/south_carolina.py
@@ -2,6 +2,7 @@
 import re
 
 import scrapy
+import json
 
 from locations.items import GeojsonPointItem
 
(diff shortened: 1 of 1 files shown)


iandees:
These look like a bunch of different kinds of places. Everything from DMV to Library to event center. I'm a little hesitant to include these because it's a collection of different places not owned/run by the same company/organization.

When you did the scraping, did you see any information that you could use to build a category so that a data user could tell the different between all the different kinds of data in here?

jleedev:
The categories are as follows:

1. State Parks
2. Libraries
3. Department of Health and Human Services (DHHS)
4. Court Houses
5. Department of Mental Health (DMH)
6. Department of Motor Vehicles (DMV)
7. SC Works

Which could either be used as straight text, e.g. `type=*` or by mapping to an appropriate OSM tag.

boomerwv1:
Yes, we can break it down into those 7 categories. We can do whatever way you think is best.

iandees:
Can you add an extra property called "category" for now and throw those in there?

locations/spiders/south_carolina.py
```diff
@@ -0,0 +1,49 @@
...
+    item_attributes = {'brand': "State of South Carolina", 'brand_wikidata': "Q53709977"}
+    allowed_domains = ["sc.gov"]
+    start_urls = (
+        'https://applications.sc.gov/PortalMapApi/api/Map/GetMapItemsByCategoryId/1,2,3,4,5,6,7',
+    )
+
+    def parse(self, response):
+        replace_list = (']', '[', '}', '"', '\\t')
+
+        zipre = response.xpath('//body//text()').extract()
```
jleedev, line 20: The response body is JSON and should be accessed with `response.json()`.

locations/spiders/south_carolina.py
```diff
@@ -0,0 +1,49 @@
...
+                    i = i.replace(', ', " ")
+                    j = i.split(',')
+                    for js in j:
+                        properties = {
+                            'ref': j[0].split(':')[1],
+                            'name': j[1].split(':')[1],
+                            'addr_full': j[6].split(':')[1] + ' ' + j[5].split(':')[1],
+                            'city': '',
+                            'state': '',
+                            'postcode': '',
```
jleedev, line 39: Should be `["Zipcode"]` of the record.

locations/spiders/south_carolina.py
```diff
@@ -0,0 +1,49 @@
...
+import re
+
+import scrapy
+
+from locations.items import GeojsonPointItem
+
+class SCSpider(scrapy.Spider):
+    download_delay = 0.2
+    name = "south_carolina"
+    item_attributes = {'brand': "State of South Carolina", 'brand_wikidata': "Q53709977"}
```
jleedev, line 11: I'd leave these off since they don't really make sense as OSM brand tags. The `@spider` attribute is sufficient to identify what these items are.

jleedev (changes requested):

boomerwv1 (commented):
Suggest changes have been made.

PR #3053, opened 2021-10-01
Cbre

boomerwv1:


locations/spiders/cbre.py
```diff
@@ -0,0 +1,45 @@
...
+                    'city': item.split(',')[-4].split('   ')[-1],
+                    'postcode': item.split(', ')[-3].split(' ')[1],
+                    'country': 'US',
+                    'phone': response.xpath('//div[@class="numbers-wrapper__numbers numbers-wrapper__numbers--office tel"]//text()').extract_first().replace('+', ''),
+                    'lat': float(item.split(',')[-2].replace('"','')),
+                    'lon': float(item.split(',')[-1].replace('"','')),
+                }
+
+                yield GeojsonPointItem(**properties)
+        except:
```
iandees, line 44: Can the try/except be removed, please? If it's catching some exception then it should be handled explicitly.

PR #3397, opened 2021-10-29
Fix spider sears

### LINKED ISSUE #3100: Spider sears is broken
During the global build at 2021-09-29-14-42-48, spider **sears** failed with **0 features** and **388 errors**.

Here's [the log](https://data.alltheplaces.xyz/runs/2021-09-29-14-42-48/logs/sears.txt) and [the output](https://data.alltheplaces.xyz/runs/2021-09-29-14-42-48/output/sears.geojson) ([on a map](https://data.alltheplaces.xyz/map.html?show=https://data.alltheplaces.xyz/runs/2021-09-29-14-42-48/output/sears.geojson))

aditya-iitd:
Fixes #3100

### DIFF
--- a/locations/spiders/sears.py
+++ b/locations/spiders/sears.py
@@ -88,15 +88,38 @@ class SearsSpider(scrapy.Spider):
         # Ensure ref is consistent across sears sites - 7 digit unique store identifier (zero padded)
         ref = str(store_data['unit']).zfill(7)
 
+        name = None
+        addr_full = None
+        city = None
+        state = None
+        postcode = None
+        phone = None
+        website = None
+
+        if 'storeName' in store_data:
+            name = store_data['storeName']
+        if 'streetAddr' in store_data:
+            addr_full = store_data['streetAddr']
+        if 'city' in store_data:
+            city = store_data['city']
+        if 'state' in store_data:
+            state = store_data['state']
+        if 'zip' in store_data:
+            postcode = store_data['zip']
+        if 'phone' in store_data:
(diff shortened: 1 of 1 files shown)


locations/spiders/sears.py
```diff
@@ -88,15 +88,38 @@ def parse_sears_outlet_store(self, response):
...
 
+        name = None
+        addr_full = None
+        city = None
+        state = None
+        postcode = None
+        phone = None
+        website = None
+
+        if 'storeName' in store_data:
```
cmecklenborg, line 99: Suggestion: Instead of all these defaults and if statements, you can use the `get` method on the store_data dictionary. E.g.:

```
properties = {
    'ref': ref,
    'name': store_data.get('storeName'),
    'addr_full': store_data.get('streetAddr'),
    ...
```

The get method returns None if the property doesn't exist (unless you specify a different default value in the call)

aditya-iitd: Done, Thanks!

cmecklenborg:
Looks like the format of the sears.com pages have changed. See https://www.sears.com/stores/florida/orlando/0001285.html for an example.

Additionally, `http://www.searshometownstores.com/sitemap-store-urls.xml` 404s, so that path can be removed from the spider.

[sears.log](https://github.com/alltheplaces/alltheplaces/files/7447443/sears.log)

aditya-iitd:
That can be a separate issue. I was just trying to fix #3100

Melaniemaire (approved):

PR #3440, opened 2022-01-21
fixing Jersey Mikes spider

boomerwv1:
I don't see a ticket for this one, but it executed without errors and 0 locations when I ran it. Site looks like it's changed a little, so I made some small changes and it's running again and returning locations.


iandees:
Thanks for the fix. I turned off the posting of tickets because it was getting pretty spammy. The weekly build still runs, though.

PR #3444, opened 2022-01-24
Add Big W spider

iandees:
Replaces https://github.com/alltheplaces/alltheplaces/pull/1643


iandees:
Timed out because the sitemap has a lot of entries.

PR #3455, opened 2022-01-29
Fix target_au

jleedev:
It appears that Akamai Ghost is happy to talk given:

1. A browser user-agent (and certain accept-* headers)
2. Over HTTP/2

Note that [HTTP/2 support in Scrapy is experimental][1].

[1]: https://docs.scrapy.org/en/latest/topics/settings.html?highlight=http2#download-handlers-base

I tested this successfully from GCP and AWS.


iandees:
Nice. Good find on Akamai.

PR #3479, opened 2022-03-08
Coles

boomerwv1:


iandees:
Can you merge in the master branch so the au searchable points are available?

boomerwv1:
@iandees  I can, but these are the same points that were added by @wessport  in https://github.com/alltheplaces/alltheplaces/pull/3477, they should match.

If you still want me to push these let me know.

PR #3488, opened 2022-03-24
Adding SpeeDee Oil Change and Auto Services to AllThePlaces

tlunderwood2014:


locations/spiders/SpeeDee.py
```diff
@@ -0,0 +1,102 @@
+# -*- coding: utf-8 -*-
```
iandees, line 1: Please keep the file names in lowercase. Can you rename this to `speedee.py`?

tlunderwood2014 (commented):
changed the file name to lowercase

iandees:
It looks like the spider isn't returning any data. Does it work for you locally?

tlunderwood2014:
> It looks like the spider isn't returning any data. Does it work for you locally?

Yes, It works locally for me.

jleedev:
Cloudflare is giving a 503 for requests originating from AWS.

PR #3492, opened 2022-03-29
Scrapy fix

### LINKED ISSUE #3491: AttributeError: 'TextResponse' object has no attribute 'body_as_unicode'
The update to Scrapy 2.6.0 (https://github.com/alltheplaces/alltheplaces/pull/3473) removed `scrapy.http.TextResponse.body_as_unicode`.

Should replace with `response.text` instead, but in many cases we should replace with `response.json()`.

jleedev:
Fixes #3491 

- Mechanically replace `json.loads(response.body_as_unicode())` with `response.json()`
- Mechanically replace `response.body_as_unicode()` with `response.text`
- Use `response.json()` in places.

### DIFF
--- a/locations/spiders/24_hour_fitness.py
+++ b/locations/spiders/24_hour_fitness.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 import scrapy
-import json
 from locations.items import GeojsonPointItem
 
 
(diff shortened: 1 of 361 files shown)


iandees:
Thanks for fixing these!

PR #3495, opened 2022-04-02
Fix jq command

jleedev:
Attempt to get latest.json and history.json working


iandees:
Good grief. I thought I had this fixed earlier. Thanks.

PR #3540, opened 2022-04-06
Fix error in chase

jleedev:
Old: 5578 features
New: 8087 features


iandees:
Guess it is running on forks that breaks auto-lint!

PR #3543, opened 2022-04-07
Add spider for Chuy's

### LINKED ISSUE #3501: Chuy's
https://www.chuys.com/locations

mackerel10:
Closes #3501

### DIFF
--- a/locations/spiders/chuys.py
+++ b/locations/spiders/chuys.py
@@ -20,11 +20,13 @@ class ChuysSpider(scrapy.Spider):
 
         for location_id in location_ids:
             try:
-                re_pattern_store_info = rf"infoWindowContent{location_id}\s\+=.+;"
+                re_pattern_store_info = (
+                    rf"infoWindowContent{location_id}\s\+=(?!.*Coming Soon).+;"
+                )
                 store_info = re.findall(re_pattern_store_info, data)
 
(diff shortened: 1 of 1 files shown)


locations/spiders/chuys.py
```diff
@@ -0,0 +1,53 @@
...
+                    "city": locality.split(", ")[0],
+                    "state": locality.split(", ")[1].split()[0],
+                    "postcode": locality.split(", ")[1].split()[1],
+                    "phone": phone,
+                    "website": website,
+                    "lat": lat,
+                    "lon": lon,
+                }
+            except Exception as e:
+                continue
```
iandees, line 51: Instead of catching Exception and continuing, please let the exception out so we can fix it.

iandees:
The output has some extra tabs in the address, can you remove those?

```
"addr:full": "10325 N. Michigan Rd.\t\t\t"
```

locations/spiders/chuys.py
mackerel10: Updated to print a statement and raise the exception

iandees: No need to print. Just let the exception get raised. Remove the try/except.

mackerel10: Done

PR #3546, opened 2022-04-08
Fix history.json

jleedev:
If the file was empty then the jq expression wouldn't work. Correctly initialize the file.

e.g.

```
¶ </dev/null jq '. += [1]' 
¶ <<<'[]' jq -c '. += [1]' 
[1]
¶ 
```


iandees:
Thanks! That's a much better way to check this.

PR #3567, opened 2022-04-16
Add Burger King UK

Cj-Malone:
Not that important, but unnecessary. The others don't have the permission.


Cj-Malone:
Whoops, branching fail

PR #3576, opened 2022-04-19
Add Wilko

Cj-Malone:
This is an easy one, get the urls from the sitemap, pull out the Microdata attributes via xpath.

There should probably be a universal Microdata parser though...

### DIFF
--- a/locations/spiders/wilko.py
+++ b/locations/spiders/wilko.py
@@ -1,0 +1,65 @@
+# -*- coding: utf-8 -*-
+from scrapy.spiders import SitemapSpider
+from locations.items import GeojsonPointItem
+
+
+class WilkoSpider(SitemapSpider):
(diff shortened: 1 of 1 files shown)


iandees:
Yep, it would be nice to have a universal microdata extractor. It seems like there are multiple methods of storing the microdata, though, so I haven't explored how to build a reusable one yet.

Cj-Malone:
I really don't know Python yet, but I'll keep it on my mind. When I wrote the code for [KAFF Places](https://gitlab.com/KAFF/places) I basically only wrote 2 spiders, Microdata and JSON-LD, I'd feed them sitemaps, regex to select the urls, and the itemtype. Sometimes I'd have to override functions when the site broke spec (or my spider couldn't cope with the edge cases of the spec), but most of the time I was basically writing config files.

PR #3577, opened 2022-04-19
Fix petsmart

jleedev:


jleedev:
Believe this is the only error, caused by an unusual redirect.

PR #3583, opened 2022-04-20
Add PureGym

Cj-Malone:
Unfortunately no coordinates from this one.

Is there a procedure for getting location from a geocoder?

Also when there aren't coordinates the GeoJSON export doesn't have `geometry` objects. I don't like this, I'd prefer setting them to 0,0 Null Island.


locations/spiders/PureGym.py
```diff
@@ -0,0 +1,51 @@
+import json
```
iandees, line 1: Can you rename the file so it's lower-cased? `puregym.py` probably

iandees:
> Is there a procedure for getting location from a geocoder?

No, we don't have a geocoder pipeline step set up yet. So far we push that on to the consumers.

> Also when there aren't coordinates the GeoJSON export doesn't have geometry objects. I don't like this, I'd prefer setting them to 0,0 Null Island.

I don't like setting fake data when we have a way to tell consumers that we don't know the geometry. Why do you prefer null island?

locations/spiders/PureGym.py
Cj-Malone: Done

Cj-Malone:
> I don't like setting fake data when we have a way to tell consumers that we don't know the geometry. Why do you prefer null island?

A lot of the GeoJSON tool chain breaks without `geometry`. I had a quick check of the spec, and I think it's required.

iandees:
[GeoJSON IETF spec](https://datatracker.ietf.org/doc/html/rfc7946#section-3.2) allows for `null` geometry:

```
   o  A Feature object has a member with the name "geometry".  The value
      of the geometry member SHALL be either a Geometry object as
      defined above or, in the case that the Feature is unlocated, a
      JSON null value.
```

jleedev:
> Unfortunately no coordinates from this one.

https://www.puregym.com/gyms/ contains the coordinates.

Cj-Malone:
@iandees Fair enough, I'll have to make some bug reports in other places.

@jleedev Thanks, new PR where I pull them out of an image url #3584

PR #3584, opened 2022-04-20
Add PureGym coords

Cj-Malone:


Cj-Malone:
Hmm, just ran it again locally and it works.

PR #3592, opened 2022-04-25
Add spider for Quest Diagnostics

jleedev:


Cj-Malone:
Did you know there is a [SitemapSpider](https://docs.scrapy.org/en/latest/topics/spiders.html#sitemapspider-examples) that can parse the sitemap for you?

Issue #2337, opened 2021-07-23
Help wanted adapting the code for lidl_uk.py to create a spider for Lidl Northern Ireland


ChrisSoderberg:
Lidl UK and Lidl Northern Ireland both use bing maps for their online store locations. I believe the author for lidl_uk.py targets bing maps directly and creates a bing maps query (https://docs.microsoft.com/en-us/bingmaps/spatial-data-services/query-api/query-by-id) to output a URL containing json data for Lidl UK branches which is then fed to the spider. I'm struggling to figure out how I might create a bing query that could do the same job for Lidl Northern Ireland

CloCkWeRX:
https://www.lidl-ni.co.uk/
https://spatial.virtualearth.net/REST/v1/data/91bdba818b3c4f5e8b109f223ac4a9f0/Filialdaten-NIE/Filialdaten-NIE?$select=*,__Distance&$filter=Adresstyp%20eq%201&key=Asz4OJrOqSHy-1xEWYGLbFhH4TnVP0LL1xgj0YBkewA5ZrtHRB2nlpfqzm1lqKPK&$format=json&jsonp=Microsoft_Maps_Network_QueryAPI_2&spatialFilter=nearby(54.5578,-5.7589,4.9864904999999995)

CloCkWeRX:
This spider already exists - 
```
class LidlNISpider(VirtualEarthSpider):
    name = "lidl_ni"
    item_attributes = LidlGBSpider.item_attributes

    dataset_id = "91bdba818b3c4f5e8b109f223ac4a9f0"
    dataset_name = "Filialdaten-NIE/Filialdaten-NIE"
```
Closable @iandees

Cj-Malone:
I don't think NI is a real country code.

Issue #1459, opened 2020-04-10
Lidl


thismakessand:
Lidl operates in 29 countries (in order by # of branches)

- [x] Germany
- [x] France
- [x] United Kingdom
- [x] Poland
- [x] Italy
- [x] Spain
- [x] Netherlands
- [x] Belgium
- [x] Romania
- [x] Portugal
- [x] Greece
- [x] Austria
- [x] Czech Republic
- [x] Finland
- [x] Hungary
- [x] Sweden
- [x] Ireland
- [x] Slovakia
- [x] Denmark
- [x] Bulgaria
- [x] Croatia
- [x] Switzerland
- [x] United States
- [x] Slovenia
- [x] Lithuania
- [x] Serbia
- [x] Cyprus
- [x] Luxembourg
- [x] Malta
- [x] Latvia
- [x] Estonia

Dimitar5555:
Sorry for the late response @Cj-Malone. Feel free to add them when you can.

Cj-Malone:
#10569

PR #3594, opened 2022-04-25
Add Lidl Ireland spiders (NI, IE)

Cj-Malone:


Cj-Malone:
Ireland can be ticked off in #1459

CloCkWeRX:
#2337

PR #3601, opened 2022-04-27
Fix errors in davids bridal

jleedev:
Exotic formatting in opening hours and blank pages with 200 status


Cj-Malone:
[These](https://github.com/alltheplaces/alltheplaces/issues?q=is%3Aissue+is%3Aopen+davids+bridal) can be closed.

PR #3613, opened 2022-05-04
Add Hungry Horse spider

Cj-Malone:


iandees:
Looks like some of these don't have a branchCode: https://data.alltheplaces.xyz/ci/2269749515/hungry_horse/log.txt

Cj-Malone:
It looks like it's just one bad [URL](https://www.hungryhorse.co.uk/pubs/west-lothian/lime-kiln/) in the sitemap. Ideally there would be a way to exclude that one from being requested, but if there is a way I don't know it.

iandees:
Maybe [`sitemap_filter()`](https://docs.scrapy.org/en/latest/topics/spiders.html#scrapy.spiders.SitemapSpider.sitemap_filter) would do it?

Cj-Malone:
Thank you, I was looking for "exclude" or "blacklist". Anyway, I've done that and I've left the `branchCode` check in in case any other URLs go bad in the future.

PR #3626, opened 2022-05-09
Ignore Robots.txt on WordPress API spiders

Cj-Malone:
Untested edit of any spiders accessing "wp-admin/admin-ajax.php" to ignore robots.txt as WordPress blocks them.

### DIFF
--- a/locations/spiders/bayshore_healthcare.py
+++ b/locations/spiders/bayshore_healthcare.py
@@ -12,6 +12,7 @@ class BayshoreHealthcareSpider(scrapy.Spider):
     name = "bayshore_healthcare"
     item_attributes = {"brand": "Bayshore Healthcare"}
     allowed_domains = ["bayshore.ca"]
+    custom_settings = {"ROBOTSTXT_OBEY": False}
 
     def start_requests(self):
         url = "https://www.bayshore.ca/wp-admin/admin-ajax.php?action=location_finder&language=en"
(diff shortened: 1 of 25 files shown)


Cj-Malone:
Looks like all these were already failing, so I'd vote for a merge anyway.

Also, I found it unexpected that the log is uploaded to a directory with the spiders file name, not name/id.
e.g
`locations/spiders/bc_pizza.py log: https://data.alltheplaces.xyz/ci/2294904372/bc_pizza/log.txt` not
`locations/spiders/bc_pizza.py log: https://data.alltheplaces.xyz/ci/2294904372/bcpizza/log.txt`

iandees:
> Also, I found it unexpected that the log is uploaded to a directory with the spiders file name, not name/id. e.g `locations/spiders/bc_pizza.py log: https://data.alltheplaces.xyz/ci/2294904372/bc_pizza/log.txt` not `locations/spiders/bc_pizza.py log: https://data.alltheplaces.xyz/ci/2294904372/bcpizza/log.txt`

This happens because the code that runs the spiders is done entirely in shell script and doesn't know the ID of the spider, only the filename. I agree that's confusing, though. Maybe at some point we can rewrite the spider runner so it has access to more information.

Cj-Malone:
> Maybe at some point we can rewrite the spider runner so it has access to more information.

Yeah, I think this will have to happen at some point. I just tried to write some code to loop through the spiders and extract the Wikidata, (I don't know Python so I may be missing something obvious) but I couldn't figure it out and used grep instead.

Also maybe the file name and the id should be the same? At least recommended for new spiders?

PR #3637, opened 2022-05-12
Improve worldcat

jleedev:
- Page through all the results instead of searching for various cities.
Did not do full run but the page says there should be 139820 results,
which is more than the 45187 seen by the current code.
- Use the worldcat registry id as ref instead of some junk


iandees:
Nice, this is a great improvement!

Issue #3585, opened 2022-04-20
Include Name Suggestion Index IDs in output?


iandees:
Should All The Places start including NSI ids in the output? or maybe including the NSI tags directly?

This would start including OSM "top level" tags like 'shop=*' or 'amenity=*', along with extra metadata.

_Originally posted by @Cj-Malone in https://github.com/alltheplaces/alltheplaces/issues/1343#issuecomment-1104434637_

iandees:
I don't really want to adopt OSM's multiple-tag system for specifying categories (amenity=restaurant, cuisine=burger means "burger restaurant").

I do see value in including category information for spider output, though. I don't know that there's a good "general purpose" set of POI categories out there, so I'm open to suggestions there!

iandees:
Another requirement for categorization would be to make it relatively easy for spider authors to find acceptable category options. Ideally the categories we pick would have an easy list to search and good documentation for what each means (for our spider authors and data consumers).

iandees:
Name Suggestion Index IDs would be good to add, but I don't know that they would help non-OSM data consumers. That's probably OK, because those data consumers could look at the data in the NSI to help map to their internal categorization scheme.

Cj-Malone:
Hmm, I'd stick with OSM over new categories as they are a (relatively) known concept, well discussed already and reasonably [documented](https://wiki.openstreetmap.org/wiki/Shops) (and that links to [shop=*](https://wiki.openstreetmap.org/wiki/Key:shop) and [amenity=*](https://wiki.openstreetmap.org/wiki/Key:amenity)).

If you decided against OSM tags, [schema.org types](https://schema.org/LocalBusiness#subtypes) are a thing. I don't know how well they cover things though, I'm an OSM person.

(I know I just recommended moving away from OSM address tags in #3579. The irony isn't lost on me.)

Cj-Malone:
For cuisine, I don't think `amenity=restaurant` and `cuisine=burger` is a massive issue. But I do dislike how OSM treats it as an array, eg `cuisine=burger;kebab`. I moved towards cuisine as a "namespace" in [KAFF Places](https://gitlab.com/KAFF/places), eg `cuisine:burger=yes` + `cuisine:kebab=yes` (backwards compatibility is really easy too).

jleedev:
[...]

Issue #3578, opened 2022-04-19
Tests


Cj-Malone:
What's the opinion on tests? In my opinion each spider should probably have a few tests to help let us know when they break.

iandees:
Generally tests are a good thing. But can we just the output of the spider to determine if it's broken? If a spider is modified then we'll run it for 1 minute here in Github and get notified for what the output looks like. On the weekly run we can see if it had any errors or if it produced any output places.

Cj-Malone:
> On the weekly run we can see if it had any errors or if it produced any output places.

What is the weekly run/how could I see the output?

If it's like #3391 the log and output give an "AccessDenied" error, and https://data.alltheplaces.xyz/ give me a 403.

iandees:
The auto-runs from Github Actions expire after 30 days, so those links will stop working after a while.

The weekly run is the data at https://alltheplaces.xyz. We recently made https://data.alltheplaces.xyz/runs/latest.json work to programmatically get the details about the most recent build and a corresponding https://data.alltheplaces.xyz/runs/history.json to get the known previous runs.

Cj-Malone:
Back to tests, I don't think the current system about subtle errors?

So if the "link discovery" stage failed to cycle through pages and only got links from the first page, you'd have a some outputted data but drastically fewer. A test to check if certain URLs were "discovered" could help.

Or if the site is updated and an xpath starts to return bogus data, or good data prefixed/suffixed with bad data. You wouldn't know until someone reported it. Tests to say we expect this data to come from this URL could help.

Cj-Malone:
The Specsavers spider is an example of a spider continuing to run without errors, but no longer returns any of the address data it originally did. It's moved from Microdata to Linked Data.

iandees:
I think in examples like that I'd rather detect changes in the amount of data returned, flagging unexpected changes in row count or completed attribute count. I worry that tests will be hard to maintain alongside the spider itself.

mjoe999:
I think the testing comes in two basic themes. To make writing spiders more simple / quicke / forwards compatible then we should attempt to push more boiler plate into common code.

I've started down the road with this pull request: https://github.com/alltheplaces/alltheplaces/pull/3648

Take the common LD JSON mark up as an example. The adherence of individual sites to a "standard" is fairly loose. Any library code starts to grow arms and legs to deal with various cases, think "beautiful soup".

[...]

PR #3648, opened 2022-05-17
[DEV] examples of a few new spiders sharing SEO / branding support ex…

mjoe999:
…tensions

This commit adds a few UK site spiders. Pandora spider is working again and extended to be global. Claire's extended to be global.

The commit is bigger perhaps than you may like as it illustrates bigger concepts at work which is the extension of common code to parse SEO data.

I will add further background to:

https://github.com/alltheplaces/alltheplaces/issues/3585
https://github.com/alltheplaces/alltheplaces/issues/3578

which where recent question issues on the project which generated some discussion.

Cheers,
Joe.

PS. I am new to all of Python, Github, scrapy and alltheplaces .. be gentle? :-)

### DIFF
--- a/locations/brands.py
+++ b/locations/brands.py
@@ -1,0 +1,58 @@
+import enum
+from locations.items import GeojsonPointItem, source_data
+
+
+class Brand(enum.Enum):
+    """
+    A wikidata Q-code helps us disambiguate "things". It helps matching with other such disambiguated data sets.
+    This enumeration helps us reference "things" in code without typing in numbers which is brittle. The enumeration
+    has both name and code point entries which allows it to be checked programmatically against data sets such as
+    wikidata itself and also the "name suggestion index".
+    """
+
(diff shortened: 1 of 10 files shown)


mjoe999:
I'm going to come back with a reworked request.

iandees:
Sorry for the lack of response and feedback here. I was on vacation and then this is a long holiday weekend so I haven't been in front of a computer as much as usual.

I like the idea of extracting out reusable and testable bits of code that individual spiders can use to make parsing pages easier and more repeatable. For example, when we know a page is using microdata, we should call a single function that extracts microdata from the page and returns that data so the spider can use it to populate an item.

I don't think I want to move the spider-specific code out of the spider itself, though. The enum and it's "item()" function in your PR here seen counterintuitive to me.

It might be easier to understand if I saw a smaller PR that was more focused? This PR seems to be proposing a pretty seismic shift in the way alltheplaces spiders work and I'd like to see smaller steps along the way if we're going to think about doing that change.

PR #3651, opened 2022-05-18
Morrisons

Cj-Malone:
Now returns 1248 locations including Morrisons Daily, Morrisons Select and Morrisons petrol stations and the normal Morrisons supermarkets.


Cj-Malone:
I've also included the OSM POI types, until something is decided #3585

PR #3653, opened 2022-05-18
Fix worldcat

jleedev:
Retry a few times when response is empty json array, and determine the number of pages in a smarter way.


jleedev:
The site is nonetheless very flaky, but this is at least verbose about it.

iandees:
Yea, I kinda regret adding this site in. I generally don't want to scrape other aggregators, but it's such a useful dataset...

PR #3654, opened 2022-05-18
Update README.md

Cj-Malone:
Try to encourage them to open an issue instead of blocking us. Ideally we'd get an issue with "Please add a crawl delay" or "Here is a more efficient API"


iandees:
Yea, good idea.

PR #3661, opened 2022-05-24
Add attribution to the preview map

Cj-Malone:
Untested


iandees:
Thanks. For posterity, I updated https://data.alltheplaces.xyz/map.html with this using:

```
AWS_PROFILE=openaddresses aws s3 cp preview/index.html s3://alltheplaces.op
enaddresses.io/map.html
```

PR #3663, opened 2022-05-25
Add Game Spider and located_in

Cj-Malone:
Ended up having to do way more cleaning that I thought with this one.

This does output errors, there are a bunch of pages (e.g. [Clydebank](https://storefinder.game.co.uk/game/stores/686/Clydebank)) that eventually timeout to a 500 error. They are listed in both the sitemap and the [human list](https://storefinder.game.co.uk/game/stores/search?letters=ABCDEFGHIJKLMNOPQRSTUVWXYZ), I've not checked if they are old stores or a temporary site issue. I've reduced the timeout so it will finish quicker, but I'd love a better solution.

I've also added `located_in` and `twitter` as defined fields.


jleedev:
The OSM key [`is_in=*`] is in use for describing these things, such as `name=Starbucks`+`is_in=Sainsbury's`.

[`is_in=*`]: https://taginfo.openstreetmap.org/keys/is_in

While this project's goal is not a direct mapping of OSM tags I think it would be suitable.

Cj-Malone:
I think `is_in` was more for administrative areas, it's deprecated according to JOSM, and the wiki has it deprecated in Netherlands. IIRC the last time "shop in a shop" was brought up on the tagging mailing list it was observed that `addr:housename` and `operator` were 2 tags being (mis)used, I don't think there was a outcome, I try not to spend any time on that list.

I've used to the OSM tags for Twitter, Facebook and image, I just don't think OSM has a comparison tag for located in.

Cj-Malone:
I think an interesting thing about `exporters.py` is that if OSM decide on a tag in the future, All The Places could map the `located_in` field to 2 tags while the first is deprecated. I fully agree with following OSM tagging as much as possible.

This PR was meant to break down @mjoe999 pull request into a smaller one. If that PR progresses, this can be closed.

PR #3667, opened 2022-05-30
some spiders sharing new SEO structured data parsing and brand support

mjoe999:
I closed my previous pull request:
https://github.com/alltheplaces/alltheplaces/pull/3648
My closure crossed with @iandees comment on it. This PR anticipates and acknowledges the dislike of the brand enum (ack also @jleedev) and replaces it with a class giving very basic brand support.
I know this PR might also be "too big" and if so instructed I will attempt to chop it up but throw it to the "crowd" now for general comment.
The one thing I do encourage is a look over some of the new spiders that have been written with said support. They are pretty "tight". I went through an exercise writing approximately 300 spiders (mostly UK) to get to this point.

### DIFF
--- a/locations/items.py
+++ b/locations/items.py
@@ -13,6 +13,7 @@ class SourceData(scrapy.Item):
     Source data associated with the creation of this POI. Pipeline code may well find it useful.
     For example, social media links may optionally be extracted from individual POI websites.
     """
+
     response = scrapy.Field()
     # The entry used to generate the POI.
     ld_json = scrapy.Field()
--- a/locations/items.py
+++ b/locations/items.py
@@ -41,7 +41,7 @@ class GeojsonPointItem(scrapy.Item):
     ref = scrapy.Field()
     brand = scrapy.Field()
     brand_wikidata = scrapy.Field()
-    image = scrapy.Field()
+    located_in = scrapy.Field()
     located_in_wikidata = scrapy.Field()
     extras = scrapy.Field()
     source_data = scrapy.Field()
--- a/locations/items.py
+++ b/locations/items.py
@@ -8,6 +8,16 @@
 import scrapy
 
 
+class SourceData(scrapy.Item):
+    """
+    Source data associated with the creation of this POI. Pipeline code may well find it useful.
+    For example, social media links may optionally be extracted from individual POI websites.
+    """
+    response = scrapy.Field()
+    # The entry used to generate the POI.
+    ld_json = scrapy.Field()
+
+
 class GeojsonPointItem(scrapy.Item):
     lat = scrapy.Field()
     lon = scrapy.Field()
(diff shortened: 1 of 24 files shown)


JoeDixon-TomTom:
I've been ill these last few weeks. Sorry for the delay. I've hopefully fixed the lint errors now.

locations/items.py
```diff
@@ -29,6 +40,49 @@ class GeojsonPointItem(scrapy.Item):
     ref = scrapy.Field()
     brand = scrapy.Field()
     brand_wikidata = scrapy.Field()
-    located_in = scrapy.Field()
```
Cj-Malone, line 32: I've got nothing against `image` being added, but I don't think `located_in` should have been removed. Mistake when merging?

JoeDixon-TomTom: I thought that would prompt a discussion. I think you will only want to be expending effort on "located in" coding for reasonably sized brands. That kind of implies a wikidata code, or certainly that one would be created if not already in existence. In which case why have the ambiguity of name and code? Having said that I will add back in further commit on branch unless you agree with my reasoning.

Cj-Malone: Urm, well having `image` twice is defiantly wrong.

For `located_in`, I can see your point, but I'm not sure.

mjoe999: ouch ... I missed the double 'image' .. to be fixed also

mjoe999: Ok so I have pushed another commit: duplicate image member removed, located_in re-instated, has_geo() check in calires spider removed.

On the latter change it is a general thing that some spiders will wish to remove pages without geo. A piece of pipeline code that looks for "remove_missing_geo_items" member in a spider may be of use but I will not overload my current commit!

locations/spiders/claires.py
```diff
@@ -1,106 +1,27 @@
...
-        ).extract()
-
-        for store_url in store_urls:
-            yield scrapy.Request(url=store_url, callback=self.parse)
+    sitemap_urls = ["https://stores.claires.com/sitemap.xml"]
+    sitemap_rules = [(".html", "parse_store")]
+
+    def parse_store(self, response):
+        for item in extract_ldjson(self.brand, response, "JewelryStore"):
+            if item.has_geo():
```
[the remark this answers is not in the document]
mjoe999: The Game sitemap includes https://storefinder.game.co.uk/game/stores/search which matches my sitemap rules. has ld+json for store type but no geo.
Londis and Premier have a few (fingers of one hand) stores that have no geo. I believe they are to be opened. The number is so small we should not worry.
Pandora has many intermediate pages which match with ld+json but have no geo.
So in all cases I think the spider code is doing what we would want it to do but as said it could be expressed more succintly.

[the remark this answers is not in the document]
mjoe999: Further commit:

So Londis and Premier no longer check for geo.

Pandora spider has a sitemap rule filter that does not pick up "intermediate" pages in store finder structure

Game: keeping approach as it is a lot clearer. For future reference you can have a sitmap rule regex with a target parse function of None, the scrapy framework will ignore such matched URLs.

mjoe999:
After a few rounds of tweaking I think the Pandora, Premier and Londis spiders illustrate pretty well the point which is trying to be achieved. They add over 8000 pretty rich POIs and are the smallest spiders in the entire project. 15 lines each of essentially boiler plate code. As I said in my original comment on this pull request then I have a whole bunch of similarly tight spiders for the UK that I can start to trickle in if we can get to agreement on the shared utility code.

mjoe999:
I've added a few more spiders, mainly major UK brands but Best Western and Body Shop more globally significant. The PR is now giving the project well over 20K new POIs. It's worth pointing out that even though this overall commit talks about SEO structured data most of the code (in seo.py) is not about microdata/ld+json etc. In fact there's less than 100 lines of code concerned directly with that. Nearly all the spiders contributed require bespoke JSON parsing from the target website. seo.py is lightening that load significantly. I'm hoping that the fact that the average length of the new/touched spiders is 25 lines of code (mainly boiler plate) is saying something?

mjoe999:
Let's be kind to ourselves and close this one.

PR #3669, opened 2022-06-03
I have added another spider which can be used to scrape shops from the GLS api. (Only For Germany)

Ebubeker:
I have managed to create this scraper also by using the searchable_places folder which gave me the coordinates in order to search everything better.

### DIFF
--- a/locations/spiders/gls-deu.py
+++ b/locations/spiders/gls-deu.py
@@ -1,0 +1,66 @@
+import scrapy
+import json
+import csv
+
+class GenspiderSpider(scrapy.Spider):
+    name = 'glsdeu'
+    allowed_domains = ['api.gls-pakete.de']
+    index = 1
+    start_urls = ['https://api.gls-pakete.de/parcelshops?version=4&coordinates=50.88246,8.91596&distance=50']
+    usedLongLat = []
+    deList = []
+
+    with open('../searchable_points/eu_centroids_40km_radius_country.csv', 'r') as openFile:
+      results = csv.reader(openFile)
+      for result in results:
+          if(result[3] == 'DE'):
+              data = {
+                  "longitude": result[2],
+                  "latitude": result[1]
+              }
+              deList.append(data)
+
+
+    def parse(self, response):
+
+        firstResults = json.loads(response.body)
+        results = firstResults['shops']
+        for result in results:
+            address = result['address']
+            coordinates = address['coordinates']
+            longitude = coordinates['longitude']
+            latitude = coordinates['latitude']
+            name = address['name']
+            state = 1
+
+            if(GenspiderSpider.usedLongLat):
+                for used in GenspiderSpider.usedLongLat:
+                    if(used['longitute'] == float(longitude) and used['latitude'] == float(latitude)):
+                        state = 0
+
+            if(state == 1):
+                dictToAdd = {
+                    "longitute": float(longitude),
+                    "latitude": float(latitude)
+                }
+
+                GenspiderSpider.usedLongLat.append(dictToAdd)
+
+                yield {
+                    "type": "Feature",
+                    "geometry": {
+                        "type": "Point",
+                        "coordinates": [float(longitude), float(latitude)]
+                    },
+                    "properties": {
+                        "name": name
+                    }
+                }
+
+
+        next_link = 'https://api.gls-pakete.de/parcelshops?version=4&coordinates=' + str(format(float(GenspiderSpider.deList[GenspiderSpider.index]['latitude']), '.5f')) + ',' + str(format(float(GenspiderSpider.deList[GenspiderSpider.index]['longitude']), '.5f')) + '&distance=50'
+
+        if GenspiderSpider.index < 16477:
+            GenspiderSpider.index += 1
+            print(GenspiderSpider.index)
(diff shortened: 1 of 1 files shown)


locations/spiders/gls-deu.py
```diff
@@ -0,0 +1,66 @@
...
+
+class GenspiderSpider(scrapy.Spider):
+    name = 'glsdeu'
+    allowed_domains = ['api.gls-pakete.de']
+    index = 1
+    start_urls = ['https://api.gls-pakete.de/parcelshops?version=4&coordinates=50.88246,8.91596&distance=50']
+    usedLongLat = []
+    deList = []
+
+    with open('../searchable_points/eu_centroids_40km_radius_country.csv', 'r') as openFile:
```
Cj-Malone, line 13: This should be in a `start_requests` function, see [this example](https://github.com/alltheplaces/alltheplaces/blob/fdf689dd0f01102349a47851bf012cd59e5246cb/locations/spiders/starbucks.py#L16).

locations/spiders/gls-deu.py
```diff
@@ -0,0 +1,66 @@
...
+        results = firstResults['shops']
+        for result in results:
+            address = result['address']
+            coordinates = address['coordinates']
+            longitude = coordinates['longitude']
+            latitude = coordinates['latitude']
+            name = address['name']
+            state = 1
+
+            if(GenspiderSpider.usedLongLat):
```
Cj-Malone, line 36: I don't understand the need for `usedLongLat`. Each set of coords should only be called once anyway,

locations/spiders/gls-deu.py
```diff
@@ -0,0 +1,66 @@
...
+
+            if(state == 1):
+                dictToAdd = {
+                    "longitute": float(longitude),
+                    "latitude": float(latitude)
+                }
+
+                GenspiderSpider.usedLongLat.append(dictToAdd)
+
+                yield {
```
Cj-Malone, line 49: There seems to be a bunch more data in the API that you could be returning: address, phone number, ref and opening hours at least.

I'd also prefer it to be a `GeojsonPointItem`

Ebubeker: I will be fixing these issues as soon as possible. Thank you so much for the feedback.

iandees:
As long as you emit something unique per store in the `ref` value of the item, the system will remove duplicates for you, so you don't have to worry about making duplicate requests.

Also, you're probably making duplicate requests because you're truncating the lat/lon values from the CSV file.

iandees (changes requested):

locations/spiders/gls-deu.py
```diff
@@ -0,0 +1,70 @@
...
+            "./locations/searchable_points/eu_centroids_40km_radius_country.csv",
+        ]
+
+        for point_file in searchable_point_files:
+            with open(point_file) as openFile:
+                results = csv.DictReader(openFile)
+                for result in results:
+                    if(result["country"] == "DE"):
+                        longitude = format(float(result["longitude"]), '.5f')
+                        latitude = format(float(result["latitude"]), '.5f')
```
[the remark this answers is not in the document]
Ebubeker: It is required to do it.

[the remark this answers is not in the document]
iandees: Instead of doing the string formatting here, please do it as part of the URL template. Instead of `{},{}`, you can use `{:0.5},{:0.5}`.

```
>>> "{},{}".format(0.4123423, 0.853592)
'0.4123423,0.853592'
>>> "{:0.5},{:0.5}".format(0.4123423, 0.853592)
'0.41234,0.85359'
```

locations/spiders/gls-deu.py
```diff
@@ -0,0 +1,57 @@
+import scrapy
+import json
+import csv
+from locations.items import GeojsonPointItem
+
+HEADERS = {"X-Requested-With": "XMLHttpRequest"}
+STORELOCATOR = "https://api.gls-pakete.de/parcelshops?version=4&coordinates={},{}&distance=40"
+
+class GenspiderSpider(scrapy.Spider):
```
iandees, line 9: The class name should be related to the spider. Can you change this to `GLSSpider` or `GeneralLogisticsSystemsSpider`?

locations/spiders/gls-deu.py
```diff
@@ -0,0 +1,57 @@
...
+    name = 'glsdeu_spider'
```
iandees, line 10: The name shouldn't include "spider". Can you change this to `gls_de`

locations/spiders/gls-deu.py
```diff
@@ -0,0 +1,57 @@
+import scrapy
```
iandees, line 1: Please rename the file `gls_de.py` to match the other country-specific spiders.

iandees:
Does this work when you run it locally? It's not running in our automatic test environment.

Ebubeker:
Yeah it works pretty well locally.

iandees:
I'm not sure why the spider isn't running, but we should get better logs when the next all_spider run happens.

Also, I'll let the CI bot automatically fix formatting issues.

iandees:
Thanks for your contribution!

PR #3670, opened 2022-06-05
Add spider for Western Union

jleedev:
This one is slow to start as it starts up and immediately enqueues half
a million scrapy Request objects in memory. Supplying
`scrapy crawl western_union -s JOBDIR=western_union-job/`
allows it to flush the Request objects to disk until they are needed and
reduces the amount of heap it needs. Might not be an issue on a machine
with sufficient RAM.

The site is happy to serve 800 pages/min, completes in under 8 hours,
produces 393778 items.

### DIFF
--- a/locations/spiders/western_union.py
+++ b/locations/spiders/western_union.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+import json
+
+from scrapy.spiders import SitemapSpider
+
+from locations.items import GeojsonPointItem
+
+
+class WesternUnionSpider(SitemapSpider):
+    name = "western_union"
+    item_attributes = {"brand": "Western Union", "brand_wikidata": "Q861042"}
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Did you see the API? It can be called with [lat/lon](https://location.westernunion.com/api/locations?country=US&lat=40.7127753lng=-74.0059728), but a [query](https://location.westernunion.com/api/locations?country=US&q=New+York&page=1) returns more results.

Cj-Malone:
I guess we've all seen now with the spiders page update, but the output varies a lot on this one. The [results](https://data.alltheplaces.xyz/runs/2022-08-20-13-32-21/stats/_results.json) confirm this is hitting the 4 hour limit.

iandees:
I'll bump the 4 hour limit up to 8.

PR #3692, opened 2022-06-19
Use HTTPS for innout spider

Cj-Malone:
This spider has been failing the weekly runs, but runs fine locally.


iandees:
This is failing because their web host is using a severely outdated SSL configuration: https://github.com/scrapy/scrapy/issues/5491

PR #3696, opened 2022-06-20
Fix: IKEA

Romeo-PHILLIPS:
After some investigation on IKEA store locators (more specifically the ones in the France https://www.ikea.com/fr/fr/meta-data/navigation/stores-detailed.json) I noticed that there was missing data in the GeoJson. Issue was that some countries don't specify the key `storePageUrl` (1 out of 36 for France). 

So instead not ingesting them, we ingest them and link it to the main store page url for that given country.

### DIFF
--- a/locations/spiders/ikea.py
+++ b/locations/spiders/ikea.py
@@ -62,8 +62,6 @@ class IkeaSpider(scrapy.Spider):
     def parse(self, response):
         data = response.json()
         for store in data:
-            if "storePageUrl" not in store:
-                continue
             opening_hours = OpeningHours()
             for day in store["hours"]["normal"]:
                 if day["open"] != "":
@@ -72,6 +70,8 @@ class IkeaSpider(scrapy.Spider):
                         day["open"],
                         day["close"],
                     )
+            split_url = response.url.split("/")
+            country_path = f"{split_url[3]}/{split_url[4]}"
             properties = {
                 "lat": store["lat"],
                 "lon": store["lng"],
@@ -80,7 +80,9 @@ class IkeaSpider(scrapy.Spider):
                 "city": store["address"].get("city"),
                 "postcode": store["address"].get("zipCode"),
                 "country": response.request.url[21:23].upper(),
-                "website": store["storePageUrl"],
+                "website": store["storePageUrl"]
+                if "storePageUrl" in store
+                else f"https://www.ikea.com/{country_path}/stores/",
                 "ref": store["id"],
                 "opening_hours": opening_hours.as_opening_hours(),
                 "extras": {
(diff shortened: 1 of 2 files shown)


iandees:
It looks like the spider is failing because not all stores have hours fields: https://data.alltheplaces.xyz/ci/2529054025/ikea/log.txt

Can you adjust to be resilient to that please?

Cj-Malone:
OK, so not it's failing in Indian stores because "24:00" isn't a valid time. We do a hacky workaround with `23:59`, so you could do it in the spider.
```patch
diff --git a/locations/spiders/ikea.py b/locations/spiders/ikea.py
index 312c6615..8d434e8b 100644
--- a/locations/spiders/ikea.py
+++ b/locations/spiders/ikea.py
@@ -70,7 +70,7 @@ class IkeaSpider(scrapy.Spider):
                     opening_hours.add_range(
                         day["day"][0:1].upper() + day["day"][1:2].lower(),
                         day["open"],
-                        day["close"],
+                        day["close"].replace("24:00", "23:59"),
                     )
             properties = {
                 "lat": store["lat"],
```

But it may be better to do it in the opening hours class, so other spiders benefit.
```patch
diff --git a/locations/hours.py b/locations/hours.py
index c1b24caa..bdd0bead 100644
--- a/locations/hours.py
+++ b/locations/hours.py
@@ -17,6 +17,8 @@ class OpeningHours(object):
             return
         if close_time.lower() == "closed":
             return
+        if close_time == "24:00":
+            close_time = "23:59"
         if not isinstance(open_time, time.struct_time):
             open_time = time.strptime(open_time, time_format)
         if not isinstance(close_time, time.struct_time):
```

locations/spiders/ikea.py
```diff
@@ -62,16 +62,16 @@ class IkeaSpider(scrapy.Spider):
...
         data = response.json()
         for store in data:
-            if "storePageUrl" not in store:
-                continue
             opening_hours = OpeningHours()
-            for day in store["hours"]["normal"]:
+            for day in store["hours"]["normal"] if "hours" in store else []:
                 if day["open"] != "":
                     opening_hours.add_range(
                         day["day"][0:1].upper() + day["day"][1:2].lower(),
```
Cj-Malone, line 69: This line can be changed to `day["day"].title(),`

Cj-Malone: My bad, it's actually `day["day"].title()[:2],` that's the whole line.

`title` gives us the first letter in uppercase, than the next letters in lower ("MON" -> "Mon"). And then we spit it to the just to 2 chars, so "Mon"->"Mo".

Cj-Malone:
LGTM

PR #3697, opened 2022-06-23
Fix Bright Horizons

Cj-Malone:


iandees:
Looks like it might be getting blocked or parsing the sitemap is taking too long. Hopefully the longer timeout on the weekly run will let it be successful.

Cj-Malone:
I had convinced myself AWS was down again, but that makes more sense. Thank you.

PR #3701, opened 2022-06-28
fix: Mcdonalds France

Romeo-PHILLIPS:
Capture d’écran 2022-06-28 à 13 43 40
Capture d’écran 2022-06-28 à 13 43 53
1506 McDonalds FR scraped.

Previous scraper URL did not longer work, might have a look to other countries if its similar.


Romeo-PHILLIPS:
I might be doing too many requests.

PR #3702, opened 2022-06-28
fix: mcdonalds usa scraper didn't use the correct order lat/long

Romeo-PHILLIPS:
`"coordinates": [-157.708106, 21.28524]}},`
First coordinate should be latitude

Capture d’écran 2022-06-28 à 18 19 10


iandees:
It looks to be correct as-is. In GeoJSON, the first number in a coordinate array is "X" or "longitude", then "Y" or "latitude". Google Maps is showing you "lat, lon". Did the spider produce data incorrectly?

PR #3703, opened 2022-06-28
Fix Asda spider

Cj-Malone:
Previously it covered 800 km of the UK, trimming some of Scotland out.


CloCkWeRX:
@Cj-Malone - question for trying to build auto detection of virtual earth spiders, how are you determining where the data is?

At the moment, https://storelocator.asda.com/ presents as a Yext store; but this spider still functions.

Cj-Malone:
https://www.lidl.co.uk/ may be easier to test

PR #3705, opened 2022-06-29
Standardize virtualearth pagination

Cj-Malone:
Also slightly improved the filters and only ask for the fields we use to shrink the response.


iandees:
Thanks for letting me use your branch to debug this issue. Found [this StackOverflow](https://stackoverflow.com/questions/68348222/aws-s3-ls-gives-error-botocore-utils-badimdsrequesterror-botocore-awsrequest-a) which suggested setting the AWS region and got it working 👍🏻 .

PR #3706, opened 2022-06-29
Feat/ingest more US mcdonalds

Romeo-PHILLIPS:
"There are 13,366 McDonald's locations in the United States as of June 20, 2022. "

I tried to tweak the search radius to find more Mcdonalds in the US here are my results:

- 50 Radius (5 minutes) = 9500 found +30%
- 25 radius (24 minutes) = 10273 found +40%
- 10 radius (around an hour) = 12514 found +71%

Currently in this PR I've set it to be 50 radius (I don't if we want to allow it to go to 10 even then length, might want to tweak the download_delay to be smaller and still make sure we don't get 403'ed). 

I think it would be a good addition because:

allplaces


Cj-Malone:
I feel like there must be a better API, but I can't find it.

Are those times with download_delay 0.2?

Romeo-PHILLIPS:
> I feel like there must be a better API, but I can't find it.
> 
> Are those times with download_delay 0.2?

Yes all at 0.2

iandees:
We have some spiders that take multiple hours, so an hour isn't too bad. Thanks for looking into this.

Romeo-PHILLIPS:
I'll lower it to a 10 mile radius then if it isn't an issue :)

locations/spiders/mcdonalds_us.py
```diff
@@ -8,13 +8,14 @@ class McDonaldsUSSpider(scrapy.Spider):
     name = "mcdonalds_us"
     item_attributes = {"brand": "McDonald's", "brand_wikidata": "Q38076"}
     allowed_domains = ["www.mcdonalds.com"]
+    download_delay = 0.2
 
     def start_requests(self):
         # Experimentally found that 250 was max results
-        base_url = "https://www.mcdonalds.com/googleappsv2/geolocation?latitude={lat}&longitude={lng}&radius=100&maxResults=250&country=us&language=en-us"
+        base_url = "https://www.mcdonalds.com/googleappsv2/geolocation?latitude={lat}&longitude={lng}&radius=50&maxResults=250&country=us&language=en-us"
```
iandees, line 15: When I test this in New York, it seems to fail with a blank response when maxResults is 250. It works with 200, so we might have to stop there. That's still not enough for the most dense areas, though.

in the future, we might add a recursive call with a smaller radius and a slightly offset lat/lon to get the rest?

Romeo-PHILLIPS: Scratch that, I seem to get more then 213 features with 25 mile radius (maxResults seems to max out there for that size) in NYC. 

10 miles still accepts the parameter to 250 (but I think it might fail if it actually returns that value) and only returns 89 for the given lat/long

Cj-Malone:
Just found a new API, no idea if it gives different data.

```bash
curl --data "grantType=client_credentials" --user "1QL44pFIYjpKqIR39rkFIwGd1XEczAeK:HodcWznEVtyk2EH1A0crCLQl7VkNl4MP" "https://eu-prod.api.mcd.com/v1/security/auth/token"

curl -H "Authorization: Bearer eyJhbGciOiJBMTI4S1ciLCJlbmMiOiJBMTI4Q0JDLUhTMjU2IiwidHlwIjoiSldUIiwia2lkIjoiMjAxNi0wOS0xOSJ9.Dhmq0rt9iNdHQDmPHnY5RPY1vrFiIgDy4eqeCcGTBR_66W_xf8of0g.zxU99r8f055AfjPAVjzDJQ.5NMUMvzIP0jG-Iq2PmKpP5ERjALmBz5fTVh8Tahh-diWwc33DYOz8conQuyMXREgl-aooSZycYg8rPMugRC7FZo-1XHRezLWBtO7djU4slBGPArSjlnwLD8_Si9tlwwYTlt7ax2mk26XJdHAQ4Yy8JFHx-09J8VwhM2tMM1j2OgM-d6PXTYW6D31sCS2JBnsBRevRKa8rrmPYa9PsKtNUmqBEt_ru4Fs26MBQJ9tMkgwDTkkfqRAUUS0LjX9b111BDIUFdY1jeuKhKL93ju_ClaxXiKqZo1bkkWGFK9-U4hkF-CJva7YZfZO7Qj-sWnWsE0Wqp-NRzM72dUq1E-57Mx9j3iXfivmizE0h7uqs0s4bFCxU3xhwuRX3NZCWNfwOP9bVro9jnpDkxhXDF-J64yrq76PBti0BhFjzeJI749CMINeGN-WPTyzRZcqaz1ukp_OlH_3dhSOONbRVpgHjV3vqjRF1kqu6LWV05Y11XJpXFTBjHIXJVui8pI8C5Uy4JoL2GqJ_4J2qzy9q-Q3VZlfkiBXP3baHsaQzId32tOLl92xxpPfzIRv5CAq7ln9jp2aev3-OlWDrbbhMw6FE4idmmIeK7WMm8ZIX33RgSQcGK0-t479YI1ltRIrNjfnELrw9CRTIDx2UVnlEzsASjQOxpmqIcBBIK5rxv6SnEeZS-Oe66VoF9mle5EdzDkRz4hzmLIsB64UaIVDjvbta1AcjnpkAHD2mm_olG-UzUg.2ywIUfU_YB4ef-K6ysBxFg" -H "mcd-clientid: 1QL44pFIYjpKqIR39rkFIwGd1XEczAeK" -H "mcd-uuid: 4ABA8315-72B9-4079-9418-339BE8B8A304" -H "Accept-Language: en-GB" "https://eu-prod.api.mcd.com/exp/v1/restaurant/location?distance=50&filter=summary&latitude=50.904&longitude=-1.163&pageSize=25"
```

Romeo-PHILLIPS:
It looks interesting, only thing is that it seems to be only for UK McDonalds

Cj-Malone:
Urm, I think there was a market parameter (maybe mcd_market?), but I for some reason I didn't post it. I'll find it in my bash history after work.

Cj-Malone:
So it was "mcd-marketid", eg `mcd-marketid: UK`, but that doesn't seem to work for US. You can see some details [here](https://config-api-dot-prd-euw-gmal-mcdonalds.appspot.com/api/config/v1/locations?key=NDM4MzU4MDk1NjA5NDI1NDk2NDEzMjkyOTkwMTIzNDg6NzIwMzl5eTc5NDBkYjkycnp6ZW4xNXFpYnZlOTRsbGtjejY3bmJ6eXlsNGU5ZTlhaHpvNGpzdzRyMjlsOWlvMQ==)

Romeo-PHILLIPS:
Hmm interesting, the only one in the request you gave is mcd-marketid: UK . USA don't have a marketid, but it also doesn't work with those that do have one, maybe something to do with the mcd-clientid/uuid?

PR #3708, opened 2022-06-29
fix: hugo boss scraper

Romeo-PHILLIPS:
https://github.com/alltheplaces/alltheplaces/issues/3202

### DIFF
--- a/locations/spiders/hugoboss.py
+++ b/locations/spiders/hugoboss.py
@@ -17,10 +17,13 @@ day_formats = {
 class HugoBossSpider(scrapy.Spider):
     name = "hugoboss"
     item_attributes = {"brand": "Hugo Boss"}
-    allowed_domains = ["production-web-hugo.demandware.net"]
+    allowed_domains = [
+        "production-web-hugo.demandware.net",
+        "production-na01-hugoboss.demandware.net",
+    ]
     download_delay = 0.5
     start_urls = (
-        "https://production-web-hugo.demandware.net/s/US/dw/shop/v16_9/stores?client_id=871c988f-3549-4d76-b200-8e33df5b45ba&latitude=37.09024&longitude=-95.71289100000001&count=200&maxDistance=100000000&distanceUnit=mi",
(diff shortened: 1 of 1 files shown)


locations/spiders/hugoboss.py
```diff
@@ -65,18 +66,18 @@ def parse(self, response):
                     "name": store["name"],
                     "opening_hours": clean_hours,
                     "website": "https://www.hugoboss.com/us/stores",
-                    "addr_full": store["address1"],
-                    "city": store["city"],
+                    "addr_full": store.get("address1"),
```
Cj-Malone, line 69: This should be street_address, not addr_full.

locations/spiders/hugoboss.py
```diff
@@ -17,12 +17,13 @@
 class HugoBossSpider(scrapy.Spider):
     name = "hugoboss"
     item_attributes = {"brand": "Hugo Boss"}
```
Cj-Malone, line 19: It would be nice to add "brand_wikidata": "Q491627" here.

locations/spiders/hugoboss.py
```diff
@@ -65,18 +66,18 @@ def parse(self, response):
                     "name": store["name"],
                     "opening_hours": clean_hours,
                     "website": "https://www.hugoboss.com/us/stores",
```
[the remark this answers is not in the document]
Cj-Malone: Adding a generic URL isn't wanted in All The Places, via item_attributes or directly added to the item. Consumers could get it programmatically from Wikidata. A specific URL is wanted, where possible.


[This url structure](https://www.hugoboss.com/us/storedetail?storeid=9925403) is easier than the one I mentioned above above.

Cj-Malone:
The API publishes `c_contactEmail`, we should collect that too.

We may also be able to use `c_categories` to generate a [clothes](https://wiki.openstreetmap.org/wiki/Key:clothes) extra.

Cj-Malone:
Anything that's not a declared field in [items](https://github.com/alltheplaces/alltheplaces/blob/master/locations/items.py) needs to be but in an "extras" object. I was thinking that "email" had been added, but it's not merged yet, so that'll have to be an extra too,

https://github.com/alltheplaces/alltheplaces/blob/9df26cb677a6481ae73f34f064ddac476df427af/locations/spiders/costacoffee_gb.py#L70-L74

Cj-Malone:
LGTM

PR #3728, opened 2022-07-06
fix: costco

Romeo-PHILLIPS:
response.json didn't manage to decode json for some reason (maybe someone has an idea), using json.loads instead.


Romeo-PHILLIPS:
Nevermind, they just actually return empty when they rate-limit.

PR #3737, opened 2022-07-15
Add Linked Data parser

Cj-Malone:


iandees:
I was thinking more along the lines of:

```
ld_json = json.loads(ld_text)
ld = LinkedDataParser()
item = ld.parse_item(ld_json)
```

But let's see how this works! Thanks for adding it.

PR #3740, opened 2022-07-17
Unify class names and filenames

Cj-Malone:
A bunch of the class names were bad, "TemplateSpider", spelling mistakes, wrong names (from copy+pasting), mixed camel case and snake case, etc.

I was also going to rename some files to include underscores for spaces, but I was getting carried away and I wasn't sure if it was wanted. eg:

```patch
--- a/locations/spiders/wellsfargo.py
+++ b/locations/spiders/wells_fargo.py
```

I've not changed any spider names in this, but they are also inconstant, but changing them could have consequences for any consumers.

I also added some more wikidata ids. The CI tests are going to be awful in this because so many of the spiders are broken.


iandees:
Whew, that's a lot of cleanup. Thanks!

PR #3741, opened 2022-07-18
Add Belhaven Pubs spider

Cj-Malone:


locations/utils.py
```diff
@@ -0,0 +1,18 @@
+import json
+
+
+def find_linked_data(response, wanted_type) -> {}:
```
iandees, line 4: I think at some point I'd like to move the linked data stuff together into a single class/file and out of here and the GeojsonPointItem. We don't need to do it now, but I'm leaving a note here for myself in the future.

Cj-Malone: We should probably do it sooner than later, just so we don't have to refactor a bunch of spiders. I'm off work tomorrow so I'll have a go.

PR #3742, opened 2022-07-19
Refactor Linked Data Parser

Cj-Malone:
Also added a few spiders, Chef & Brewer, Farmhouse Inns and Go Outdoors.


Cj-Malone:
We should consider setting a default [download_delay](https://docs.scrapy.org/en/latest/topics/settings.html?highlight=delay#download-delay). Maybe 1 second would be polite? Individual spiders could still increase or decrease it.

iandees:
Yea, a default sounds good. There's a max run time for each spider in a global run that might need to be tweaked, but something between 0.75 and 1.0 seems like a pretty polite thing to default to.

locations/spiders/farmhouse_inns.py
```diff
@@ -0,0 +1,19 @@
...
+        "brand_wikidata": "Q105504972",
+        "country": "GB",
+    }
+    sitemap_urls = ["https://www.farmhouseinns.co.uk/xml-sitemap"]
+    sitemap_rules = [
+        (r"https:\/\/www\.farmhouseinns\.co\.uk\/pubs\/([-\w]+)\/([-\w]+)\/$", "parse")
+    ]
+
+    def parse(self, response):
+        yield LinkedDataParser.parse(response, "BarOrPub")
```
iandees, line 19: Nice :+1:

iandees:
Thanks, this looks great.

PR #3744, opened 2022-07-19
Fix Casey's General Store

iandees:


locations/spiders/caseys_general_store.py
```diff
@@ -1,134 +1,28 @@
...
-            }
+        item = LinkedDataParser.parse(response, "ConvenienceStore")
 
-            try:
-                h = self.parse_hours(store["storeOpenHours"])
-                if h:
-                    properties["opening_hours"] = h
-            except:
-                pass
+        item["ref"] = item["website"].split("/")[-1]
```
Cj-Malone, line 26: We could also do a generic fallback ref = url [here](https://github.com/alltheplaces/alltheplaces/blob/a6d70ecd2afb0b1c6b235b12fa0a1d75fb3d0582/locations/linked_data_parser.py#L84) too

Cj-Malone: Yeah, having a specific value is better, but I think we may end up with a bunch of spiders manually setting ref = url, if so handling it here would be easier. But we'll see, maybe most linked data will use branchCode or (at)id so it wont be necessary anyway.

PR #3746, opened 2022-07-21
Fix/panda express update

Romeo-PHILLIPS:
923 -> 2231
Added opening hours


iandees:
Thanks for cleaning this up! In the future, it'd be great to find a way to do this without using the points file. There's a [location directory](https://www.pandaexpress.com/locations) to spider over that would put less load on their servers and go faster.

Cj-Malone:
You can probably do a [CrawlSpider](https://github.com/alltheplaces/alltheplaces/blob/c1bd45953887db04a6b05d8dc5b016d5eb0666e9/locations/spiders/my_dentist_gb.py#L10) on the directory too, so you don't have to worry about xpathing the state and city urls yourself.

PR #3754, opened 2022-07-31
fix Tractor Supply spider

jakewarren:
### Changes
- apply Akamai Ghost bypass from https://github.com/alltheplaces/alltheplaces/pull/3455


iandees:
@jakewarren do you know what else needs to be installed to support HTTP/2 in Scrapy?

jakewarren:
> @jakewarren do you know what else needs to be installed to support HTTP/2 in Scrapy?

I don't, both this spider and the `target_au` spider were working out of the box for me 🤷

jakewarren:
I am going to close this PR and instead attempt to implement it as a sitemap spider, should be significantly faster. Not sure if the HTTP/2 bypass will be needed or not but at least we already have that figured out. 

https://www.tractorsupply.com/sitemap_stores.xml

PR #3757, opened 2022-08-01
Add Matalan, Oak Furnitureland and Optical Express spiders

Cj-Malone:
I can't remember the exact numbers, but the Optical Express opening hours fuckery got us the vast majority of the opening hours. Going forward I wont be putting that much effort into bad opening hours.

### DIFF
--- a/locations/hours.py
+++ b/locations/hours.py
@@ -122,6 +122,9 @@ class OpeningHours(object):
             for rule in rules:
                 days, time_ranges = rule.split(" ", 1)
 
+                if time_ranges.lower() == "closed":
+                    continue
+
                 for time_range in time_ranges.split(","):
                     start_time, end_time = time_range.split("-")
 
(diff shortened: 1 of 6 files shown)


iandees:
They should pay us to fix their opening hours data 😄

Cj-Malone:
I've thought the same before, how often do we come across malformed websites or data? We [compare the data to OSM](https://osm.mathmos.net/chains/), to see if OSM is valid, but I'd love to report broken source data back to them. Even just saying "your lat/lons are shit, here are OSMs".

iandees:
Yea, it would be great to generate a report like Chain Reaction at the end of weekly run and then point people at it...

My experience is that businesses don't generally care enough about their business listings to take action on feedback, but it would be good to know about the errors generally.

Cj-Malone:
Yeah that's my experience too, optimistically I hope they don't care because we aren't talking to the right people. First level customer service at these places answering emails are far from the people creating the website, or the marking people who are trying to advertise there locations.

I don't even try and report errors anymore, maybe a generated report with big red writing may help, but I don't know.

Back to OSM, Chain Reaction is exactly the way I'd marry this data with OSM, and I'd love for it to go global. But I think @rjw62 has said it's not in his scope.

PR #3758, opened 2022-08-02
Try changing the homedepot spider to sitemap + linked data

iandees:


jakewarren:
@iandees unfortunately Home Depot doesn't store the coordinates in the JSON-LD data so the lat/lon isn't being pulled into the produced GeoJSON. 

it also doesn't look like the store hours are being parsed but that is much lower priority.

Cj-Malone:
I hit a site that didn't have the coords in the ld, but did in the html. I added an extra [if statement](https://github.com/alltheplaces/alltheplaces/blob/d27ca11c7bfc2dca66018414fdd20eb7837595cf/locations/spiders/optical_express.py#L67) so hopefully when the site is updated, the spider doesn't break. I hope the new ld parser becomes flexible enough so that the spiders can survive minor website changes.

PR #3765, opened 2022-08-04
convert Tractor Supply to sitemap spider

jakewarren:
### Changes
- convert to a sitemap spider to increase scraping speed
- apply Akamai Ghost bypass from https://github.com/alltheplaces/alltheplaces/pull/3455


jakewarren:
So while installing `h2` I noticed that pipenv was upgrading all of the dependencies. Trying to avoid that I then tried `pipenv install --keep-outdated h2` but that switch seemed to have no effect. 

Next I tried just manually adding the h2 hashes to the Pipfile.lock file but got a `Your Pipfile.lock (8f22f2) is out of date. Expected: (f58731).` error. I'm not all that familiar with pipenv so I just committed the entire changeset for Pipfile.lock. I don't know if that is acceptable or not.

jleedev:
- Correct invocation is `pipenv install h2 --keep-outdated`, although the difference is very minor and there's little harm in allowing pipenv to update the whole lockfile.
- Rather than installing `h2` directly, Scrapy says ([1]) to install `Twisted[http2]`.
- It bears repeating that Scrapy's HTTP/2 support is still experimental. In particular, it seems that connection errors cause exceptions that lead Scrapy to abandon requests entirely rather than retry them. So this does produce some results, but it would be ideal to add a comment or file an issue explaining the situation.

[1]: https://docs.scrapy.org/en/latest/topics/settings.html?highlight=http2#download-handlers-base

Example of error:

image

jleedev:
To be more specific, `target_au` works because it's only making like 10 requests, but for a longer running job, the inevitable connection failures can cause an arbitrary number of requests to be abandoned.

jakewarren:
Thanks for the research! So sounds like we need to shelve this until better support is added to scrapy or we find another bypass. Unfortunately I haven't had any luck by trying variations of headers and user agents.

PR #3771, opened 2022-08-08
Fix fcbanking

jleedev:
- Somebody decided that the //-style comments in a json document should
  now have \r line terminators
- Modernize code

### DIFF
--- a/locations/spiders/fcbanking.py
+++ b/locations/spiders/fcbanking.py
@@ -1,11 +1,13 @@
 # -*- coding: utf-8 -*-
 import json
 import re
+import urllib.parse
 
 from scrapy.spiders import SitemapSpider
 
-from locations.linked_data_parser import LinkedDataParser
+from locations.hours import OpeningHours
 from locations.items import GeojsonPointItem
+from locations.linked_data_parser import LinkedDataParser
 
 
 class FcBankingSpider(SitemapSpider):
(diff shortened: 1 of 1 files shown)


locations/spiders/fcbanking.py
```diff
@@ -2,50 +2,32 @@
...
-        }
 
-        return GeojsonPointItem(**properties)
+        ldjson = response.xpath('//script[@type="application/ld+json"]/text()').get()
+        ldjson = ldjson.replace("\r", "\n")
+        data = json.loads(re.sub(r"^//.*$", "", ldjson, flags=re.M))
+        item = LinkedDataParser.parse_ld(data)
+        item["lat"] = lat
+        item["lon"] = lon
+        item["ref"] = item["website"] = response.url
```
Cj-Malone, line 31: I now think we should try and trim websites before using them as a ref. eg instead of `https://www.fcbanking.com/branch-locations/ohio/ashtabula/park-avenue-office/4200-park-ave/` we just use `ohio/ashtabula/park-avenue-office/4200-park-ave/`

That way, if they change the url structure to `https://www.fcbanking.com/branchs/ohio/ashtabula/park-avenue-office/4200-park-ave/` in the future we'd keep the same ref. Ideally they'd publish there own id though.

locations/spiders/fcbanking.py
```diff
@@ -2,50 +2,32 @@
...
+        item["opening_hours"] = "; ".join(data["openingHours"])
```
Cj-Malone, line 32: There opening hours look structured, just with a am/pm format. We should be able to parse them fairly easily.

Cj-Malone:
Do a .replace to get proper AM/PM from a.m./p.m., then '%p' should read it

PR #3772, opened 2022-08-08
Fix Walmart

Cj-Malone:
Move to CrawlSpider. The sitemap has bad urls and old stores that soft 404.

```patch
+https://www.walmart.com/store/3081/sacramento-ca
-https://www.walmart.com/store/3081-sacramento-ca
```

I don't know if this will work without being banned, I ran it with a 2.5 delay and got 1,384 stores before being blocked.

### DIFF
--- a/locations/spiders/walmart.py
+++ b/locations/spiders/walmart.py
@@ -1,19 +1,29 @@
 # -*- coding: utf-8 -*-
 import json
 import re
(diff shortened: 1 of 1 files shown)


Cj-Malone:
It takes a while to get started because it pages through states and cities before doing any stores.

PR #3777, opened 2022-08-10
Add costcutter_gb (1472 stores)

mjoe999:
So costcutter_gb uses ld+json but lists all the brands it stocks rather than itself, it does so in a way causing the library parser to error (exception). You see how I have subverted this is the spider code.

However, @Cj-Malone it does beg the question as to why we would decode the brand data? The wikidata code / brand name that by convention we put in the spider is likely the correct brand reference. My reading of the code paths is that picking up the brand from ld+json will stop it being copied from the spider "item_attributes".

### DIFF
--- a/locations/linked_data_parser.py
+++ b/locations/linked_data_parser.py
@@ -90,15 +90,6 @@ class LinkedDataParser(object):
         if item["ref"] is None:
             item["ref"] = ld.get("@id")
 
-        if ld.get("brand"):
-            if isinstance(ld["brand"], str):
-                item["brand"] = ld["brand"]
-            elif (
-                ld["brand"].get("@type") == "Brand"
-                or ld["brand"].get("@type") == "Organization"
-            ):
-                item["brand"] = ld["brand"].get("name")
-
         return item
 
     @staticmethod
--- a/locations/spiders/costcutter_gb.py
+++ b/locations/spiders/costcutter_gb.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+import scrapy
+from locations.linked_data_parser import LinkedDataParser
+
+
+class CostcutterGBSpider(scrapy.spiders.SitemapSpider):
+    name = "costcutter_gb"
+    item_attributes = {
+        "brand": "Costcutter",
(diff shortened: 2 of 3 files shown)


Cj-Malone:
I like the idea of trusting the brand string to them, but in reality... it's not that great.

We should either, fix the crash by throwing away brand arrays (or using the first time), or by removing the brand parsing. Like I said, I like the idea of trusting upstream to get there brand right, but I've not got a strong opinion.

mjoe999:
My vote:

- remove brand parse code from linked_data_parser.py
- add a log.warn to parse code for match on "closed" in name .... helps  spider writers
- I will search for and remove "closed" name items in costcutter_gb.py

If the log.warn had been in prior to writing the spider then the hope would be that I had noticed and taken action. So thanks CJ for your eagle-eyes.

If others agree I will make the change in this PR.

iandees:
I would prefer that we don't filter on "(closed)" in the linked data parser or globally. If we filter that out, it should happen in the spider itself, since the way they mark a place as closed is specific to the store locator we're scraping.

Similarly, if the store locator does weird things with brand, we should override the default behavior like Joe is doing here. We should probably add a comment about why we're doing that, though.

mjoe999:
Chatted to @Cj-Malone , dropped brand parsing from LD parser. Leave brand specification to spiders for now as otherwise unclear side affects with item["brand"] vs <spider>.item_attributes["brand"]

Cj-Malone:
> is specific to the store locator we're scraping

While it is, I think having "closed" in the name will be relatively common. But we can re visit it if we see a pattern emerging.

Costcutter is unfortunately quite bad at this. Not only do they still have the old stores listed, with ld, but also only _some_ of the closed ones have "closed" in the name. I then thought we could filter any out without `opening_hours`, but it isn't perfect, there are some open shops without opening hours. I don't think we can be perfect with Costcutter.

> add a log.warn to parse code for match on "closed" in name .... helps spider writers

This feels reasonable. But it does raise the question of where this code should go, if we just put it in the ld parser only that benefits. Is a pipeline the right place?

PR #3786, opened 2022-08-11
Add riversland_gb spider (~250 clothes stores)

mjoe999:


locations/spiders/riverisland_gb.py
```diff
@@ -0,0 +1,38 @@
...
+        )
+    ]
+
+    def parse_func(self, response):
+        ld = LinkedDataParser.find_linked_data(response, "ClothingStore")
+        if ld:
+            item = LinkedDataParser.parse_ld(ld)
+            item["lat"] = float(ld["latitude"])
+            item["lon"] = float(ld["longitude"])
+            item["ref"] = response.url
```
Cj-Malone, line 30: We should just use the unique part of the URL, so in this case `response.url.split("/")[-1]`

PR #3791, opened 2022-08-12
Add jdsports_gb (400 sports goods stores)

mjoe999:


locations/spiders/jdsports_gb.py
```diff
@@ -0,0 +1,28 @@
...
+    ]
+    download_delay = 0.2
+
+    def parse_store(self, response):
+        if "-soon" in response.url:
+            self.logger.warning("ignoring store opening soon %s", response.url)
+        else:
+            item = LinkedDataParser.parse(response, "Store")
+            if item:
+                item["ref"] = response.url
```
iandees, line 27: Just use the unique part of the URL:

```suggestion
                item["ref"] = response.url.strip('/').split('/')[-1]
```

locations/spiders/jdsports_gb.py
```diff
@@ -0,0 +1,28 @@
...
+    start_urls = ["https://www.jdsports.co.uk/store-locator/all-stores/"]
+    rules = [
+        Rule(
+            LinkExtractor(allow="store-locator/"), callback="parse_store", follow=False
+        )
+    ]
+    download_delay = 0.2
+
+    def parse_store(self, response):
+        if "-soon" in response.url:
```
Cj-Malone, line 22: Can this be done as a 'deny' in the LinkExtractor? That way we'd skip the http request.

locations/spiders/jdsports_gb.py
```diff
@@ -0,0 +1,28 @@
...
+            self.logger.warning("ignoring store opening soon %s", response.url)
+        else:
+            item = LinkedDataParser.parse(response, "Store")
+            if item:
+                item["ref"] = response.url.strip("/").split("/")[-1]
+                yield item
```
Cj-Malone, line 28: Unimportant, but since this function only ever returns 1 item, this doesn't need to be a 'yield'

Cj-Malone (commented):
For future us, the ld data here gives us `"country": "UNITED KINGDOM"`, we should probably ensure `country` is an ISO country code. In this case "GB".

locations/spiders/jdsports_gb.py
mjoe999: There's just 2 of them. It's clearer in code I think.

Cj-Malone (approved):

PR #3793, opened 2022-08-12
Update PR workflow to use latest ubuntu image.

### LINKED ISSUE #3792: Upgrade workflow from ubuntu-18.04
“The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead. For more details see https://github.com/actions/virtual-environments/issues/6002 ”

https://github.com/alltheplaces/alltheplaces/blob/7cad1a82c4c107ac08ccfd68f26dcd9f2e7a5da3/.github/workflows/spider-pr.yaml#L11

The other workflows use `ubuntu-latest`.

iandees:
Fixes #3792

### DIFF
--- a/.github/workflows/spider-pr.yaml
+++ b/.github/workflows/spider-pr.yaml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   test-spider:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     env:
       AWS_KEY: QUtJQTRLMllUQzIzRzVZRkxENkg=
       AWS_SEC: b1RmNXNxc0VSNlNKdFNhbUdlcjJwZHR6YXQwNXpRRDRPZUtoVkpPSA==


.github/workflows/spider-pr.yaml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   test-spider:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     env:
       AWS_KEY: QUtJQTRLMllUQzIzRzVZRkxENkg=
       AWS_SEC: b1RmNXNxc0VSNlNKdFNhbUdlcjJwZHR6YXQwNXpRRDRPZUtoVkpPSA==
```
Cj-Malone, line 14: Is this an actual secret?

iandees: Yes, it's an encoded secret. I do this so that Github Actions triggered by users submitting pull requests from their fork will still upload to S3. Github doesn't export secrets in these situations, and that's why the commenting fails.

Issue #3676, opened 2022-06-08
Vision Express


Cj-Malone:
https://www.visionexpress.com/opticians

Cj-Malone:
Added in #3799

PR #3799, opened 2022-08-16
Add visionexpress_gb spider (544 optician stores)

mjoe999:
Required some fixes to the Google URL parse code.

@Cj-Malone : fixes above are for when query parameter required was first in the param list on the link!


locations/spiders/visionexpress_gb.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def extract_google_position(item, response):
+    for link in response.xpath("//img/@src").extract():
+        if link.startswith("https://maps.googleapis.com/maps/api/staticmap"):
+            item["lat"], item["lon"] = url_to_coords(link)
+            return
+
+
+def set_located_in(item, located_in, located_in_wikidata):
```
Cj-Malone, line 14: This is now in 2 spiders, should it be moved to `items.py`?

Cj-Malone:
I've checked the new `url_to_coords`, but assuming you ran `pytest` and everything still works, LGTM.

locations/spiders/visionexpress_gb.py
mjoe999: It's on my list but not for this commit.

mjoe999:
Fixes #3676

locations/spiders/visionexpress_gb.py
iandees: I still don't understand this pattern. It doesn't make sense to extract two assignments into a function. Is there extra functionality you're thinking would go in here in the future?

I could see an argument to be made for better readability, but the misdirection makes it harder to follow than putting the assignments in line in my opinion.

PR #3800, opened 2022-08-16
Add monsoon_gb (~100 clothing stores)

mjoe999:


locations/spiders/monsoon_gb.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def parse(self, response):
+        item = LinkedDataParser.parse(response, "ClothingStore")
+        if item:
+            item["website"] = response.url
+            item.update(self.MONSOON)
+            yield item
+            if "Accessorize" in item["name"]:
+                # There is a co-located Accessorize store, this is how they work!
+                item = LinkedDataParser.parse(response, "ClothingStore")
```
Cj-Malone, line 26: Do we have to re create `item` and set the website, or can we use the original?

also we probably want `located_in` on this item.

mjoe999: You would have to "copy" the original, I have seen some of this behaviour. "deepcopy" I think is more safe / future proof but I am not an expert.

I will set "located_in" but of course that is not quite correct. It should be "located_with" as they many times have two small stores side by side. As you can see I made the comment in the code. I often wonder why people these days write so few comments .... perhaps I start to understand!

mjoe999:
Forget about this one.

PR #3802, opened 2022-08-16
Add spider for Fandango

jleedev:


mjoe999:
https://www.wikidata.org/wiki/Q5433722 good to link?

mjoe999:
Got it!

jleedev:
This spider produces 2,767 items with 108 distinct values for "brand". The latest run contains 1,112 items for Regal and AMC combined. So that's 1655 items and 106 spiders that would not be duplicated by this; half of these brands have less than 5 items.

jleedev:
Looking a bit more closely, almost all of these coordinates are really quite far off, and I think this quality issue is a reflection of this dataset as an aggregator. Makes sense to close.

PR #3803, opened 2022-08-16
Add generic parser

### LINKED ISSUE #3571: Poundland
[Store finder](https://www.poundland.co.uk/store-finder/)
[Store page](https://www.poundland.co.uk/store-finder/store_page/view/id/456/) includes a interesting JSON blob

[API](https://www.poundland.co.uk/rest/poundland/V1/locator) unsure how to call this, it returns JSON or XML.

### LINKED ISSUE #3716: Pret
I've previously scraped [this website](https://locations.pret.co.uk/), but the data is awful, they have a bunch of fake stores. This [API](https://api1.pret.com/v1/shops) is presumably better.

Cj-Malone:
The idea here is that the vast majority of the keys are the same. Just like the Linked Data parser, it isn't perfect, there will often have to be fixes before or after in spider code. I've written Poundland and Pret spiders while writing the `DictParser`. They both have clean up, Poundland has a better `ref` then it's `id`, and Pret has bad street addresses.

The spider code also handled extra data + opening hours. I don't think it's feasible to have a generic opening hours parser.

Fixes #3571, Fixes #3716

### DIFF
--- a/locations/dict_parser.py
+++ b/locations/dict_parser.py
@@ -39,13 +39,18 @@ class DictParser(object):
             item["street_address"] = ", ".join(item["street_address"])
 
         item["city"] = DictParser.get_first_key(address, ["city", "town"])
-        item["state"] = DictParser.get_first_key(address, ["state"])
+        item["state"] = DictParser.get_first_key(address, ["state", "region"])
         item["postcode"] = DictParser.get_first_key(
             address, ["postCode", "post_code", "postalCode"]
         )
         item["country"] = DictParser.get_first_key(address, ["country", "countryCode"])
 
-        item["phone"] = DictParser.get_first_key(obj, ["phone", "telephone", "tel"])
+        contact = DictParser.get_first_key(obj, ["contact"])
+
+        if not contact or not isinstance(contact, dict):
+            contact = obj
+
+        item["phone"] = DictParser.get_first_key(contact, ["phone", "telephone", "tel"])
 
         item["ref"] = DictParser.get_first_key(obj, ["ref", "id", "store_id", "shopNumber", "slug"])
 
--- a/locations/dict_parser.py
+++ b/locations/dict_parser.py
@@ -34,11 +34,8 @@ class DictParser(object):
         )
         item["street"] = DictParser.get_first_key(address, ["street", "streetName"])
         item["street_address"] = DictParser.get_first_key(
-            address, ["streetAddress", "street_address", "line1", "line"]
+            address, ["streetAddress", "street_address", "line1"]
         )
-
-        if isinstance(item["street_address"], list):
-            item["street_address"] = ", ".join(item["street_address"])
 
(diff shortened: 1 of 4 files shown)


locations/dict_parser.py
```diff
@@ -0,0 +1,71 @@
+from locations.items import GeojsonPointItem
+
+
+class DictParser(object):
```
iandees, line 4: I'm ok with including this, but I worry that we'll be endlessly updating/adding if-checks every time we want to use it with a new spider. If that ends up happening, I'd rather see us keep that tedious work in the spider code instead of trying to modify this parser.

jleedev: +1, there's clear value in handling standards-compliant schema markup, but less so in handling e.g. streetAddress, street_address, line1, line, and whoops we also need locationStreet1, street1, addressLine1, address1_line1, and addr1.

locations/spiders/poundland.py
```diff
@@ -0,0 +1,61 @@
+import scrapy
+
+from locations.dict_parser import DictParser
+from locations.hours import OpeningHours
+
+
+class PoundlandSpider(scrapy.Spider):
+    name = "poundland"
+    poundland = {"brand": "Poundland", "brand_wikidata": "Q1434528"}
```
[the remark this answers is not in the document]
Cj-Malone: I don't know exactly what you mean because the only think bing `.copy()`ed is a `GeojsonPointItem`, not brand stuff.

I can drop the `pep` and `poundland` fields though and just do it in `parse` if you'd prefer.

locations/dict_parser.py
Cj-Malone: I don't think `DictParser` should be updated with every new spider, edge cases should be fixed in spider code, not here. We'll probably workout some bumps early on like with the `LinkedDataParser`, but after that it should stay pretty static.

`line` -> `street_address` is clearly a stretch because I was writing the Poundland spider at the same time. I'll remove that. If I was being cleaver I should have looked through existing spiders and pulled out the most common fields, but instead I just did from memory.

I think we could also do some fancy things with camelCase vs snake_case in `get_first_key` like I'm already doing with upper/lower case to avoid the list being repetitive.

locations/spiders/poundland.py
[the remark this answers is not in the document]
iandees: > I can drop the pep and poundland fields though and just do it in parse if you'd prefer.

Yes please, that would help with readability.

locations/dict_parser.py
mjoe999: I agree that street_address and it's many friends is a stretch, if you see my earlier attempt at similar:

https://github.com/alltheplaces/alltheplaces/pull/3667/files#diff-4e15618ba17581552d29b623a4d025d66598a7b51be5ae1bede253c8a9dae551

then you will see that it does not attempt to so that field. Several hundred spiders say that there is good mileage in the other fields though and CJ's comments above still apply.

PR #3806, opened 2022-08-17
Update 99centonly to use LinkedData

iandees:


locations/linked_data_parser.py
```diff
@@ -74,6 +74,7 @@ def parse_ld(ld) -> GeojsonPointItem:
                 item["country"] = addr.get("addressCountry") or addr.get(
                     "addresscountry"
                 )
+                item["phone"] = addr.get("telephone")
```
Cj-Malone, line 77: I kinda think this is one of the errors that should be fixed in spider, not ld. But I have done the exact fix in a spider already https://github.com/alltheplaces/alltheplaces/blob/d27ca11c7bfc2dca66018414fdd20eb7837595cf/locations/spiders/optical_express.py#L65 and [["address"]["telephone"]](https://github.com/alltheplaces/alltheplaces/search?q=%5B%22address%22%5D%5B%22telephone%22%5D) seems fairly common, so no objection from me.

iandees: Yea, I was waffling on this – especially after I keep bringing it up. It seems fairly common like you point out, but definitely non-obvious and not "standards compliant".

PR #3808, opened 2022-08-19
Add a Microdata parser.

jleedev:
To address:

- [ ] Handle more of the w3c test suite, an itemprop can be absolute and itemtype can be empty. Not sure if anything needs this.
- [ ] Ensure we're sufficiently sensible in the face of itemref.
- [ ] "find the properties of the item" can produce a "microdata error", while "get the object for an item" can produce the string "ERROR". Is there a difference? `<test0085>` produces the latter. The w3c test suite doesn't seem to produce the former.
- [ ] It's probably fine to leave all values untyped?

Converts the HTML5 Microdata directly to JSON-LD in a similar format to (nearly) all of the JSON-LD seen in the wild.

Steps of this process:

1. Run the algorithm to extract the microdata given at <https://html.spec.whatwg.org/multipage/microdata.html>, _not_ the one at <https://www.w3.org/TR/microdata-rdf/>. [Note]

2. Convert the HTML Microdata JSON output to a JSON-LD format.

   While the algorithm to extract the microdata is implemented more or less to spec, this is entirely ad-hoc. If the root item seems to be in the schema.org namespace, produce a correct-looking output.
   Flatten lists of property values if there is only one element.
   Duplicated string values are removed; this is a side effect of what would happen if you actually ran this through an RDF processor.

3. Take the JSON-LD produced and hang it off of the document, then invoke the Linked Data parser as usual. Has the advantage that the special cases here can be reused, such as "phone a property of address", or "multiple addresses for some reason", and so on.

The default entry point I expect will be this one:

    MicrodataParser.convert_to_json_ld(response)
    LinkedDataParser.parse(response, "LocalBusiness")
    # etc.

but the class exposes methods to stop at either of the steps, whether to do more preprocessing, or if that's desired for the output.

Include three spiders to illustrate some possible usage patterns:

- Fix fifththirdbank
- Fix sunglasshut (address and geo are included twice; the
  LinkedDataParser already knows to just look at the first address.)
- Add spider for northwest_bank (insert a hack in the markup to recover
  two different types of openingHours mixed together)

[Note]

Why? Microdata to RDF refers to the HTML Microdata specification _anyway_, for instance when getting the properties of the item, so I found it less reliable as a specification. I think in particular it was unclear when to recursively produce an item when an itemprop has an itemscope. And the output includes typed values and triples, both of which are unnecessary here: Typing all values as strings is sufficient, and the tree structure produced by the HTML Microdata algorithm is exactly what we need, not the graph.

If you want to invoke an RDF library, dear reader, then you need to correctly distinguish between strings and IRI's, be prepared to prevent the library from crawling arbitrary JSON-LD context documents it finds referenced somewhere, and you need to learn how to write a JSON-LD frame in order to produce a JSON-LD document that looks generally familiar. And distinguish between string values with and without a language code. Maybe you could wire up the rest of the application to the rdflib Python API, or write everything in SPARQL. Let me know if you find that fun.

### DIFF
--- a/locations/spiders/sunglass_hut.py
+++ b/locations/spiders/sunglass_hut.py
@@ -1,54 +1,25 @@
-import scrapy
+# -*- coding: utf-8 -*-
 import re
+
+import scrapy
+
 from locations.items import GeojsonPointItem
+from locations.linked_data_parser import LinkedDataParser
+from locations.microdata_parser import MicrodataParser
 
 
-class SunglassHutSpider(scrapy.Spider):
+class SunglassHutSpider(scrapy.spiders.SitemapSpider):
     name = "sunglass_hut"
     item_attributes = {"brand": "Sunglass Hut", "brand_wikidata": "Q136311"}
     allowed_domains = ["stores.sunglasshut.com"]
-    start_urls = ("https://stores.sunglasshut.com/",)
-
-    def parse_opening_hours(self, opening_hours):
-        return ";".join(opening_hours)
+    sitemap_urls = [
+        "https://stores.sunglasshut.com/robots.txt",
+    ]
 
     def parse(self, response):
-        urls = response.xpath(
-            '//a[@class="c-directory-list-content-item-link" or @class="c-location-grid-item-link"]/@href'
-        ).extract()
-        # If cannot find 'c-directory-list-content-item-link' or 'c-location-grid-item-link' then this is a store page
-        if len(urls) == 0:
-            properties = {
-                "addr_full": response.xpath(
-                    'normalize-space(//div[@class="c-location-info"]//span[@itemprop="streetAddress"]/text())'
-                ).extract_first(),
-                "city": response.xpath(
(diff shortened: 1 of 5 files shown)


iandees:
This is great. Thanks for adding this. Did you want to keep fixing spiders in here or should I merge this?

jleedev:
Let's merge it!

The code could use some polish, but I've documented the edge cases that I'm at least aware of, and in practice it handles the cases that actually matter.

locations/spiders/sunglass_hut.py
```diff
@@ -1,54 +1,25 @@
...
-                ),
-            }
-            yield GeojsonPointItem(**properties)
-        else:
-            for path in urls:
-                yield scrapy.Request(response.urljoin(path), callback=self.parse)
+        MicrodataParser.convert_to_json_ld(response.selector)
+        ld = LinkedDataParser.find_linked_data(response, "LocalBusiness")
+        if ld is not None:
+            ld["geo"] = ld["geo"][0]
```
Cj-Malone, line 23: I think we can move this into the LinkedData parser, I think it's going to happen a lot now.

PR #3810, opened 2022-08-19
Add bestwestern spider (~3,500 hotels)

mjoe999:
Required a few tweaks to the DictParser. Prefer country code over country.


locations/spiders/bestwestern.py
```diff
@@ -0,0 +1,65 @@
...
+            my_brand = None
+            for brand in self.brands:
+                if my_name.startswith(brand["brand"].lower()):
+                    my_brand = brand
+                    break
+            if not my_brand:
+                # For the most part these are hotels not owned by BW, but part of a "collection"
+                return
+            item = DictParser.parse(summary)
+            set_brand(item, my_brand)
```
jleedev, line 50:
```suggestion
            item.update(my_brand)
```

mjoe999: Totally fair comment but I would like to leave as is for now as "from_wikidata()" and "set_brand()" are things I would like to see a more structural "fix" for and when I fix one occurrence I would fix all. grep "from_wikidata()" at least lets me find them readily at a later date.

iandees: I'm having a hard time seeing the utility of these setter functions. What's the benefit over just including the data in `item_attributes` or setting them as part of the `parse()` function?

mjoe999: When there are multiple brands then item_attributes does not cut it. The only benefit for now is that I can port my spiders in easily and then we can apply a more structural clean up. Refactoring method calls is far easier than unpicking a set of field assignments.

locations/spiders/bestwestern.py
```diff
@@ -0,0 +1,65 @@
...
+            json_data = json.loads(html.unescape(data_hotel_details))
+            summary = json_data["summary"]
+            my_name = summary["name"].lower()
+            my_brand = None
+            for brand in self.brands:
+                if my_name.startswith(brand["brand"].lower()):
+                    my_brand = brand
+                    break
+            if not my_brand:
+                # For the most part these are hotels not owned by BW, but part of a "collection"
```
jleedev, line 47: Best Western doesn't own any hotels, these are all franchisees.

Even if some boutique hotels might additionally have their own websites, I suspect that BW serves as the official booking site, and it makes sense to include them.

locations/spiders/bestwestern.py
iandees: If item_attributes doesn't cut it, then we should be setting the details in the parse function. As it stands, the setter function is obfuscating the field assignments and makes it harder to follow.

locations/spiders/bestwestern.py
mjoe999: If it quacks like a Best Western then I produce the POI from the spider i.e. does the hotel have a great big board outside saying "Best Western"! The "collections" hotels are available standalone in this case:
https://www.tripadvisor.co.uk/Hotel_Review-g189541-d227260-Reviews-The_Huxley_Copenhagen-Copenhagen_Zealand.html
They do not have a board outside.

locations/spiders/bestwestern.py
mjoe999: Let me ask the question then. This PR #3667 was too much too soon. But hey now you have LD, microdata and "dict" parsers. Would you now be ready for something like:

https://github.com/alltheplaces/alltheplaces/blob/e993696dc443d0a5de4328993193ad16cc621f90/locations/brands.py

... coming in as part of this PR. I tjink brand/wikidata are central concepts. They deserve a bit of first class support which will mean less typing and I hope clearer code all round.

iandees: My feedback there is the same as on this PR. I don't see the benefit of extracting brand and brand_wikidata into separate objects or functions. The spider parse functions should be where that happens. It's easier to read and update when it's all in the one spider python file.

mjoe999:
Ok, another try at this. Chopped down to be like my other accepted PRs.

PR #3811, opened 2022-08-19
Add spider for National Tire and Battery (ntb)

jleedev:


Cj-Malone:
Yeah I commented on the wrong PR there. We need to remove the broken duplicate spider.

PR #3812, opened 2022-08-19
Fix biglots

jleedev:


Cj-Malone:
Is this different from the `national_tire_battery` spider?

jleedev:
> Is this different from the `national_tire_battery` spider?

Assume you meant that on #3811; nope, I just didn't see it.

(Argument in favor of "spider `name` should be the domain name.)

PR #3814, opened 2022-08-19
Add bodyshop spider (~930 global stores)

mjoe999:


locations/spiders/thebodyshop.py
```diff
@@ -0,0 +1,56 @@
...
+                    "https://api.thebodyshop.com/rest/v2/thebodyshop-{}/stores/{}".format(
+                        self.parse_pages[key], store_id
+                    ),
+                    headers={"Accept": "application/json"},
+                    callback=self.parse_json,
+                    cb_kwargs=dict(html_response=response),
+                )
+
+    def parse_json(self, response, html_response):
+        store = json.loads(response.text)
```
jleedev, line 48:
```suggestion
        store = response.json()
```

mjoe999: Done

locations/spiders/thebodyshop.py
```diff
@@ -0,0 +1,56 @@
...
+        store["location"] = store["geoPoint"]
+        item = DictParser.parse(store)
+        item["ref"] = store["name"]
+        item["name"] = store["displayName"]
+        item["website"] = html_response.url
+        item["street_address"] = store["address"].get("line1")
```
Cj-Malone, line 54: Address/line1 should already be parsed as street_address via the DictParser.

Cj-Malone:
Is this spider pulling all the "store-details" html pages, then filtering some away, then fetching the json?

We should definitely move the filtering before pulling the html page, and we can probably skip the html page entirely.

mjoe999:
The pages it does not look at are a very small number. I have to get the page to get the ID to get the JSON.

mjoe999:
Ok everyone you are correct. It could be in a filter function for the site map. Can I say though that a great deal extra is possible when you have a page "response" in your hands. This is future code.

mjoe999:
I've tried to "optimise". The code is far too involved given the way the sitemap spider works as far as I can see. This is good to go as it stands.

mjoe999:
OK, I've fixed it so that it does half the HTTP requests.

PR #3815, opened 2022-08-19
Don't crash on missing telephone property

jleedev:


locations/linked_data_parser.py
```diff
@@ -76,7 +76,8 @@ def parse_ld(ld) -> GeojsonPointItem:
                 )
                 item["phone"] = addr.get("telephone")
 
-        item["phone"] = ld.get("telephone").strip()
+        if "telephone" in ld:
+            item["phone"] = ld["telephone"].strip()
```
iandees, line 80: Python recently introduced new syntax for this sort of thing:

```suggestion
        if tele := ld.get("telephone"):
            item["phone"] = tele.strip()
```

Cj-Malone:
We should probably do that on everything.

PR #3825, opened 2022-08-22
Add pytest workflow

Cj-Malone:
So, I've actually no idea what I'm doing here, with Pipfile or GH workflows.

The idea is that any PR that changes core code should have the tests ran against them.


iandees:
Looks good to me.

PR #3828, opened 2022-08-23
Move centroid points support to library code, modify some spiders to use

mjoe999:
Move a method into library code.

Picked a few first spiders to use it. Both have had fixes applied to existing small problems .... that's how it goes .... you will see more White Castle locations slide in as they say.

### DIFF
--- a/locations/geo.py
+++ b/locations/geo.py
@@ -1,3 +1,4 @@
+import csv
 import math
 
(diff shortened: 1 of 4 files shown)


iandees:
I'm going to let the errors bubble up. It will only stop the spider when running on Github during PRs. The weekly run does not stop when an error occurs.

PR #3829, opened 2022-08-23
Fix pytest

Cj-Malone:
Like I said last time, don't really know what I'm doing. But hopefully this fixes the workflow.


Cj-Malone:
Thank you

PR #3833, opened 2022-08-24
Fixed the Aldi_Nord_De &  Rossmann_De

Ebubeker:
I know third party api isn't a good choice but there is no other way. The website code is in javascript so I cannot scrape any data from it. The only way is through the api. At least it can do the job for some time. This is all I could do.


Ebubeker:
I also fixed a small issue in the rossmann_de scraper

Romeo-PHILLIPS:
API seems to also include opening hours and phone number: https://uberall.com/api/storefinders/ALDINORDDE_UimhY3MWJaxhjK9QdZo3Qa4chq1MAu/locations/all

Romeo-PHILLIPS:
Also I don't know if its an issue that their data is also licensed (https://uberall.com/images/webpages/tcs/20220908_UB-GmbH-TC-EN_web-partner.pdf), so it might be an issue to convert it into Creative Commons’ CC-0?

PR #3835, opened 2022-08-24
Add Coop spiders

Cj-Malone:


Cj-Malone:
`midcounties_cooperative` and `scotmid` are failing because they need a DictParser fix from #3831 or #3810.

PR #3845, opened 2022-08-26
Fix Carrefour Argentina, it's a lot less beefy!

mjoe999:
Old spider was not returning results.


locations/dict_parser.py
```diff
@@ -34,7 +34,14 @@ def parse(obj) -> GeojsonPointItem:
...
         item["street"] = DictParser.get_first_key(address, ["street", "streetName"])
         item["street_address"] = DictParser.get_first_key(
-            address, ["streetAddress", "street_address", "addressLine1", "line1"]
+            address,
+            [
+                "streetAddress",
+                "street_address",
+                "addressLine1",
+                "line1",
+                "addressLineOne",
```
iandees, line 43: Jeez how many ways can people write this?! 😄

Cj-Malone: A never ending battle.

Also ["address_1"](https://github.com/alltheplaces/alltheplaces/blob/1747b4b4acf78478b8b5823c02b46ca125c99064/locations/spiders/festival_foods.py#L28), but I don't think it'll be that common.

PR #3853, opened 2022-08-26
Fix ihg_hotels (~6000 hotels)

mjoe999:
I think the microdata parser needs Python > 3.8.10. One brand within IHG has microdata. I will move forward Python next week when I upgrade Ubuntu. Take this in and I will catch the TODO soon enough?


mjoe999:
There is no TODO left on this PR now.

PR #3856, opened 2022-08-28
Fix Walmart Spider

Cj-Malone:
This moves the Walmart Spider to `__NEXT_DATA__` data since the old `window.__WML_REDUX_INITIAL_STATE__` is gone.

GH will fail because CrawlSpider will crawl the directory before doing the store pages. I've not ran it against the whole of the US, but it works on Alabama.

### DIFF
--- a/locations/spiders/walmart.py
+++ b/locations/spiders/walmart.py
@@ -1,8 +1,9 @@
 # -*- coding: utf-8 -*-
 import json
-import re
 
-from locations.items import GeojsonPointItem
+from locations.dict_parser import DictParser
+from locations.hours import OpeningHours
+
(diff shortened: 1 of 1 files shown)


jleedev:
Could try setting these in `custom_settings` https://docs.scrapy.org/en/latest/faq.html#does-scrapy-crawl-in-breadth-first-or-depth-first-order

Cj-Malone:
Probably better as an actual setting so all spiders benefit.

Cj-Malone:
I can't get the desired results with the settings, and TBH I don't think it's a big deal. Running in GH actions isn't a normal situation for a spider.

I ran this with `download_delay = 3` and got blocked before it finished, hopefully 5 gets us through.

PR #3866, opened 2022-08-29
Fix spar_gb, causes a small collateral stir in "geo" functions.

mjoe999:
So SPAR GB which I recently converted to "sitemap" decided to change their back-end. Such is life.
So back to the old approach except this time use a more structural approach to getting UK outward postal codes.
Grab a public data file which also includes lat, lon etc of said postcodes.
Add a wrapper function in geo.py Also, found an eequivalent for the US, see new README.md.
The US function was tested with porting "vetco" to it. It works (a few more stores in fact!).
Later can port the other US ones and drop "us_zcta.csv"!
germany_centroids_80km_radius_country.csv was also dropped, the description poorly described the contents!
norma_de was the only one using it. Ported to use centroids and now finds more stores!


mjoe999:
dict_parser changes removed, hope that is clearer

locations/searchable_points/postcodes/README.md
```diff
@@ -0,0 +1,18 @@
...
+## Postal code data
+
+Postal code data can be very useful for driving store locator query
+interfaces which require discrete area parameters. These could be
+postal codes or lat/lon.
+
+There are various open data releases of postal code data. Available
+here:
+
+**uszips.csv** "BASIC" download from [simplemaps](https://simplemaps.com/data/us-zips)
```
iandees, line 11: This file is licensed CC-BY 4.0. We would need to provide attribution somewhere, maybe in this code? I haven't had a chance to dig into what would be required.

mjoe999: I had noted that an added the README.md in the same directory. I will add more links in the appropriate code fragment.

mjoe999: Added more. I believe we were good before .... but even better now.

mjoe999: Marking this as unresolved after having a quick read up on GitHub etiquette ..... new to GitHub you see!

PR #3871, opened 2022-08-30
Compute key variations for DictParser

Cj-Malone:
This avoids us having to write "lon", "Lon", "LON" for every key

### DIFF
--- a/locations/dict_parser.py
+++ b/locations/dict_parser.py
@@ -123,23 +123,19 @@ class DictParser(object):
 
     @staticmethod
     def get_variations(key):
-        results = [key]
+        results = set(key)
 
         lower = key.lower()
-        if not lower in results:
-            results.append(lower)
+        results.add(lower)
 
         upper = key.upper()
-        if not upper in results:
-            results.append(upper)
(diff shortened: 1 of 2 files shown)


locations/dict_parser.py
```diff
@@ -147,12 +116,77 @@ def parse(obj) -> GeojsonPointItem:
...
-            elif obj.get(key.upper()):
-                return obj[key.upper()]
+            variations = DictParser.get_variations(key)
+            for variation in variations:
+                if obj.get(variation):
+                    return obj[variation]
+
+    @staticmethod
+    def get_variations(key):
+        results = [key]
```
iandees, line 126: Can you use a `set()` instead of an array? That'll save all the `if not x in results`.

Cj-Malone: I didn't know that was a thing, thank you.

Cj-Malone:
Some of these conversions were much easier to do in regex substitution, but unfortunately Pythons `re.sub` doesn't support case transformations like the PHP implementation of regex, which is the default on [regex101](https://regex101.com/)

PR #3872, opened 2022-08-30
Fuzzy's Taco Shop scraper.

youngpm:
Yeah the LinkedData stuff is awesome!


youngpm:
Ah never mind, this is good, the test just time outs at 2 minutes since this scrape takes a while.

PR #3873, opened 2022-08-31
Add The Gym Group Spider

Cj-Malone:


mjoe999:
Rename to the_gym_group_gb.py? Not a panacea but makes things a lot clearer when there may be thousands of spiders! Also allows easy filter such as "_gb" to look see what we have in this area or other territories?

Brands which cross territories are not necessary to qualify with _<country_code>

It's obviously not a panacea but does make things clearer.

Cj-Malone:
Or potentially `locations/spiders/GB/the_gym_group.py`, but I've no strong opinion on it.

PR #3877, opened 2022-09-01
Fix BMO Haris Spider

Cj-Malone:
The old spider has been inconstant because it was hitting the 4 hour limit, the limit has now been increased but this would still hit it. approx 40,000 in the US with a 0.5 second delay would still run for around 5 days.

This gives us 43,477 across US and CA in about 15 minuets. And technically we could drop the 0.5 delay.

### DIFF
--- a/locations/dict_parser.py
+++ b/locations/dict_parser.py
@@ -24,7 +24,7 @@ class DictParser(object):
         "locality",
     ]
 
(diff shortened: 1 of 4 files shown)


Cj-Malone:
I don't intend to fix Rite Aid as part of this PR. This is ready for review.

PR #3881, opened 2022-09-02
Fix footlocker, expand to be global, 3K+ stores

mjoe999:


locations/spiders/footlocker.py
```diff
@@ -1,109 +1,46 @@
...
+        "https://stores.footlocker.nl/sitemap.xml",
+        "https://stores.footlocker.no/sitemap.xml",
+        "https://stores.footlocker.co.nz/sitemap.xml",
+        "https://stores.footlocker.pl/sitemap.xml",
+        "https://stores.footlocker.pt/sitemap.xml",
+        "https://stores.footlocker.se/sitemap.xml",
+        "https://stores.footlocker.co.uk/sitemap.xml",
+    ]
+    sitemap_rules = [
+        ("/en/", "parse"),
```
iandees, line 34: Does this list need to include all the domains in the sitemap list?

mjoe999: The "sitemap_urls" list is comprehensive as far as I can see.
The "sitemmap_rules" list is saying for the English speaking countries use whatever pages they produce. For non English speaking countries they publsih both an English version and "local". We take the English languag page rather than the local language so as not to duplicate.

iandees: Ah that makes sense, thanks for the explanation!

PR #3882, opened 2022-09-02
Improve? Fix? Walgreens spider

mjoe999:
While investigating why no output in last few runs then
had cause to simplify spider. Slowed down the request rate
as they definitely throttle. Let's see what happens.


mjoe999:
Ok so I have seen the truncated Walgreen output JSON now. You will know it has been timing out. The 2 second download interval proposed runs at a steady rate for me. Shorter download intervals get heavily throttled by their end .... zero downloads over a longish period.

PR #3893, opened 2022-09-04
Fix marathon (~6K sites), CSVFeedSpider fun

mjoe999:


locations/spiders/marathon.py
```diff
@@ -1,39 +1,23 @@
...
+        "country": "US",
+    }
+    start_urls = [
+        "https://www.marathonbrand.com/content/includes/mpc-brand-stations/SiteList.csv"
+    ]
+
+    def parse_row(self, response, row):
+        if row["Status"] == "Open":
+            row["street_address"] = row["Address"]
+            del row["Address"]
```
iandees, line 20:
```suggestion
            row["street_address"] = row.pop("Address")
```

iandees:
I wonder if it needs a different User-Agent?

mjoe999:
How annoying. It was working for me as it was using a local cache of the page from months ago.

Cj-Malone:
I just came across this, the [file](https://www.marathonbrand.com/content/includes/mpc-brand-stations/SiteList.csv) is now base64'ed but it also isn't accessible via curl or scrapy...

PR #3905, opened 2022-09-07
Fix simply_self_storage

jleedev:


Cj-Malone:
The phone number seems to be an array, following OSM tagging it would be `number 1; number 2`

Cj-Malone (approved):
LGTM, thank you Josh

PR #3910, opened 2022-09-08
Fix iron_mountain (500 US + CA locations)

mjoe999:


locations/spiders/iron_mountain.py
```diff
@@ -1,86 +1,18 @@
...
-            postcode=address["post_code"],
-            lat=float(address["lat"]),
-            lon=float(address["lng"]),
-            phone=address["local_phone"],
-            website=address["url"].replace("\\", ""),
-            country=response.meta.get("country"),
-            ref=address["url"].replace("\\", ""),
-        )
+        item = OpenGraphParser.parse(response)
+        if item["lat"]:
```
Cj-Malone, line 17: I think this would error if lat isn't set. Did you mean .get?

Are there many pages in the sitemap without an lat? Should we do a sitemap filter instead?

mjoe999: No to the first (look at DictParser logic)
No to the second.
And so no to the third.

I'll tell you what we want .... what we really really want:

item.has_position()

then we could save on these enjoyable discussions ... I believe I wrote that once somewhere.

iandees: The DictParser will set "lat" to `None` if it can't find a latitude, so it should be good to check like this.

PR #3913, opened 2022-09-08
Fix bananarepublic, also does CA stores (436 in total)

mjoe999:


locations/spiders/bananarepublic.py
```diff
@@ -1,120 +1,25 @@
...
-            "postcode": zip_code,
-            "country": store_addr["countryCode"],
-            "website": response.url,
-        }
-
-        yield GeojsonPointItem(**properties)
+        if item := LinkedDataParser.parse(response, "ClothingStore"):
+            if item.get("lat"):
+                item["name"] = "Banana Republic"
+                item["ref"] = response.url
```
Cj-Malone, line 20: I would prefer we just use the unique store id from the URL as the ref.

mjoe999: Really .... do you think it is unique across their estate? That is kind of asking for trouble. "ref" I think is primarily for duplicate removal and if they do not supply then the clue is in the name of URL. I think LinkedDataParser and friends if they cannot get a "ref" supplied by the website should use the response.url ... or at least the path part.

locations/spiders/bananarepublic.py
```diff
@@ -1,120 +1,25 @@
...
-            "state": store_addr["stateProvinceCode"],
-            "postcode": zip_code,
-            "country": store_addr["countryCode"],
-            "website": response.url,
-        }
-
-        yield GeojsonPointItem(**properties)
+        if item := LinkedDataParser.parse(response, "ClothingStore"):
+            if item.get("lat"):
+                item["name"] = "Banana Republic"
```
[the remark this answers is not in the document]
mjoe999: Trivial.

locations/spiders/bananarepublic.py
Cj-Malone: > do you think it is unique across their estate?

Yes. And I think when we're writing a spider we should look for a uid. Ideally that would be published in the data, but if not, we should find one.

> That is kind of asking for trouble. [snip] cannot get a "ref" supplied by the website should use the response.url 

We've already ran into trouble using the whole URL as the ref. #3788

> or at least the path part

That's what I'm asking for, the uid is in the path.

It's really not that uncommon for these websites to ignore everything except the uid too. e.g. Holland and Barratt [NL store](https://www.hollandandbarrett.nl/winkels/genk-9033/), take the uid (`genk-9033`) and add checkout the [GB website](https://www.hollandandbarrett.com/stores/genk-9033/). Using the whole url (or even the whole path) as a ref, this gives us 2 stores, not de duped. Use the uid and one gets removed as a duplicate.

mjoe999: There is no perfect solution.

In this case we use direct store sitemap URL ... which is preferred. So having to even fart around with a "ref" at all is somewhat obtuse.

The code as it stands works.

Cj-Malone: In this case, we can't copy the uid from one site to another. It's not the end of the world to use the whole url as the ref.

PR #3916, opened 2022-09-08
Fix Gucci, worldwide ~450 stores

mjoe999:
If you like discussions about what constitues and individual store then this one is just for you.

### DIFF
--- a/locations/spiders/gucci.py
+++ b/locations/spiders/gucci.py
@@ -56,4 +56,7 @@ class GucciSpider(scrapy.spiders.SitemapSpider):
             yield scrapy.Request(sitemap_url, self._parse_sitemap)
 
     def parse(self, response):
-        yield LinkedDataParser.parse(response, "Store")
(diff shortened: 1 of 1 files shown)


locations/spiders/gucci.py
```diff
@@ -1,84 +1,59 @@
...
-                hours = self.parse_hours(days, hours)
-
-                if hours:
-                    properties["opening_hours"] = hours
-
-            except:
-                pass
-
-            yield GeojsonPointItem(**properties)
+        yield LinkedDataParser.parse(response, "Store")
```
iandees, line 59: Looks like they don't include lat/lon in the linked data JSON, but they do include it in the HTML at `section[@id="store-locator"]`

iandees: Yep, some pipeline steps that insert statistics about missing or invalid important (but not required) fields like lat/lon along with format checks on opening_hours and website would be helpful.

PR #3919, opened 2022-09-08
Add USC spider (300+ clothing stores)

mjoe999:


iandees:
Looks like they use one of the anti-bot things. I'll merge anyway and we can tackle it later.

mjoe999:
It is a two way thing. Some US/CA sites fail for us in the UK but are good for you. And vice-versa. Once again a more structural approach needed on infra side ... but that it is a whole other step. Myself and CJ could talk about running certain spiders this side of the Atlantic in the short term, but this requires dialogue.

Also the fact that you do not set a default download_delay in the project means that you are hammering a lot of sites. At the very least I would set one even if you don't use it for your weekly run. Remember any Tom, Dick or Harry can download the project and give it a bad name. I think some sites are blocking you for this.

iandees:
It's pretty easy to detect a naive scraper like ours. All they have to do is block AWS IPs (which is what's happening here) or block user-agents. There are things we can do to avoid this (pay for international or residential proxies, run some spiders on our own machines), but we seem to be making good progress still so I don't know that it's worth it.

PR #3920, opened 2022-09-08
Add webuyanycar_gb (500 used car dealers)

mjoe999:
Anyone getting tired off:

**from locations.linked_data_parser import LinkedDataParser
from locations.microdata_parser import MicrodataParser
MicrodataParser.convert_to_json_ld(response)
LinkedDataParser.parse(response, "BLAH"):**

appearing everwhere, and could possibly envisage:

**from locations.microdata_parser import MicrodataParser
MicrodataParser.parse(response, "BLAH")**

.. being added to do this!

### DIFF
--- a/locations/spiders/webuyanycar_gb.py
+++ b/locations/spiders/webuyanycar_gb.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+from scrapy.spiders import CrawlSpider, Rule
+from scrapy.linkextractors import LinkExtractor
+from locations.linked_data_parser import LinkedDataParser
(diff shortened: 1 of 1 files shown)


iandees:
I think if someone wanted to write a "SitemapMicrodataSpider" and a "SitemapLinkedDataSpider" that did all the boilerplate for you, that would be pretty cool...

It would be a composition of the existing "SitemapSpider" and the parsing, with hooks for customizing the entity type and adding extra bits to the item when the parser doesn't quite pick it all up automatically.

mjoe999:
Possibly so. But baby steps surely? :-)

PR #3931, opened 2022-09-11
Use sitemap instead of location search

jleedev:


locations/spiders/freddys.py
```diff
@@ -10,61 +11,51 @@
...
-                        lng=point["longitude"],
-                        today=today,
-                        nextweek=nextweek,
-                    ),
-                    callback=self.parse_stores,
+        response.selector.remove_namespaces()
+        for url in response.xpath("//loc/text()").extract():
+            if m := re.search("/location/([^/]+)/", url):
+                slug = m[1]
+                url = f"https://nomnom-prod-api.freddys.com/restaurants/byslug/{slug}?nomnom=calendars&nomnom_calendars_from={today}&nomnom_calendars_to={nextweek}&nomnom_exclude_extref=999"
```
iandees, line 30: Nice. I was looking for something like this but couldn't find it 👍🏻

PR #3934, opened 2022-09-11
Update Potbelly

iandees:


iandees:
It looks like Freddy's uses the same company as Potbelly, so I fixed Freddy's hours parsing too.

PR #3941, opened 2022-09-11
Validation pipeline

iandees:


locations/settings.py
```diff
@@ -23,7 +23,7 @@
 # Crawl responsibly by identifying yourself (and your website) on the user-agent
 USER_AGENT = f"Mozilla/5.0 (X11; Linux x86_64) {BOT_NAME}/{locations.__version__} (+https://github.com/alltheplaces/alltheplaces; framework {scrapy.__version__})"
 
-ROBOTSTXT_USER_AGENT = BOT_NAME
+ROBOTSTXT_USER_AGENT = USER_AGENT
```
Cj-Malone, line 26: I don't know enough about it to be certain, but I think it was right before. The User agent token/ROBOTSTXT_USER_AGENT is different to the Full user agent string/USER_AGENT. https://developers.google.com/search/docs/advanced/crawling/overview-google-crawlers

iandees: Yea, I accidentally `git add location/`'d and added a bunch of local changes to this PR that I didn't mean to. I'll undo this.

PR #3948, opened 2022-09-12
MOD Pizza spider.

youngpm:
I had to tweak the linked_data_parser as the geo tag was nested under the "location", not sure if this was a good idea though...  If there's a better way, let me know.

### DIFF
--- a/locations/linked_data_parser.py
+++ b/locations/linked_data_parser.py
@@ -42,8 +42,12 @@ class LinkedDataParser(object):
     def parse_ld(ld) -> GeojsonPointItem:
         item = GeojsonPointItem()
 
-        if geo := ld.get("geo"):
-            if isinstance(ld["geo"], list):
+        if (
+            (geo := ld.get("geo"))
+            or "location" in ld
+            and (geo := ld["location"].get("geo"))
+        ):
+            if isinstance(ld.get("geo"), list):
                 geo = geo[0]
 
             if geo.get("@type", "GeoCoordinates") in (
@@ -121,7 +125,6 @@ class LinkedDataParser(object):
     @staticmethod
     def parse(response, wanted_type) -> GeojsonPointItem:
         ld_item = LinkedDataParser.find_linked_data(response, wanted_type)
-
         if ld_item:
             item = LinkedDataParser.parse_ld(ld_item)
 
--- a/locations/linked_data_parser.py
+++ b/locations/linked_data_parser.py
@@ -47,7 +47,7 @@ class LinkedDataParser(object):
             or "location" in ld
             and (geo := ld["location"].get("geo"))
         ):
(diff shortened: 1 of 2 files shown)


locations/linked_data_parser.py
```diff
@@ -42,8 +42,12 @@ def find_linked_data(response, wanted_type) -> {}:
     def parse_ld(ld) -> GeojsonPointItem:
         item = GeojsonPointItem()
 
-        if geo := ld.get("geo"):
-            if isinstance(ld["geo"], list):
+        if (
```
Cj-Malone, line 45: I'm not sure this is part of the spec, or a common problem. If not I'd prefer it moved into spider code. We'd have to stop using `LinkedDataParser.parse` instead use `LinkedDataParser.find_linked_data`, fix the result, then use `LinkedDataParser.parse_ld`, [eg](https://github.com/alltheplaces/alltheplaces/blob/ebca0409ede42f60c1c1f4a91a029817edeb9c89/locations/spiders/sunglass_hut.py#L18)

If it is common and we stick with this we need to fix `if isinstance(ld.get("geo"), list):` to `if isinstance(geo, list):`

locations/spiders/modpizza.py
```diff
@@ -0,0 +1,33 @@
...
+        item = LinkedDataParser.parse(response, "FastFoodRestaurant")
+
+        # The address isn't in the microdata.
+        item["street_address"] = response.css(
+            ".Address-field.Address-line1::text"
+        ).get()
+        item["city"] = response.css(".Address-city::text").get()
+        item["state"] = response.css(".Address-field.Address-region::text").get()
+        item["postcode"] = response.css(".Address-field.Address-postalCode::text").get()
+        item["country"] = "US"
```
Cj-Malone, line 31: For static values we put them in `item_attributes`, so `item_attributes = {"brand": "MOD Pizza", "brand_wikidata": "Q19903469", "country": "US"}`

locations/linked_data_parser.py
youngpm: @Cj-Malone I ran it through the [schema.org validator](https://validator.schema.org/#url=https%3A%2F%2Flocations.modpizza.com%2Fusa%2Fky%2Flexington%2F2217-war-admiral-way) and it reported no errors, so I am guessing it is valid?  

Happy to fix it up either way!

Cj-Malone: OK, yeah you're right, it's to spec. [location](https://schema.org/location) can be a [Place](https://schema.org/Place) and that can contain geo.

So we just need to do:

```patch
-if isinstance(ld.get("geo"), list):
+if isinstance(geo), list):
```

Cj-Malone (approved):
LGTM, thank you.

PR #3952, opened 2022-09-12
Allow more day values in OH

Cj-Malone:
Most spiders that have opening hours have to do some kind of sanitation on the day, most of it could be shared code.

### DIFF
--- a/locations/hours.py
+++ b/locations/hours.py
@@ -13,6 +13,16 @@ DAYS_FULL = [
     "Saturday",
     "Sunday",
 ]
+
+DAYS_DE = {
+    "Mo": "Mo",
+    "Di": "Tu",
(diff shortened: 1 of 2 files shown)


jleedev:
Also have wochentag in adeg.py. https://github.com/alltheplaces/alltheplaces/blob/46f9ad3dab3e043def1e56ad9cfc3f166514b652/locations/spiders/adeg.py#L9-L17

jleedev:
Although "Do" is Sunday in Spanish, Thursday in German, so translating days without knowledge of the language is probably ill-advised.

Cj-Malone:
> Or, we make the days map mandatory (we can default to English), and only check for one language.

I've done that, I think the overlap would cause too much trouble.

PR #3957, opened 2022-09-13
Update costacoffee_gb; introduce H3 hex grid

jleedev:
Backend returns all results in a 5km radius; therefore use resolution 6 with an average edge length of ~3.22 km.

We cover the UK with 8626 cells, discarding 18975 items and scraping 13476.

I still want to check in the code to generate the boundaries, as well as write a fancy technique for cases where the search results are limited by a maximum count as well as by radius, but this is still usable as is.

Will attempt to go in detail later, but briefly:

- The existing centroids*.csv files are developed in Mercator projection, which means the points are actually much closer together than the user bargains for. Nunavut is 2 times as large as Ontario and should not contain 10 times as many points; Alaska is 2 times as large as Texas, not 8.
- The code that produced these points was not checked in, making it cumbersome to add additional regions.

It might be nice to be able to produce a point grid at _any_ required resolution, but spherical math is difficult, and I find value in H3 being robust and thoroughly tested. Back of the envelope I think if I made a grid of points exactly 5 km apart, I could cover the UK with just 3600 cells, but I can get the h3 grid entirely for free.

Note: using prerelease version of h3 purely because they renamed some functions.


Cj-Malone:
I've found a better API for Costa, with 13,660 items. I'll do a PR for that.

No opinion on the H3 stuff, it's a bit over my head, but the output looks good.

jleedev:
Ah, well then.

I was eyeing this for Starbucks and USPS collection boxes, which attempt to drill down to denser locations, incorrectly, and Allpoint ATM search.

mjoe999:
Having a robust interface for giving a point grid at various resolutions within a territory would be good. The existing point files are not ideal as you say. Also needed and it is closely related would be a scatter point grid at maximum inter-point distance radially out from point to a limit, this would enable:

for x, y in point_locations("FR", 200000)      ..... 200km

for city in city_locations("FR", 200000):         .... cities with pop > 0.2M (wehave this i/f)
    for radial_locations(city, 10000, 1000):      ..... takes lat/lon from city, radial area 10Km, generate points max distance 1KM

I think this gives a reasonable toolkit for spider writers where one does not have a CJ to sort it out :-)

mjoe999:
But I forgot to say ... very interesting on the H3 side etc ... I keep learning.

PR #3958, opened 2022-09-14
Another Costa PR

### LINKED ISSUE #3805: Only about half of stores returned by costacoffee_gb spider
https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/costacoffee_gb.py

I think the Costa Coffee GB spider is only returning around half of the Costa stores. (My attempt to match the non-express stores from AllThePlaces with OSM data has found only about half of the ~1800 OSM objects with brand:wikidata=Q608845 are matched. See https://osm.mathmos.net/chains/Q608845/ )

For example, the Diss Mere Street store at IP22 4AD is not in the AllThePlaces output: store https://www.costa.co.uk/locations/store-locator/map?latitude=52.3749&longitude=1.1089399999999898&open=1

From the comment at line 17 of costacoffee_gb.py:
`# TODO: Can't figure out how to return more than 5 miles`
it seems you may already be aware of this, but I couldn't see an issue about it, so I thought it best to create one.

Cj-Malone:
No more lat lon search, resulting in 13,661 items in 20 seconds.

Fixes #3805

### DIFF
--- a/locations/pipelines.py
+++ b/locations/pipelines.py
@@ -175,7 +175,10 @@ class CheckItemPropertiesPipeline(object):
         if opening_hours := item.get("opening_hours"):
             if not isinstance(opening_hours, str):
                 spider.crawler.stats.inc_value("atp/field/opening_hours/wrong_type")
-            elif not self.opening_hours_regex.match(opening_hours):
+            elif (
+                not self.opening_hours_regex.match(opening_hours)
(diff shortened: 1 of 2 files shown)


mjoe999:
Excellent CJ. Also excellent Josh, I will put my thoughts on my PR.

PR #3883, opened 2022-09-02
Add insights command to start looking at output sanity

mjoe999:
Now that we have one command why not another? For the most part spiders are looked
at in terms of how many POIs they produce week to week. Sure, they may have been
well examined for output quality initially but this can and does decay over time. Also
would it not be good to get some insights into output quality while in development (and
after) with a little tooling help, in addition to the Mk. 1 eyeball.

So here is a first offering. Can be extended, changed, thrown away! But at least
we are on a path to better output. There are many extensions to "insights" that I can
think of, but for another day. As it is non-mutating to the operation of the spiders I
hope it is not contentious.

As you know I think the brand wikidata is a key component. I have added NSI as a
GIT submodule. It is one of those occasions where this works well as they publish
their output to the repo (dist/*.json). This enables a cheap check of sanity e.g.
if a food chain with 1,000 outlets is not in NSI then it may suggest that either
the Q-code in the spider is wrong/in-appropriate or that an entry could be added
to NSI or even wikidata, plus if there are so many POIs how is it being handled in
OSM itself, you'd think there would be some entries!

Some example usage, with ALP GeoJSON expanded result dump:

$ pipenv run scrapy insights --country-codes ~/Downloads/ouput/
$ pipenv run scrapy insights --wikidata-codes ~/Downloads/ouput/
$ pipenv run scrapy insights --country-codes ~/Downloads/ouput/mcdonalds*

For the NSI/wikidata stuff to function you must:

$ git submodule init

on first pull, and:

$ git submodule update --progress

... initially and periodically. Or least that's how I think it goes!

PR #3959, opened 2022-09-14
Add country codes to spiders

Cj-Malone:
I used insights from #3883 to find spiders without a country code. This is just the top ones, that were easy to fix, there is a long tale.


Cj-Malone:
This failed because of [Chase](https://data.alltheplaces.xyz/ci/3052298058/chase/log.txt) failing, I don't intend to fix that in this PR.

PR #3964, opened 2022-09-16
Add vapestore Spider

Cj-Malone:
Locations wont be extracted on this one until #3963 is merged and adds support for this url format.


Cj-Malone:
It's failing because of the null lat+longs, also fixed in that PR.

PR #3969, opened 2022-09-22
Add StructuredDataSpider

Cj-Malone:
`StructuredDataSpider` can be inherited by any spider, including `SitemapSpider` and `CrawlSpider`s.

It gives us `parse_sd` which does the normal Microdata/LinkedData stuff, using a `wanted_types` attribute. It also does some smarts with getting refs from the url (only when none is set from source), it will try to grab a regex group from `rules` or `sitemap_rules`. If not it falls back to the full url. It can also try to pull phone and email addresses from the response.

There is also `inspect_item` callback where we can do clean up in spider.

### DIFF
--- a/locations/spiders/webuyanycar_gb.py
+++ b/locations/spiders/webuyanycar_gb.py
@@ -1,28 +1,32 @@
 # -*- coding: utf-8 -*-
+from locations.google_url import extract_google_position
+from locations.spiders.vapestore_gb import clean_address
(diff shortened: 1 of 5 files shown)


Cj-Malone:
[webuyanycar_gb](https://data.alltheplaces.xyz/ci/3104842049/webuyanycar_gb/log.txt) is failing, works fine locally so presumably network related.

PR #3993, opened 2022-09-23
Add spider for Condado Tacos

jleedev:


Cj-Malone:
You've missed `inspect_item`. From site map rules you can go straight to `parse_sd` And do the minor clean up in `inspect_item`.

PR #3994, opened 2022-09-23
Add comment explaining method purpose

jleedev:


Cj-Malone (approved):
Assuming `yield from` works as we'd expect with multiple, or no items. This LGTM, thanks.

PR #4016, opened 2022-09-26
Fix Aldi IE/GB Spider

Cj-Malone:


Cj-Malone:
`download_delay = 10` isn't technically required, but is requested by robots.txt.

PR #4017, opened 2022-09-27
Fix, to some level, Marriott (~7K sites)

mjoe999:
I say fix but who knows what state it was in before. The Marriott estate is complex as hopefully is clear from the code and comments. One thing for sure, would not have liked to do this without:

$ pipenv run scrapy nsi --name protea
"Protea Hotels", "Q17092570"
       -> https://www.wikidata.org/wiki/Q17092570
       -> https://www.wikidata.org/wiki/Special:EntityData/Q17092570.json
       -> South African-based hotel chain
       -> http://protea.marriott.com/
       -> item_attributes = {"brand": "Protea Hotels", "brand_wikidata": "Q17092570"}

### DIFF
--- a/locations/spiders/marriott_hotels.py
+++ b/locations/spiders/marriott_hotels.py
@@ -1,122 +1,81 @@
-import json
-import re
 import scrapy
-
-from scrapy.selector import Selector
-from locations.items import GeojsonPointItem
+from locations.dict_parser import DictParser
 
 
+# See https://www.marriott.com/marriott-brands.mi for an explanation of the world of Marriott brands.
+# Of the sites below Ritz Carlton is probably the most important.
+# TODO: ["Ritz-Carlton", "Q782200"] will needs its own site spider?
+# TODO: ["JW Marriott Hotels", "Q1067636"] will need its own site spider?
+# TODO: ["Renaissance Hotels", "Q2143252"] will need its own site spider?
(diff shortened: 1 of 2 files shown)


locations/spiders/marriott_hotels.py
```diff
@@ -1,122 +1,81 @@
...
-            yield GeojsonPointItem(**properties)
-
-    def parse(self, response):
-        xml = Selector(response)
-        xml.remove_namespaces()
-
-        urls = xml.xpath("//loc/text()").extract()
-        for url in urls:
-            yield scrapy.Request(response.urljoin(url), callback=self.parse_hotel)
+        letters = "ABCDEFGHIJKLMNOPQRSTUVXYZ"
```
iandees, line 54: Instead of making 676 requests for all combinations of the alphabet, can we use the `my_brands` dict that you put together and make requests for just the brands we know exist?

JoeDixon-TomTom: You could and that is what my code used to do until my interest was raised when I ported it and found additional codes. Future movements in the endpoint would show themselves as ERROR logs. Otherwise we will never see them. It's perhaps worth a comment in the code. Your call of course.

JoeDixon-TomTom: @iandees if you tell me which way you wish me to go on this I will do it.

iandees: I think it's ok. Especially with the download_delay 👍🏻

PR #4028, opened 2022-09-28
fix: AutoNation brand incorrect

Romeo-PHILLIPS:
Capture d’écran 2022-09-28 à 16 43 51

Auto Nation scraper was setting as a brand what brands the dealership was selling, which isn't the brand of the store locator.


Cj-Malone:
Urm, this is valid OSM tagging, but it is a dual use of the tag. I'd support moving it to an `extra`, maybe `sells`?

Romeo-PHILLIPS:
> Urm, this is valid OSM tagging, but it is a dual use of the tag. I'd support moving it to an `extra`, maybe `sells`?

I could agree with this yea

Cj-Malone (approved):
LGTM, thank yoh

PR #4049, opened 2022-10-01
Update docs

Cj-Malone:
I'm not good at this stuff, so I'm fine if there is a bunch of criticism, or someone else decides to rewrite it.


iandees:
That's a great update, thanks!

PR #4064, opened 2022-10-04
fixing Boston Pizza spider

boomerwv1:


boomerwv1:
Ok, probably wasn't caught up.

Thanks.

PR #4072, opened 2022-10-10
New command to compare ATP output with OSM & NSI

mjoe999:
`$ pipenv run scrapy insights --atp-nsi-osm <geojson output dir> --outfile <json output file>`

Command gets wikidata tag counts from OSM taginfo. We already have the code to pull in NSI data and also stream in features from an ATP run geojson dump. This command produces a JSON file of what there is in ATP/NSI/OSM per unique wikidata code. The weekly run should publish this json to data.alltheplaces.xyz.

There is a commit coming to alltheplaces.xyz for a web (datatables) page that consumes it:

https://github.com/alltheplaces/alltheplaces.xyz/pull/23

Ian can decide on the specific plumbing?


mjoe999:
Hi @iandees ,
I don't know if you missed this or not. The contribution is not to core code but rather sits out in command land. The function if integrated is useful I think. It picks up various anomalies in ATP vs NSI .... pointing to a possible fix in one or the other. It also points us at ATP spiders that need more work to get more items. It also points us at significant brands for which we do not have a spider. Is it possible for you to give your initial thoughts?
Cheers,
Joe.

iandees:
Hi! Yep, I saw it. Thanks for the contribution, of course.

I haven't had a chance to fully understand what is going on between the two changes and I want to figure it out before merging either one. It would be helpful if you could give a high level overview of how you envision this working and what the goal is when proposing large-ish changes.

mjoe999:
I will add some more commentary to the code. In particular this commit. But essentially the command forms a dict keyed by Q-code which has interesting info about that Q-code from the OSM tag count perspective if any, NSI name and description perspective if any and ATM download count and brand name perspective if any. This data structure can be output as a JSON for consumption by the viewer HTML contributed to alltheplaces.xyz. I think it's useful, see the example screen shot in the alltheplaces.xyz commit. I would anticipate you run this command in your weekly run script and upload the JSON output to alltheplace.xyz for consumption by the contributed datatables page. While it may be considered a large change it is not to core library behaviour. Cheers, Joe.

mjoe999:
Hi @iandees ,
I've submitted a few changes which hopefully clarify what's going on further.
Cheers,
Joe.

locations/commands/insights.py
```diff
@@ -127,6 +146,69 @@ def check_wikidata_codes(self, args):
...
+            if not count:
+                count = 0
+            r["atp_count"] = count + 1
+            if brand:
+                r["atp_brand"] = brand
+
+        # Write a JSON format output file which is datatables friendly.
+        for_datatables = {"data": list(wikidata_dict.values())}
+        with open(outfile, "w") as f:
+            f.write(json.dumps(for_datatables))
```
iandees, line 210: You don't have to json.dumps first.

```suggestion
            json.dump(for_datatables, f)
```

iandees (approved):

PR #4076, opened 2022-10-11
Fix officedepot; add json comments mode

jleedev:
Add an option to handle ld-json with comments (and trailing commas), using a [json5] library to support a relatively standardized version of this format.

[json5]: https://json5.org/


jleedev:
It's possible that this parameter would be better handled via kwargs, or an instance variable.

iandees:
We should probably just switch over to json5 altogether at some point.

PR #4078, opened 2022-10-12
Add Urban Outfitters spider

### LINKED ISSUE #669: Urban Outfitters
https://www.urbanoutfitters.com/stores/

tonyskapunk:
Uses a long delay due to CDN blocking
Only scrapes US stores as non-US stores are not consistent.

Fixes #669


tonyskapunk:
That [403 in the log](https://data.alltheplaces.xyz/ci/3231663217/urban_outfitters/log.txt) is due to the CDN blocking the request, likely other requests already coming from the same IP :crying_cat_face:

mjoe999:
Hi Tony,
Using a SitemapSpider and a StructuredDataSpider I think would go a long way to compress your contribution. US stores: https://www.urbanoutfitters.com/store_sitemap.xml
Sometimes it's possible to make the code really tight e.g.
https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/smashburger.py
Not saying that this would be the case here though but worth a play?
Cheers,
Joe.

tonyskapunk:
@mjoe999 thanks for the suggestion, I'll definitely look at that alternative.

@Cj-Malone, thanks as well for the info. I'll take a look and update this PR.

locations/spiders/urban_outfitters.py
```diff
@@ -0,0 +1,14 @@
...
+
+from locations.structured_data_spider import StructuredDataSpider
+
+
+class UrbanOutfitters(SitemapSpider, StructuredDataSpider):
+    download_delay = 10
+    name = "urban_outfitters"
+    item_attributes = {"brand": "Urban Outfitters", "brand_wikidata": "Q3552193"}
+    allowed_domains = ["www.urbanoutfitters.com"]
+    sitemap_urls = ["https://www.urbanoutfitters.com/robots.txt"]
```
Cj-Malone, line 12: Sorry, I've confused things here. Sometimes we need to scrape multiple sitemaps, usually because there is a sitemap per region. Having a closer look the sitemaps are split by language. We only need the default [English](https://www.urbanoutfitters.com/store_sitemap.xml) one, not the others eg [French](https://www.urbanoutfitters.com/fr-fr/store_sitemap.xml).

```suggestion
    sitemap_urls = ["https://www.urbanoutfitters.com/store_sitemap.xml"]
```

locations/spiders/urban_outfitters.py
```diff
@@ -0,0 +1,14 @@
...
+    sitemap_rules = [("", "parse_sd")]
+    sitemap_follow = ["/sitemapindex.xml", "/store_sitemap.xml", "/stores/"]
```
Cj-Malone, line 14: `sitemap_follow` defines the sitemaps which should be parsed. So `/store_sitemap.xml` is the valid rule, but because we're now pointing directly at it. It's unneeded.
```suggestion
```
`/stores/` is the rule for the page urls that should be parsed from the sitemap. So it could be `sitemap_rules = [("/stores/", "parse_sd")]`, however in this case it's not necessary because all the pages in that sitemap are stores.

Cj-Malone (commented):
Having a look at a [Structured Data valadator](https://validator.schema.org/#url=https%3A%2F%2Fwww.urbanoutfitters.com%2Fstores%2Furban-outfitters-walnut-street), we can see the source data looks pretty good. However it looks like `url` points to a internal testing domain, we'll have to override that, also it's missing opening hours. Missing opening hours isn't a deal breaker, but since you've already written code to parse them we may as well copy it over.

```py
        item["website"] = response.url
        # TODO: Opening hours.
        yield item
```

This is looking good, I appreciate the time you've taken.

tonyskapunk:
> Having a look at a [Structured Data valadator](https://validator.schema.org/#url=https%3A%2F%2Fwww.urbanoutfitters.com%2Fstores%2Furban-outfitters-walnut-street), we can see the source data looks pretty good. However it looks like `url` points to a internal testing domain, we'll have to override that,

Nice catch, sure, I'll get it fixed.

>  also it's missing opening hours. Missing opening hours isn't a deal breaker, but since you've already written code to parse them we may as well copy it over.

True, I'll add it back.

Thanks for the snippet :100: !!

I also noted some issues with the name that was considered previously I'll get that fixed too.

locations/spiders/urban_outfitters.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class UrbanOutfitters(SitemapSpider, StructuredDataSpider):
+    download_delay = 10
+    name = "urban_outfitters"
+    item_attributes = {"brand": "Urban Outfitters", "brand_wikidata": "Q3552193"}
+    allowed_domains = ["www.urbanoutfitters.com"]
+    sitemap_urls = ["https://www.urbanoutfitters.com/store_sitemap.xml"]
+    sitemap_rules = [("", "parse_sd")]
+
```
Cj-Malone, line 15:
```suggestion
    user_agent = (
        "Mozilla/5.0 (X11; Linux x86_64; rv:105.0) Gecko/20100101 Firefox/105.0"
    )

```

This seems to get us unblocked.

locations/spiders/urban_outfitters.py
```diff
@@ -0,0 +1,79 @@
...
+        for h in raw_hours:
+            entry = {}
+            hrs = " ".join(h.split(" ")[1:])
+            entry["name"] = h.split(" ")[0].strip(":")
+            entry["openHour"] = hrs.split("-")[0].strip()
+            entry["closeHour"] = hrs.split("-")[1].strip()
+            hours.append(entry)
+
+        oh = self.parse_hours(hours)
+        item["opening_hours"] = oh
```
Cj-Malone, line 32: This is erroring on [some stores](https://www.urbanoutfitters.com/stores/rotterdam). I'm happy for you to just wrap it in a try statement, or if you want to try and normalize them, you can.

tonyskapunk: Let me take a look to it and see if I can figure out something, I'll ask for help if I get stuck with it. Thanks!

tonyskapunk: I've fixed the formatting on the hours, and now is looking better; whenever you have some time, please take a look. Thanks!

Cj-Malone (approved):
Did you run it through without any errors? If so it's good for me.

PR #4083, opened 2022-10-13
Add stats for NSI matching

Cj-Malone:
So this started out as me trying to add categories from NSI, and hopefully we'll end up there. But I think we need some stats first.

- `atp/nsi/brand_missing` for times when NSI is missing an entry with the Wikidata ID
- `atp/nsi/perfect_match` for when there is a single NSI match for the Wikidata ID
- `atp/nsi/multiple_hits` for when there are multiple NSI matches with the Wikidata ID

You can see the commit log for some ideas I've had to filter the multiple NSI matches, but ultimately I want to see the stats and identify the specific spiders before I go to far there. This is the TODO on line 229.

Removing line 236 would apply the NSI tags to the output, not overriding any tags set by the spider, applied to the item or extras depending. I'd be happy for us to start doing this right away, but I'm also happy to wait until after the weekly run so we can see the stats and discus it properly.

### DIFF
--- a/ci/fetch_all_stats.sh
+++ b/ci/fetch_all_stats.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+BUILD_ID=$1
+
+if [ -z "${BUILD_ID}" ]; then
+    (>&2 echo "Missing BUILD_ID")
+    (>&2 echo "Try '$0 2022-10-08-13-31-39'")
+    exit 1
+fi
+
+mkdir -p "${BUILD_ID}"
+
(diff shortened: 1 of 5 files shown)


iandees:
Let's see what happens when we add the tags from NSI. Remove that early return :)

Cj-Malone:
2022-10-15-13-31-49
  | atp/nsi/brand_missing | atp/nsi/perfect_match | atp/nsi/multiple_hits | atp/field/brand_wikidata/missing
-- | -- | -- | -- | --
Spider Count | 110 | 376 | 157 | 63
Item Count | 410,433 | 510,180 | 704,516 | 31,851

Cj-Malone:
Unless there is any objections, tomorrow I'll work on filtering NSI matches by country code, it should get us some easy wins like Subway were there is 3 NSI entries to one QID.

PR #4091, opened 2022-10-14
Run the insights command as part of weekly build

iandees:
This runs the command added in #4072 after all the spiders are done, saves it in the stats/ dir, and includes the URL in the build summary JSON.


iandees:
Does this look ok, @mjoe999?

mjoe999:
That looks just fine. I think your run this weekend will yield some interest for various of us!

Cj-Malone (approved):

PR #4096, opened 2022-10-15
765: Adding spider: Crate And Barrel

### LINKED ISSUE #765: Crate And Barrel
https://www.crateandbarrel.com/stores/list-state/retail-stores

kumanoit:
Fixes #765

### DIFF
--- a/locations/spiders/crateandbarrel.py
+++ b/locations/spiders/crateandbarrel.py
@@ -15,8 +15,11 @@ class CrateAndBarrelSpider(CrawlSpider, StructuredDataSpider):
     rules = [
         Rule(
             LinkExtractor(allow=r"stores\/list-state\/retail-stores\/([a-zA-Z]{2})$"),
+        ),
+        Rule(
+            LinkExtractor(allow=r"stores\/.+\/.+$"),
             callback="parse_sd",
-        )
+        ),
     ]
 
     def post_process_item(self, item, response, ld_data, **kwargs):
--- a/locations/spiders/crateandbarrel.py
+++ b/locations/spiders/crateandbarrel.py
@@ -1,0 +1,29 @@
+import json
+
+from scrapy.spiders import CrawlSpider, Rule
+from scrapy.linkextractors import LinkExtractor
+
+from locations.linked_data_parser import LinkedDataParser
+
+
+class CrateAndBarrelSpider(CrawlSpider):
+    name = "crateandbarrel"
+    allowed_domains = ["www.crateandbarrel.com"]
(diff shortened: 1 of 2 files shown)


locations/spiders/crateandbarrel.py
```diff
@@ -0,0 +1,29 @@
...
+
+class CrateAndBarrelSpider(CrawlSpider):
+    name = "crateandbarrel"
+    allowed_domains = ["www.crateandbarrel.com"]
+    item_attributes = {"brand": "crateandbarrel", "brand_wikidata": "Q5182604"}
+    start_urls = ["https://www.crateandbarrel.com/stores/list-state/retail-stores"]
+    rules = [
+        Rule(
+            LinkExtractor(allow=r"stores\/list-state\/retail-stores\/([a-zA-Z]{2})$"),
+            callback="parse",
```
JoeDixon-TomTom, line 17: Hi, I think if you changed the callback to "parse_sd" and inherited from StructuredDataSpider, bit like this one:
https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/smashburger.py
.. then you would not need the parse() method below at all as parse_sd in the superclass would do it all for you. I hope so anyway! :-)
Cheers,
Joe.

kumanoit: Thank you @JoeDixon-TomTom 
Worked like a charm :)

locations/spiders/davidstea.py
```diff
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 from scrapy.spiders import SitemapSpider
 
-from locations.hours import OpeningHours
```
kumanoit, line 4: This was unused so removed it.
This spider is working well after removing above import.

locations/spiders/crateandbarrel.py
Cj-Malone: The issue now is that you're parsing the state URLs, and setting the ref it it's URL. So when a [state](https://www.crateandbarrel.com/stores/list-state/retail-stores/ca) has multiple stores all the stores get the same ref, and removed as duplicates.

```patch
-        Rule(
-            LinkExtractor(allow=r"stores\/list-state\/retail-stores\/([a-zA-Z]{2})$"),
-            callback="parse_sd",
-        )
+        Rule(
+            LinkExtractor(allow=r"stores\/list-state\/retail-stores\/([a-zA-Z]{2})$"),
+        ),
+        Rule(
+            LinkExtractor(allow=r"stores\/.+\/.+$"),
+            callback="parse_sd",
+        )
```

kumanoit: Thank you..
changes incorporated

PR #4098, opened 2022-10-15
Fix website in StructuredDataSpider

Cj-Malone:
This used to be set by LinkedDataParser.parse but that's no longer used.

### DIFF
--- a/locations/linked_data_parser.py
+++ b/locations/linked_data_parser.py
@@ -164,7 +164,7 @@ class LinkedDataParser(object):
             if isinstance(item["website"], list):
                 item["website"] = item["website"][0]
 
-            if item["website"] is None or item["website"] == "":
+            if not item["website"]:
                 item["website"] = response.url
             elif item["website"].startswith("www"):
                 item["website"] = "https://" + item["website"]
--- a/locations/linked_data_parser.py
+++ b/locations/linked_data_parser.py
@@ -161,11 +161,10 @@ class LinkedDataParser(object):
         if ld_item:
             item = LinkedDataParser.parse_ld(ld_item)
 
-            if item["website"] is None:
-                item["website"] = response.url
-            elif item["website"] == "":
-                item["website"] = response.url
-            elif item["website"][0] == "/":
+            if isinstance(item["website"], list):
+                item["website"] = item["website"][0]
+
+            if item["website"] is None or item["website"] == "":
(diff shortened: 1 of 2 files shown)


locations/linked_data_parser.py
```diff
@@ -161,11 +161,10 @@ def parse(response, wanted_type, parse_json5=False) -> GeojsonPointItem:
         if ld_item:
             item = LinkedDataParser.parse_ld(ld_item)
 
-            if item["website"] is None:
-                item["website"] = response.url
-            elif item["website"] == "":
-                item["website"] = response.url
-            elif item["website"][0] == "/":
+            if isinstance(item["website"], list):
```
iandees, line 164: Check that there's something in the list before indexing into it:

```suggestion
            if isinstance(item["website"], list) and item["website"]:
```

locations/linked_data_parser.py
```diff
@@ -161,11 +161,10 @@ def parse(response, wanted_type, parse_json5=False) -> GeojsonPointItem:
...
+                item["website"] = item["website"][0]
+
+            if item["website"] is None or item["website"] == "":
```
iandees, line 167: Same as:

```suggestion
            if not item["website"]:
```

locations/structured_data_spider.py
```diff
@@ -85,6 +85,15 @@ def parse_sd(self, response):
                     if item["ref"] is None:
                         item["ref"] = response.url
 
+                item["website"] = None
+                if isinstance(item["website"], list):
```
iandees, line 89:
```suggestion
                if isinstance(item["website"], list) and item["website"]:
```

locations/structured_data_spider.py
```diff
@@ -85,6 +85,15 @@ def parse_sd(self, response):
...
+                    item["website"] = item["website"][0]
+
+                if item["website"] is None or item["website"] == "":
```
iandees, line 92:
```suggestion
                if not item["website"]:
```

locations/linked_data_parser.py
Cj-Malone: `parse_ld` always populates `website`, even if it's None. `isinstance(None, list)` works as expected. Can update if you want though.

iandees: No prob.

PR #4101, opened 2022-10-16
feat: Add Ted's Montana Grill spider

### LINKED ISSUE #3522: Teds Montana Grill
https://www.tedsmontanagrill.com/locations.html

rubyruins:
Fixes #3522.

### DIFF
--- a/locations/spiders/teds_montana_grill.py
+++ b/locations/spiders/teds_montana_grill.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+import scrapy
+from locations.items import GeojsonPointItem
+
+
+class TedsMontanaGrillSpider(scrapy.Spider):
+    name = "teds_montana_grill"
+    allowed_domains = ["www.tedsmontanagrill.com"]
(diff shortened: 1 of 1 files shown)


locations/spiders/teds_montana_grill.py
```diff
@@ -0,0 +1,39 @@
...
+        ref = location.xpath(".//a/@href").extract_first()
+        city = location.xpath(".//h5/text()").extract_first()
+        phone = (
+            location.xpath(".//p")[-1]
+            .xpath(".//span/a/@href")
+            .extract_first()
+            .replace("tel:", "")
+        )
+        address = location.xpath(".//p")[-1].xpath("text()").extract()
+        addr_full = ", ".join([line.strip() for line in address[:-1]])
```
iandees, line 29: Could get the first line of this and set it as the `street_address` in addition to joining it into `addr_full`?

rubyruins: Makes sense. It turns out that `addr_full` really was the `street_address`, so I have added the `street_address` field and included it with the `postcode` as `addr_full`.

PR #4102, opened 2022-10-16
Add Bashas' spider

### LINKED ISSUE #3506: Bashas'
https://www.bashas.com/locations/

cmecklenborg:
Fixes #3506

### DIFF
--- a/locations/spiders/bashas.py
+++ b/locations/spiders/bashas.py
@@ -1,0 +1,56 @@
+import html
+import json
+import scrapy
+
+from locations.dict_parser import DictParser
+from locations.hours import OpeningHours
+
+
+class BashasSpider(scrapy.spiders.SitemapSpider):
+    name = "bashas"
+    item_attributes = {
+        "brand": "Bashas'",
+        "brand_wikidata": "Q4866786",
+    }
+
+    download_delay = 0.2
+    sitemap_urls = ["https://www.bashas.com/wpsl_stores-sitemap.xml"]
+    sitemap_rules = [("/stores/", "parse_store")]
+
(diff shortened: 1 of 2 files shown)


cmecklenborg:
Because of non-ascii characters in the store names, the geojson output contains escaped unicode characters under the "name" attribute, e.g. `Bashas\u2019 Supermarket: Power & McDowell`. Is this a concern? If so, what is the preferred way to normalize these characters?

iandees:
> the geojson output contains escaped unicode characters under the "name" attribute

That's ok. JSON parsers will unescape those into unicode codepoints properly.

locations/spiders/bashas.py
```diff
@@ -0,0 +1,56 @@
...
+    def parse_store(self, response):
+        script = response.xpath(
+            "//script[@id='wpsl-js-js-extra']/text()"
+        ).extract_first()
+        open_brace = script.find('"locations":[{') + 13
+        close_brace = script.rfind("}]}") + 1
+        store_json = json.loads(script[open_brace:close_brace])
+        hours = response.xpath("//table[@class='wpsl-opening-hours']//text()").extract()
+
+        item = DictParser.parse(store_json)
```
iandees, line 29: It looks like the DictParser is picking up an address and setting it as `addr_full` when it should be `street_address`. Can you add a `item["street_address"] = item.pop("addr_full", None)` somewhere after this line?

cmecklenborg: I made this change. I noticed the same problem was happening with a spider I added yesterday (Gelsons), so I made the change there too.

Issue #889, opened 2018-05-02
Ridley's Family Markets


maxerickson:
https://shopridleys.com/choose.php

PR #4109, opened 2022-10-17
feat: Ridley's Family Markets

Romeo-PHILLIPS:
Fixes https://github.com/alltheplaces/alltheplaces/issues/889

### DIFF
--- a/locations/spiders/ridleys.py
+++ b/locations/spiders/ridleys.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+import scrapy
+import json
+from locations.items import GeojsonPointItem
+
+
+class Ridleys(scrapy.Spider):
+    name = "ridleys"
+    item_attributes = {"brand": "Ridley's Family Markets"}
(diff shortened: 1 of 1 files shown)


locations/spiders/ridleys.py
```diff
@@ -0,0 +1,29 @@
...
+                "street_address": data.get("streetAddress"),
+                "name": data.get("storeName"),
+                "city": data.get("city"),
+                "postcode": data.get("zip"),
+                "state": data.get("state"),
+                "website": data.get("rio_url"),
+                "phone": data.get("phone"),
+                "lat": data.get("lat"),
+                "lon": data.get("lon"),
+                "country": data.get("US"),
```
iandees, line 26: You probably meant for this to just be `"US"`. You might also consider putting it up in `item_attributes`.

```suggestion
                "country": "US",
```

Romeo-PHILLIPS: Good catch thanks !

PR #4129, opened 2022-10-20
Filter potential NSI matches by country code

### LINKED ISSUE #3795: Enforce 2 digit country code
As noted in https://github.com/alltheplaces/alltheplaces/pull/3791#pullrequestreview-1071740124, the spiders are currently inconstant in the country output.

I think we all agree that we should follow [OSM](https://wiki.openstreetmap.org/wiki/Key:addr:country), and use "ISO 3166-1 alpha-2".

I can envision a wrapper function, like `ensure_country_code("United Kingdom")` that returns "GB". I'm leaning towards this being in spider code (and shared parsers like the ld one) not a pipeline.

Cj-Malone:
Also add a pipeline to try and clean up countries codes. Fixes #3795

### DIFF
--- a/locations/pipelines.py
+++ b/locations/pipelines.py
@@ -9,6 +9,7 @@ import re
 
 from scrapy.exceptions import DropItem
 
+from locations.commands.insights import CountryUtils
 from locations.name_suggestion_index import NSI
 
 
@@ -44,6 +45,18 @@ class ApplySpiderLevelAttributesPipeline(object):
         for (key, value) in item_attributes.items():
             if item.get(key) is None:
                 item[key] = value
+
+        return item
+
+
+class CountryCodeCleanUpPipeline(object):
+    def __init__(self):
+        self.country_utils = CountryUtils()
+
+    def process_item(self, item, spider):
+        if country := item.get("country"):
+            if clean_country := self.country_utils.to_iso_alpha2_country_code(country):
+                item["country"] = clean_country
 
         return item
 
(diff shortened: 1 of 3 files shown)


Cj-Malone:
ALDI South is a decent example, it has 2 entries in NSI with that [QID](https://nsi.guide/index.html?t=brands&k=shop&v=supermarket&tt=Q41171672).

There is still a couple issues I can think of: Multiple types for the correct establishment eg [bank vs atm](https://nsi.guide/index.html?t=brands&tt=Q627381). Multiple entries for the correct, but different brands eg [Starbucks vs Starbucks Reserve](https://nsi.guide/index.html?t=brands&k=amenity&v=cafe&tt=Q37158). And also `operator` drama, we have a bunch of scrapers with `brand_wikidata`, but NSI has them as `operator:wikidata`.

But for now, lets see the numbers from the next weekly run.

Cj-Malone:
2022-10-22-13-31-48
  | atp/nsi/brand_missing | atp/nsi/perfect_match | atp/nsi/cc_match | atp/nsi/match_failed | atp/field/brand_wikidata/missing
-- | -- | -- | -- | -- | --
Spider Count | 112 | 387 | 64 | 124 | 62
Item Count | 310,260 | 510,800 | 172,775| 517,270 | 24,196

Item Count is actually not a great indicator at the moment, western_union is +-200,000 items each run, and spiders are rapidly dying.

I don't know if I'll get around to altering the pipeline before the next weekly, but I will for the following. + the TODO list is in the comment above if anyone else wants a crack at it.

PR #4135, opened 2022-10-21
feat: Add Safelite spider

### LINKED ISSUE #1571: Safelite
https://www.safelite.com/store-locator/store-locations-by-state

rubyruins:
Fixes #1571.

### DIFF
--- a/locations/spiders/safelite.py
+++ b/locations/spiders/safelite.py
@@ -1,0 +1,103 @@
+# -*- coding: utf-8 -*-
+import scrapy
+from locations.items import GeojsonPointItem
+from locations.hours import OpeningHours, DAYS_FULL
+
+
+class SafeliteSpider(scrapy.Spider):
+    name = "safelite"
+    allowed_domains = ["www.safelite.com"]
+    item_attributes = {"brand": "Safelite", "brand_wikidata": "Q28797369"}
+    start_urls = ("https://www.safelite.com/store-locator/store-locations-by-state",)
+
+    def parse(self, response):
+        state_urls = response.xpath(
+            './/div[@class="state-store-list"]/a/@href'
+        ).extract()
+        for state_url in state_urls:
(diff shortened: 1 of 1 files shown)


locations/spiders/safelite.py
```diff
@@ -0,0 +1,174 @@
...
+                .xpath(".//p/text()")
+                .extract_first()
+            )
+            addr_full = addr_full.split(",")
+            street_address = ", ".join(addr_full[:-2])
+            city = addr_full[-2].strip()
+            state_postcode = addr_full[-1].strip()
+            state, postcode = state_postcode.strip().split(" ")
+            state = state.strip()
+            state = STATES.get(state, state)
```
iandees, line 103: Same thing on state name expansion here.

locations/spiders/safelite.py
```diff
@@ -0,0 +1,174 @@
...
+            addr_full = ", ".join(addr_full)
```
iandees, line 104: Don't need addr_full with the other info you're adding.

locations/spiders/safelite.py
```diff
@@ -0,0 +1,174 @@
...
+            lat = response.xpath(
+                './/div[@class="store-map"]/@data-start-lat'
+            ).extract_first()
+            lon = response.xpath(
+                './/div[@class="store-map"]/@data-start-lon'
+            ).extract_first()
+        except:
```
iandees, line 111: If there's an error, we should let the exception bubble up instead of outputting some empty data.

rubyruins: The exception is caused in cases where the store does not have any location data and only opening hours data. However, the HTML structure for this data is pretty much the same as the location data, without differences in any classes or tags. Hence, the only alternative here was to put it in a try catch block since pages without location selectors would fail the split clause and directly pick up the store timings.

PR #4140, opened 2022-10-24
Standardise Browser User Agents

Cj-Malone:


Cj-Malone:
farmershomefurniture is crashing on Opening Hours, I'll sort it in another PR, don't want to make GitHub run the CI again. dominos_pizza_gb only works in GB, the others are slow starters.

The idea behind this is primally for update ability, we're ignoring them in spiders, hopefully we'll ignore them less here. Also Black butchers some of them, this looks better.

PR #4147, opened 2022-10-27
Adding spider for Pet Valu

boomerwv1:


mjoe999:
Hi Boomer,

The API call you have is efficient, but StructuredDataSpider and:
https://store.petvalu.ca/sitemap.xml
... would give you a lot richer data with less code?
If you do want to stick with the API call then item = DictParser.parse(json_entry)
should help with less code and more data extracted.
Up to you!
Cheers,
Joe.

boomerwv1:
Thanks @mjoe999, went with DictParser and it seems to have worked.

locations/spiders/petvalu.py
```diff
@@ -0,0 +1,24 @@
...
+    item_attributes = {"brand": "Pet Valu", "brand_wikidata": "Q58009635"}
+    allowed_domains = ["store.petvalu.ca"]
+    start_urls = (
+        "https://store.petvalu.ca/modules/multilocation/?near_location=toronto&threshold=40000000000000&geocoder_components=country:CA&distance_unit=km&limit=20000000000&services__in=&language_code=en-us&published=1&within_business=true",
+    )
+
+    def parse(self, response):
+        data = response.json()
+        for i in data["objects"]:
+            item = DictParser.parse(i)
```
Cj-Malone, line 23: DictParser isn't perfect, we have to clean up something before, or after it. In this case, `street` is actually the `street_address`. DictParser can't handle both `addr_full` and smaller address fields, so we could set it after:

```suggestion
            i["street_address"] = i.pop("street")
            item = DictParser.parse(i)
            item["addr_full"] = i["formatted_address"]
```

Cj-Malone (commented):
There is also some other nice to haves in the source data like `location_url` and `hours` if they are easy enough to parse.

boomerwv1:
@Cj-Malone , updated to include the location URLS.

Cj-Malone (approved):

PR #4151, opened 2022-10-28
extend country code clean up pipeline

mjoe999:
CJ's recent commit to try and standardise country code fields to ISO2 alpha as best we can has been extended further with code that will set the country code from our spider naming convention if appropriate and also as a final resort the item website URL.

As part of this CountryUtils has been pulled out of the insights.py command file and put in "core" ATP library code where it should be.

This means we will get more POIs with country codes which means that the NSI categorisation pipeline code should pick up more items.

Also means that going forward people in many cases will not need to set the country code field "by hand" in the spider.

### DIFF
--- a/locations/commands/insights.py
+++ b/locations/commands/insights.py
@@ -1,9 +1,9 @@
 import json
-import geonamescache
 import os
 import requests
 from collections import Counter
 from locations.name_suggestion_index import NSI
+from locations.country_utils import CountryUtils
 from scrapy.commands import ScrapyCommand
 from scrapy.exceptions import UsageError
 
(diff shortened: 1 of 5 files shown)


iandees:
Hmm, I'll merge this in and see how it goes, but I don't think the website or spider name _always_ correspond directly with what country code we should set in the output data.

Cj-Malone:
I feel like we've been pretty good for country codes, in spider names (except US) lately, and item_atteibutes, but I'll put it on my todo list to check the output after the next weekly.

mjoe999:
@iandees thanks for that, and I do agree with your concern. I'd say we go more steps forward than back though.
Down the road it would be nice to have a check of the country code against the lat/lon for sanity. Yes you would think that lat/lon would be able to give us country code but there are many lat/lon errors in scraped data also. The two way check could prove useful.

I was thinking of doing a PR with a few small doc/README clarifications on practice e.g. spider naming convention. Sound like a good thing to do?

PR #4169, opened 2022-11-01
Fix Intermarche Spider

Cj-Malone:
@Romeo-PHILLIPS perhaps you could provide some insight on this one? Particularly around `modelLabel` at line 54+, and maybe [NSI](https://nsi.guide/index.html?t=brands&tt=Intermarch%C3%A9) isn't perfect?

### DIFF
--- a/locations/spiders/intermarche.py
+++ b/locations/spiders/intermarche.py
@@ -1,47 +1,72 @@
-# -*- coding: utf-8 -*-
-import re
-
 import scrapy
+from scrapy import Request
 
-from locations.items import GeojsonPointItem
-from locations.hours import OpeningHours
+from locations.dict_parser import DictParser
+from locations.hours import DAYS, OpeningHours
 
 
 class IntermarcheSpider(scrapy.Spider):
     name = "intermarche"
     allowed_domains = ["intermarche.com"]
-    # TODO: Needs work in the spider I would think.
-    # TODO: item_attributes = {"brand": "Intermarché", "brand_wikidata": "Q3153200"}
-    # TODO: item_attributes = {"brand": "Intermarché Express", "brand_wikidata": "Q98278043"}
+    INTERMARCHE_SUPER = {"brand": "Intermarché Super", "brand_wikidata": "Q3153200"}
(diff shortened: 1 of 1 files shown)


Romeo-PHILLIPS:
The NSI seems correct as this super-market chain offers gas stations (not all of them, those that have `"entityType": "GAS_STATION"`). 

For the brand names they should all be `Intermarché` CONTACT/DRIVE/SUPER/HYPER are just `store_types` to reflect on size of the store and what they do.

`Retrait La Poste` is weird because in some cases it just looks like a post office but nothing to do with Intermarché.

And `Réservé Soignants` are just drive through stores reserved for medical workers.

Cj-Malone:
NSI fixed: https://github.com/osmlab/name-suggestion-index/pull/7305

Romeo-PHILLIPS:
I think you might want to convert `SUPER ALIMENTAIRE` to -> `Intermarché Super`

Cj-Malone:
Yeah, I'm currently pushing `SUPER ALIMENTAIRE` and `SUPER GENERALISTE` [through](https://github.com/alltheplaces/alltheplaces/pull/4169/files#diff-1d96f33816271a731982a2a2b1872787c5a42060e4b01129de9fff44f2f0ea3aR70), because I didn't know the difference. Happy to remove it if the difference is little, or correct the output if need be.

Romeo-PHILLIPS:
> Yeah, I'm currently pushing `SUPER ALIMENTAIRE` and `SUPER GENERALISTE` [through](https://github.com/alltheplaces/alltheplaces/pull/4169/files#diff-1d96f33816271a731982a2a2b1872787c5a42060e4b01129de9fff44f2f0ea3aR70), because I didn't know the difference. Happy to remove it if the difference is little, or correct the output if need be.

Yeah they both fall into `Intermarché Super` brand

Cj-Malone:
Are they both "supermarkets" or is one closer to a hardware store, or "houseware"?

https://wiki.openstreetmap.org/wiki/Key:shop

Romeo-PHILLIPS:
> Are they both "supermarkets" or is one closer to a hardware store, or "houseware"?
> 
> https://wiki.openstreetmap.org/wiki/Key:shop

Both supermarkets

PR #4175, opened 2022-11-01
fix: extra_space_storage

Romeo-PHILLIPS:


Cj-Malone (approved):
LGTM.

If this keeps happening we should think about pointing `sitemap_urls` to `robots.txt` with decent `sitemap_rules` or `sitemap_follows` it shouldn't slow the spiders down too much, and may be more flexible to these kind of changes.

PR #4176, opened 2022-11-02
Fix extended_stay_america

Cj-Malone:


locations/spiders/extended_stay_america.py
```diff
@@ -1,52 +1,16 @@
-# -*- coding: utf-8 -*-
-import json
-import re
+from scrapy.spiders import SitemapSpider
 
-import scrapy
+from locations.structured_data_spider import StructuredDataSpider
+from locations.user_agents import BROSWER_DEFAULT
```
jleedev, line 4: > BROSWER

Cj-Malone: Well... fuck

Cj-Malone: I've fixed it here, and I'll put it on my TODO list to fix it in the other spiders

Cj-Malone:
Looks like the GitHub CI ip is now being blocked by them, we can hope that it's better on the AWS server for the weekly, but it's probably one that needs to be split into several proxies.

PR #4181, opened 2022-11-03
Add NaPTAN spider

Cj-Malone:
So NaPTAN is a gov dataset of public transit in England, Wales and Scotland. I don't know if it's in scope for ATP but OSM UK is considering a Quarterly Project using the data, so I wrote it for that, and here seems like a decent home for it.

BTW `HTTPCACHE_ENABLED` and `HTTPCACHE_EXPIRATION_SECS` were useful while testing this, I don't know if it makes sense to be set in `settings.py`, or for me to continue to use it locally. I don't think it'd be affect the weeklies as there is no persistence between them.

### DIFF
--- a/locations/spiders/naptan_gb.py
+++ b/locations/spiders/naptan_gb.py
@@ -5,76 +5,22 @@ from locations.items import GeojsonPointItem
 
 class NaptanGBSpider(Spider):
     name = "naptan_gb"
+    # Smaller areas can be tested: eg "&atcoAreaCodes=010"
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Yeah so it includes airports, train stations, bus stations, and individual stops. In terms of numbers, the individual stops are obviously the most common...

Cj-Malone:
I've dropped the individual stops, I'll keep a [local branch](https://github.com/Cj-Malone/alltheplaces/tree/naptan_gb) for OSMUK

PR #4187, opened 2022-11-06
More docs and one associated spider edit

mjoe999:
This writing game is harder work than code! You get to the point of boring yourself. Once again I have expressed some sentiments that might not be agreed with or are sloppily worded etc. @iandees: sorry to make you read this! We can chop it however you like but I think we start to have the some reasonably useful "bones".


iandees:
Thanks for writing the docs!

PR #4189, opened 2022-11-07
Add spider for Macy's Backstage

jleedev:
Many are departments of Macy's stores, but Wikipedia claims that at least 9 are freestanding.


iandees:
@Cj-Malone looks like the structured data spider is creating an array for street_address on this one. Can you spot an easy way to fix that globally or should it happen on this spider specifically?

jleedev:
Context:

```
 ¶ <macys_backstage.geojson jq '.features[].properties."addr:street_address"' | tail -n 20
"401 S Lake Ave"
"72780 Highway 111"
[
  "Unit 2000",
  "9301 Tampa Ave"
]
"341 Sun Valley Mall"
[
  "Ste 0",
  "555 Broadway St"
]
"2559 El Camino Real"
"21900 Vanowen St"
"200 E Cypress Ave"
"200 Brea Mall"
"2601 Ming Ave"
```

Probably similar to the phone number treatment FWIW.

Cj-Malone:
That should work, I did a couple other fixes at the same time, hope neither of you mind.

Cj-Malone:
But also, I don't think this is to spec, and this is the first time I've see it, so arguably it should be fixed in `pre_process_data`.

But it works, and I've no strong opinion right now.

jleedev:
Agree that it's not obvious what to do here based on what the specifications suggest. Might prefer a semicolon. But it's a good improvement.

When faced with an array value, tippecanoe json encodes it, which is suboptimal on that end, so this is also better.

It might be somewhat non obvious what's going on when debugging a site with all the automatic translations going on; some sort of debug logging flag might help. Perhaps as a flag to the `scrapy sd` command to verbosely say "here's the original microdata, here's every block of json-ld we considered, here's the item before pipelining, here's the finished item."

Cj-Malone:
> Might prefer a semicolon

I'd agree for most values, we want to match the OSM for multiple values (semicolon seperated), but in this case for the street address it's a single value with extra details. I prefer a comma. I guess that's an argument it move it into spider code, because the next instance of this occurring might be multiple values, but I don't have a strong opinion on that, we can fix it when/if we need to.

> "here's the original microdata, here's every block of json-ld we considered, here's the item before pipelining, here's the finished item."

Yeah, I agree that could be helpful.

> The `schema:Place` type might be a bad default

Unfortunately I don't think the default `wanted_types` will ever be perfect... I've seen sites with a small generic `LocalBusiness`, then a detailed `ConvenienceStore`, the current default `wanted_types` isn't great for that either, but `LocalBusiness` is so common we save more by leaving it in. I don't know how common `Place` is, so I don't really have a strong opinion on it, but I also don't think it's realistic to expect this to work perfectly in every situation.

> produces two items and you have to figure out why

Yeah, I'd agree, even just printing the type with it would help.

Cj-Malone:
@iandees anything need doing to get this one merged?

PR #4194, opened 2022-11-08
Add uBreak iFix Spider

Cj-Malone:
I wanna rewrite the structured data opening hours parser at some point, so extra tests or documented cases of things not being parsed when maybe they should be would be appreciated.


iandees:
Whenever I think of the opening hours problem I think that maybe we could convince @albarrentine to write a system like libpostal that tries to parse arbitrary structs/strings into a consistent opening hours output.

PR #4199, opened 2022-11-09
Convert JDSports GB Spider StructuredDataSpider

Cj-Malone:
It's broken in the weekly, but worked fine locally, presumably they block AWS ips. `download_delay` increased to the default 1 seconds, but still don't really expect this to work in the weeklies.


iandees:
What do you think about adding a "requires_proxy" field in the spider? Either a boolean value or a list of country codes for proxy source. Then in the future we can use that to send requests through a proxy if we go that route.

Cj-Malone:
Yeah, I agree in principle, the implementation will probably have to change once we implement it, so for now I'll just add `requires_proxy = True`.

PR #4205, opened 2022-11-11
Add spider for Coop Vitality (Switzerland)

brawer:


brawer:
Sorry about the lint error; fixed.

Cj-Malone:
It's a shame their sitemap doesn't include stores, because they have good [structured data available](https://validator.schema.org/#url=https%3A%2F%2Fwww.coopvitality.ch%2Fde%2Fcoop-vitality-crissier).

But anyway, this is a great first spider, thank you. And the lint fix was appreciated too.

You're clearly an OSM mapper with this tagging :p, we try to stay pretty close to OSM tags but don't always go all the way. eg we often give up with unit, housenumber/name and street and just use `street_address`. I think our `name` is pretty much OSMs `branch` tag too, or it could be with a little bit of cleaning. IMO data consumers should probably be using `brand` over `name` in most rendering cases, same goes for OSM consumers.

PR #4208, opened 2022-11-14
Add spider for Valora brands

brawer:


locations/spiders/valora.py
```diff
@@ -0,0 +1,217 @@
...
+        "ServiceStoreDB": ("ServiceStore DB", "Q84482517", {}),
+        "Spettacolo": ("Caffè Spettacolo", "Q111728781", {}),
+        "U-Store": ("U-Store", "Q113290511", {}),
+        # However, the AllThePlaces pipeline does not apply any NSI rules
+        # if multiple rules match with conflicting tags. Perhaps the NSI data
+        # is buggy when there’s such a conflict; anyhow, by making an explicit
+        # decision here, we can work around this.
+        # See https://github.com/osmlab/name-suggestion-index/issues/7400.
+        "k kiosk": ("k kiosk", "Q60381703", {"shop": "kiosk;newsagent"}),
+        "Press & Books": ("Press & Books", "Q100407277", {"shop": "books;newsagent"}),
```
Cj-Malone, line 33: I think I'd prefer you to pick one shop type for each of these and use it via [categories](https://github.com/alltheplaces/alltheplaces/blob/master/locations/categories.py).

Cj-Malone:
> the AllThePlaces pipeline does not apply any NSI rules if multiple rules match with conflicting tags. Perhaps the NSI data is buggy when there’s such a conflict

In this case, I think NSI probably has duplicates and one of each should be removed.

However it is valid for NSI to have multiple entries for one wikidata id, and it's ATP that need better matching. I started the NSI matching relatively recently but haven't done anything in a couple weeks. My current [TODO list](https://github.com/alltheplaces/alltheplaces/pull/4129#issuecomment-1285400462) if you have any input on it.

Cj-Malone (approved):
I feel better about the shop categories, thank you.

The data output is great, thanks for all the effort. You seem to be more of a python person than me, so I can't really make specific comments on the code. However I'm slightly worried, that a) you're gonna burnout with the data cleaning, and 2) when this spider breaks (all spiders eventually break) it'll be hard for someone else fix it.

Maybe "less is more", "perfect is the enemy of good" etc.

brawer:
> Maybe "less is more", "perfect is the enemy of good" etc.

Agreed. So what shall we do about the current code?

Cj-Malone:
Ian will make the final decision, but I expect it'll be fine.

iandees:
As CJ said, thanks for the contribution. I particularly appreciate the comments in here!

PR #4219, opened 2022-11-16
Additional burger king restaurants (BE, LU, ES, PT, IT)

jancallewaert:


Cj-Malone:
The black test is a linting one, don't worry about it. If it isn't fixed before this is merged, a follow up PR will happen automatically.

PR #4226, opened 2022-11-17
Add base Woosmap Spider

Cj-Malone:
I started the day off thinking that Woosmap was third party data. But it's not, it's first party data on a hosted store finder. And I found a better end point that we don't have to mess around with lat/lons or grids.

I want everyones opinions on some kind of base spiders, specifically for third party store finders. @jleedev @iandees @mjoe999 You can checkout the specific implementation, but I just rushed it, it's the idea that really want your opinions on.

### DIFF
--- a/locations/spiders/e_leclerc.py
+++ b/locations/spiders/e_leclerc.py
@@ -1,85 +1,17 @@
-import scrapy
-
-from locations.hours import OpeningHours
-from locations.items import GeojsonPointItem
+from locations.storefinders.woosmap import WoosmapSpider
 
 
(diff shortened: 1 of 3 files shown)


iandees:
I think this is great. It's along the same lines as the structured data parsers.

mjoe999:
Concur. Looks good, and glad to see you have not put it in the locations/ top level which is starting to look a little crowded (for good reasons)!

Romeo-PHILLIPS:
This is very cool good job

Cj-Malone:
I'm glad everyone likes it. If I get time tomorrow I'll work on a Uberall one.

PR #4229, opened 2022-11-18
Use an offline reverse geocoder to assign country codes

brawer:
Called as a last resort when none of the other methods worked.

After this change, the `ich_tanke_strom` spider emits features with the following `addr:country` tags. From inspecting a few items, the country codes seem to get assigned correctly.

`{'CH': 4465, 'DE': 180, 'AT': 97, 'SK': 62, 'IT': 16, 'LI': 13, 'FR': 12, 'PL': 6, 'FI': 4, 'SH': 3, 'IL': 3, 'HU': 3, 'NL': 3, 'UA': 2, 'LT': 1, 'AU': 1, 'BE': 1, 'CY': 1, 'LA': 1}`


locations/pipelines.py
```diff
@@ -75,6 +76,15 @@ def process_item(self, item, spider):
                 spider.crawler.stats.inc_value("atp/field/country/from_website_url")
                 item["country"] = country
 
+            # Still no country set, try an offline reverse geocoder.
+            lat, lon = item.get("lat"), item.get("lon")
+            if lat != None and lon != None:
```
Cj-Malone, line 81: Can this be condensed to:

```suggestion
            if lat and lon:
```

(I've not actually checked this)

brawer: That would exclude anything on the equator or the prime meridian.

Cj-Malone: I think that's quite an edge case, but yeah lets not make a regression.

mjoe999:
It's funny how life goes sometimes. I started the day with a problem to solve: Given a list of locations / cities that are known to be in a single country .... but which one?

I was going to do something with geonames and capitals etc. This will be way better I hope. I'm glad I got diverted and then saw this before I did anything :-)

Good find I think.

PR #4234, opened 2022-11-18
Add spider for Stadt Zürich, 2425 places

brawer:
There’s a _lot_ more data that could be converted here, but let’s start small.

### DIFF
--- a/locations/spiders/stadt_zuerich_ch.py
+++ b/locations/spiders/stadt_zuerich_ch.py
@@ -112,6 +112,8 @@ class StadtZuerichCHSpider(scrapy.Spider):
         return {}
 
     def parse_bicycle_parking(self, p):
+        # For bicycle and motorcycle parkings, the data feed puts the
+        # feature type into the name; the parkings have no real names.
         amenity = {
             "Motorrad": "motorcycle_parking",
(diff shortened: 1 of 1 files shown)


locations/spiders/stadt_zuerich_ch.py
```diff
@@ -0,0 +1,201 @@
...
+        return oh.as_opening_hours()
+
+    def parse_park(self, p):
+        return {"leisure": "park"}
+
+    def parse_phone(self, p):
+        phone = (p.get("tel") or "").replace(" ", "")
+        if phone.startswith("0"):
+            phone = "+41" + phone[1:]
+        return phone
```
mjoe999, line 173: This is great and diligent. I've always kind of resisted doing the phone "clean" as I feel a more structural pipeline approach will do it for us. This needs knowing the country code. But hey, you're already guiding towards a happier place in this context with:
https://github.com/alltheplaces/alltheplaces/pull/4229

brawer: Once we know the country code (after #4229 is merged), normalizing phone numbers would actually be quite easy. I’ve added a TODO comment.

PR #4246, opened 2022-11-22
Implement the "continue" part of "there is a microdata error; continue".

jleedev:
Add a test case that was triggering this and a reasonable enough parse.

(Observation: The validator.schema.org parser creates an event with the same location property twice.)


Cj-Malone (approved):
Thank you, this indeed fixes my crash.

PR #4253, opened 2022-11-23
Add Uberall base spider

Cj-Malone:


iandees:
Idea: a command line helper that you pass a brand webpage to and it tries to figure out if it could use one of these well-known spider types.

Cj-Malone:
Yeah it would probably get more useful when there are more of these common spiders.

I've been wondering if we could do a big comparison to NSI to see how many entries we don't have spiders for.

Or how many could we generate automatically? With common spiders, or Structured Data Spiders?

For the [perfect case](https://github.com/alltheplaces/alltheplaces/blob/812f9fabc99b25cac8f68535c1bddd1d445208f6/locations/spiders/macys_backstage.py), if NSI or Wikidata had some kind of `matchUrls` then we could drop the `sitemap_rules`, in most cases `sitemap_urls` is documented in robots.txt, or guessable. We can extend the default `wanted_types`, or could NSI house it, like `matchTags`?

PR #4280, opened 2022-12-05
Add nhs_england_gb (~15,000 locations)

mjoe999:
So the NHS is big. Very big. Just doctors and dentists in England add 15K sites.
One contentious change is to structured_data_spider.py
What was there before is obviously wrong.
But us my change good? @Cj-Malone ?


Cj-Malone:
I can't remember that, but yeah obviously wrong. Maybe it was meant to be if website is not none, but even then, isinstance none list, works as you'd expect.

So yeah, LGTM.

PR #4291, opened 2022-12-08
Drop empty microdata objects

### LINKED ISSUE #4284: lloyds_bank_gb spider doesn't capture addresses
I'm not sure how easy / feasible this would be to fix, but the lloyds_bank_gb spider doesn't currently pick up the postcode or address for each branch. I don't know the details of what the spidering framework can detect by default, but the address data does seem to be included in the HTML in some sort of structured way.

Cj-Malone:
Fixes #4284

### DIFF
--- a/locations/microdata_parser.py
+++ b/locations/microdata_parser.py
@@ -255,6 +255,10 @@ def convert_item(item):
     if itemid := item.get("id"):
         ld["@id"] = itemid
     if len(item["properties"].items()) == 0:
+        # Guard against goofy use of microdata such as:
+        # <a itemscope itemprop=address itemtype=PostalAddress>...</a>
+        # Which produce an item with no properties and make it difficult to
+        # parse the correct item later down. See test_multiple_addresses
         return
     for k, v in item["properties"].items():
         ld[k] = filter(
--- a/locations/microdata_parser.py
+++ b/locations/microdata_parser.py
@@ -254,8 +254,12 @@ def convert_item(item):
     # Properties is a list; if its length is 1 then flatten, else don't.
     if itemid := item.get("id"):
         ld["@id"] = itemid
+    if len(item["properties"].items()) == 0:
+        return
     for k, v in item["properties"].items():
-        ld[k] = [convert_item(val) if isinstance(val, dict) else val for val in v]
+        ld[k] = filter(
+            None, [convert_item(val) if isinstance(val, dict) else val for val in v]
+        )
         ld[k] = remove_duplicates(ld[k])
         if len(ld[k]) == 1:
             ld[k] = ld[k][0]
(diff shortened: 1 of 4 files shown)


Cj-Malone:
hmmm, @jleedev implementation is wrong, or the test?

Adding a name to the test keeps it from being dropped.
```patch
diff --git a/tests/test_microdata_parser.py b/tests/test_microdata_parser.py
index f02da4aa..7a3c3cc5 100644
--- a/tests/test_microdata_parser.py
+++ b/tests/test_microdata_parser.py
@@ -1,3 +1,4 @@
+
 from parsel.selector import Selector
 
 from locations.microdata_parser import MicrodataParser
@@ -7,7 +8,7 @@ def test_itemref():
     src = """
 <main itemscope itemtype="http://schema.org/GroceryStore">
     <article itemref="schema-location" itemprop="event" itemscope itemtype="http://schema.org/Event">
-        <div id="schema-location" itemprop="location" itemscope itemtype="http://schema.org/Place"></div>
+        <div id="schema-location" itemprop="location" itemscope itemtype="http://schema.org/Place"><span itemprop="name">Test</span></div>
     </article>
 </main>
     """
@@ -25,7 +26,7 @@ def test_itemref():
                                 "location": [
                                     {
                                         "type": ["http://schema.org/Place"],
-                                        "properties": {},
+                                        "properties": {"name": ["Test"]},
                                     }
                                 ]
                             },
@@ -39,5 +40,5 @@ def test_itemref():
     assert ld == {
         "@context": "https://schema.org",
         "@type": "GroceryStore",
-        "event": {"@type": "Event", "location": {"@type": "Place"}},
+        "event": {"@type": "Event", "location": {"@type": "Place", "name": "Test"}},
     }
```

locations/microdata_parser.py
```diff
@@ -254,8 +254,12 @@ def convert_item(item):
     # Properties is a list; if its length is 1 then flatten, else don't.
     if itemid := item.get("id"):
         ld["@id"] = itemid
+    if len(item["properties"].items()) == 0:
+        return
```
jleedev, line 258:
```suggestion
    if len(item["properties"].items()) == 0:
        # Guard against goofy use of microdata such as:
        # <a itemscope itemprop=address itemtype=PostalAddress>...</a>
        # Which produce an item with no properties and make it difficult to
        # parse the correct item later down.
        return
```

jleedev:
> hmmm, @jleedev implementation is wrong, or the test?
> 
> Adding a name to the test keeps it from being dropped.

Makes sense, yep.

Could also add the problematic snippet from https://branches.lloydsbank.com/aldershot/115-victoria-road if we want to.

Cj-Malone:
I don't have an explanation to why all the tests are failing on GitHub now. They work locally.

jleedev (approved):

iandees:
> I don't have an explanation to why all the tests are failing on GitHub now. They work locally.

Looks like something in pytest is using a Python 3.11 feature that doesn't exist in the version that setup-python creates by default.

jleedev:
Changing `pipenv install` to `pipenv sync` would prevent that.

PR #4293, opened 2022-12-09
fix Averitt spider

abbensid2:
This is a fix for the Averitt spider

### DIFF
--- a/locations/spiders/averitt.py
+++ b/locations/spiders/averitt.py
@@ -1,69 +1,118 @@
-import re
-
 import scrapy
-
 from locations.items import GeojsonPointItem
 
 
 class AverittSpider(scrapy.Spider):
     name = "averitt"
     item_attributes = {"brand": "Averitt Express", "brand_wikidata": "Q4828320"}
-    allowed_domains = ["averittexpress.com"]
+    allowed_domains = ["averitt.com"]
     start_urls = [
-        "https://www.averittexpress.com/resources/facility-locator",
+        "https://www.averitt.com/locations",
     ]
 
     def parse(self, response):
         states = response.xpath('//select[@name="state"]/option/@value').extract()
         for state in states:
-            if state == "0":
+            if state == "State":
                 continue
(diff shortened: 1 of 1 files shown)


locations/spiders/averitt.py
```diff
@@ -1,69 +1,118 @@
...
                 callback=self.parse_state,
             )
 
     def parse_state(self, response):
         centers = response.xpath(
-            '//section[contains(@class, "content-main")]//div[@class="grid"]//a/@href'
+            '//*[@id="hs_cos_wrapper_widget_1613430615107_"]/b/a/@href'
         ).extract()
         for center in centers:
             yield scrapy.Request(response.urljoin(center), callback=self.parse_center)
```
Cj-Malone, line 28: This page detection could probably be simplified by a SitemapSpider.
https://www.averitt.com/sitemap.xml

locations/spiders/averitt.py
```diff
@@ -1,69 +1,118 @@
...
-            "state": response.xpath(
-                'normalize-space(//span[@class="region"]//text())'
-            ).extract_first(),
-            "postcode": response.xpath(
-                'normalize-space(//span[@class="postal-code"]//text())'
-            ).extract_first(),
-            "website": response.url,
-            "lat": float(lat) if lat else None,
-            "lon": float(lon) if lon else None,
+            "ref": "_" + ref_map_url,
```
Cj-Malone, line 108: I'm not in love with using this Google id as our ref. I doubt there is any real issue with it, but we could just use `response.url`

Cj-Malone: Also, we could potentially use `extract_google_position` to get the coords. It's not perfect but it may work on this site.

locations/spiders/averitt.py
abbensid2: You are right, I fixed it, thanks.

Cj-Malone (approved):
I don't like the xpaths, but it's better than a broken spider. Thank you

PR #4298, opened 2022-12-09
City of Zürich: Add street lamps

brawer:


iandees:
I need to think about this one. I don't know that street lamps count as "places" to me.

Cj-Malone:
I don't have a major opinion on this, but I think historically ATP has very much been about shops and amenities, not just any data. I think now we've got categorisation the differentiation is less important, ATP could become a nice place for ingesting, technically ingesting and standardising the data, but also ingesting to the OSM community.

brawer:
Personally, I think there’s a need in the overall OSM “ecosystem” for some infrastructure to process third-party feeds and websites, that is: fetching the data from upstream sources; converting it to a standard format (to GeoJSON); normalizing the schema (to OSM tags); caching the results; and ultimately finding the differences to OSM and doing something about them. Of course this piece of infrastructure doesn’t have to be AllThePlaces, but like the Zürich example shows, scraping store finders isn’t really very different from interfacing with a municipal GIS. Admittedly, the world’s cities publish all kinds of datasets, including trees, railway tracks, street geometry, roof shapes, and various other data — there’s hardly any limit. Personally I’d hope that ATP could become the place to convert all these feeds (otherwise, one would eventually come up with an ATP clone, meaning a fair amount of duplication). But I can also understand if you consider such things excessively out of scope; there’s always the risk of losing focus.

iandees:
Yea, this is interesting. I started ATP to do exactly what you describe: collect locations, normalize to GeoJSON, and make it easy to merge that GeoJSON into OSM. But the political aspect of bringing data into OSM became so off-putting to me that I decided to focus on this as a dataset by itself. It became useful to some people in that form, but now we have a chance to bring it back to OSM and expand to other types of features.

Your point about interacting with GIS is poignant. Our sister project, OpenAddresses, has gotten a little stale and it could be that we should fold the address/building/parcel data sources into ATP. We could write Scrapy spiders that do the GIS data downloading. We'd have to reconcile how the OA source JSON converts to ATP spiders, but they are essentially the same thing expressed in a different way.

🤔

Cj-Malone:
I've come to the same conclusion with regards to imports or bulk edits in OSM, I get away with a fair bit in GB but if I did it elsewhere there'd be drama.

It would be interesting to work with local chapters or communities to get them data, and help them work with it themselves.

I'd be up for helping write some OA type spiders.

brawer:
So what’s the verdict?

iandees:
Let's do it.

iandees:
Speaking from the perspective of a data consumer (we use ATP at work), since we are adding non-place data we'll need to think of a way to let data consumers discern between the two. Any ideas for how that might work?

brawer:
> [S]ince we are adding non-place data we'll need to think of a way to let data consumers discern between the two. Any ideas for how that might work?

What's the definition of a place? Any feature tagged as `amenity` or `shop`?

iandees:
A "place" is a place where people congregate to do business.

OSM doesn't really have a set of tags that match this sort of thing, but things with `amenity`, `shop`, `tourism`, or `leisure` tend to fall into what I would call a place.

We could suggest that data consumers filter on OSM tags, but most of the ATP data doesn't have OSM tags/categorization associated with it yet.

brawer:
> most of the ATP data doesn't have OSM tags

Ouch, can this be fixed? Is the problem that spiders don't emit `brand:wikidata`, or is NSI missing rules for those brands?

iandees:
It can be, but we decided on using OSM tags after we had 1000+ spiders and we haven't gone back and made sure they're all getting tags.

Cj-Malone:
> It can be, but we decided on using OSM tags after we had 1000+ spiders and we haven't gone back and made sure they're all getting tags.

I think one of the reasons that swayed you to OSM tags was that we already had Wikidata ids, so we get a bunch from NSI for free. You might think it's worse than it actually is.

> most of the ATP data doesn't have OSM tags

I don't have exact numbers, but more than half do. 1,161,541 out of 1,870,706 matched against NSI (atp/nsi/perfect_match, atp/nsi/cc_match, atp/nsi/category_march), so they definitely have categories.

> Ouch, can this be fixed? Is the problem that spiders don't emit brand:wikidata, or is NSI missing rules for those brands?

There is 473,865 that failed to match against NSI (atp/nsi/match_failed), half of them are fucking Western Union. Maybe we should manually flag them as `amenity=money_transfer`? The other half need better matching, or we need to manually apply categories.

141,582 items don't have Wikidata (atp/field/brand_wikidata/missing), we should add Wikidata to them. Although some of them do have categories like 12,873 from the Church of England spider. And there are some that don't need Wikidata, like the GBFS spider with 77,434 items.

And there are 69,121 items with Wikidata ids that aren't in NSI (atp/nsi/brand_missing). Some of them need NSI entries creating, some of them are too small to bother, some of them need different ids. And at least 18,671 (NHS) of them have categories anyway.

brawer:
See #4510 for a proposed PR that would count the number of items by category, including an `atp/category/missing` counter that would measure the size of the problem discussed here.

iandees:
Thanks for digging into it and finding numbers. Our wikidata coverage is much better than I thought. Thank you all for doing the work of adding that data!

I think the answer to my question is that data consumers should filter our data to match their use. In the past, I think my assumption was that all of the ATP data was "place-like", but that hasn't been the case since we added things like USPS collection boxes, light posts, or even libraries (some consumers might only want chain businesses). Adding wikidata and NSI IDs and then OSM tags means that consumers should use one of those sets of categorization to filter to the data they want to consume.

This should probably go in the README or something at some point.

PR #4300, opened 2022-12-10
fix Bcpizza spider

abbensid2:
This is a fix for the bcpizza spider

### DIFF
--- a/locations/spiders/bc_pizza.py
+++ b/locations/spiders/bc_pizza.py
@@ -1,84 +1,26 @@
 import re
-import scrapy
 from datetime import datetime
 from locations.hours import OpeningHours
 from locations.items import GeojsonPointItem
+from geonamescache import GeonamesCache
+from scrapy.spiders import CrawlSpider, Rule
+from scrapy.linkextractors import LinkExtractor
 from locations.google_url import extract_google_position
 
-# source: https://gist.github.com/rogerallen/1583593
-us_state_to_abbrev = {
-    "Alabama": "AL",
-    "Alaska": "AK",
-    "Arizona": "AZ",
-    "Arkansas": "AR",
-    "California": "CA",
-    "Colorado": "CO",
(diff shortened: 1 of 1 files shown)


locations/spiders/bc_pizza.py
```diff
@@ -1,95 +1,136 @@
...
+    "Wisconsin": "WI",
+    "Wyoming": "WY",
+    "District of Columbia": "DC",
+    "American Samoa": "AS",
+    "Guam": "GU",
+    "Northern Mariana Islands": "MP",
+    "Puerto Rico": "PR",
+    "United States Minor Outlying Islands": "UM",
+    "U.S. Virgin Islands": "VI",
+}
```
Cj-Malone, line 67: Semi recently I tried to remove these state arrays from spiders, swapping most of them to `GeonamesCache().get_us_states()`. In this case, I don't think it's that important to go from state name to state code, but if it was, it should be done in pipeline code so all US spiders benefit, not just this one.

locations/spiders/bc_pizza.py
```diff
@@ -1,95 +1,136 @@
...
-            lat = float(i["lat"])
-            lon = float(i["lng"])
-            hours = i["hours"]
-            hours = self.convert_hours(re.findall(regex_hours, hours))
-            phone = i["phone"]
-            website = i["facebook_url"]
-            addr_full = "{} {}, {} {}".format(street, city, state, postcode)
+        urls = response.xpath("//a[@rel='bookmark']/@href").extract()
+        for url in urls:
+            yield scrapy.Request(url, self.parse_locations)
```
Cj-Malone, line 79: This one could be a [sitemap spider](https://bc.pizza/portfolio-sitemap.xml), or a [CrawlSpider](https://docs.scrapy.org/en/latest/topics/spiders.html#crawlspider). Just a recommendation from me though, not a requirement because of how simple the current method is.

PR #4305, opened 2022-12-11
Add Ford dealerships

iandees:


JoeDixon-TomTom:
ford_dealers_us.py? There will be many more Ford dealerships world-wide.

iandees:
> ford_dealers_us.py? There will be many more Ford dealerships world-wide.

Yep, good call. I was going to try adding other countries but got distracted figuring out why this one wasn't returning any results. I'll make it US-specific for now.

PR #4321, opened 2022-12-12
Fix Brewdog Spider

abbensid2:
This is a fix for the brewdog spider


locations/spiders/brewdog.py
```diff
@@ -1,53 +1,41 @@
...
-            opening_hours=opening_hours.as_opening_hours(),
-        )
+        data = response.xpath('//script[@id="__NEXT_DATA__"]/text()').get()
+        data_json = json.loads(data)
+        bars = data_json["props"]["pageProps"]["content"]
+        item = GeojsonPointItem()
+        for bar in bars:
+            item["ref"] = bar["sys"]["id"]
+            item["name"] = bar["fields"]["name"]
+            item["addr_full"] = bar["fields"]["address"] if "address" in bar["fields"].keys() else None
```
Cj-Malone, line 23: This sort of thing can be avoided with `.get`. eg:
```python
item["addr_full"] = bar["fields"].get("address")
```
Or we can also pass a default, like a empty dict for nested ones.
```python
item["lat"] = bar["fields"].get("location", {}).get("lat")
```

Cj-Malone:
Note it's actually the ford spider failing, it's caught up in this PR because of the pre-commit.ci linting.

locations/spiders/brewdog.py
```diff
@@ -1,53 +1,40 @@
...
-            addr_full=addr,
-            lat=lat,
-            lon=lon,
-            opening_hours=opening_hours.as_opening_hours(),
-        )
+        data = response.xpath('//script[@id="__NEXT_DATA__"]/text()').get()
+        data_json = json.loads(data)
+        bars = data_json["props"]["pageProps"]["content"]
+        item = GeojsonPointItem()
+        for bar in bars:
```
Cj-Malone, line 20: I think the item should be inside the loop

```suggestion
        for bar in bars:
            item = GeojsonPointItem()
```

locations/spiders/brewdog.py
```diff
@@ -1,53 +1,40 @@
...
+            item["ref"] = bar["sys"].get("id")
+            item["name"] = bar["fields"].get("name")
+            item["addr_full"] = bar["fields"].get("address")
+            item["phone"] = bar["fields"].get("contactNumber")
+            item["email"] = bar["fields"].get("contactEmail")
+            item["lat"] = bar["fields"].get("location", {}).get("lat")
+            item["lon"] = bar["fields"].get("location", {}).get("lon")
```
Cj-Malone, line 27: Since the current code has been written this isn't important, but for future we have `DictParser.parse` that can do most of the common fields for us automatically, I think it can do all of these.

```suggestion
            item = DictParser.parse(bar["fields"])
            item["ref"] = bar["sys"]["id"]
```

Cj-Malone (approved):
LGTM, thanks

PR #4323, opened 2022-12-13
Specify which JSON parser to use instead of using a flag

iandees:
Also fixes the Ford spider by using the "chompjs" JSON parser.


iandees:
Er, I merged this a little faster than I meant. @Cj-Malone, what do you think of the json_parser="json" change in this to allow chompjs as an option?

Cj-Malone:
Not tested it, but yeah I agree in principal. The try except probably needs changing, and I think there are other spiders that use the changed methods, they might need to be fixed. I'll check both of them later today.

PR #4329, opened 2022-12-13
Fix Tesla spider

jancallewaert:
The address of the stores seems to have moved since it was implemented.


Cj-Malone (approved):
A CrawlSpider should probably be used here. But this is a quick fix, I'm fine leaving it as is.

Thank you

Issue #4338, opened 2022-12-14
Tag datasets with machine-readable metadata


brawer:
Should spiders be able to declare dataset attributes (“metadata”), in particular for licensing?

ATP could export them as FeatureCollection properties. For data consumers, this would mean they could ingest ATP data while filtering out any datasets whose license they don’t like. Without machine-readable licensing data, consumers might be more reluctant to consume ATP data. Of course, most websites don’t actually declare any license; data consumers need to decide for themselves what they want to do in that case. Still, there is plenty of open data around with a clear license.

A scraper could optionally declare its dataset attributes like this:

```python
class ExampleSpider(scrapy.Spider):
    name = "example"
    allowed_domains = ["data.example.org"]
    dataset_attributes = {
        "license": "Creative Commons Zero",
        "license:website": "https://example.org/legal/",
        "license:wikidata": "Q6938433",
    }
```

The scraped GeoJSON file might look like this:

```json
{"type":"FeatureCollection","features":[
{"type": "Feature", "id": "RMOK0O_OUPw-GrmVyGJ_i6M266o=", "properties": {"name": "A", ...}},
{"type": "Feature", "id": "jy5en3B92fHIc70K8Gq2pGYCpKw=", "properties": {"name": "B", ...}}],
"properties":{"license":"Creative Commons Zero","license:website":"https://example.org/legal/","license:wikidata":"Q6938433"}}
```

I made an experimental implementation to see how this would fit into the ATP/Scrapy framework; pull request coming.

mjoe999:
I was thinking of suggesting that we add license tags to the output but here you are with a proposal. Happy days.

So there's a few things going on here. I guess we have typically normalised the POI / dataset attributes within each "Feature". It's simple but bloated. I don't know how fussy GeoJSON clients are but I would think/hope not, so no problem to have such attributes common as you suggest. A small discussion I found here is of interest:

https://gis.stackexchange.com/questions/25279/is-it-valid-to-have-a-properties-element-in-an-geojson-featurecollection

I would suggest that the dataset attributes where written before the Feature elements? That way for a huge file you can stream process the JSON and get the meta-data before the items.

I guess the other thing I would suggest but not necessary at this stage is that we have a licence.py or similar which enumerates the common licences? That way in the spider we could have:

**licence = Licence.CREATIVE_COMMONS_ZERO**

Spiders would still be free to add their own licence terms but in most instances this enumeration would suffice. A bit like the categories.

[...]

PR #4339, opened 2022-12-14
Allow spiders to tag output with machine-readable metadata

brawer:
For the motivation of this change, see
https://github.com/alltheplaces/alltheplaces/issues/4338.

### DIFF
--- a/locations/exporters.py
+++ b/locations/exporters.py
@@ -103,7 +103,7 @@ class GeoJsonExporter(JsonItemExporter):
 
     def export_item(self, item):
         if spider_name := item.get("extras", {}).get("@spider"):
-            if self.spider_name == None:
+            if self.spider_name is None:
                 self.spider_name = spider_name
                 self.write_geojson_header()
             if spider_name != self.spider_name:
--- a/locations/exporters.py
+++ b/locations/exporters.py
@@ -96,12 +96,28 @@ class LineDelimitedGeoJsonExporter(JsonLinesItemExporter):
 class GeoJsonExporter(JsonItemExporter):
     def __init__(self, file, **kwargs):
         super().__init__(file, **kwargs)
-        self.spider_names = set()
+        self.spider_name = None
+
+    def start_exporting(self):
+        pass
+
+    def export_item(self, item):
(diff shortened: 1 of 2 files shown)


iandees:
Can you think of a way to move the FeatureCollection properties to the top of the file?

brawer:
I tried, but didn’t manage. The problem is that Scrapy doesn’t pass the spider to exporters — if I understand correctly, the framework is built for multiple spiders running in parallel within in the same process, so theoretically (but not with ATP) a single exporter instance may combine data from many different spiders into one single file. On the other hand, surely there must be _some_ way to find the name of the currently running spider? Maybe you’ll see have an idea how to do this?

brawer:
> Can you think of a way to move the FeatureCollection properties to the top of the file?

Buffering the data into a temp file? If we do that, we could also compute a `bbox` for the FeatureCollection, and sort the features by `id`; this would stabilize the file contents across weekly runs. Not sure if anyone cares, but why not.

brawer:
Ah, perhaps the exporter could overwrite `export_item()` and delay writing the header when it receives the first item to export. I’ll give that a try, but it’ll have to wait a couple days.

iandees:
Unrelated to this change, but the exporter needs to be looked at again because when a spider emits an exception and crashes, the exporter doesn't get the `export_finished()` call and the resulting GeoJSON file is broken as a result. Maybe we could write to CSV and then some external process could convert to GeoJSON + add the feature collection properties?

brawer:
> Unrelated to this change, but the exporter needs to be looked at again because when a spider emits an exception and crashes, the exporter doesn't get the `export_finished()` call and the resulting GeoJSON file is broken as a result. Maybe we could write to CSV and then some external process could convert to GeoJSON + add the feature collection properties?

Here’s a proposal to handle this particular problem in a slightly different way, which would also help with other sources of flakiness: https://github.com/alltheplaces/alltheplaces/issues/4351

brawer:
> Can you think of a way to move the FeatureCollection properties to the top of the file?

Yes, after some back and forth I figured out a way. Please take another look.

PR #4341, opened 2022-12-14
Add spider for crocs_es

abbensid2:


iandees:
It looks like this is still data for Spain, but there's some errors in the lat/lon. Can you change it back to crocs_es?

abbensid2:
> It looks like this is still data for Spain, but there's some errors in the lat/lon. Can you change it back to crocs_es?

how I can do it?

iandees:
Ah nevermind, I misunderstood one of the commits. Looks good.

abbensid2:
> Ah nevermind, I misunderstood one of the commits. Looks good.

on the previous commit I tried to merge the two spiders crocs_es and crocs_eu but it doesn’t give the same results

PR #4343, opened 2022-12-14
Fix CreditUnion spider

jancallewaert:
The lat and lon of the Credit Unions have disappeared from their site. I don't know if we allow a spidered POI without lat/lon.

### DIFF
--- a/locations/spiders/creditunion.py
+++ b/locations/spiders/creditunion.py
@@ -2,63 +2,10 @@ import logging
 import re
 
 import scrapy
+from geonamescache import GeonamesCache
 
 from locations.hours import OpeningHours
 from locations.items import GeojsonPointItem
-
-STATES = [
-    "AL",
-    "AK",
-    "AZ",
-    "AR",
-    "CA",
-    "CO",
-    "CT",
-    "DC",
-    "DE",
-    "FL",
-    "GA",
-    "HI",
-    "ID",
-    "IL",
-    "IN",
-    "IA",
-    "KS",
-    "KY",
-    "LA",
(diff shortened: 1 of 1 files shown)


Cj-Malone:
We do collect things even if we can't get coordinates. They could be geo coded by a data consumer, or potentially at some point by ATP.

Cj-Malone:
Would you mind replacing the state array with `GeonamesCache().get_us_states()` please.

locations/spiders/creditunion.py
```diff
@@ -107,24 +107,29 @@ def parse_bank(self, bank):
         except IndexError:
             phone = ""
 
-        href = bank.xpath('.//div[@class="location-results__share"]/div/a/@href').extract_first()
-        lat, lon = re.search(r"lat=(.*?)&lng=(.*?)&", href).groups()
         name = bank.xpath(".//h3/text()").extract_first()
-        ref = name + "-" + "-".join(address[0].split(" "))
+        ref = "-".join(name.split(" ")) + "-" + "-".join(address[0].split(" "))
```
iandees, line 111: All the Places generates IDs based on a hash of the `ref` and the spider name, so we prefer to keep the `ref` consistent where possible. Unless there's a specific reason you made this change, can we roll this particular line back to keep the name consistent?

```suggestion
        ref = name + "-" + "-".join(address[0].split(" "))
```

JanCallewaert-TomTom: Fixed

Cj-Malone:
Thank you

PR #4345, opened 2022-12-14
Fix Shell and BP for open hours and more

mjoe999:
Shell and BP spiders were not handling opening hours comprehensively.

Added an OSM fuel type enumeration to categories.py to save finger fumbles etc

street_address handled correctly.

next stop is exxonmobil.py similar clean up and add a whole bunch of Esso stations.


mjoe999:
Hi @iandees , if you're thinking on this one I guess one reason could be that the commit changes things such that we only add in the fuel tags if we have a positive hit. At the minute I think things are brittle enough that things change at the website and we start to say things like "fuel:diesel"="no" when that is not the case. That's the kind of thing that can upset sites so I would rather be cautious?

iandees:
Thanks for pinging me on this Joe. My concern is around the `apply()` method on an enum. It's odd to me that a function on an enum would apply its values to another object. There's a bit too much indirection. The spider code should be the only place doing the work of deciding when to apply the attributes. I'm thinking about a better way to arrange that.

mjoe999:
Fair enough. Me and my enums :-)

I could suggest an apply_yes_no() method in categories.py:

`apply_yes_no(Fuel.DIESEL, item, condition, apply_negative=False)
`
The first parameter could either be a string and an enum item which the method would dig into to get the value.

There are enough yes/no tags in OSM that it would be of reasonable general use.

mjoe999:
Pushed a change as per your prod.

PR #4353, opened 2022-12-15
Add spider for Kate spade

abbensid2:


Cj-Malone:
The `katespade` spider probably need deleting too, I assume they are the same brand.

PR #4366, opened 2022-12-17
Add spider for mercure

abbensid2:


Cj-Malone:
I think this data was already included in the Accor spider

abbensid2:
> I think this data was already included in the Accor spider

Yes ~99% data are included, So I delete the spider

PR #4377, opened 2022-12-19
Add spider for remax_de

abbensid2:


locations/spiders/remax_de.py
```diff
@@ -0,0 +1,37 @@
...
+    name = "remax_de"
+    item_attributes = {
+        "brand": "RE/MAX",
+        "brand_wikidata": "Q965845",
+    }
+    allowed_domains = ["remax.com/"]
+
+    def start_requests(self):
+        url = "https://wp.ooremax.com/wp-json/eapi/v1/agencies?per_page=24&page={}"
+        for i in range(11):
```
Cj-Malone, line 17: Hardcoding the amount of pages is going to backfire at some point when they open enough new locations. I see you've added some padding, currently there is only 9 pages in use. And let's be honest, by the time they expand to need more than 11 pages the API will have probably changed. But for future let's not hardcode pages, it could have been something like:
```
if len response != per_page:
  request page +1
```

PR #4385, opened 2022-12-20
Add spider for samsonite_eu

abbensid2:


eisams:
This spider is wrong, e.g., the stores in Norway are not official Samsonite stores but dealers carrying the Samsonite brand

E.g., these are POIs that should be imported: https://www.morris.no

PR #4391, opened 2022-12-20
Add spider for Belgian stores of Carrefour

jancallewaert:


jancallewaert:
The website exists in French and Dutch. I chose a default. What's the usual convention for this?

iandees:
The default is ok, as long as there aren't duplicates.

PR #4401, opened 2022-12-21
Fix Lids

jancallewaert:
There's still a problem with performing the request. I believe it's caused by the cookies, but I don't know how to fix it 😞

### DIFF
--- a/locations/spiders/lids.py
+++ b/locations/spiders/lids.py
@@ -1,48 +1,30 @@
-import datetime
-
 import scrapy
 
-from locations.hours import OpeningHours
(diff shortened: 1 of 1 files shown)


iandees:
If you pull down the changes I made here, it looks like the issue was related to user-agent. If you want to try fixing it go ahead, otherwise let me know and I can have a crack at it.

jancallewaert:
> If you pull down the changes I made here, it looks like the issue was related to user-agent. If you want to try fixing it go ahead, otherwise let me know and I can have a crack at it.

Thanks for the help. It is working now 😄

PR #4402, opened 2022-12-21
Add spider for yves_rocher_it

abbensid2:


iandees (commented):
There's an existing "Uberall" spider helper. Example is here: https://github.com/alltheplaces/alltheplaces/blob/9e4b3cf71ba7ab831bbf46070e444b71309e955f/locations/spiders/aldi_nord_nl.py

Can you switch to using that instead?

PR #4404, opened 2022-12-21
Storing starbucks ownership code

reuning:
The Starbucks JSON provides an ownership code that was not currently captured. I added the info to the `extras` field. 

I'm not sure if a rationale is necessary here, but this is useful information because the current unionization effort in the US has focused on corporate owned Starbucks and in general there tends to be different rules for non-corporate owned Starbucks.


reuning:
After some searching, it looks like this failed because the points it goes through before being cutoff are all in areas without starbucks. Not sure if this is a big deal. I'm happy to do anything else here.

iandees:
Yep, that sounds like the right reason for why the CI failed.

Thanks for making this change!

PR #4420, opened 2022-12-23
update spider aldi_sud_gb

abbensid2:


abbensid2:
<google-sheets-html-origin>

Country | Before updating | After update
-- | -- | --
GB | 969 | 992

PR #4421, opened 2022-12-23
update spider aldi_sud_ie

abbensid2:


abbensid2:
<google-sheets-html-origin>

Country | Before updating | After update
-- | -- | --
IE | 148 | 153

PR #4429, opened 2022-12-25
fix and update spider canadiantire

abbensid2:


iandees:
Looks like the spider is getting blocked: https://data.alltheplaces.xyz/ci/3777737894/canadiantire/log.txt

abbensid2:
> Looks like the spider is getting blocked: https://data.alltheplaces.xyz/ci/3777737894/canadiantire/log.txt

It must be the IP of the server that blocks maybe !
on my computer, it runs normally

PR #4439, opened 2022-12-26
Fix and update spider coffeetime

abbensid2:


Cj-Malone:
Should you be paging through?

Also this would be good as shared code if you wanted to write it as a store finder spider.

PR #4458, opened 2022-12-27
Fix and update spider goodlifefitness

abbensid2:


locations/spiders/goodlifefitness.py
```diff
@@ -1,36 +1,37 @@
...
-                "city": store["City"],
-                "state": store["Province"],
-                "postcode": store["PostalCode"],
-                "phone": store["Phone"],
-                "lat": float(store["Lat"]),
-                "lon": float(store["Long"]),
-            }
-
-            yield GeojsonPointItem(**properties)
+        template = "https://www.goodlifefitness.com/content/experience-fragments/goodlife/header/master/jcr:content/root/responsivegrid/header.GetClubsWithDetails.{}.false.true.20221127.json"
```
Cj-Malone, line 18: I'm a little consigned that that's a date, and the spider will either break or return old data

abbensid2: Since I’m looking for the id on the sitemap, I think it’s okay?

Cj-Malone: Well the date is in the template url, hopefully closed locations would get dropped from the sitemap and then then the spider wouldn't scrape them. But would the API give data for new locations but with an old date from before they existed?

I think we're OK for now, we can revisit it when the spider breaks. However if I was writing this spider, I'd have just done it as a StructuredDataSpider (+ xpath the lat/lon), I know that's less metadata right now, but I think it's less likely to break, and I think they might add more data to that overtime, and then we'd get it automatically.

abbensid2: Yes, I agree, I'm going to change it tonight

locations/spiders/goodlifefitness.py
```diff
@@ -1,36 +1,37 @@
...
+        url_id = re.findall("[0-9]+", response.url)[0]
+        yield scrapy.Request(url=template.format(url_id), callback=self.parse_club)
```
Cj-Malone, line 20: Also note, the current process is: get the sitemap -> get the html page -> get the API

We can skip getting the html page if we rewrite the url in `sitemap_filter`.

PR #4475, opened 2022-12-29
Fix and update spider jcpenney

abbensid2:


abbensid2:
@Cj-Malone 
infinite loop, would you know why?

abbensid2:
> @Cj-Malone infinite loop, would you know why?

The same address has six services (6 names) !

Cj-Malone:
OK, so we get items from all of the following:
```
https://www.jcpenney.com/locations/fl/miami/clothing-stores-miami-fl-2478.html
https://www.jcpenney.com/locations/fl/miami/home-store-miami-fl-2478.html
https://www.jcpenney.com/locations/fl/miami/jewelry-miami-fl-2478.html
https://www.jcpenney.com/locations/fl/miami/mattresses-miami-fl-2478.html
https://www.jcpenney.com/locations/fl/miami/salon-miami-fl-2478.html
https://www.jcpenney.com/locations/fl/miami/sephora-miami-fl-2478.html
```

- If we know we only want one, we could change the linkextractor to be more specific, eg including `clothing-stores` in the regex. But that might miss stores if they don't have a clothing department.
- It looks like they share a ref, `2478`, so we could set the ref in the spider, then scrapy will throw away duplicates. This might backfire if the first department it gets has different metadata like opening hours, or phone numbers.
- Or potentially we could collect them all, the different departments have different metadata, and we do like data.

Cj-Malone:
Also this should probably be a `StructuredDataSpider`.

abbensid2:
> * Or potentially we could collect them all, the different departments have different metadata, and we do like data.

I'm all right with that

Cj-Malone:
I am too, but we'll see what @iandees opinion is.

I'd like to see each POI categorised though.

iandees:
Since these aren't separate stores (i.e. a Starbucks in a Target), they're individual departments/themes inside a single store, I'm inclined to only collect one record per physical store and add attributes that say what departments are available.

It seems like each store's main page starts with `clothing-stores`, so the link extractor could key off that.

iandees:
Why isn't this matching with https://nsi.guide/index.html?t=brands&k=shop&v=department_store#jcpenney-592fe0

Cj-Malone:
It can't differentiate from the [opticians](https://nsi.guide/index.html?t=brands&k=shop&v=optician&tt=Q920037#jcpenneyoptical-5b6831), if we manually set the category it'll then be able match and get the other NSI tags.

Cj-Malone (approved):
The NSI matching could do with improving, I wish there ids were stable. But this works for now.

PR #4479, opened 2022-12-29
Update spider remax de

abbensid2:
@Cj-Malone  that's good for you


locations/spiders/remax_de.py
```diff
@@ -35,3 +34,8 @@ def parse(self, response):
             item["opening_hours"] = oh.as_opening_hours()
 
             yield item
+
+        page = int(re.findall("\d+$", response.url)[0]) + 1
```
iandees, line 38:
```suggestion
        page = int(re.findall(r"\d+$", response.url)[0]) + 1
```

PR #4480, opened 2022-12-29
update spider goodlifefitness

abbensid2:
@Cj-Malone that's good for you!
We missed opening_hours


Cj-Malone (approved):
So this is RE: https://github.com/alltheplaces/alltheplaces/pull/4458#discussion_r1059000052

IMO the benefits of Structured Data out way the opening hours in this case.

PR #4504, opened 2022-12-31
Add categories to more spiders

Cj-Malone:


Cj-Malone:
I cancelled the spider test because I feel bad for GitHub actions and we know these spiders work, I got the stats from the output of last weeks run.

iandees:
Maybe I should adjust the spider tester to not run when more than 2 or 3 spiders changed.

Cj-Malone:
I don't know if its worth the effort, the multiple spider PRs aren't very often and I now know I can cancel the actions. But if you want to go to the effort, by all means do.

PR #4521, opened 2023-01-01
Fix and update spider metrodiner

abbensid2:


locations/spiders/metrodiner.py
```diff
@@ -1,83 +1,35 @@
...
-                )
-
-        return opening_hours.as_opening_hours()
-
-    def parse_location(self, response):
-        data = response.xpath('//div[@class="site-header__top-wrapper container"]/div/script/text()').re_first(
-            r"locations:\s*(\[.*\])"
+    def post_process_item(self, item, response, ld_data):
+        ldjson = json.loads(
+            response.xpath('//script[@type="application/ld+json"]/text()[contains(.,\'"Restaurant"\')]').get()
```
Cj-Malone, line 22: `ldjson` is unnecessary because `ld_data` is already passed as an argument,

Cj-Malone: Cleaned in #4536

PR #4523, opened 2023-01-01
Support features with non-point geometry

brawer:


brawer:
Confused about the GitHub action failure. Is it related to this change?

locations/exporters.py
```diff
@@ -138,7 +138,8 @@ def _get_serialized_fields(self, item, default_value=None, include_empty=None):
                 )
             except ValueError:
                 logging.warning("Couldn't convert lat (%s) and lon (%s) to string", lat, lon)
-
+        elif geometry := item.get("geometry"):
+            feature.append(("geometry", geometry))
```
iandees, line 142: Can you change the lat/lon stuff above to set the `geometry` property instead of adding to the feature directly?

iandees:
> Confused about the GitHub action failure. Is it related to this change?

No, it's unrelated to your change. It seems like the Github API sometimes responds with something unexpected when trying to list the files changed in a pull request. I haven't figured out why, though.

locations/exporters.py
brawer: Done.

PR #4548, opened 2023-01-03
Add Electrify America

iandees:


locations/spiders/electrify_america.py
```diff
@@ -0,0 +1,31 @@
...
+
+            extras = dict()
+            extras["capacity"] = item.get("evseMax")
+
+            if access := item.get("type"):
+                if access == "PUBLIC":
+                    extras["access"] = "public"
+
+            if "extras" not in feature:
+                feature["extras"] = dict()
```
Cj-Malone, line 28: We do this alot. Maybe it'd be better to always populate extras as a empty dict?

iandees: Yea, that probably makes sense.

Cj-Malone: OK, I'll do a PR tomorrow, unless you do it before. I've also been thinking about `opening_hours` we should probably allow that to be `OpeningHours`, and then do `as_opening_hours()` on export, maybe init that too.

PR #4552, opened 2023-01-03
Fix and update spider penske

abbensid2:


locations/spiders/penske.py
```diff
@@ -1,114 +1,23 @@
...
-            urls = response.xpath('//section[@class="locations-by-city"]/ul/li/a/@href').extract()
+    sitemap_urls = ["https://www.pensketruckrental.com/sitemap.xml"]
+    sitemap_rules = [(r"/locations/us/[-\w]+/[-\w]+/[0-9]+/$", "parse_sd")]
+    wanted_types = ["LocalBusiness"]
 
-        if not urls:
-            urls = response.xpath('//a[contains(@class,"location-link")]/@href').extract()
-            is_store = True
+    def sitemap_filter(self, entries):
+        yield from entries
```
iandees, line 17: No need to have `sitemap_filter()` here if you're not doing any filtering.

locations/spiders/penske.py
```diff
@@ -1,114 +1,23 @@
...
 
-        for url in urls:
+    def post_process_item(self, item, response, ld_data):
+        item["ref"] = re.findall("[0-9]+", response.url)[0]
+        item.pop("email")
```
iandees, line 21: `pop()` will throw an exception is the key doesn't exist. Add `None` default to prevent that exception.

```suggestion
        item.pop("email", None)
```

PR #4555, opened 2023-01-03
Fix and update spider pigglywiggly

abbensid2:


iandees:
It looks like the Piggly Wiggly Midwest stores have their own store locator at https://www.shopthepig.com/. Would you mind adding that one with the same wikidata ID, too?

abbensid2:
> at https://www.shopthepig.com/

it is not possible scrapy the website !

Screenshot 2023-01-03 at 18 22 31

iandees:
Thanks for checking.

PR #4602, opened 2023-01-08
Finish renaming of `GeojsonPointItem` to `Feature`

brawer:
After this change, `git grep GeojsonPointItem` finds no matches in the AllThePlaces codebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ You can contact us in several ways:
 
 ## Adding a spider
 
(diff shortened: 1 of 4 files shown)


Cj-Malone:
I'd be tempted to keep `GeojsonPointItem` around for a little while longer, just because of the open PRs and every so often we get a PR from an old base. I'm glad you changed the docs though.

iandees:
> I'd be tempted to keep GeojsonPointItem around for a little while longer

I'm OK with removing it and keeping an eye out for changes that attempt to reintroduce it. Could we use a `DeprecationWarning` to make it a little more obvious?

mjoe999:
I think just remove it. Any pain is small.

PR #4606, opened 2023-01-09
tighten up top level README, defer more to ./docs

mjoe999:
I think this gives a slightly cleaner GitHub home page given all the additions etc over the last few months?


iandees:
Generally, I think a shorter README is OK, but I want to read through all the docs and this change to make sure we're not missing anything.

mjoe999:
I've had another look and think that `./ docs/GETTING_STARTED.md` could be pulled into the `./README.md ` I have submitted as it is fresh and I think tight? Give me a nudge if you agree with this?
There are a few other things that need saying like `requires_proxy` but that would be another PR.

mjoe999:
@iandees : i've pulled up the docs/GETTING_STARTED into the top level README which I think makes it better?

the changed spiders are simply unused imports which my git hooks like to do.

PR #4607, opened 2023-01-09
add some more SHOP_* categories to core enumeration

mjoe999:
Added stronger consistency to Categories.SHOP_* (see new test)
Added some of the more significant missing shop categories from https://nsi.guide/?t=brands
Can imagine down the road doing more with: https://cdn.jsdelivr.net/npm/@openstreetmap/id-tagging-schema@latest/dist/taginfo.min.json
.. but for now do not want to get totally dependent upon NSI and this whole category thing is an evolution anyway.
@iandees : if you like this then I would apply same consistency to AMENITY_* categories?


iandees:
I appreciate making the enum names consistent, but I don't think we need to add category enums if spiders aren't using them.

JoeDixon-TomTom:
Up to you but I only did the SHOP_* with high counts i.e. optimistic that someone will come along and contribute and not need to add to core file.

PR #4612, opened 2023-01-09
Fix browser typeo

Cj-Malone:
BROSWER_DEFAULT -> BROWSER_DEFAULT

I introduced this spelt wrong in #4140 and Josh pointed it out in #4176


iandees:
Wow, lots of eyes looked at that and didn't notice 😆 , thanks Josh!

PR #4615, opened 2023-01-09
Implement addr:suburb for Australian addresses

davidhicks:
Refer to https://en.wikipedia.org/wiki/Suburbs_and_localities_(Australia) for a description of how places in Australia have a "Suburb" subdivision below "City".
    
For example, in the state of Victoria (1st level), a town named Omeo (2nd level) exists. There are no suburbs of Omeo.
    
For example, in the state of Victoria (1st level), a city named Melbourne (2nd level) exists. Within Melbourne, a suburb named Carlton (3rd level) exists.
    
From the description of https://schema.org/addressLocality and Australian tagging guidelines at https://wiki.openstreetmap.org/wiki/Australian_Tagging_Guidelines/Australian_features#Addresses it is more likely that addr:suburb would be the corresponding OpenStreetMap key for addressLocality.
    
Per Australian tagging guidelines at https://wiki.openstreetmap.org/wiki/Australian_Tagging_Guidelines/Australian_features#Addresses, addr:city is not used within Australia. iD editor does not ask users to enter addr:city but rather asks users to enter addr:suburb for Australian addresses.

### DIFF
--- a/DATA_FORMAT.md
+++ b/DATA_FORMAT.md
@@ -31,6 +31,7 @@ Each GeoJSON feature will have a `properties` object with the following keys:
 | `addr:housenumber`    | No  | The house number part of the address.
 | `addr:street`         | No  | The street name.
 | `addr:street_address` | No  | The street address, including street name and house number and/or name.
+| `addr:suburb`         | No  | The suburb part of the address.
 | `addr:city`           | No  | The city part of the address.
 | `addr:state`          | No  | The state or province part of the address.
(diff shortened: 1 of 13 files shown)


iandees:
All the Places is "inspired by" OSM tagging, but doesn't promise to follow OSM tagging guidelines. To me, the fact that the OSM community in Australia has decided to use `addr:suburb` instead of `addr:city` is a peculiarity of OSM, and I'd rather have consistent output in our project.

PR #4626, opened 2023-01-10
Fix Federal Savings Bank

jancallewaert:
Fix spider, no wikidata entry exists.

### DIFF
--- a/locations/categories.py
+++ b/locations/categories.py
@@ -74,6 +74,8 @@ class Categories(Enum):
     SHOP_WATCHES = {"shop": "watches"}
     SHOP_WHOLESALE = {"shop": "wholesale"}
 
+    OFFICE_FINANCIAL = {"office": "financial"}
+
     ATM = {"amenity": "atm"}
     BANK = {"amenity": "bank"}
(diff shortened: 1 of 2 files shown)


jancallewaert:
> I've made a [Wikidata entry](https://www.wikidata.org/wiki/Q116161036). The category is a weird one, the [website](https://www.thefederalsavingsbank.com/about/our-locations/) says only 2 are banks, the rest are just for mortgages. I think I'd suggest an `office` tag over `amenity=bank`, maybe [office=mortgage_broker](https://taginfo.openstreetmap.org/tags/office=mortgage_broker) but it's got really low usage.
> 
> Also this is to small for NSI, so we do need to apply a category here.



> `office`

I added the wikidata entry. I found examples of [office=financial](https://taginfo.openstreetmap.org/tags/office=financial) for mortgage lending, e.g. https://www.openstreetmap.org/way/452314095. I propose to use that one. Some also have [shop=money_lender](https://wiki.openstreetmap.org/wiki/Tag:shop=money%20lender?uselang=en), e.g. https://www.openstreetmap.org/node/4730509037. That seems less appropriate according to the description on the wiki.

Cj-Malone (approved):
Sounds good to me.

PR #4630, opened 2023-01-11
Chargepoint

iandees:


iandees:
If anyone's itching for something to do, this one could get an extra eye on it to figure out how to reduce duplicates. Also, the EVgo, Electrify America, and this spider all know more about what types of connectors and the power available but I didn't figure out how to add that to the data.

PR #4631, opened 2023-01-11
Fix walmart spider

jancallewaert:
Fixing spider by getting back as a SitemapSpider. I know it has been a sitemap spider in the past and was converted in a crawler. However, it seems a crawler doesn't work correctly while a sitemaps spider for me does work...

### DIFF
--- a/locations/spiders/walmart.py
+++ b/locations/spiders/walmart.py
@@ -15,6 +15,7 @@ class WalmartSpider(SitemapSpider):
     sitemap_urls = ["https://www.walmart.com/sitemap_store_main.xml"]
     sitemap_rules = [("", "parse_store")]
     custom_settings = {"AUTOTHROTTLE_ENABLED": True, "USER_AGENT": BROWSER_DEFAULT}
+    requires_proxy = True
 
     def store_hours(self, store):
         if store.get("open24Hours") is True:
--- a/locations/spiders/walmart.py
+++ b/locations/spiders/walmart.py
@@ -4,6 +4,7 @@ from scrapy.spiders import SitemapSpider
 
 from locations.dict_parser import DictParser
 from locations.hours import OpeningHours
+from locations.user_agents import BROWSER_DEFAULT
 
 
 class WalmartSpider(SitemapSpider):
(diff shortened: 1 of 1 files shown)


JanCallewaert-TomTom:
Is it possible the Github Action is getting blocked? Locally, it runs just fine.

jancallewaert:
Is the Github Action blocked by the website? Locally, it works fine :-(.

Cj-Malone:
IIRC when this was changed the issue wasn't the finding the store links, but getting blocked. The first x pages worked.

Maybe we should try [AUTOTHROTTLE_ENABLED](https://docs.scrapy.org/en/latest/topics/autothrottle.html#autothrottle-enabled), but I think this one will always cause issues in the current environment.

jancallewaert:
> IIRC when this was changed the issue wasn't the finding the store links, but getting blocked. The first x pages worked.
> 
> Maybe we should try [AUTOTHROTTLE_ENABLED](https://docs.scrapy.org/en/latest/topics/autothrottle.html#autothrottle-enabled), but I think this one will always cause issues in the current environment.

Okay, maybe we should then close this PR and think about a more general solution. I reckon there are more spiders that suffer from this.

Is there any way to see the output of the weekly runs per spider? I don't have time to write new ones but fixing spiders I can do. If I see the output of a spider, it might be easier for me to decide whether it's worth the effort to look at the spider.

Cj-Malone:
> think about a more general solution. I reckon there are more spiders that suffer from this.

I think we've pretty much decided the solution is to set up a proxy/residential ips. It's just not been implemented yet. We are flagging the spiders that need it as `requires_proxy = True` though.

> Is there any way to see the output of the weekly runs per spider? I don't have time to write new ones but fixing spiders I can do. If I see the output of a spider, it might be easier for me to decide whether it's worth the effort to look at the spider.

In theory the [log](https://data.alltheplaces.xyz/runs/2023-01-07-13-32-28/logs/walmart.txt) but in this case it's blank. I think the issue there is related to that log only showing "errors", and scrapy/atp not classing this as an error.

Cj-Malone (approved):
We at least got some [POIs](https://data.alltheplaces.xyz/map.html?show=https://data.alltheplaces.xyz/ci/4026224846/walmart/output.geojson) with this combination. We'll have to wait and see for the weekly.

Cj-Malone:
Finally got Walmart POIs! Thank you.

PR #4639, opened 2023-01-13
Sort same-day opening hours

brawer:
There had already been a unit test for sorting multiple opening hours on the same day, but the implementation had not actually sorted the hours. Before this change, the results thus got formatted in pseudo-random order depending on the value of Python’s built-in hash function. Accidentally, the existing test case happened to give the expected sorting order.

### DIFF
--- a/tests/test_opening_hours.py
+++ b/tests/test_opening_hours.py
@@ -99,6 +99,11 @@ def test_multiple_times():
     o.add_range("Mo", "13:00", "17:30")
     assert o.as_opening_hours() == "Mo 08:00-12:00,13:00-17:30"
 
+    o2 = OpeningHours()
+    o2.add_range("Tu", "09:00", "12:00")
+    o2.add_range("Tu", "15:00", "17:00")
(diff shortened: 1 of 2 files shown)


tests/test_opening_hours.py
```diff
@@ -99,6 +99,11 @@ def test_multiple_times():
     o.add_range("Mo", "13:00", "17:30")
     assert o.as_opening_hours() == "Mo 08:00-12:00,13:00-17:30"
 
+    o2 = OpeningHours()
+    o2.add_range("Tu", "09:00", "12:00")
+    o2.add_range("Tu", "15:00", "17:00")
```
Cj-Malone, line 104: Should the test be the other way?
```suggestion
    o2.add_range("Tu", "15:00", "17:00")
    o2.add_range("Tu", "09:00", "12:00")
```

brawer: Added another test case. The `o2` case is interesting because it fails with the current main branch, ie. without the `sorted()` call that this pull request adds to the opening hours implementation.

PR #4649, opened 2023-01-14
Have du report in bytes instead of kb

iandees:
The history.json file is getting the wrong size for the output because `du` is reporting in blocks of kilobytes. Switch that to bytes to report apparent size.

### DIFF
--- a/ci/run_all_spiders.sh
+++ b/ci/run_all_spiders.sh
@@ -127,12 +127,12 @@ if [ ! $retval -eq 0 ]; then
 fi
 
 (>&2 echo "Saving embed to https://data.alltheplaces.xyz/runs/latest/info_embed.html")
-OUTPUT_FILESIZE=$(du "${SPIDER_RUN_DIR}/output.zip"  | awk '{ print $1 }')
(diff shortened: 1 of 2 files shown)


ci/run_all_spiders.sh
```diff
@@ -127,7 +127,7 @@ if [ ! $retval -eq 0 ]; then
 fi
 
 (>&2 echo "Saving embed to https://data.alltheplaces.xyz/runs/latest/info_embed.html")
-OUTPUT_FILESIZE=$(du "${SPIDER_RUN_DIR}/output.zip"  | awk '{ print $1 }')
+OUTPUT_FILESIZE=$(du -b "${SPIDER_RUN_DIR}/output.zip"  | awk '{ print $1 }')
 OUTPUT_FILESIZE_PRETTY=$(echo "$OUTPUT_FILESIZE" | awk '{printf "%0.1f", $1/1024}')
```
Cj-Malone, line 131:
```suggestion
OUTPUT_FILESIZE_PRETTY=$(echo "$OUTPUT_FILESIZE" | awk '{printf "%0.1f", $1/1024/1024}')
```

Or numfmt, but you'll have to change the HTML too.

PR #4662, opened 2023-01-18
Fix Goodstart spider

JanCallewaert-TomTom:
Fixing the Goodstart spider.

Some remarks: 

- The changes to the website removed the original ref. The new version has an `id` field that is taken by the `DictParser`, but it's different compared to older scraped results from https://data.alltheplaces.xyz/runs/2022-12-10-13-32-28/output/goodstart.geojson. Does it make sense to try to recreate them?
- The JSON result contains a suburb, but not a city. PR #4615 seems to state that suburb is not used in our model. What should I do with this?
- A center can have different services: kindergarten-preschool, nursery, toddler, ... I used the `amenity=childcare` since I can find this the most on OSM. It might make sense to go more granular, but I guess I can't add multiple amenities to a single place.

### DIFF
--- a/locations/categories.py
+++ b/locations/categories.py
@@ -83,6 +83,7 @@ class Categories(Enum):
     BUREAU_DE_CHANGE = {"amenity": "bureau_de_change"}
     CAFE = {"amenity": "cafe"}
     CHARGING_STATION = {"amenity": "charging_station"}
+    CHILD_CARE = {"amenity": "childcare"}
     CLINIC_URGENT = {"amenity": "clinic", "healthcare": "clinic", "urgent_care": "yes"}
     COFFEE_SHOP = {"amenity": "cafe", "cuisine": "coffee_shop"}
     COMPRESSED_AIR = {"amenity": "compressed_air"}
--- a/locations/spiders/goodstart.py
+++ b/locations/spiders/goodstart.py
@@ -25,6 +25,7 @@ class GoodstartSpider(scrapy.Spider):
             item["image"] = "https://www.goodstart.org.au" + i["imageUrl"]
             item["addr_full"] = i["fullAddress"]
             item["street_address"] = i["address"]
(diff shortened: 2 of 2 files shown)


Cj-Malone:
> * The changes to the website removed the original ref. The new version has an `id` field that is taken by the `DictParser`, but it's different compared to older scraped results from https://data.alltheplaces.xyz/runs/2022-12-10-13-32-28/output/goodstart.geojson. Does it make sense to try to recreate them?

That's fine, constant refs are nice, but it's not always possible.

> * The JSON result contains a suburb, but not a city. PR [Implement addr:suburb for Australian addresses #4615](https://github.com/alltheplaces/alltheplaces/pull/4615) seems to state that suburb is not used in our model. What should I do with this?

In my opinion, just put it in `city`/`addr:city`. Supporting all possible address formats isn't viable here, and it's barley viable in OSM with people converting every address.

I think of ATPs `city` as "addressLocality", and `state` as `addressRegion` if this [helps](https://schema.org/PostalAddress). We also semi recently added `street_address` because the practicality of splitting that into `housenumber` and `street` in every spider... just wasn't going to work.

ATPs output isn't "import ready" for OSM, anyone wishing to do that will have to process the data some more. ATP is close to OSM, but not perfect.

> * A center can have different services: kindergarten-preschool, nursery, toddler, ... I used the `amenity=childcare` since I can find this the most on OSM. It might make sense to go more granular, but I guess I can't add multiple amenities to a single place.

We don't really support multiple categories on a POI, nor does OSM. Technically we can, eg: `amenity=pub` + `tourism=hotel` or `shop=car;boat` but it's poorly supported everywhere. I'd like to see both OSM and ATP improve on this one, but I doubt it'll happen anytime soon. Your compromise here seems like the best option for now.

Cj-Malone (approved):

PR #4664, opened 2023-01-18
Add spider for relay_fr

abbensid2:


locations/spiders/relay_fr.py
```diff
@@ -0,0 +1,39 @@
...
+            for day in days.xpath("//tr"):
+                if day.xpath("./td[2]/text()").get() == "Closed":
+                    continue
+                oh.add_range(
+                    day=day.xpath("./td[1]/text()").get(),
+                    open_time=day.xpath(".//time/text()").get().split(" - ")[0],
+                    close_time=day.xpath(".//time/text()").get().split(" - ")[1],
+                )
+
+            item["opening_hours"] = oh.as_opening_hours()
```
Cj-Malone, line 37: Just an FYI, since a few days ago we can now do `item["opening_hours"] = oh`. `.as_opening_hours()` is then called in pipeline code, this allows us to skip the validation checks as we know it'll be a valid opening hours string.

matkoniecz:
https://www.relay.com/ seems down, also on checking with a regular browser

not yet sure how to fix https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/relay_fr.py

is this brand still existing?

abbensid2:
Hi @matkoniecz,
The Relay brand still exists and the relay_fr.py spider works fine, but it requires a French IP address to run successfully

matkoniecz:
I opened https://github.com/alltheplaces/alltheplaces/issues/8701 as spider when run by ATP in fact fails. Not sure how to solve it given #6433

PR #4668, opened 2023-01-19
Fix Jumbo Argentina spider

JanCallewaert-TomTom:
There's no wikidata entry. Jumbo Argentina is, in fact, a brand from [Cencosud](https://www.wikidata.org/wiki/Q1053372) and not from the Dutch supermarket chain [Jumbo](https://www.wikidata.org/wiki/Q2262314). However, that's not present in Wikidata. Should this be added to Wikidata and NSI?

### DIFF
--- a/locations/spiders/jumbo_ar.py
+++ b/locations/spiders/jumbo_ar.py
@@ -5,9 +5,8 @@ from locations.dict_parser import DictParser
 
 
 class JumboARSpider(scrapy.Spider):
-    name = "jumbo_argentina"
-    item_attributes = {"brand": "Jumbo", "brand_wikidata": "Q116259940"}
-    allowed_domains = ["www.jumbo.com.ar"]
+    name = "jumbo_ar"
+    item_attributes = {"brand": "Jumbo", "brand_wikidata": "Q6310864"}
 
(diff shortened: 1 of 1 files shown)


Cj-Malone:
I just made a [Wikidata entry](https://www.wikidata.org/wiki/Q116259940). IMO every spider should have Wikidata. I think NSI actually says brands with more than 10 locations can be included, but I wouldn't really add anything with less than ~100 locations.

JanCallewaert-TomTom:
> I just made a [Wikidata entry](https://www.wikidata.org/wiki/Q116259940). IMO every spider should have Wikidata. I think NSI actually says brands with more than 10 locations can be included, but I wouldn't really add anything with less than ~100 locations.

Thanks, added the wikidata code.

UKChris-osm:
> I think NSI actually says brands with more than 10 locations can be included.

The automatic planet data scan will capture brands with 50 or more locations mapped globally, or transit / operators with 10 or more locations mapped globally, but this isn’t a solid rule, just a random ballpark amount. Anything less can be considered, especially if there is an indication these will grow in number.

Cj-Malone (approved):

PR #4679, opened 2023-01-23
Fix and update spider whsmith_gb

abbensid2:


Cj-Malone:
So the thing that changed here was that they updated there robots.txt. We're not hiding, I hope they'd reach out it they wanted to talk.

PR #4700, opened 2023-01-31
First go at Denmans GB scraper (74 locations, but enables many similar ones)

peternewman:
From https://www.denmans.co.uk/den/store-finder

### DIFF
--- a/locations/spiders/denmans_gb.py
+++ b/locations/spiders/denmans_gb.py
@@ -1,0 +1,45 @@
+from datetime import datetime
+
+from scrapy import Spider
+
+from locations.dict_parser import DictParser
+from locations.hours import DAYS, OpeningHours
+
+
+class DenmansGBSpider(Spider):
+    name = "denmans_gb"
+    item_attributes = {"brand": "Denmans", "brand_wikidata": "Q1"}
+    # Seems to return all stores regardless of lat long as long as it's in the UK?
+    start_urls = ["https://www.denmans.co.uk/den/store-finder/findNearbyStores?latitude=51&longitude=-0"]
+
+    def parse(self, response):
+        for store in response.json()["results"]:
+            store["address"]["street_address"] = ", ".join(
+                filter(
+                    None,
+                    [
+                        store["address"].pop("line1"),
+                        store["address"].pop("line2"),
+                        store["address"].pop("line3"),
+                    ],
+                )
+            )
+            item = DictParser.parse(store["address"])
+            item["phone"] = store["address"]["phone"]
+            item["name"] = store["displayName"]
+            item["ref"] = store["name"]
+            #https://www.denmans.co.uk/den/Bradley-Stoke-Bristol/store/1AR
+            item["website"] = f'https://www.denmans.co.uk/den/{store["address"]["town"].replace(" ", "-")}/store/{store["name"]}'
+            item["opening_hours"] = self.decode_hours(store)
+            yield item
+
+    @staticmethod
+    def decode_hours(store):
+        oh = OpeningHours()
+        for r in store["openingHours"]["rexelWeekDayOpeningList"]:
+            oh.add_range(
+                DAYS[r["weekDay"]],
+                r["openingTime"]["formattedHour"],
+                r["closingTime"]["formattedHour"],
+            )
+        return oh.as_opening_hours()
--- a/locations/spiders/denmans_gb.py
+++ b/locations/spiders/denmans_gb.py
@@ -22,12 +22,8 @@ class DenmansGBSpider(Spider):
                     ],
                 )
             )
-            store["address"]["country"] = store["address"]["country"]["isocode"]
+            store["ref"] = store.pop("name")
             item = DictParser.parse(store)
-            item["name"] = store["displayName"]
-            item["ref"] = store["name"]
-            # We can either pass in store, and it picks up the lat/long, or store["address"] and it gets the email and phone
-            item["email"] = store["address"]["email"]
             item["phone"] = store["address"]["phone"]
             # e.g. https://www.denmans.co.uk/den/Bradley-Stoke-Bristol/store/1AR
             item[
@@ -50,4 +46,4 @@ class DenmansGBSpider(Spider):
                 r["closingTime"]["formattedHour"],
                 time_format="%I:%M %p",
             )
-        return oh.as_opening_hours()
+        return oh
--- a/locations/spiders/denmans_gb.py
+++ b/locations/spiders/denmans_gb.py
@@ -7,7 +7,7 @@ from locations.hours import OpeningHours
 class DenmansGBSpider(Spider):
     name = "denmans_gb"
     item_attributes = {"brand": "Denmans", "brand_wikidata": "Q116508855"}
-    # This seems to return all stores regardless of lat long as long as it's in the UK?
+    # This seems to return all stores regardless of lat-long; as long as it's in the UK?
     start_urls = ["https://www.denmans.co.uk/den/store-finder/findNearbyStores?latitude=51&longitude=-0"]
 
     def parse(self, response):
--- a/locations/spiders/denmans_gb.py
+++ b/locations/spiders/denmans_gb.py
@@ -3,7 +3,7 @@ from datetime import datetime
 from scrapy import Spider
 
 from locations.dict_parser import DictParser
-from locations.hours import DAYS, OpeningHours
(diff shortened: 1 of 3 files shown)


locations/spiders/denmans_gb.py
```diff
@@ -0,0 +1,47 @@
...
+
+from scrapy import Spider
+
+from locations.dict_parser import DictParser
+from locations.hours import DAYS, OpeningHours
+
+
+class DenmansGBSpider(Spider):
+    name = "denmans_gb"
+    item_attributes = {"brand": "Denmans", "brand_wikidata": "Q1"}
```
peternewman, line 11: Will need a Wikidata ID from https://github.com/osmlab/name-suggestion-index/issues/7729

locations/spiders/denmans_gb.py
```diff
@@ -0,0 +1,47 @@
...
+
+    @staticmethod
+    def decode_hours(store):
+        oh = OpeningHours()
+        for r in store["openingHours"]["rexelWeekDayOpeningList"]:
+            oh.add_range(
+                r["weekDay"],
+                r["openingTime"]["formattedHour"],
+                r["closingTime"]["formattedHour"],
+            )
```
Cj-Malone, line 49: So this is crashing on days off.

`openingTime`+`closingTime` are Null/None on closed days. You could do an `if r["closed"]` before accessing them, or you could do `r.get("openingTime", {}).get("formattedHour")`, which would return None on closed days and `add_range` would throw it away.

Also I think you'll need to add a custom time format: `time_format="%I:%M %p"`

locations/spiders/denmans_gb.py
```diff
@@ -0,0 +1,50 @@
...
+                    [
+                        store["address"].pop("line1"),
+                        store["address"].pop("line2"),
+                        store["address"].pop("line3"),
+                    ],
+                )
+            )
+            item = DictParser.parse(store["address"])
+            item["phone"] = store["address"]["phone"]
+            item["contact:email"] = store["address"]["email"]
```
Cj-Malone, line 27: It's just `email`, you can see the full list [here](https://github.com/alltheplaces/alltheplaces/blob/master/locations/items.py).

But in this case, `DictParser` should get the email and phone, so you can delete these lines.
```suggestion
```

peternewman: Thanks, I've just opened https://github.com/alltheplaces/alltheplaces/pull/4702 to fix the docs!

peternewman:
I assume the trick is then to leverage this to do basically the same thing for:
https://www.rexel.co.uk/uki/store-finder
https://www.ideal.co.nz/nzi/store-finder/
https://www.idealelectrical.com/aie/store-finder
https://www.jrt.com.au/ajt/store-finder
https://www.learsmith.com.au/als/store-finder
https://www.rexel.com.au/are/store-finder
https://www.nedco.ca/cnd/store-finder
https://atlantic.rexel.ca/cra/store-finder
https://www.westburne.ca/cwr/store-finder
etc?

Is it worth doing anything clever to subclass it, or just duplicating the file and editing as necessary?

Cj-Malone:
> Is it worth doing anything clever to subclass it, or just duplicating the file and editing as necessary?

We have ["storefinders"](https://github.com/alltheplaces/alltheplaces/tree/master/locations/storefinders), assuming most [rexel](https://www.rexel.com/en/countries-and-banners/) stores use the same API, you could write one. Or there could be a monolithic rexel spider that did all the brands/countries.

I might lean towards the latter, but both options are valid, and both options are better than copy and pasting this file 15 times.

locations/spiders/denmans_gb.py
```diff
@@ -0,0 +1,53 @@
...
+                        store["address"].pop("line3"),
+                    ],
+                )
+            )
+            store["address"]["country"] = store["address"]["country"]["isocode"]
+            item = DictParser.parse(store)
+            item["name"] = store["displayName"]
+            item["ref"] = store["name"]
+            # We can either pass in store, and it picks up the lat/long, or store["address"] and it gets the email and phone
+            item["email"] = store["address"]["email"]
```
peternewman, line 30: This one could be fixed by adding email1 to the DictParser, as they duplicate it in another field, or both could try obj, then fall back to obj["address"] if no results...

locations/spiders/denmans_gb.py
```diff
@@ -0,0 +1,53 @@
...
+                filter(
+                    None,
+                    [
+                        store["address"].pop("line1"),
+                        store["address"].pop("line2"),
+                        store["address"].pop("line3"),
+                    ],
+                )
+            )
+            store["address"]["country"] = store["address"]["country"]["isocode"]
```
peternewman, line 25: Is it worth trying ["country"]["isocode"] more generally as this pattern seems to be present in a few scrapers?

Cj-Malone: I agree with this. There might be a slight chance it breaks another spider, but that's why I prefer to clean up data before it goes into DictParser, not after. We can keep an eye on it with the next weekly.

peternewman:
> > Is it worth doing anything clever to subclass it, or just duplicating the file and editing as necessary?
> 
> We have ["storefinders"](https://github.com/alltheplaces/alltheplaces/tree/master/locations/storefinders), assuming most [rexel](https://www.rexel.com/en/countries-and-banners/) stores use the same API, you could write one.

There seem to be at least a few different flavours in use, e.g. the US ones work in a different way:
https://www.rexelusa.com/locations/ny/jamestown/3906

> Or there could be a monolithic rexel spider that did all the brands/countries.

I wondered if that might get a bit confusing especially hitting multiple URLs and different NSI brands.
 
> I might lean towards the latter, but both options are valid, and both options are better than copy and pasting this file 15 times.

Yeah makes sense. Are you happy for the suggested improvements to DictParser above then to simplify this (and potentially a few other spiders) further?

locations/spiders/denmans_gb.py
Cj-Malone: Contact details in an address object does seem to be relatively common. So it'd be nice to support it, but I think the most common situation is it being in the root (`obj`). I'm nervous about this one because I don't think anyone looks at the stats of contact details so it we lost a bunch nobody would notice.

Adding `email1` seems like a clear win though.

peternewman: Understood, I'd think things like none/str checks would catch this, I've left it with just email1 for now anyway.

Cj-Malone:
> I wondered if that might get a bit confusing especially hitting multiple URLs and different NSI brands.

We have a bunch of spiders that output multiple brands eg: kroger, bestwestern, hilton, ihg_hotels, inditex etc. The NSI stuff works fine with it. I haven't looked at rexel enough to check which is better in this instance. I'm sure either will be fine.

locations/spiders/denmans_gb.py
```diff
@@ -22,12 +22,8 @@ def parse(self, response):
                     ],
                 )
             )
-            store["address"]["country"] = store["address"]["country"]["isocode"]
+            store["ref"] = store.pop("name")
```
peternewman, line 25: That's a neat solution!

Cj-Malone: It would have been better if I had changed the website at the same time 🤦

It works because DictParser will still get the name from the displayName field.

locations/spiders/denmans_gb.py
```diff
@@ -0,0 +1,49 @@
...
+                    [
+                        store["address"].pop("line1"),
+                        store["address"].pop("line2"),
+                        store["address"].pop("line3"),
+                    ],
+                )
+            )
+            store["ref"] = store.pop("name")
+            item = DictParser.parse(store)
+            item["phone"] = store["address"]["phone"]
```
peternewman, line 27: Should we move this before DictParser then?

peternewman:
Is it worth merging this now, and then me reworking it for some of the other sites as I imagine at least some will need wikidata and stuff.

locations/spiders/denmans_gb.py
Cj-Malone: I don't think it's massively important either way.

Cj-Malone (approved):
Sounds good to me

peternewman:
> Sounds good to me

Great, I'm happy for you to merge and/or approve workflows whenever you're happy then.

Cj-Malone:
Only @iandees can merge, but after this is I will no longer need to approve your GitHub actions.

PR #4702, opened 2023-01-31
Fix the email format info

peternewman:
As per https://github.com/alltheplaces/alltheplaces/pull/4700/files#r1091696935


peternewman:
Wouldn't the easiest solution just be a column for the internal name where it's different, minimal changes and no need to describe each thing again?

Cj-Malone:
Yeah it'd be easiest and quickest, but this file is for data consumers, when the other docs are for contributors. But then again if we had 2 docs the descriptions would have to be in 2 places.

No strong opinion from me.

Issue #4614, opened 2023-01-09
New spider suggestion: TSB (UK High Street Bank)


rjw62:
### Brand name

TSB

### Wikidata ID

Q7671560

### Store finder url(s)

https://branches.tsb.co.uk/sitemap.xml

Individual branch pages can be distinguished from locality pages by the fact that the former end in ".html".

Cj-Malone:
It looks like standard Microdata too, so it can be a StructuredDataSpider (and SitemapSpider for page discovery). Although the Microdata has 2 types so we'd have to define them with `wanted_types = [["BankOrCreditUnion", "FinancialService"]]`. (This is a bug, it's not how I meant to implement it)

We can check it with:
`$ pipenv run scrapy sd https://branches.tsb.co.uk/aberdeen/aberdeen-st-nicholas.html --wanted-types=BankOrCreditUnion,FinancialService`

```python
{'city': None,
 'email': None,
 'extras': {},
 'image': 'https://branches.tsb.co.uk/aberdeen/aberdeen-st-nicholas.html',
 'lat': '57.147410168890175',
 'lon': '-2.0984620085982897',
 'name': 'Aberdeen, 7-21a St Nicholas St',
 'opening_hours': 'Mo-Tu 09:00-17:00; We 10:00-17:00; Th-Fr 09:00-17:00; Sa '
                  '09:00-15:00',
 'phone': '01224 369984',
 'postcode': 'AB10 1HE',
 'ref': 'https://branches.tsb.co.uk/#13597885',
 'state': None,
 'street_address': '7-21a St Nicholas St Aberdeen',
 'website': 'https://branches.tsb.co.uk/aberdeen/aberdeen-st-nicholas.html'}
```

davidhicks:
Implemented and working per https://data.alltheplaces.xyz/runs/2023-05-31-05-12-59/stats/tsb_gb.json

PR #4705, opened 2023-01-31
Add spider for TSB (UK high street bank)

rjw62:
Should implement issue #4614 .

I don't understand why, but I seemed to need
`wanted_types = ["BankOrCreditUnion", "FinancialService"]`
rather than
`wanted_types = [["BankOrCreditUnion", "FinancialService"]]`
as was suggested in the issue thread. With the wanted_types line either missing or being the latter above, I would get no results.

Running locally, the spider returns 264 branches, which is consistent with the 220 branches quoted at https://en.wikipedia.org/wiki/TSB_Bank_(United_Kingdom), plus around 43 "popup" locations referred to at https://www.tsb.co.uk/news-releases/tsb-to-launch-pop-ups-across-uk/ .

### DIFF
--- a/locations/spiders/tsb_gb.py
+++ b/locations/spiders/tsb_gb.py
@@ -1,0 +1,16 @@
+from scrapy.spiders import SitemapSpider
+
+from locations.categories import Categories
+from locations.structured_data_spider import StructuredDataSpider
+
+
+class TSBGB(SitemapSpider, StructuredDataSpider):
+    name = "tsb_gb"
+    item_attributes = {
+        "brand": "TSB",
+        "brand_wikidata": "Q7671560",
+        "extras": Categories.BANK.value,
+    }
+    sitemap_urls = ["https://branches.tsb.co.uk/sitemap.xml"]
+    sitemap_rules = [(r"https:\/\/branches\.tsb\.co\.uk\/[-\w]+\/[-\/\w]+\.html$", "parse_sd")]
+    wanted_types = ["BankOrCreditUnion", "FinancialService"]


rjw62:
> I think my original intention was (lists of) single types to match any of the site types, but if someone gives a (list of) multiple types then all they types needed to be present

That was what I deduced from the code I looked at. So I'm not sure what's actually going on.

> The only thing is the image is actually the page url. Somewhere in the Microdata conversion `<link itemprop="image" href="">` gets turned into `"image": ["https://branches.tsb.co.uk/birmingham/birmingham-great-barr.html"]`. I don't know if that's technically right in HTML land. Here we'll have to override `pre_process_data` or `post_process_item` and set the image to None.

I think it may be technically correct, but we should certainly strip out those image links. Any preferences for which of `pre_process_data` or `post_process_item` to use? I guess removing empty image tags before processing is a bit cleaner (and more future proof) than removing them afterwards. Though the test for afterwards could be just to remove the image link if it matches the page URL.

Cj-Malone:
No real preference but "Though the test for afterwards could be just to remove the image link if it matches the page URL." sounds good to me. It means going forward if they fix the markup we should start getting the image without any updates.

locations/spiders/tsb_gb.py
```diff
@@ -14,3 +14,8 @@ class TSBGB(SitemapSpider, StructuredDataSpider):
     sitemap_urls = ["https://branches.tsb.co.uk/sitemap.xml"]
     sitemap_rules = [(r"https:\/\/branches\.tsb\.co\.uk\/[-\w]+\/[-\/\w]+\.html$", "parse_sd")]
     wanted_types = ["BankOrCreditUnion", "FinancialService"]
+
+    def post_process_item(self, item, response, ld_data, **kwargs):
+        if item["image"] == item["website"]:
+            item["image"] = None
+        yield from self.inspect_item(item, response)
```
Cj-Malone, line 21: We don't really do `inspect_item` any more.

```suggestion
        yield item
```

Cj-Malone (approved):
LGTM, thank you.

FYI scrapy can cache stuff, it is really helpful with spiders that pull a bunch of pages. Something like: `pipenv run scrapy crawl -s HTTPCACHE_ENABLED=True -s HTTPCACHE_EXPIRATION_SECS=604800` gives you 7 days of cache.

PR #4707, opened 2023-01-31
Try hitting individual countries/regions to fix Abercrombie & Fitch

peternewman:


Cj-Malone:
The old url, or even https://www.abercrombie.com/api/ecomm/a-uk/storelocator/search?shopRegion works in my browser. I wonder if this is just a run time issue for the US AWS server does the weekly release.

PR #4708, opened 2023-01-31
Convert DenmansGBSpider to our base RexelSpider

peternewman:


locations/spiders/denmans_gb.py
```diff
@@ -1,49 +1,12 @@
...
-        for r in filter(lambda x: (not x["closed"]), store["openingHours"]["rexelWeekDayOpeningList"]):
-            oh.add_range(
-                r["weekDay"],
-                r["openingTime"]["formattedHour"],
-                r["closingTime"]["formattedHour"],
-                time_format="%I:%M %p",
-            )
-        return oh
+    def parse_item(self, item, feature, **kwargs):
+        yield item
```
Cj-Malone, line 12: The idea behind the `parse_item` in the other spiders is that it's a place the spider can clean up or parse extra data. eg accor adds brand details and url, while leroy_merlin does nothing.

This "template function" should be in the shared code, then denmans_gb can override it if needed.

peternewman: Oops yeah, that's now done!

Cj-Malone (approved):

Issue #4587, opened 2023-01-06
Anaconda (Australian outdoor goods retailer)


davidhicks:
### Brand name

Anaconda

### Wikidata ID

Q105981238

### Store finder url(s)

https://www.anacondastores.com/sitemap/store/store-sitemap.xml

davidhicks:
For future reference, I also found https://www.anacondastores.com/rest/v1/cart/cncStoreList/?locationQuery=2000&latitude=&longitude= which appears to return store details for up to the 3 closest stores to a post code in Australia. Not very useful from what I can tell, but there could be other REST API calls hidden away that return all store details in one response.

PR #4723, opened 2023-02-04
Add Anaconda (AU) (78 stores)

davidhicks:
Closes issue #4587


davidhicks:
Just a quick note that when I run the spider I get 78 stores returned. The pre-commit checks return less than this figure of 78, citing 'timeout' as the reason. Perhaps the servers the checks are being run from are being throttled and an increased delay should be added to this spider?

iandees:
The Github Actions run of the spider is configured to time out after 60 seconds. It's meant to make sure the spider is running and might return some records, not to run the whole thing. That's why a "timeout" still gives a successful build.

PR #4724, opened 2023-02-06
Fix Buckle spider by forcing json load

JanCallewaert-TomTom:


Cj-Malone:
I think json5 should parse this? If so converting this to a StructuredDataSpider with `json_parser = "json5"` may be better than manipulating the raw json.

JanCallewaert-TomTom:
> I think json5 should parse this? If so converting this to a StructuredDataSpider with `json_parser = "json5"` may be better than manipulating the raw json. Or `json_parser = "chompjs"`, in a worst case scenario.

You're right. Prefect and much cleaner spider 😄

Cj-Malone (approved):
LGTM, thanks

PR #4730, opened 2023-02-07
Fix Edward Jones spider

JanCallewaert-TomTom:
Changes in the website removed the `lat/lon` from the pages. There's some workaround to contact an API (`https://www.edwardjones.com/api/v3/financial-advisor/results?q=" + postcode + "&distance=50&distance_unit=mi&page=1&matchblock=&searchtype=2`), but it's quite cumbersome to get it right and not get blocked.

I'll mark the PR as a draft and think further on it.

### DIFF
--- a/locations/spiders/edwardjones.py
+++ b/locations/spiders/edwardjones.py
@@ -1,5 +1,3 @@
-import json
-
 import requests
 from scrapy.spiders import SitemapSpider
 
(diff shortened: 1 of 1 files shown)


locations/spiders/edwardjones.py
```diff
@@ -1,39 +1,45 @@
...
-            yield scrapy.Request(url, callback=self.parse_location, meta=meta)
-        yield self.get_page(1 + response.meta["page"])
-
-    def parse_location(self, response):
-        ld = LinkedDataParser.find_linked_data(response, "LocalBusiness")
-        del ld["openingHoursSpecification"]
-        item = LinkedDataParser.parse_ld(ld)
-        item["lat"] = response.meta["lat"]
-        item["lon"] = response.meta["lon"]
+            response = requests.get(url, headers=headers, data={}).json()
```
iandees, line 31: Please use `yield` to generate a new request rather than using `requests` to  make HTTP calls inside the parse function.

JanCallewaert-TomTom: Updated the code. Thanks for the feedback.

PR #4738, opened 2023-02-08
Add Fastbreak Spider

Cj-Malone:
https://github.com/osmlab/name-suggestion-index/issues/7759

### DIFF
--- a/locations/storefinders/storerocket.py
+++ b/locations/storefinders/storerocket.py
@@ -7,10 +7,10 @@ from locations.items import Feature
 
 
 class StorerocketSpider(Spider):
-    id = ""
+    storerocket_id = ""
 
     def start_requests(self):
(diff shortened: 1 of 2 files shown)


locations/storefinders/storerocket.py
```diff
@@ -0,0 +1,35 @@
+from scrapy import Spider
+from scrapy.http import JsonRequest
+
+from locations.dict_parser import DictParser
+from locations.hours import OpeningHours, sanitise_day
+from locations.items import Feature
+
+
+class StorerocketSpider(Spider):
+    id = ""
```
iandees, line 10: Maybe this should be `storerocket_id` to make it clearer about how the ID is used?

iandees: Yea, I think the primary use will be in other spiders that subclass from the `StorerocketSpider`, in which case it's helpful to have the extra info.

PR #4741, opened 2023-02-08
Fix The North Face Spider

Cj-Malone:


iandees:
Some entries have `addr:state` set to "TNF RETAIL". Can you spot a way to avoid that?

Cj-Malone:
Probably just remove it in parse_item, I'll sort it tomorrow.

PR #4764, opened 2023-02-09
Add additional Storepoint spiders

davidhicks:


locations/spiders/digi_telecommunications_my.py
```diff
@@ -0,0 +1,37 @@
...
+from locations.storefinders.storepoint import StorepointSpider
+
+
+class DiGiTelecommunicationsMYSpider(StorepointSpider):
+    name = "digi_telecommunications_my"
+    item_attributes = {"brand": "DiGi Telecommunications", "brand_wikidata": "Q3268530"}
+    key = "1624a4f7eafca0"
+
+    def parse_item(self, item, location: {}, **kwargs):
+        item["addr_full"] = item.pop("street_address")
```
Cj-Malone, line 13: If this is common across all StorepointSpider spiders, it should go in `StorepointSpider`. Same goes for the opening hours.

locations/spiders/digi_telecommunications_my.py
```diff
@@ -0,0 +1,37 @@
...
+        item["country"] = "MY"
```
Cj-Malone, line 14: This isn't actually necessary, we get it from the spider name now: digi_telecommunications_**my**

locations/spiders/digi_telecommunications_my.py
davidhicks: It does indeed look common with street_address and addr_full, as well as the opening hours format. I'll change the Storepoint spider.

davidhicks:
Cleaned up and reorganised the patch set with changes made to address feedback.

locations/spiders/digi_telecommunications_my.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class DiGiTelecommunicationsMYSpider(StorepointSpider):
+    name = "digi_telecommunications_my"
+    item_attributes = {"brand": "DiGi Telecommunications", "brand_wikidata": "Q3268530"}
+    key = "1624a4f7eafca0"
+
+    def parse_item(self, item, location: {}, **kwargs):
+        if item["phone"] is not None:
+            item["phone"] = "+60-" + re.sub(r"^601", "1-", re.sub(r"^03-", "3-", item["phone"]))
```
iandees, line 13: You shouldn't need to do any phone parsing/manipulation. We normalize the phone number during the scraping process.

davidhicks: Thanks, I've removed it. I think at one point I was getting errors about some invalid phone numbers and therefore tried to rewrite the numbers manually. I must have confused spiders as it isn't needed for DiGi.

PR #4766, opened 2023-02-09
Fix Boots TH spider

JanCallewaert-TomTom:
There are also addresses available in Thai through a different URL. Do we localize the names?


iandees:
We don't explicitly support localized names, but if there are multiple languages you can add them to `extras`. I don't think we need to scrape other URLs to pull down localized names at this point.

Issue #4757, opened 2023-02-09
keke: use StorepointSpider


davidhicks:
The current KekeSpider (keke.py) uses a SitemapSpider that seemingly can now be replaced with a single request via the StorepointSpider using application key [158d5359b0129d](https://api.storepoint.co/v1/158d5359b0129d/locations) that is found at https://www.kekes.com/locations

Cj-Malone:
In general my preferred order of spiders is:

1. StructuredDataSpider, this data is specifically published to be scraped, it's as close as we can get to having permission.
2. API spiders, well structured data, but no docs.
3. HTML scrapers with xpaths everywhere, just so fragile.

That's my order, @iandees may order it differently. And people may value the cost/efficiency StructuredDataSpider vs api spiders differently.

Specifically for Keke, it's a mix of 1 and 3. So for me it'd come down to the details, how much data can we get from each, and the perceived stability of those xpaths. I've not looked at the details, but I trust you or someone else to ultimately either rewrite the spider or close this issue.

iandees:
Yep, generally I agree with this, with the caveat that (2) is split into "first party APIs" and "third party APIs". When a first-party API is available, I would prefer that over the data available over third party APIs.

Cj-Malone:
If you mean first/third party APIs as in, who hosts them, I'm not sure I care that much any more. I used to, but it's clear this data is from the brands, it's first party data. Maybe we should document it, I'll do a PR tomorrow to add dataset attributes.

If you mean first/third party as in data. I'm still strongly against it, I don't even think it should be "4" in that list. We shouldn't collect it.

iandees:
I just mean that if a business runs their own location API service and then also sends it to where2getit or something, we should prefer the business's own location API service.

I agree that we should not include data aggregators.

Cj-Malone:
I agree in principle, but in reality it's about maintainability. So I guess I wouldn't replace a first party api with a third party one, but writing a new one, I'd lean towards using the third party one, assuming the data is just as good.

It might have been storepoint, but I think it's storerocket, they actually advertise to potential customers that they'll use the data directly from your Google sheet, so brands can avoid duplicate datasets.

We're probably splitting hairs anyway because how many brands have multiple apis? I expect it's not a lot.

iandees:
> how many brands have multiple apis? I expect it's not a lot

Yep, agree.

davidhicks:
[...]

PR #4773, opened 2023-02-10
Document apis in the output

Cj-Malone:
Briefly mentioned in https://github.com/alltheplaces/alltheplaces/issues/4757#issuecomment-1424639884

Similar to StructuredDataSpider, and the licenses by @brawer we want to give data consumers details about the source of the data so they can make a decision.

### DIFF
--- a/locations/spiders/fastbreak_us.py
+++ b/locations/spiders/fastbreak_us.py
@@ -1,13 +1,20 @@
+from locations.hours import OpeningHours, sanitise_day
 from locations.items import Feature
-from locations.storefinders.storerocket import StorerocketSpider
+from locations.storefinders.storerocket import StoreRocketSpider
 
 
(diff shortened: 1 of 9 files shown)


davidhicks:
The current set of store finders are all hosted externally by third parties (SaaS).

What dataset_attributes would be used, if any, for self-hosted store finders such as https://github.com/alltheplaces/alltheplaces/pull/4768

Cj-Malone:
In theory Where2GetItSpider can be self hosted, although I've not seen an example of it. I don't know about store finders that can only be self hosted, I'm not suggesting we go back and add this data to every spider.

iandees:
`source=api` feels a bit generic to me. I don't have a great suggestion for a replacement, though, so let's go with this and see how it works.

PR #4790, opened 2023-02-12
Switch from using Sitemap to 'Full Store List' web page for Specsavers

rjw62:
This spider previously used the Specsavers' sitemap at https://www.specsavers.co.uk/sitemap.xml but by comparing the data with OSM, I noticed a number of stores were missing from the sitemap. Specsavers also has a 'Full Store List' web page at https://www.specsavers.co.uk/stores/full-store-list which appears to be more complete.

I'm not sure if how I've coded it is the best way to do it, but when run locally the new spider appears to work, and returns 913 stores, compared to 838 from the current spider at the last ATP run.

This seems to be a big improvement. However, there's still at least one store that's missing from the Full Store List page: https://www.specsavers.co.uk/stores/addlestone .

### DIFF
--- a/locations/spiders/specsavers_gb.py
+++ b/locations/spiders/specsavers_gb.py
@@ -1,19 +1,20 @@
-from scrapy.spiders import SitemapSpider
-
+import scrapy
+import re
 from locations.structured_data_spider import StructuredDataSpider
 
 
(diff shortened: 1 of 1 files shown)


Cj-Malone (approved):
The preferred way would be a CrawlSpider, but this is good enough for me. Thank you.

PR #4792, opened 2023-02-13
Add National Seating & Mobility Spider

Cj-Malone:
https://github.com/osmlab/name-suggestion-index/issues/7771

### DIFF
--- a/locations/storefinders/store_locator_plus.py
+++ b/locations/storefinders/store_locator_plus.py
@@ -1,0 +1,31 @@
+from scrapy import Spider
+from scrapy.http import JsonRequest
+
+from locations.dict_parser import DictParser
+from locations.items import Feature
+
+
+class StoreLocatorPlusSpider(Spider):
+    dataset_attributes = {"source": "api", "api": "storelocatorplus.com"}
+
+    slp_dataset = None
+    slp_key = None
+
+    def start_requests(self):
+        yield JsonRequest(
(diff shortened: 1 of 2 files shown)


locations/storefinders/store_locator_plus.py
```diff
@@ -0,0 +1,31 @@
...
+        yield JsonRequest(
+            url=f"https://dashboard.storelocatorplus.com/{self.slp_dataset}/wp-json/myslp/v2/locations-map/search?action=csl_ajax_onload&api_key={self.slp_key}"
+        )
+
+    def parse(self, response, **kwargs):
+        if not response.json()["data"]["success"]:
+            return
+
+        for location in response.json()["data"]["response"]:
+            location["street_address"] = ", ".join(filter(None, [location.pop("address2"), location.pop("address")]))
```
iandees, line 24: I think you might want to flip `address` and `address2` here? I see a couple results with e.g. "Suite, Address" instead of "Address, Suite"

Cj-Malone: Oh I miss read you, this is what I intended. In GB I'd we have the small part of the address first. "Unit x, y Street", "Flat z, w Street". I'm happy to conform the the US format though.

It's like your weird date formats, you're just mixing the order up :p

iandees: 😄

PR #4808, opened 2023-02-14
Always output geometry

Cj-Malone:
Weirdly not outputting `geometry` is against the spec, it wants `"geometry": null` when there isn't any.


iandees:
For future reference, the [GeoJSON RFC 3.2](https://www.rfc-editor.org/rfc/rfc7946#section-3.2) says:

>    o  A Feature object has a member with the name "geometry".  The value
      of the geometry member SHALL be either a Geometry object as
      defined above or, in the case that the Feature is unlocated, a
      JSON null value.

Cj-Malone:
Yeah and geojson.io and geojsonlint.com both give errors without it.

PR #4810, opened 2023-02-14
Switch to filesystem cache

Cj-Malone:
As mentioned offline, slightly worried about the sqlite backend for the weekly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,3 +98,4 @@ ENV/
 *zip
 commands.txt
 http_cache.sqlite
+http_cache/
--- a/locations/__init__.py
+++ b/locations/__init__.py
@@ -1,1 +1,5 @@
+import requests_cache
+
 __version__ = "1.0"
+
+requests_cache.install_cache(expire_after=60 * 60 * 24 * 3, backend="filesystem")
--- a/locations/name_suggestion_index.py
+++ b/locations/name_suggestion_index.py
@@ -1,5 +1,4 @@
 import requests
-import requests_cache
 
 
 class Singleton(type):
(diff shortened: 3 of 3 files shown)


iandees:
I don't think Scrapy uses the `requests` library to make HTTP requests, so I don't think we should worry about how `scrapy` and `requests` caching interact. Is that your worry?

Cj-Malone:
It's not a worry that scrapy and requests_cache will clash, it's a worry that requests_cache and requests_cache will clash. We're running it 1,200+ times, I'm worried one of those time the sqlite database will be locked.

iandees:
Got it 👍🏻

jleedev:
This says that the sqlite database is only used for redirects. https://github.com/requests-cache/requests-cache/blob/main/docs/user_guide/backends/filesystem.md

But if you're worried about sqlite concurrency, add `wal=True`. https://github.com/requests-cache/requests-cache/blob/09df7d15f6d403b0e8412cc914bea226a3abe9ed/requests_cache/backends/sqlite.py#L44

Cj-Malone:
By default it uses the [sqlite backend](https://requests-cache.readthedocs.io/en/stable/modules/requests_cache.backends.filesystem.html), now we're on the filesystem backend, but it still uses some sqllite... so I'm still concerned·

I can actually POC this now (thanks [SO](https://stackoverflow.com/questions/14272715/how-can-i-lock-a-sqlite-database#answer-14274452)) with: `$ sqlite3 http_cache/redirects.sqlite`
```sql
PRAGMA locking_mode = EXCLUSIVE;
BEGIN EXCLUSIVE;
```
run a spider
`sqlite3.OperationalError: database is locked`

So if a spider crashes, the db will stay locked and all future spiders will fail to start.

But a spider has to crash, with the db locked, that's only used for redirects, when we're only using it for nsi and taginfo... So it's a bit of an edge case. But hoping that we don't get redirects isn't a good place to be.

My current proposal: wrap `install_cache` in a try, best case scenario nothing goes wrong, the try is never used, worst case scenario we get a locked db and all future spiders run without a cache. It's not ideal, but it's better than not running all future spiders entirely.

jleedev:
Right, so I'd try `backend=FileCache(wal=True)`, or `backend=SQLiteCache(wal=True)`.

Cj-Malone:
I did, it still threw an exception. Try out my PR, we've got a couple days before the weekly is ran.

PR #4811, opened 2023-02-14
Speed up preview map; modernize code

jleedev:
It turns out that creating hundreds of thousands of images in a web page isn't the best rendering technique.

Incorporate Supercluster (https://github.com/mapbox/supercluster) when the number of features exceeds a threshold (500) to allow interactively browsing a gigantic collection of points. The cluster styling is per their sample code (after the Leaflet MarkerCluster style).

Other approaches that worked, but not as well:

- The Leaflet MarkerCluster (https://github.com/Leaflet/Leaflet.markercluster) includes nice touches like animating a circle of unclustered markers when they overlap, but it was significantly slower.
- It is actually possible to get the map to render without clustering, by setting { preferCanvas: true }, and asking L.geoJson to create an L.circleMarker for point features. This eliminates the memory usage of creating a huge number of marker icons in the DOM, but still slows down on large collections.

Also bump the version of Leaflet (1.2.0 to 1.9.3), remove JQuery, and generally improve the page appearance.

### DIFF
--- a/preview/index.html
+++ b/preview/index.html
@@ -1,56 +1,104 @@
+<!doctype html>
 <html>
     <head>
-        <link rel="stylesheet" href="https://unpkg.com/leaflet@1.2.0/dist/leaflet.css"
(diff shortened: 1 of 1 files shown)


iandees:
Thanks!

iandees:
I believe I have to manually copy this to an S3 bucket, so it might not show up immediately when I merge.

PR #4814, opened 2023-02-15
Add Brumby's Bakeries (AU) (85 stores)

### LINKED ISSUE #4800: Brumby's Bakeries (AU/NZ bakery chain)
### Brand name

Brumby's

### Wikidata ID

Q4978794

### Store finder url(s)

https://www.brumbys.com.au/store-locator/

There is an embedded <script> with a storesObj object that can be parsed. Individual store pages do not have opening hours, so parsing this storesObj object is most efficient without limiting data captured.

davidhicks:
Source data is very messy and the clean-up is very painful. This spider is likely going to be fairly fragile and require fairly regular updates to match whatever mess will come next in the source data.

Closes #4800


Cj-Malone:
TBH I don't understand CDATA, why it exists, or how we should be working with it. But we can ignore it with regex. Then chompjs will parse it for us, but it turns out "stores" is a string, not a json array. So we can load that with normal json.

```python
        data_raw = response.xpath('//script[@id="theme_app_js-js-extra"]/text()').get()
        if m := re.search(r"var storesObj =\s*({.+});\s*var", data_raw):
            stores = json.loads(chompjs.parse_js_object(m.group(1))["stores"])
```

I've not looked at the source data for the opening hours, but I would ask, "is it worth it?"

davidhicks:
Improvements made to use chompjs for data extraction, and to also use add_ranges_from_string from https://github.com/alltheplaces/alltheplaces/pull/4893 to significantly simplify opening hours extraction.

davidhicks:
Ready for merge?

iandees:
Yep, thanks for the ping.

PR #4816, opened 2023-02-15
Add Harvey Norman (AU/NZ) (239 stores)

davidhicks:


davidhicks:
Hmmm geo blocked it would appear--HTTP 406 error being returned by the CI check.

davidhicks:
requires_proxy = True added. Thanks for confirming it it a geoblock

Cj-Malone:
Wouldn't we want it to be `requires_proxy = "AU"`?
True would be we'll get the country from the spider name, but in this case it isn't in the name, so it'd fall back to US.

Cj-Malone:
We may as well document it while we think about it, otherwise when the proxies are finally set up we'd still have to go through every spider to check it.

PR #4825, opened 2023-02-17
feat: Q8

Romeo-PHILLIPS:


locations/spiders/q8.py
```diff
@@ -0,0 +1,37 @@
...
+    def parse(self, response, **kwargs):
+        for store in response.json().get("Stations").get("Station"):
+            opening_hours = OpeningHours()
+            website = store.get("NodeURL")
+            ohs = store.get("OpeningHours")
+            for day, hours in ohs.items():
+                if hours == "":
+                    continue
+                hours = hours.split("-")
+                opening_hours.add_range(day, hours[0], hours[-1])
```
Cj-Malone, line 21: You're missing some data with multiple open periods in a day, eg: `"Wednesday": "08:00-12:00 13:00-19:00"`
Something like this should work:
```suggestion
                for period in hours.split(" "):
                    start_time, end_time = period.split("-")
                    opening_hours.add_range(day, start_time, end_time)
```

PR #4827, opened 2023-02-17
feat: Action

Romeo-PHILLIPS:


iandees:
It looks like the store query will sometimes return an empty JSON. Can you try handling that case so it doesn't cause an error?

PR #4835, opened 2023-02-20
feat: Handelsbanken SE

Romeo-PHILLIPS:


locations/spiders/handelsbanken_se.py
```diff
@@ -0,0 +1,51 @@
...
+from locations.categories import Categories, apply_category
+from locations.hours import DAYS, OpeningHours
+from locations.items import Feature
+
+
+class HandelsbankenSESpider(scrapy.Spider):
+    name = "handelsbanken_se"
+    item_attributes = {"brand": "Handelsbanken", "brand_wikidata": "Q1421630"}
+    start_urls = [
+        "https://locator.maptoweb.dk/handelsbanken.com/locator/points/where/CountryCode/eqi/se?callback=jQuery1820675693055071215_1676888222517&_=1676888222528"
```
Cj-Malone, line 14: FYI, the `_=rand` parameter is a "cache buster" sites and random noise to the url to prevent browsers cacheing the data. It's not relevant to us.

The `callback=func_name` is why you're not getting real json back. It's to run a JavaScript function.

The end of the url `where/CountryCode/eqi/se` looks like a query to only select SE stores. Trimming it at least give us a few `"countryCode":"DK"`. I've not looked at the data properly though.

https://locator.maptoweb.dk/handelsbanken.com/locator/points

PR #4836, opened 2023-02-20
feat: Spar NL

Romeo-PHILLIPS:
I need help with this one, I'm using SitemapSpider x StructuredDataSpider but its not managing to get the coordinates. Even though I see them available it doesn't seem able to extract them. I even tried to force set them in `post_process_item` but doesn't work

### DIFF
--- a/locations/spiders/spar_nl.py
+++ b/locations/spiders/spar_nl.py
@@ -1,0 +1,10 @@
+from scrapy.spiders import SitemapSpider
+
+from locations.structured_data_spider import StructuredDataSpider
+
+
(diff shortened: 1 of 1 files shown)


Cj-Malone:
I've not debugged properly, just a quick look at the data. I expect the issue will be re the comma instead of the period. `"geo": { "@type": "GeoCoordinates", "latitude": "51,983034", "longitude": "5,6476909" }` vs `"geo": { "@type": "GeoCoordinates", "latitude": "51.983034", "longitude": "5.6476909" }`

I imagine this will happen semi often, so I'm open for changes for `LinkedDataParser` to fix it.

Cj-Malone (approved):

PR #4838, opened 2023-02-20
feat: Burger King SE

Romeo-PHILLIPS:


Romeo-PHILLIPS:
Question, is it normal that the CI gets 63 items but I get 122 in local?

iandees:
> is it normal that the CI gets 63 items but I get 122 in local?

Yes. The CI will stop at ~60 seconds and give a timeout, whereas you probably let yours run longer and get more stores locally.

PR #4855, opened 2023-02-22
feat: BMW Global

Romeo-PHILLIPS:
I have some questions with this one. The API returns POIs that each have the attribute `key` as an ID for a dealership, but the dealership can appear multiple time with a different `category` which represents the brand that is sold there, like:

```
{
                "key": "08904_3",
                "name": "Westcoast BMW",
                "countryCode": "AU",
                "country": "Australia",
                "state": "WA",
                "city": "Wangara",
                "postalCode": "6065",
                "street": "17 Luisini Rd",
                "lng": 115.828932,
                "dist": -1.0,
                "oh": null,
                "category": "BMW",
                "attributes": {
                    "distributionPartnerId": "08904",
                    "distributionCode": "EH",
                    "outletId": "3",
                    "ownerId": "08904",
                    "homepage": "http://www.westcoastbmw.com.au",
                }
}
```
and 

```
{
                "key": "08904_3",
                "name": "Westcoast BMW",
                "countryCode": "AU",
                "country": "Australia",
                "state": "WA",
                "city": "Wangara",
                "postalCode": "6065",
                "street": "17 Luisini Rd",
                "additionalStreet": null,
                "settlement": null,
                "lat": -31.795432,
                "lng": 115.828932,
                "dist": -1.0,
                "oh": null,
                "category": "MINI",
                "attributes": {
                    "distributionPartnerId": "08904",
                    "distributionCode": "EH",
                    "outletId": "3",
                    "ownerId": "08904",
                    "homepage": "http://www.westcoastbmw.com.au",
                 }
}
```
Multiple brands (from the same groups) can share the same dealership ID.
So currently in the code, if a dealership contains multiple brand, I set the brand to be "BMW". But I'm not sure if this is correct.

For example using this query `https://www.bmw.com/c2b-localsearch/services/api/v4/clients/BMWSTAGE2_DLO/DE/pois?brand=BMW_BMWI_BMWM&cached=off&language=en&lat=0&lng=0&maxResults=7000&showAll=true&unit=km&country=AU` for all Australia POIs we have a dealership that sells every brand from the group but on the MINI website it is considered a separate POI owned by MINI (https://www.mini.com.au/find-dealers)
Capture d’écran 2023-02-22 à 11 59 43

My question is, should I keep my current rule in place or should I set the key to be: `key = "f{brand}-{key}` to match what the other brands from the group have in place.

We can also filter brands in the request by adding the `category` parameter in the URL like `category=BMW`.


Cj-Malone:
I don't think the current OSM system for dealerships is a good one. We (miss)use the `brand`/`brand:wikidata` as the brands of cars sold, not the brand of the POI. I don't have a solution to that. Maybe `sells:cars:BMW=yes` or `sells:Q26678=yes`.

As for ATP, this seems like the best compromise. Thank you.

PR #4858, opened 2023-02-22
schnucks: fix and rename to schnucks_us

davidhicks:
Returns 114 locations.


iandees:
FYI, there's no need to force-push changes. It all gets squashed to a single commit anyway by Github.

iandees:
Is there another Schnuck's you're planning on adding? Since spider names are used as part of the ID for each entry, we try not to rename spiders unless there's a need to.

davidhicks:
> Is there another Schnuck's you're planning on adding? Since spider names are used as part of the ID for each entry, we try not to rename spiders unless there's a need to.

No, I just renamed it to make it clear it's a US brand and to allow item["country"] to self-populate from the spider name. Should I undo the renames? For new spiders, should they have a country suffix if the brand is only in one country?

iandees:
Don't worry about the renames, sorry. I see that you're fixing them at the same time. I thought it was only a file rename.

New spiders should have country code suffixes, yea.

PR #4859, opened 2023-02-22
take5: fix and rename to take_5_us

davidhicks:
Returns 1238 locations.


iandees:
Looks like it's giving us a 403: https://data.alltheplaces.xyz/ci/4243248698/take_5_us/log.txt

Probably because it's blocking AWS IP ranges or user agents. You could try changing the user agent?

davidhicks:
No luck with a different user agent. It appears to be the AWS IP ranges that are being blocked. I'm not sure requires_proxy = "US" is the right answer here either

PR #4869, opened 2023-02-23
bankofhawaii: fix and rename to bank_of_hawaii_us

davidhicks:
Includes branches and ATMs in Hawaii for a total of 258 locations.


locations/spiders/bank_of_hawaii_us.py
```diff
@@ -0,0 +1,37 @@
...
+    start_urls = ["https://www.boh.com/bohapi/get-locations?lat=21.33&lng=-157.845934&radius=250"]
+
+    def parse(self, response):
+        for location in response.json()["locations"]:
+            item = DictParser.parse(location)
+            item["lat"] = location["geocode"]["latitude"]
+            item["lon"] = location["geocode"]["longitude"]
+            oh = OpeningHours()
+            for day_name, day in location["operationalHours"]["hours"].items():
+                if day["isClosed"] == True:
```
iandees, line 22:
```suggestion
                if day["isClosed"] is True:
```

or

```suggestion
                if day["isClosed"]:
```

locations/spiders/bank_of_hawaii_us.py
```diff
@@ -0,0 +1,37 @@
...
+                    continue
+                oh.add_range(day_name.title(), day["hours"][0]["openTime"], day["hours"][0]["closeTime"], "%H:%M:%S")
```
Cj-Malone, line 24: Is there ever `day["hours"][1]`? Should we loop it?

locations/spiders/bank_of_hawaii_us.py
davidhicks: Thanks, fixed

locations/spiders/bank_of_hawaii_us.py
davidhicks: I haven't seen multiple hours ranges, but it doesn't hurt to add the logic for it so I've fixed it :)

Issue #4854, opened 2023-02-22
screwfix_gb spider failing


rjw62:
In the last couple of weeks the screwfix_gb spider has stopped working. It looks like Screwfix have changed their store page URLs, and probably also the formatting of the pages. I can only half fix things though.

First the regex in the line
```rules = [Rule(LinkExtractor(allow=r"\/stores\/(\d+)\/.+$"), callback="parse_sd")]```
needs to be changed to account for the new URLs. The store ID is now two uppercase letters and a number, so something like
```rules = [Rule(LinkExtractor(allow=r"\/stores\/([A-Z][A-Z][0-9])\/.+$"), callback="parse_sd")]```
would work.

However, the StructuredDataParser then fails to extract the data from the store pages. Running
```pipenv run scrapy sd https://www.screwfix.com/stores/AC3/whitby --wanted-types=HardwareStore```
returns
```failed to decode structured data```

The data looks like it's there, e.g. at https://www.screwfix.com/stores/AC3/whitby we have
```<script type="application/ld+json">{&quot;@context&quot;:&quot;https://schema.org&quot;,&quot;@type&quot;:&quot;HardwareStore&quot;,&quot;@id&quot;:&quot;https://www.screwfix.com/stores/AC3/whitby&quot;,&quot;name&quot;:&quot;Screwfix Whitby&quot;,&quot;address&quot;:{&quot;@type&quot;:&quot;PostalAddress&quot;,&quot;addressCountry&quot;:&quot;United Kingdom&quot;,&quot;addressLocality&quot;:&quot;Whitby&quot;,&quot;addressRegion&quot;:&quot;Yorkshire&quot;,&quot;postalCode&quot;:&quot;YO22 4PZ&quot;,&quot;streetAddress&quot;:&quot;Unit 1&quot;,&quot;telephone&quot;:&quot;03330 112 112&quot;},&quot;geo&quot;:{&quot;@type&quot;:&quot;GeoCoordinates&quot;,&quot;latitude&quot;:54.4725969,&quot;longitude&quot;:-0.6019658}}</script>```
I don't know if it's correct HTML or not, but I'm guessing that the `&quot;` encoding may be messing things up for the parser. I'm afraid it's currently beyond me to work out how we should fix this.

FYI, decoding the `&quot;`s in the above, and pretty-printing the JSON, gives
```
{
	"@context": "https://schema.org",
	"@type": "HardwareStore",
[...]

PR #4871, opened 2023-02-24
Fix broken Screwfix spider

rjw62:
See discussion in #4854.

### DIFF
--- a/locations/spiders/screwfix_gb.py
+++ b/locations/spiders/screwfix_gb.py
@@ -10,5 +10,5 @@ class ScrewfixGBSpider(CrawlSpider, StructuredDataSpider):
     allowed_domains = ["www.screwfix.com"]
     download_delay = 0.1
     start_urls = ["https://www.screwfix.com/stores/all"]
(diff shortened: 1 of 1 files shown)


locations/spiders/screwfix_gb.py
```diff
@@ -1,14 +1,27 @@
...
+from locations.linked_data_parser import LinkedDataParser
 
 
-class ScrewfixGBSpider(CrawlSpider, StructuredDataSpider):
+class ScrewfixGBSpider(CrawlSpider):
     name = "screwfix_gb"
     item_attributes = {"brand": "Screwfix", "brand_wikidata": "Q7439115"}
     allowed_domains = ["www.screwfix.com"]
-    download_delay = 0.1
+    download_delay = 1
```
Cj-Malone, line 14:
```suggestion
```

We don't need to define the default, this way we can change it later without having to change spiders.

Cj-Malone (approved):
LGTM, thanks

PR #4880, opened 2023-02-24
feat: Main Event Update

Romeo-PHILLIPS:


Cj-Malone:
Looks like this loses coords, we should still collect them in pre/post processing.

locations/spiders/main_event.py
```diff
@@ -1,72 +1,18 @@
...
-                    day=spec["dayOfWeek"][0:2],
-                    open_time=spec["opens"],
-                    close_time=spec["closes"],
-                )
+    sitemap_urls = ["https://www.mainevent.com/sitemap.xml"]
+    sitemap_rules = [(r"\/locations\/", "parse_sd")]
 
-        return opening_hours.as_opening_hours()
+    def post_process_item(self, item, response, ld_data, **kwargs):
+        item["ref"] = item["name"]
```
Cj-Malone, line 17: I'd prefer the URL as a ref, it's less likely to have a collision

iandees:
Looks like it isn't grabbing lat/lon anymore. Can you pull that back in?

Romeo-PHILLIPS:
@iandees @Cj-Malone the website not longer seems to give out lat/long. Previous way seemed to hook into a google maps scripts but this has seemed to have been replaced by a [screenshot](https://images.ctfassets.net/3dar4x4x74wk/6m0ftZmtCU8nqUm9wc9vrw/1509b23a2fba3fa23c1c16b26c75df0e/ABQ_MAP.png) of the location on a map. 

Looking at the run history can also see it hasn't been pulling lat longs

iandees:
It seems like there's a latitude and longitude in a big JSON blob on the page. I wonder if it's easy enough to get that.

Cj-Malone:
Good old regex: `\\\"latitude\\\":(-?\d+\.\d+),\\\"longitude\\\":(-?\d+\.\d+),`

PR #4896, opened 2023-02-28
feat: Scania

Romeo-PHILLIPS:


locations/spiders/scania.py
```diff
@@ -0,0 +1,136 @@
...
+            yield Feature(
+                {
+                    "ref": store.get("scaniaId"),
+                    "country": store.get("domicileCountry").get("countryCode"),
+                    "name": address_details.get("addressee"),
+                    "street_address": postal_address.get("street").get("streetName").get("value"),
+                    "phone": address_details.get("fixedPhoneNumber").get("subscriberNumber"),
+                    "email": address_details.get("electronicMailAddress"),
+                    "postcode": postal_address.get("postalCode"),
+                    "city": postal_address.get("city").get("value"),
```
iandees, line 131: This seems to include the city and the state together in one string. Can you split them apart?

Romeo-PHILLIPS: This seems to only be the case for USA, do you want me to add a condition for it?

PR #4915, opened 2023-03-02
Add a test for Belgium phone number cleanup

iandees:
I noticed that the phone numbers in #4912 were not getting parsed properly. I'm adding a test here to see why.

### DIFF
--- a/.github/workflows/pytest-pr.yaml
+++ b/.github/workflows/pytest-pr.yaml
@@ -5,6 +5,8 @@ on:
     paths:
     - '.github/workflows/*.yaml'
     - 'locations/*.py'
+    - 'tests/*.py'
+
 
 jobs:
   pytest:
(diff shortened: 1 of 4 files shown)


iandees:
Hmm, this was supposed to fail.

iandees:
@Cj-Malone or @brawer, the suzuki_be spider here is getting country code from the spider name, but the phone number parser is saying the phone number is invalid even though an example number from the spider parses fine in the test I added. Any ideas why this might be happening?

Cj-Malone:
The spider is failing because I added a split phone by "/" then process each independently. In which case, neither is valid. When we encounter an invalid number we put it back in the list and let consumers make a decision.

Anecdotally I think split by "/" was the wrong call, I think I've seen more phone numbers formatted like so than multiple numbers separated by "/". No data, but I'd be OK with removing that.

Not throwing bad phone numbers away. I can't remember which "side" I was on, and so no strong opinion now. I think we can assume the phonenumbers library is good enough so I'd be OK throwing the others away, but I am a little cautious we would throw away "bad" data that is fine, we should look at the data. (I very much doubt I'm gonna look at the data)

No idea why the test is actually working. It looks like it should be failing to me. I get home tomorrow so I can have a proper look then.

Cj-Malone:
So the CI is having a moment right now, but I think the reason you've got those ✔️ is because the tests weren't actually ran.

iandees:
Oh that'll do it.

PR #4918, opened 2023-03-02
Add Caffe Nero GB Spider

### LINKED ISSUE #4917: Caffe Nero GB
### Brand name

Caffe Nero

### Wikidata ID

Q675808

### Store finder url(s)

https://caffenero.com/uk/stores/

Looking at the source code of the above URL, line 295 appears to show a JavaScript data object called `storesData` which appears to contain store data for many (or all) of the GB stores, including the URL for each one.

Could this variable be extracted and broken up by a spider?

Cj-Malone:
Fixes #4917

### DIFF
--- a/locations/spiders/caffe_nero_gb.py
+++ b/locations/spiders/caffe_nero_gb.py
@@ -1,0 +1,30 @@
+import chompjs
+from scrapy import Spider
+
+from locations.categories import Categories, apply_category
+from locations.dict_parser import DictParser
+
+
+class CaffeNeroGBSpider(Spider):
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Well done for the quick addition @Cj-Malone 👍

Just out of interest, the base map linked to via the bot above seems very old, could it be updated to use the newest base map?

iandees:
> the base map linked to via the bot above seems very old, could it be updated to use the newest base map?

It uses the Stamen tiles simply because it's a better background to highlight the places on top than the OSM.org tiles. It would be great to have a better map if you want to update it.

PR #4919, opened 2023-03-03
feat: Add spider for jack_and_jones

abbensid2:


Cj-Malone:
Looks like lat/lon are flipped.

Issue #4310, opened 2022-12-11
Tapi Carpets results include closed branch holding pages


rjw62:
In the results returned by the tapi_gb spider there are a number of "branches" that I am assuming have previously closed, but the company has kept the web page and sitemap links, presumably for SEO purposes. The web pages for these zombie branches look different from the regular branch pages, but still contain some bare-bones JSON-LD data for the "branch".

Example real branch page: https://www.tapi.co.uk/stores/ashton-under-lyne

Example zombie branch pages: https://www.tapi.co.uk/stores/altrincham & https://www.tapi.co.uk/stores/bury

I'm not sure of the best way to detect and remove these zombie branches. They seem to have an empty openinghours array, and empty address strings apart from "postalCode": "SE6 2QU". However, that postcode is used for a genuine branch https://www.tapi.co.uk/stores/catford so you can't just remove all the branches with that postcode.

Cj-Malone:
As usual I have opinions.

- [ ] We need docs for brands, not just us. We need a document to encourage brands to enable us to scrape them, to follow schema.org, to not block us with Cloudflare or similar. To tell them why they should care.
- [ ] We need a way to better highlight source errors, like your Chain Reaction tool, but more targeted on the source data not OSM. Noting issues with opening hours, coords, missing details, phone number issues and other stuff we clean in pipelines. This needs to be pretty and specific.
- [ ] We need a way to communicate with the brands. Clearly messaging customer service or admin@tld isn't going to work. I would love @schemaorg to require, (or strongly recommend) data providers put in contact details for the "data owner" who can fix the source errors, but even that spec was written today (it may already be possible via `contactPoint` or `accountablePerson`), it'd still take years to get adoption. We need a solution for today.

Cj-Malone:
#8767

PR #4930, opened 2023-03-06
Add Specsavers (SE) (118 stores)

davidhicks:


Cj-Malone:
A) This really annoys me that they have broken Microdata. How hard is it for these organisations? And how can we contact them? https://github.com/alltheplaces/alltheplaces/issues/4310#issuecomment-1345656688
B) Is it possible to combine a bunch of these into a global spider?

davidhicks:
> A) This really annoys me that they have broken Microdata. How hard is it for these organisations? And how can we contact them? [#4310 (comment)](https://github.com/alltheplaces/alltheplaces/issues/4310#issuecomment-1345656688) B) Is it possible to combine a bunch of these into a global spider?

I tried MicrodataParser and gave up quickly because they're only using it for postcode and city fields, hence it's easier and cleaner to just extract these two fields with XPath similar to all the other fields we need to obtain.

You're right that many of these European Specsavers spiders are similar in that generally it's the same XPath queries to extract information from each page. There are a few minor quirks/differences between countries but that could be resolved with some if statements in a combined European Specsavers spider. These quirks, as well as the 3-4 different styles of spiders needed for various Specsavers country websites, led me to just create one spider per country as I wouldn't be surprised to see Specsavers websites eventually consolidate one-by-one to use of e.g. ld+json (as is the case for e.g. specsavers_au) or yext JSON API data (for e.g. specsavers_nz).

PR #4893, opened 2023-02-28
OpeningHours: introduce add_ranges_from_string

davidhicks:
add_ranges_from_string is added as a parser function designed to help translate from opening hours provided as human readable text in a fuzzy format.

This function is intended to add opening hours ranges based on the following type of ranges that may be found in source data:

1. Monday-Wednesday: 5pm - 7pm
2. Monday-Wednesday 08:00-14:00
3. Monday to Tuesday: 15:00:01 to 16:35
4. Thurs 2PM-6:30PM
5.  Fri    9am  -  11am
6. Weekends: 8:00 AM to 6:00 PM
7. Monday to Thursday 7am to 7pm, Friday 7am to 10pm, Weekends CLOSED

This function works with a source that may mix and match from the type of examples listed above. For example, if a source uses both 12h and 24h times to describe opening hours.

This function does not currently support extraction of multiple opening hours on the same day of a week. Only the first opening time of each day is extracted. This function should therefore not be used if the source data has the potential to list multiple opening hours for the same day.

This function does not currently support named times of each day such as 'Midday' and "Midnight". It is possible to replace these prior to the input string being provided to this function.

This function allows specification of days and day ranges in langauges other than the default English.

PR #4934, opened 2023-03-06
Add Schnitz (AU) (75 stores)

davidhicks:


locations/spiders/schnitz_au.py
```diff
@@ -0,0 +1,53 @@
...
+            .replace(" pm", "PM")
+            .split()
+        )
+        hours_raw = [hours_raw[n : n + 3] for n in range(0, len(hours_raw), 3)]
+        for day in hours_raw:
+            if day[0] == "Jan26":
+                continue
+            if day[1] == "0:00AM" and day[2] == "0:00AM":
+                continue
+            oh.add_range(DAYS_EN[day[0]], day[1], day[2], "%I:%M%p")
```
Cj-Malone, line 50: Another none standard `day[0]` such as "Jan25" would crash this and we'd lose the POI. A reminder for us all that we'll probably get a bunch less POIs around christmas.

Untested, but something like the below using `sanatise_day` might help:
```suggestion
        for day_raw, start_time, end_time in hours_raw:
            if day := sanatise_day(day_raw):
                if start_time == "0:00AM" and end_time == "0:00AM":
                    continue
                oh.add_range(day, start_time, end_time, "%I:%M%p")
```

davidhicks: Ideally I'd like to use https://github.com/alltheplaces/alltheplaces/pull/4893 because it appears they're not using a structured datetime notation, and I wouldn't be surprised to see "Weekdays: ..." or "Sat-Sun: ..." in the future or other free text names. It's a little bit overkill for this situation, but makes the code easier to read versus keeping track of and iterating over a list or dictionary of raw day range names and hour ranges.

PR #4936, opened 2023-03-06
big_boy: fix and rename to big_boy_us

davidhicks:
Depends on https://github.com/alltheplaces/alltheplaces/pull/4893 as the opening hours for this store are very messy and include multiple formatting styles, day ranges, some typos, etc.


davidhicks:
Hmmm, the relatively unstable extract_google_position strikes again.

Can the function be replaced with a regular expression that extracts latitudes and longitudes with `r"@(\d+),(\d+),\d+z"` or returns nil if no match is found. If this regular expression doesn't find the latitude/longitude, there is no point trying a very naive `split` on the URL to try and find these values by assumed location.

davidhicks:
Branch is a mess, will send a new request.

PR #4938, opened 2023-03-07
feat: audi smart

charlesdassonville:


locations/spiders/audi.py
```diff
@@ -0,0 +1,131 @@
...
+        "UKR",
+        "USA",
+        "UZB",
+        "ZAF",
+    ]
+
+    def start_requests(self):
+        for iso_3 in self.countries:
+            iso_2 = CountryUtils().to_iso_alpha2_country_code(country_str=iso_3)
+            for lat, lon in point_locations("eu_centroids_120km_radius_country.csv", iso_2):
```
iandees, line 98: If we're only iterating through EU points, will we fetch any US data?

locations/spiders/smart.py
```diff
@@ -0,0 +1,91 @@
...
+                "ref": row.get("baseInfo").get("externalId"),
+                "name": row.get("baseInfo").get("name1"),
+                "brand": self.name,
+                "city": row.get("address")["city"],
+                "state": row.get("address").get("region").get("region"),
+                "street_address": row.get("address").get("line1"),
+                "postcode": row.get("address").get("zipcode"),
+                "country": row.get("address").get("country"),
+                "lat": float(lat) if lat else None,
+                "lon": float(lon) if lon else None,
```
iandees, line 84: The GeoJSON output pipeline step does this for you, so your spider only needs to set lat/lon.

locations/spiders/audi.py
```diff
@@ -0,0 +1,131 @@
...
+    def parse(self, response):
+        data = response.json()
+        store_data = data.get("data").get("dealersByGeoLocation").get("dealers")
+        count = data.get("data").get("dealersByGeoLocation").get("meta").get("resultCount")
+        assert count < 100
+        for store in store_data:
+            properties = {
+                "ref": store.get("dealerId"),
+                "name": store.get("name"),
+                "addr_full": " ".join(store.get("address")),
```
Cj-Malone, line 119: My personal preference:

```suggestion
                "addr_full": ", ".join(store.get("address")),
```

But I'm open for a discussion if anyone thinks otherwise.

locations/spiders/audi.py
```diff
@@ -0,0 +1,131 @@
...
+                    },
+                    meta={"country_code": iso_2},
+                )
+
+    def parse(self, response):
+        data = response.json()
+        store_data = data.get("data").get("dealersByGeoLocation").get("dealers")
+        count = data.get("data").get("dealersByGeoLocation").get("meta").get("resultCount")
+        assert count < 100
```
Cj-Malone, line 114: Is this meant to stay in?

locations/spiders/audi.py
```diff
@@ -0,0 +1,131 @@
...
+                    data={
+                        "operationName": "Dealer",
+                        "variables": {"country": iso_3, "lat": float(lat), "lng": float(lon)},
+                        "query": "query Dealer($country: String, $services: String, $lng: Float!, $lat: Float!) {\n  dealersByGeoLocation(\n  limit: 100\n    country: $country\n    serviceFilter: $services\n  lat: $lat\n lng: $lng\n) {\n    dealers {\n      ...FragmentDealerFields\n      __typename\n    }\n    meta {\n      resultCount\n      __typename\n    }\n    __typename\n  }\n}\n\nfragment FragmentDealerFields on Dealer {\n  dealerId\n  vCardUrl\n matPrimaryCode\n  kvpsPartnerKey\n  name\n  address\n  latitude \n  longitude\n houseNumber\n  street\n region\n  city\n  zipCode\n  services\n  distance\n  phone\n  additionalData {\n    displayName\n    locationName\n    __typename\n  }\n  openingHours {\n    departments {\n      id\n      departmentName\n      openingHours {\n        id\n        open\n        timeRanges {\n          openTime\n           \n          __typename\n        }\n        __typename\n      }\n      __typename\n    }\n    __typename\n  }\n  __typename\n}\n",
```
Cj-Malone, line 105: So this query:
```graphql
query Dealer($country: String $services: String $lng: Float! $lat: Float!) {
    dealersByGeoLocation(
        limit: 100
        country: $country
        serviceFilter: $services
        lat: $lat
        lng: $lng) {
            dealers {
                ...FragmentDealerFields
                __typename
            }
            meta {
                resultCount
                __typename
            }
            __typename
        }
    }

    fragment FragmentDealerFields on Dealer {
        dealerId
        vCardUrl
        matPrimaryCode
        kvpsPartnerKey
        name
        address
        latitude 
        longitude
        houseNumber
        street
        region
        city
        zipCode
        services
        distance
        phone
        additionalData {
            displayName
            locationName
            __typename
        }
        openingHours {
            departments {
                id
                departmentName
                openingHours {
                id
                open
                timeRanges {
                    openTime
                    __typename
                }
                __typename
            }
            __typename
        }
        __typename
    }
    __typename
}
```

The first thing that jumps out is there is only `openTime` adding `closeTime` would get you it, then we could collect the opening times.

And after a little poke around, `dealersByMarket` may be best for us, `A-GBR` A being Aldi, GBR being GB. Not tested with other countries.

```graphql
query Dealer {
    dealersByMarket(market: "A-GBR") {
        dealers {
            dealerId
            name
            services
            latitude
            longitude
            phone: phoneInternational
            fax: faxInternational
            email
            url
            operator: chainId
            country
            openingHours {
                departments {
                    departmentName
                    openingHours {
                        timeRanges {
                            openTime
                            closeTime
                        }
                    }
                }
            }
            address
            houseNumber
            street
            city
            zipCode
        }
    }
}
```

Cj-Malone: Actually I think the proper way to do it is:
```graphql
query Dealer ($market: Market!) {
    dealersByMarket(market: $market) {
        dealers {
            dealerId
            name
            services
            latitude
            longitude
            phone: phoneInternational
            fax: faxInternational
            email
            url
            operator: chainId
            country
            openingHours {
                departments {
                    departmentName
                    openingHours {
                        timeRanges {
                            openTime
                            closeTime
                        }
                    }
                }
            }
            address
            houseNumber
            street
            city
            zipCode
        }
    }
}
```
With variables:
```json
{"market": "A-GBR"}
```
But it my first day with graphql so I could be wrong.

Cj-Malone: And it looks like we can get all the markets with:

```graphql
query Dealer {
    marketInfo {
        markets {
            market
            scope
        }
    }
}
```

Scope is either `public`, `kdw` or `bad`, not sure on the meaning.

locations/spiders/audi.py
charlesdassonville: no, i removed it.
however how should we raise a warning if the number of pois is higher than 100 ?

charlesdassonville: cause in this case, we will miss some

Cj-Malone: Urm, `self.logger` would be the way, but I think count is the amount of results sent, so it was never going to be over 100, and I don't know if you are dropping POIs anyway as you are doing a request every 120km.

But the `dealersByMarket` returns 113 for "A-GBR" and 1,237 for "A-DEU" so it needs to be removed when switching to that query.

Pipfile
```diff
@@ -30,3 +30,4 @@ scrapy-playwright = "*"
 requests-cache = "*"
 requests = "*"
 pycountry = "*"
+pandas = "*"
```
iandees, line 33: I would prefer not to add a huge dependency like `pandas`, if possible. The filtering you're doing with the CSV can be done with built-in Python code.

iandees (commented):
Can you write up a comment or some other documentation describing how you created the h3.csv file? If we ever want to re-create or update the file, we'll need to know how to do it.

Pipfile
Cj-Malone: I'm a little confused as why it's in this PR as it's not used in these spiders?

I'm not against it being included, IIRC @jleedev was thinking about h3 a while ago.

locations/spiders/audi.py
```diff
@@ -0,0 +1,55 @@
...
+class AudiSpider(Spider):
+    name = "audi"
+    item_attributes = {
+        "brand": "Audi",
+        "brand_wikidata": "Q23317",
+    }
+    graphql_url = "https://dev-dealer-graphql.apps.emea.vwapps.io/"
+
+    def start_requests(self):
+        market_request = requests.post(
```
Cj-Malone, line 17: I'd vote against `requests` being used here, I'd yield a `JsonRequest`, let scrapy do it's thing then another callback to yield the market requests.

Pipfile
charlesdassonville: @Cj-Malone  i created this function and file for having a complete world wide grid in several sizes, to use it for audi. But it appears after that audi scrapping can be done differently (cf commits). 
I still think a world wide grid in different sizes can be useful (instead of several files for each countries or eu only)
i can move it into another PR if you want

charlesdassonville: @iandees we can do the filtering manually but is it a good practice ? and it might take longer and take more memory. 
we can discuss this in another PR if you want. i'll remove the commits from this one

locations/spiders/audi.py
charlesdassonville: updated :-)

Cj-Malone:
Audi isn't getting a category. Maybe just `apply_category(Categories.SHOP_CAR, properties)` or maybe we need to set it based on the `services` returned?

It seems the Smart spider is never getting more than 10 POIs per country. I can't see a good solution to that.

PR #4944, opened 2023-03-07
Add KFC (MY) (747 stores)

davidhicks:


locations/spiders/kfc_my.py
```diff
@@ -0,0 +1,29 @@
...
+from locations.hours import DAYS, OpeningHours
+
+
+class KFCMYSpider(scrapy.Spider):
+    name = "kfc_my"
+    item_attributes = {"brand": "KFC", "brand_wikidata": "Q524757"}
+    allowed_domains = ["kfc.com.my"]
+    start_urls = [
+        "https://kfc.com.my/graphql?query=query+allLocation%7BallLocation%7Blocations%7Baddress+city+code+coleslaw+country+created+curbside+delivery_close+delivery_open+delivery_tier+dinein+drivethru+gesStoreId+is_breakfast+is_delivery+is_selfcollect+lat+launch_date+legacy_store_id+locationId+long+name+phone+riderType+selfcollect_close+selfcollect_open+selfcollect_tier+polygon+smartbox+state+storeCmgId+storeName+updated+zip+disabled_skus+is_breakfast+coleslaw+drivethru+smartbox+dinein+__typename%7D__typename%7D%7D&operationName=allLocation&variables=%7B%7D"
+    ]
```
Cj-Malone, line 13: So I've been playing with graphql today. This request is actually:

```graphql
query allLocation {
    allLocation {
        locations {
            address
            city
            code
            coleslaw
            country
            created
            curbside
            delivery_close
            delivery_open
            delivery_tier
            dinein
            drivethru
            gesStoreId
            is_breakfast
            is_delivery
            is_selfcollect
            lat
            launch_date
            legacy_store_id
            locationId
            long
            name
            phone
            riderType
            selfcollect_close
            selfcollect_open
            selfcollect_tier
            polygon
            smartbox
            state
            storeCmgId
            storeName
            updated
            zip
            disabled_skus
            is_breakfast
            coleslaw
            drivethru
            smartbox
            dinein
            __typename
        }
    __typename
    }
}
```

We can remove any fields from that which we don't need. Removing `disabled_skus` halves the response size. `coleslaw` is listed twice :p

 We can also rename fields here, instead of cleaning up after dictparser we could:
```patch
-            locationId
+            ref: locationId
```

davidhicks: Fixed :)

Cj-Malone (approved):
Good enough for me thanks.

My comments, just comments, no changes necessary:
- I've started using `JsonRequest` more, even for requests like this when it's not needed, IMO it's polite to send the Accept header when we want JSON.
- We have `add_days_range` which would stop you computing the open and close time 7 times.
- I'd have collected drive through status
- `_typename` isn't necessary in the query, we can drop it like any other field.
- We seem to have been POSTing to qraphql most frequently, I don't know if its technically better, but I like patterns.
- We can now output `opening_hours` as `OpeningHours` not just the string. When we use the class we can skip validation in the pipeline.
- Callback `parse` is the default, you don't need to define it here.
- We could alias `locationId` to `ref` in the query, then we wouldn't need to clean it up after DictParser. `ref: locationId`

davidhicks:
Thanks for the helpful feedback! It'll help me improve spiders in the future.

I've implemented everything recommended except that method="POST" doesn't work for this GraphQL endpoint.

PR #4946, opened 2023-03-08
Collect mcdonalds drive through details

Cj-Malone:
And modernise opening hours


davidhicks:
It turns out the ap-prod server that libmaccas-rs uses is just for Australia. There are seemingly other servers available that may return location details for other countries, and at least some of these servers are listed at https://github.com/paulbtw/mcbrokenio/blob/master/README.md

Cj-Malone:
McDonalds really is a sea of APIs. They should write a spider for us.

PR #4949, opened 2023-03-08
fix: Piti bateau

charlesdassonville:


locations/spiders/petit_bateau.py
```diff
@@ -7,5 +7,5 @@ class PetitBateauSpider(SitemapSpider, StructuredDataSpider):
     name = "petit_bateau"
     item_attributes = {"brand": "Petit Bateau", "brand_wikidata": "Q3377090"}
     sitemap_urls = ["https://stores.petit-bateau.com/sitemap.xml"]
-    sitemap_rules = [(r"https://stores\.petit-bateau\.com/(?:\w+)/(?:\w+)/(?:\w+)/(?:\w+)", "parse_sd")]
+    sitemap_rules = [(r"https://stores.petit-bateau.com/", "parse_sd")]
```
davidhicks, line 10: There should be some check done to filter out URLs such as https://stores.petit-bateau.com/austria and https://stores.petit-bateau.com/fr/ukraine which aren't store pages. Additionally, we shouldn't be trying to request and parse the same page multiple times but in different languages? Additionally, it should always be necessary to escape the periods in the domain name, as well as the forward slash in the path.

Perhaps the regular expression should be:
`r"https:\/\/stores\.petit-bateau\.com\/[a-z\-]{3,}\/[^\/]+`

davidhicks (changes requested):

locations/spiders/petit_bateau.py
charlesdassonville: The new re gives me around 300 pois.. there should be more than 3k.

PR #4950, opened 2023-03-08
Add Amasty Store Locator store finder and some example store spiders

davidhicks:


iandees:
Thanks for including the doc on the spider helper.

PR #4951, opened 2023-03-08
feat: Intersport NL

Romeo-PHILLIPS:


locations/spiders/intersport_nl.py
```diff
@@ -0,0 +1,51 @@
...
+                    "street": store.get("address1"),
+                    "phone": store.get("phone"),
+                    "email": store.get("email"),
+                    "postcode": store.get("postalCode"),
+                    "city": store.get("city"),
+                    "state": store.get("state"),
+                    "country": store.get("countryCode"),
+                    "website": store.get("link"),
+                    "lat": store.get("latitude"),
+                    "lon": store.get("longitude"),
```
davidhicks, line 48: DictParser.parse(store) would simplify field extraction here?

Romeo-PHILLIPS: Yes, its just a personal preference of mine to be able to quickly see what data a spider can pull without having to run it. I went ahead and used a DictParser

PR #4953, opened 2023-03-09
 Add KFC (JP) (1165 stores)

davidhicks:


locations/dict_parser.py
```diff
@@ -9,30 +9,39 @@ class DictParser:
...
+        # EN
         "address-region",
         "region",
         "state",
         "state-province",
         "province",
         "state-code",
         "county",
+        # JP
+        "都道府県",
```
iandees, line 44: Can you add same-line comments with English translations for these?

iandees:
Thansk!

PR #4961, opened 2023-03-12
Add Repco (AU/NZ) (385 stores)

davidhicks:


iandees:
It looks like this is blocking cloud service provider IP's. Can you add a `requires_proxy = "AU"`  like [this one](https://github.com/alltheplaces/alltheplaces/blob/caa7fb33bea542d24c0a7d958fa23f332de5983e/locations/spiders/kfc_ca.py#L15)?

davidhicks:
Done. I wasn't sure if requires_proxy is allowed to be a list of country codes, or a comma separated list, or just a single country code as a string. Ideally I think it could be a string (for a single country) or a list of country codes if multiple may be permitted?

iandees:
Yea, we haven't implemented the code that would actually use it, but I think a string for one and an array of strings for multiple options seems sensible.

PR #4987, opened 2023-03-15
Add Choices Flooring (NZ) (7 stores)

davidhicks:


Cj-Malone:
7 is quite low... I don't know if we have a hard rule, but unless there is an alternate motive, I think this is to small to bother with.

davidhicks:
> 

I wouldn't usually care for a store with 7 locations other than this choices_floor_nz spider complements choices_flooring_au so that the total number of stores for the whole brand are included.

iandees:
I'm ok with spiders that produce small numbers of places. My only concern would be increased maintenance burden, but I'm not too worried about that right now. If it breaks, someone can come along and fix it.

PR #4992, opened 2023-03-15
Add Reece brands (AU/NZ) (561 stores)

davidhicks:


iandees:
Can you apply a category? Maybe `doityourself`?

davidhicks:
I have to catch up on recent additions needing to be pushed to NSI.

Is it OK for me to add to https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/shop/trade.json rather than assign a category within ATP?

iandees:
Yep, that's a good way of doing it (so OSM folks would get the brand too). @Cj-Malone has made changes like this before, too, so it might be worth looking at some of his pull requests there.

Cj-Malone:
I'd always prefer us to put categories in NSI, obviously OSM also benefits immediately, but we also benefit in the long term if it's decided to change the category. Just like DIY vs Trade above.

I'm a maintainer of NSI now so I'll review and merge it when it's ready. NSI has lots of docs, but essentially you're just adding a name, locations and the tags [eg](https://github.com/osmlab/name-suggestion-index/commit/08b70de1bc29314046f5bf824241d43e572ecd07).

davidhicks:
I've added some changes to manually add categories for different Reece branded stores.

PR #5007, opened 2023-03-21
Add Pepperoni's (US) (20 stores)

davidhicks:


davidhicks:
NSI merge request to allow ATP to obtain categories for this spider: https://github.com/osmlab/name-suggestion-index/pull/7896

PR #5009, opened 2023-03-21
Enhance OpeningHours add_ranges_from_string and use it with Store Rocket store finder

davidhicks:
Store Rocket store finder can now automatically extract opening hours using an enhanced add_ranges_from_string function in OpeningHours that now supports named days (e.g. midday) and multiple time ranges on a single day (e.g. Mon-Wed 08:00-13:00,14:00-18:00).

### DIFF
--- a/locations/hours.py
+++ b/locations/hours.py
@@ -512,7 +512,12 @@ class OpeningHours:
                                 self.add_range(d, start_time, end_time, time_format)
 
     def add_ranges_from_string(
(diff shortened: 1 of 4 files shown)


iandees:
I agree with flake8 here: that function is a little too complex. Can you break it up into functions in any way so it's easier to follow along with what's going on?

iandees:
Do you think this is still useful, @davidhicks? Can you rebase onto something newer or close it please?

davidhicks:
Yep this is still useful. It's just a bit of work to get that mega function split up into useful sub-functions. I think we'd just end up with a bunch of functions like `create_regexes_for_extracting_day_ranges()`, `create_regexes_for_extracting_time_ranges()`, etc. I'll have another look to see how the readability of the function can be improved.

PR #5022, opened 2023-03-21
Add Soul Pattinson Chemist (AU) (44 stores)

davidhicks:


locations/spiders/soul_pattinson_chemist_au.py
```diff
@@ -0,0 +1,21 @@
...
+
+class SoulPattinsonChemistAUSpider(Spider):
+    name = "soul_pattinson_chemist_au"
+    item_attributes = {"brand": "Soul Pattinson Chemist", "brand_wikidata": "Q117225301"}
+    start_urls = ["https://soulpattinson.com.au/wp-admin/admin-ajax.php?action=store_search&autoload=1"]
+
+    def parse(self, response):
+        for location in response.json():
+            item = DictParser.parse(location)
+            item["name"] = location["store"].replace("&#038;", "&")
```
Cj-Malone, line 15: Is `html.unescape` a better solution?

```suggestion
            item["name"] = html.unescape(location["store"])
```

davidhicks: Thanks. It was just the ampersand symbol that is escaped but I've switched over to use the more generic html.unescape method.

PR #5032, opened 2023-03-22
Deichmann: fix spider

davidhicks:


locations/spiders/deichmann.py
```diff
@@ -13,29 +14,23 @@ class DeichmannSpider(SitemapSpider, StructuredDataSpider):
...
 
     def post_process_item(self, item, response, ld_data, **kwargs):
         if "dosenbach.ch" in response.url:
             item.update(self.DOSENBACH)
 
+        item["ref"] = item["ref"].split("#")[1]
+
+        item["opening_hours"] = OpeningHours()
+        item["opening_hours"].from_linked_data(ld_data)
+
```
Cj-Malone, line 30: Since you've fixed `from_linked_data` is this still needed?

```suggestion
```

davidhicks: Thanks I fixed it. I forgot to remove that after fixing from_linked_data to work with international words for "CLOSED"

PR #5039, opened 2023-03-22
aramark: fix and rename to aramark_uniform_services_us

davidhicks:
The old aramark spider has been broken for some time due to the company switching over to a new GraphQL-based store locator.

This new spider uses a paginated GraphQL query approach to extract 297 locations from the US for Aramark Uniform Services and Aramark Cleanroom Services.

Aramark is present in other countries and provides other services (separately branded) both in America and elsewhere, hence this spider has been renamed to allow additional Aramark owned brands to be added. It does not appear from a quick look that GraphQL is being used on other Aramark country websites hence separate spiders would need to be created for any country websites that contain a store locator.


davidhicks:
Corresponding NSI change to add categories: https://github.com/osmlab/name-suggestion-index/pull/7903

PR #5041, opened 2023-03-23
blue_pearl_pet_hospital: fix and rename to bluepearl_pet_hospital_us

davidhicks:


locations/spiders/bluepearl_pet_hospital_us.py
```diff
@@ -0,0 +1,28 @@
...
+    allowed_domains = ["bluepearlvet.com"]
+    sitemap_urls = ["https://bluepearlvet.com/hospital-sitemap.xml"]
+    sitemap_rules = [(r"\/hospital\/[\w\-]+\/$", "parse_sd")]
+    wanted_types = ["LocalBusiness"]
+
+    def post_process_item(self, item, response, ld_data):
+        name_html = Selector(text=ld_data["name"])
+        item["name"] = " ".join((" ".join(name_html.xpath("//text()").getall())).split()).replace(
+            " - BluePearl Pet Hospital", ""
+        )
```
Cj-Malone, line 20: I've not looked at the data, but would `normalize-space` help here? Something like:
```suggestion
        item["name"] = name_html.xpath("normalize-space(/text())").get().replace(" - BluePearl Pet Hospital", "")
```

davidhicks: I don't think this is possible with XPath 1.0 that Scrapy's choice of XML library lxml supports. normalize-space only accepts a single string as an argument where we have a list of strings that need to be concatenated and then have whitespace normalised.

In XPath 2.0 the string-join function could potentially make this proposal work, similar to:
item["name"] = name_html.xpath("normalize-space(string-join(//text(), ' '))").get().replace(" - BluePearl Pet Hospital", "")

Cj-Malone: Ah I got annoyed with the XPath 1.0 thing recently, but I can't remember the details now. Thanks for checking.

Cj-Malone: I think it was we can't do a case insensitive things.

PR #5044, opened 2023-03-23
brico_io_it: fix spider

davidhicks:


locations/spiders/brico_io_it.py
```diff
@@ -1,38 +1,44 @@
...
-            item["website"] = f'https://www.bricoio.it{location["URL"]}'
+            item = DictParser.parse(location)
+            item["ref"] = location["code"]
+            item["name"] = location["alias"]
+            address_fields = {k: v for k, v in DictParser.parse(location["address"]).items() if v}
+            print(address_fields)
+            item["street_address"] = " ".join(
+                filter(None, [location["address"]["street1"], location["address"]["street2"]])
+            )
+            item.update(address_fields)
```
Cj-Malone, line 28: Why do we have to call DictParser a second time on the address?

Just looked at the data, not ran it, but I think this should work:

```suggestion
            item["street_address"] = ", ".join(
                filter(None, [location["address"]["street1"], location["address"]["street2"]])
            )
            item["country"] = location["address"]["idCountry"]
```

locations/spiders/brico_io_it.py
```diff
@@ -1,38 +1,44 @@
...
             item["opening_hours"] = OpeningHours()
-            for rule in location["Orari"]["FasceOrarie"]:
-                day = sanitise_day(rule["Giorno"][:3], DAYS_IT)
-                for time in rule["Orario"].split(" / "):
-                    start_time, end_time = time.split("-")
-                    item["opening_hours"].add_range(day, start_time, end_time, time_format="%H.%M")
+            for timeslot in location["hours"]["timeslots"]:
+                open_time = str(timeslot["from"]["hours"]) + ":" + str(timeslot["from"]["minutes"]).zfill(2)
+                close_time = str(timeslot["to"]["hours"]) + ":" + str(timeslot["to"]["minutes"]).zfill(2)
+                item["opening_hours"].add_range(day_list[timeslot["dayOfWeek"]], open_time, close_time)
```
Cj-Malone, line 42: This is an off by 1 thing? How about:

```suggestion
                item["opening_hours"].add_range(DAYS[timeslot["dayOfWeek"] - 1], open_time, close_time)
```

Or maybe `+ 1` depending on the direction.

locations/spiders/brico_io_it.py
davidhicks: Thanks, fixed. I hadn't even realised I left a print() in there as well!

locations/spiders/brico_io_it.py
davidhicks: Thanks for the simplification, fixed it too. It works with -1 only because index 0 (Monday) minus 1 is index -1, which is index 6 (Sunday). I can't see why we'd ever care about +1 and need to use something like deque to rotate all elements of the list because it'd mean Tuesday is the start of the week someplace.

PR #5057, opened 2023-03-27
Add pipeline to format state codes and/or geocode state for US and CA

Cj-Malone:


Cj-Malone:
Changed to https://github.com/thampiman/reverse-geocoder as it includes state data, also faster apparently.

Cj-Malone:
It doesn't feel faster...

PR #5058, opened 2023-03-27
Add SweetIQ store locator and sample brands which use this store locator

davidhicks:


davidhicks:
Two brands needed to be added to NSI. Pull request waiting to be pulled: https://github.com/osmlab/name-suggestion-index/pull/7926

locations/storefinders/sweetiq.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def parse(self, response):
+        for location in response.json()["features"]:
+            if location["properties"]["isPermanentlyClosed"]:
+                return
+
+            item = DictParser.parse(location["properties"])
+            item["ref"] = location["properties"]["branch"]
+            item["lat"] = location["geometry"]["coordinates"][1]
+            item["lon"] = location["geometry"]["coordinates"][0]
```
Cj-Malone, line 44: FYI, assuming this is GeoJSON, we can just pass the `geometry` through. Then in the admittedly unlikely event that give us a Polygon, we wouldn't crash.

```suggestion
            item["geometry"] = location["geometry"]
```

locations/storefinders/sweetiq.py
```diff
@@ -0,0 +1,78 @@
...
+                "CASH ONLY": PaymentMethods.CASH,
+                "CASH": PaymentMethods.CASH,
+                "Cash": PaymentMethods.CASH,
+                "DEBIT": PaymentMethods.DEBIT_CARDS,
+                "Debit": PaymentMethods.DEBIT_CARDS,
+                "DISCOVER": PaymentMethods.DISCOVER_CARD,
+                "Discover": PaymentMethods.DISCOVER_CARD,
+                "MASTERCARD": PaymentMethods.MASTER_CARD,
+                "Mastercard": PaymentMethods.MASTER_CARD,
+                "MasterCard": PaymentMethods.MASTER_CARD,
```
Cj-Malone, line 66: If we normalise the key we could shrink this down, right?
```suggestion
                "MASTERCARD": PaymentMethods.MASTER_CARD,
````

```python
                if payment_method.upper() in payment_methods.keys():
```

locations/storefinders/sweetiq.py
davidhicks: Thanks, that works. I've changed it.

locations/storefinders/sweetiq.py
davidhicks: Improved this too, thanks!

PR #5061, opened 2023-03-28
feat: bestwestern world

charlesdassonville:
how to add a delay between requests to avoid CAPTCHA?


Cj-Malone:
`download_delay` controls that. The default now is 1 second, if we're getting stopped it should be increased to at least that, or higher if needed.

PR #5075, opened 2023-03-29
feat: mephisto

mickaelabdealy:
Hello @Cj-Malone 
I'm using `AmastyStoreLocatorSpider` but it looks like it' running using my personal localization. Do you how I can fake localization or just run it worldwide ?


Cj-Malone:
Localization as in language? Maybe http headers? There is a setting like DEFAULT_REQUEST_HEADERS, and you can override it in custom_settings. (I'm on my phone so I may have misspelt)

Otherwise @davidhicks might be able to help, I think he wrote this.

davidhicks:
If you remove the path from `allowed_domains`  (this variable should never have a path included) you get:

2023-03-30 12:51:07 [scrapy.downloadermiddlewares.redirect] DEBUG: **Redirecting (302)** to <GET https://www.mephisto.com**/en/**amlocator/index/ajax/> from <GET https://www.mephisto.com/amlocator/index/ajax/>
2023-03-30 12:51:07 [scrapy.core.engine] DEBUG: Crawled (200) <GET https://www.mephisto.com**/en/**amlocator/index/ajax/> (referer: None)

If you want to avoid that redirect, or force the use of "https://www.mephisto.com/be-fr/amlocator/index/ajax/", then remove `allowed_domains = ...` and set `start_urls = ["https://www.mephisto.com/be-fr/amlocator/index/ajax/"]` instead.

I've checked en vs. be-fr pages and don't see how it'd change the output of the spider though. It seems to mostly just change the localization of labels in the popup_html message when a user clicks a marker on the map, but the storefinder doesn't care about these labels in popup_html anyway.

mickaelabdealy:
> Using `allowed_domains` like that seems wrong to me, it's unexpected and potentially harmful if a spider wanted to pull data from a different domain like a CDN. But I guess that's exactly why `start_urls` overrides the behaviour. You've already thought about this :p

I found out that there is a parameter for the page. And they are storing the page_count [here](https://www.mephisto.com/be-fr/points-de-vente/).

davidhicks:
I thought I replied again yesterday but it looks like I accidentally closed tabs without submitting the reply.

From investigations I did, the locale (e.g. en or be-fr) only changes labels, particularly in popup_html that is shown to users when they click on a map marker, but the spider doesn't use these labels anyway so you could set the locale to "asouidhoqwaiusdh" and the spider will keep working just fine.

The page number does seem to matter for **some** Amasty store locators, but seemingly none of the ones ATP has committed prior to Mephisto. For previously committed Amasty store locators, it appears all results are returned without paging being used, and thus ?p=2 or ?p=1283721837 could be used and it has no impact.

https://www.la-z-boy.com.au/storelocator is an Amasty store locator, but there is no pager UI shown to users and no element I could find on the store locator page for determining how many pages of results exist in total. Thus it appears that a spider for this store would need to start at ?p=1 and increment until the first duplicate result is found. With all Amasty store locators, silly page numbers like ?p=1293879128367 still return results, but the behaviour seems variable. For Mephisto, ?p=1000 through ?p=1003 might return stores 1 through 10, then ?p=1004 and ?p=1005 returns stores 11 through 20, then ?p=1006 again returns stores 1 through 10. For La-Z-Boy, every page after the last just keeps returning the same last page of results.

Cj-Malone:
@AmastyLtd-zz any guidance on reliably paging through the store locator Ajax API?

@davidhicks the second docs like seem to imply there is a stable set of fields in the popup HTML. Did you consider parsing them in the shared code?
photo
country
state
city
zip
address
phone
email
website
description
lat
lng

PR #5077, opened 2023-03-29
Fix Noodles and Company opening hours

Cj-Malone:


Cj-Malone:
hmm, is indeed weird.

PR #5086, opened 2023-04-01
Feat/nh hotels

mickaelabdealy:
Hello @Cj-Malone 
I don't know why but the when I run the spider it does not seem to reach the url `"https://www.nh-hotels.com/rest/datalayer/hotelPage/" + store_id`. When I used `scrapy shell` with the same url, It worked and got data. Do you have any idea what is happening ?

One example [url](https://www.nh-hotels.com/rest/datalayer/hotelPage/13024).

Thanks in advance.


Cj-Malone:
Works for me, and [GitHub CI](https://data.alltheplaces.xyz/map.html?show=https://data.alltheplaces.xyz/ci/4584294452/nh_hotels/output.geojson). Maybe they had blocked you?

locations/spiders/nh_hotels.py
```diff
@@ -0,0 +1,37 @@
...
+        yield Request(url=url, callback=self.parse_site)
+
+    def parse_site(self, response):
+        data = response.json()
+        item = Feature()
+        item["ref"] = data["id"]
+        item["name"] = data["name"]
+        item["city"] = data["city"]
+        item["country"] = data["country"]
+        item["street_address"] = data["address"]["street"]
```
Cj-Malone, line 28: Looks like this sometimes has a trailing "," `"addr:street_address": "Onderlangs, 10,"`. `·strip(",")` might be enough, or `clean_address(data["address"]["street"])`

locations/spiders/nh_hotels.py
```diff
@@ -0,0 +1,37 @@
...
+        item["postcode"] = data["address"]["postalCode"]
+        item["addr_full"] = f"""{item["street_address"]}, {item["postcode"]} {item["city"]}, {item["country"]}"""
```
Cj-Malone, line 30: Same issue with the street address, we get ",," sometimes. `"addr:full": "Onderlangs, 10,, 6812 CG Arnhem, NL",`.

`clean_address([item["street_address"], item["postcode"], item["city"], item["country"]])` should help.

However I'm not sure I agree with computing `addr_full` in spider. If it's helpful to data consumers, we should just add a pipeline to do it to all of the spiders automatically.

locations/spiders/nh_hotels.py
```diff
@@ -0,0 +1,37 @@
...
+    allowed_domains = ["nh-hotels.com"]
+    sitemap_urls = ["https://www.nh-hotels.com/hotels-sitemap.xml"]
+    user_agent = BROWSER_DEFAULT
+
+    def parse(self, response):
+        store_id = response.xpath(
+            '//section/div/button[@class="btn btn-primary js-track-book-hp js-track-book-hpGA4"]/@data-id'
+        ).get()
+        url = "https://www.nh-hotels.com/rest/datalayer/hotelPage/" + store_id
+        yield Request(url=url, callback=self.parse_site, cb_kwargs={"website": response.url})
```
Cj-Malone, line 20: `//@data-id` would probably work here, and is less likely to break when they change the website style. Actually the following is probably the safest:
```suggestion
        if store_id := response.xpath("//@data-id").get():
            url = "https://www.nh-hotels.com/rest/datalayer/hotelPage/" + store_id
            yield Request(url=url, callback=self.parse_site, cb_kwargs={"website": response.url})
```

locations/spiders/nh_hotels.py
```diff
@@ -0,0 +1,37 @@
...
+        item["city"] = data["city"]
+        item["country"] = data["country"]
+        item["street_address"] = clean_address(data["address"]["street"])
+        item["postcode"] = data["address"]["postalCode"]
+        item["lat"] = data["address"]["latitud"]
+        item["lon"] = data["address"]["longitud"]
+        item["email"] = data["contact"]["mail"]
+        item["phone"] = data["contact"]["phone"]
+        item["website"] = website
+
```
Cj-Malone, line 36: So we aren't getting a category on this. The quick answer would be:

`apply_category(Categories.HOTEL, item)`

But it seems there are several NH brands: NH Collection, NH Hotels, Nhow and Tivoli at least. It'd be nice to correctly brand the output if possible.

PR #5088, opened 2023-04-02
Feat/cora be, lu and fr

mickaelabdealy:


locations/spiders/cora_fr.py
```diff
@@ -0,0 +1,16 @@
...
+    item_attributes = {"brand": "Cora", "brand_wikidata": "Q686643"}
+    key = "woos-789d2f65-5cee-39a3-9e38-a31bd63a6b57"
+    custom_settings = {"DEFAULT_REQUEST_HEADERS": {"Origin": "https://www.cora.fr/"}}
+
+    def parse_item(self, item, feature, **kwargs):
+        # Filter only on CORA shop and not affiliated shop
+        if item["name"].startswith("CORA"):
+            item["website"] = "https://www.cora.fr" + item["website"]
+            item["lat"] = item["geometry"]["coordinates"][0]
+            item["lon"] = item["geometry"]["coordinates"][1]
```
iandees, line 15: There's no need to convert from GeoJSON geometry to lat/lon.

locations/spiders/cora_fr.py
```diff
@@ -0,0 +1,16 @@
...
+
+class CoraFRSpider(WoosmapSpider):
+    name = "cora_fr"
+    item_attributes = {"brand": "Cora", "brand_wikidata": "Q686643"}
+    key = "woos-789d2f65-5cee-39a3-9e38-a31bd63a6b57"
+    custom_settings = {"DEFAULT_REQUEST_HEADERS": {"Origin": "https://www.cora.fr/"}}
+
+    def parse_item(self, item, feature, **kwargs):
+        # Filter only on CORA shop and not affiliated shop
+        if item["name"].startswith("CORA"):
```
iandees, line 12: Instead of wrapping this whole thing in the `if` block, return early if the item doesn't match the desired name:

```suggestion
        if not item["name"].startswith("CORA"):
            return
```

iandees (changes requested):

iandees (approved):

PR #5093, opened 2023-04-03
jax_tyres_and_auto_au: fix spider

davidhicks:
With upcoming public holidays in Australia this week, this spider is broken due to unhandled day names such as "Good Friday". Unfortunately for such weeks with a public holiday, there is no place on the website showing what the usual hours would have been. Thus the hours for this spider will constantly fluctuate depending on whether the given week has a public holiday.

### DIFF
--- a/locations/spiders/jax_tyres_and_auto_au.py
+++ b/locations/spiders/jax_tyres_and_auto_au.py
@@ -37,7 +37,7 @@ class JaxTyresAndAutoAUSpider(Spider):
         day_names = response.xpath('//div[contains(@class, "pgStLcSn-openingHours")]/div[1]/p/text()').getall()
         hour_ranges = response.xpath('//div[contains(@class, "pgStLcSn-openingHours")]/div[2]/p/text()').getall()
         for i in range(1, 7, 1):
-            if hour_ranges[i].upper() == "CLOSED":
+            if not any(char.isdigit() for char in hour_ranges[i]):
                 continue
             open_time = hour_ranges[i].split(" - ")[0]
             close_time = hour_ranges[i].split(" - ")[1]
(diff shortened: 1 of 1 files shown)


locations/spiders/jax_tyres_and_auto_au.py
```diff
@@ -37,7 +37,7 @@ def parse(self, response):
         day_names = response.xpath('//div[contains(@class, "pgStLcSn-openingHours")]/div[1]/p/text()').getall()
         hour_ranges = response.xpath('//div[contains(@class, "pgStLcSn-openingHours")]/div[2]/p/text()').getall()
         for i in range(1, 7, 1):
-            if hour_ranges[i].upper() == "CLOSED":
+            if not any(char.isdigit() for char in hour_ranges[i]):
                 continue
             open_time = hour_ranges[i].split(" - ")[0]
             close_time = hour_ranges[i].split(" - ")[1]
```
Cj-Malone, line 43: This would also crash if another holiday used a number in there? eg "Good Friday 7th" Or if the times weren't padded by a space, just "07:45 AM-05:00 PM".

It'd be safer to use regex here:
```suggestion
            if m := re.match(r"(\d\d:\d\d [AP]M)\s*-\s*(\d\d:\d\d [AP]M)", hour_ranges[i]):
                properties["opening_hours"].add_range(day_names[i], m.group(1), m.group(2), "%I:%M %p")
```

locations/spiders/jax_tyres_and_auto_au.py
```diff
@@ -37,7 +37,7 @@ def parse(self, response):
         day_names = response.xpath('//div[contains(@class, "pgStLcSn-openingHours")]/div[1]/p/text()').getall()
         hour_ranges = response.xpath('//div[contains(@class, "pgStLcSn-openingHours")]/div[2]/p/text()').getall()
         for i in range(1, 7, 1):
```
Cj-Malone, line 39: FYI `zip` would make this more readable, at least to me, but I don't know if it has any other pros/cons.

```python
        for day, hours in zip(day_names, hour_ranges):
```

locations/spiders/jax_tyres_and_auto_au.py
davidhicks: Thanks for the suggestion, that's a much better approach. I've fixed it :)

Cj-Malone (approved):

PR #5099, opened 2023-04-03
ra_sushi: fix and rename to ra_sushi_us

davidhicks:


davidhicks:
Pull request to add this brand to NSI: https://github.com/osmlab/name-suggestion-index/pull/7951

PR #5113, opened 2023-04-04
ross_dress: fix and split into ross_dress_us and ross_dress_gu

### LINKED ISSUE #4750: ross_dress: use new Where2GetIt spider
RossDressSpider (ross_dress.py) already uses a Where2GetIt start URL, but the spider can now be simplified with the introduction of the Where2GetIt spider in https://github.com/alltheplaces/alltheplaces/pull/4741

The application key for W2GI appears to be 097D3C64-7006-11E8-9405-6974C403F339 as obtained from https://hosted.where2getit.com/rossdressforless/locator.html

@Cj-Malone

davidhicks:
Ross Dress is now using Where2GetIt. As there are locations in mainland USA (1701) and Guam (3), two spiders are needed as the Where2GetIt spider only finds locations within a radius of 5000 (probably km) and Guam is much further from mainland USA than 5000km.

Fixes #4750


davidhicks:
@iandees I have done some digging around with the Where2GetIt store finder and plan to rewrite it per https://github.com/alltheplaces/alltheplaces/issues/5119

The current method used by the Where2GetIt store finder is inefficient and error-prone, and appears limited to a radius of 5000km (going beyond this causes no response to be received from the API). The proposed rewritten spider would return all stores for both USA and Guam at once in a single response:

```
curl -s 'https://hosted.where2getit.com/rossdressforless/rest/getlist' -X POST -H 'Accept: application/json' -H 'Content-Type: application/json' --data-raw '{"request":{"appkey":"097D3C64-7006-11E8-9405-6974C403F339","formdata":{"objectname":"Locator::Store"}}}' | jq '.response.collection[].country' | sort | uniq -c
```

PR #5116, opened 2023-04-04
Add Black Sheep Coffee Spider

Cj-Malone:
https://github.com/osmlab/name-suggestion-index/pull/7964


Cj-Malone:
Note to self, do a git pull before coding, not before commit. :p

PR #5120, opened 2023-04-04
feat: add new spider apcoa_be

elsa-alves-monteiro:


locations/spiders/apcoa_be.py
```diff
@@ -0,0 +1,24 @@
...
+    allowed_domains = ["www.apcoa.be"]
+    start_urls = ["https://www.apcoa.be"]
+    rules = [
+        Rule(LinkExtractor(allow=r"https://www\.apcoa\.be/parkings-per-stad/[-\w]+/$")),
+        Rule(
+            LinkExtractor(allow=r"https://www\.apcoa\.be/parkings-per-stad/[-\w]+/[-\w]+/$"),
+            callback="parse_sd",
+        ),
+    ]
+    download_delay = 0.5
```
iandees, line 20: Our default download delay is ~1 sec. Do we need to go faster than that?

PR #5126, opened 2023-04-06
Feat/maxmara

elsa-alves-monteiro:


locations/spiders/maxmara.py
```diff
@@ -0,0 +1,45 @@
...
+                item["ref"] = store_info["name"]
+                item["name"] = store_info["displayName"]
+                item["phone"] = store_info["phone1"]
+                item["addr_full"] = store_info["formattedAddress"]
+
+                oh = OpeningHours()
+                for day, hours in store_info.get("openingHours").items():
+                    for chunk in hours:
+                        open_at = chunk.split(" - ")[0].replace(".", ":")
+                        close_at = chunk.split(" - ")[1].replace(".", ":")
```
Cj-Malone, line 41: This calls `split` and `replace` 2 times per day, so 14 times each per POI. Multiply that by the POI count and we can get a big number.

If we flip it around we only have to call each once, so half the calls:
```suggestion
                        open_at, close_at = chunk.replace(".", ":").split(" - ")
```

PR #5128, opened 2023-04-06
feat: add essentiel-antwerp

mickaelabdealy:


locations/spiders/essentiel_antwerp.py
```diff
@@ -0,0 +1,29 @@
...
+from locations.items import Feature
+
+
+class EssentielAntwerpSpider(scrapy.Spider):
+    name = "essentiel_antwerp"
+    item_attributes = {"brand": "Essentiel-Antwerp"}
+    start_urls = ["https://www.essentiel-antwerp.com/be_fr/storelocator"]
+
+    def parse(self, response):
+        data = response.xpath(".").get()
```
Cj-Malone, line 15: There is `response.text` for when it's a text response, or `response.body` for none text.
```suggestion
        data = response.text
```

locations/spiders/essentiel_antwerp.py
```diff
@@ -0,0 +1,29 @@
...
+        stores = json.loads("[" + re.findall(r"(stores:\[)(.*?)(])", data)[0][1] + "]")
```
Cj-Malone, line 16: Meddling with the regex a bit:

- `findall` gets all the instances, but then you are only using `[0]`, `search` should just give us the first.
- You are matching 3 groups, `stores:[`, the stores, and `]`, then adding `[` and `]` back. We can just include them in the match.

```suggestion
        stores = json.loads(re.search(r"stores:(\[.*\])", data).group(1))
```

locations/spiders/essentiel_antwerp.py
```diff
@@ -0,0 +1,29 @@
...
+            item["ref"] = store["storelocator_id"]
+            item["phone"] = store["phone"]
+            item["country"] = store["country"]
+            item["state"] = store["state"]
+            item["postcode"] = store["zipcode"]
+            item["lat"] = float(store["latitude"])
+            item["lon"] = float(store["longtitude"])
+            item["street_address"] = store["address"]
+            item["city"] = store["city"]
+            item["name"] = "Essentiel-Antwerp" + " ".join(store["rewrite_request_path"].split("-"))
```
Cj-Malone, line 28: I don't like the idea of producing our own store names. But I'll let @iandees decide on that.

**If** we do it, it shouldn't be all lowercase:
```suggestion
            item["name"] = "Essentiel-Antwerp" + store["rewrite_request_path"].replace("-", " ").title()
```

locations/spiders/essentiel_antwerp.py
```diff
@@ -0,0 +1,29 @@
...
+            yield item
```
Cj-Malone, line 29: Lets collect the website:
```suggestion
            item["website"] = f'https://www.essentiel-antwerp.com/eu/{store["rewrite_request_path"]}/'
            yield item
```

locations/spiders/essentiel_antwerp.py
```diff
@@ -0,0 +1,29 @@
...
+import re
+
+import scrapy
+
+from locations.items import Feature
+
+
+class EssentielAntwerpSpider(scrapy.Spider):
+    name = "essentiel_antwerp"
+    item_attributes = {"brand": "Essentiel-Antwerp"}
```
Cj-Malone, line 11:
```suggestion
    item_attributes = {"brand": "Essentiel-Antwerp", "brand_wikidata": "Q117456339"}
```

PR #5137, opened 2023-04-11
feat: add first hotels

mickaelabdealy:


iandees:
It looks like the individual hotel pages could be parsed using the StructuredDataSpider. Take a look at other spiders like [this one](https://github.com/alltheplaces/alltheplaces/blob/77a065830706faf45669956ee2b69e94b9a33ca5/locations/spiders/mitre_10_au.py) to see how you can pull URLs and pass off to `parse_sd()` to get the structured data.

PR #5165, opened 2023-04-16
Add StockInStore store finder and many new brands that use this store finder

davidhicks:


Cj-Malone:
@davidhicks where did you find the list of sites using this store finder/do we know if they are still active and up to date?

[Intersport AU website](https://intersport.com.au/store-locator/) says 78 POIs. This gives 68.

davidhicks:
@Cj-Malone 

https://www.stockinstore.com/clients/ was used to find clients. Then the client website was checked for either (a) store locator page or (b) "find stores with this item in stock" widgets. Once one of these were found (in that order of priority), requests to stockinstore's API were observed.

Perhaps for this store, their database of stores for the stock locator (with stockinstore) doesn't align with the database of stores in their store locator page/widget.

Their store locator page/widget is: https://intersport.com.au/wp-admin/admin-ajax.php?action=store_search&lat=-25.2744&lng=133.77514&max_results=25&search_radius=100&autoload=1 which is provided by https://github.com/wp-plugins/wp-store-locator/blob/master/frontend/wpsl-ajax-functions.php#L15 instead of stockinstore's API.

PR #5168, opened 2023-04-16
Add City Beach (AU) (67 stores)

davidhicks:


locations/spiders/city_beach_au.py
```diff
@@ -0,0 +1,27 @@
...
+
+class CityBeachAUSpider(Spider):
+    name = "city_beach_au"
+    item_attributes = {"brand": "City Beach", "brand_wikidata": "Q16958619"}
+    allowed_domains = ["www.citybeach.com.au"]
+    start_urls = [
+        "https://www.citybeach.com/on/demandware.store/Sites-CityBeachAustralia-Site/default/Stores-FindStores?showMap=true&radius=10000&postalCode=0870"
+    ]
+
+    def start_requests(self):
```
iandees, line 16: No need for `start_requests` since you're not doing anything special with `start_urls`

davidhicks: Yep the _Accept_ header doesn't matter so the following examples return the same JSON blob:
curl 'https://www.citybeach.com/on/demandware.store/Sites-CityBeachAustralia-Site/default/Stores-FindStores?showMap=true&radius=15.0&postalCode=2000&radius=15 -H 'Accept: application/json, text/javascript, */*; q=0.01'
curl 'https://www.citybeach.com/on/demandware.store/Sites-CityBeachAustralia-Site/default/Stores-FindStores?showMap=true&radius=15.0&postalCode=2000&radius=15 -H 'text/html, application/xhtml+xml, application/xml;q=0.9, image/webp, */*;q=0.8'

I recall reading in another comment thread I can't find a preference from @Cj-Malone (I think) towards using _JsonRequest_ when the expected result is a JSON blob. Was there a reason for this, or plain old default _Request_ can be used when the server doesn't care about the _Accept_ header?

iandees: Gotcha. I don't think it's necessary, but I see what you're doing. I'll merge and we can adjust later if it's a problem.

PR #5181, opened 2023-04-17
tapetro: fix and rename to travelcenters_of_america_us

davidhicks:
Note the API request is very slow to respond, but I didn't observea timeout.

If anyone can be bothered, it would be possible to supplement API data with additional fields of data (e.g. opening hours, services available, more granular fuel type data) that could be scraped off the item["website"] of each location.


davidhicks:
> Can you please run `pipenv uninstall xlrd` to update the `Pipfile.lock`?

Done, thanks :)

iandees:
Thanks!

PR #5186, opened 2023-04-17
supercheap_auto: fix opening hours for NZ locations

davidhicks:
This spider was returning errors for each NZ location due to a few problems in handling of opening hours.

OpeningHours.add_ranges_from_string() is now used instead to simplify this spider.


davidhicks:
@iandees Should be ready to merge.

PR #5187, opened 2023-04-17
Add categories to systeme_u spider

Cj-Malone:
This seems to work occasionally on the weekly, but I didn't manage to get it to work with Mozilla VPN in FR. Maybe a residential proxy would?

In theory this should be a SitemapSpider+StrucuturedDataSpider, but I didn't want to change too much while I can't test it.

Anyway I've added brands and categories, hopefully it works at some point. Looking at the sitemap there should be 1088 shops and 686 stations.


Cj-Malone:
Lets see if the FR community tell me off for [this](https://www.openstreetmap.org/changeset/135026700). I do it in GB.

PR #5194, opened 2023-04-21
Switch to reverse_geocoder single threaded mode

Cj-Malone:
By default, reverse_geocoder uses multiple threads since we only ever ask for 1 POI at a time, it wastes a lot of CPU time setting them up.

```python
from scrapy.spiders import Spider

class Test2Spider(Spider):
    name = "test2"
    start_urls = ["data:,"]

        for i in range(10000):
            yield Feature(ref=str(i), lat=45, lon=-90)
```
mode=2 (default) took 721 seconds
mode=1 took 2.734139 seconds


iandees:
Wow, good catch.

Cj-Malone:
Well I felt it [weeks ago](https://github.com/alltheplaces/alltheplaces/pull/5057#issuecomment-1486874410), but procrastinated so I don't deserve any credit :p

PR #5241, opened 2023-05-01
fix insights to used paged OSM taginfo API

mjoe999:
"insights" at https://www.alltheplaces.xyz/wikidata.html is broken due to the taginfo API requiring paged access. This PR fixes that and adds a few more improvements:

- additionally support insights generation from ATP ZIP files (less clumsy in many cases)
- allow filters to be set so that certain spiders can be ignored when building the output
- use ijson dependency for iterating large JSON files
- add more fields to the output --atp-nsi-osm JSON (countries per QID, spiders contributing to each country)

The JSON output while extended is still compatible with the wikidata.html page at https://www.alltheplaces.xyz/. It is also compatible with https://github.com/alltheplaces/alltheplaces.xyz/pull/29.


iandees:
Sorry to have forgotten about this, @mjoe999. In the future, it's helpful if you can split the changes up into smaller PRs. It's easier and quicker to review.

mjoe999:
No problem @iandees . I will thin this down (since you have done some of the work) and go again :-) It's all about getting to a position where wikidata.html will light up with even more useful information (hopefully).

iandees:
This one is all good. I'll merge it now. I only mentioned it because I think that's why I forgot to finish the review and merge.

Issue #4878, opened 2023-02-24
General clean up


Cj-Malone:
These are kind of personal TODOs, but not necessary immediately and anyone is welcome to get involved. Also feel free to add to the list.

# Convert spiders to StructuredDataSpider

Existing LinkedData and Microdata spiders should get converted to StrucutredDataSpider.

Some good places to get started:
`git grep "application/ld+json"`
`git grep "itemprop"`

# Convert spiders to SitemapSpider

A bunch of older spiders parse sitemaps manually, we should move to the standard code where possible.

`git grep "/loc/text()"`

# Convert spiders to CrawlSpider

I'm not sure we could do an easy grep for this one, but CrawlSpider and LinkExtracter will probably be more robust and break less.

# `addr_full` vs `street_address`

A bunch of old spiders still use `addr_full` when they should be using `street_address`.

`git grep "addr_full"`

Cj-Malone:
# Opening hours validation

Not all spiders output OSM valid opening hours.

Fetch/check stats scripts looking at `atp/field/opening_hours/invalid`

Cj-Malone:
# Download delay

`download_delay` should only be defined when breaking with the default. Most cases of `download_delay = 1` are probably wrong and make it harder to change the default later on.

`git grep -E "download_delay = 1(\.0)?$"`

# User agents

Any none default user agents (Browser user agents) should probably not be defined in spider, but using `BROWSER_DEFAULT`

`git grep "Mozilla/5"`

Cj-Malone:
# No category

All spiders should have categories.

`$ ./check_stats.sh 2023-02-18-13-32-45 atp/category/missing`

Cj-Malone:
# HTML encoded fields

Spiders should output valid text, not `&amp;`

`$ scrapy insights --html-encoding file/files`

Cj-Malone:
> # Opening hours validation
> 
> Not all spiders output OSM valid opening hours.
> 
> Fetch/check stats scripts looking at `atp/field/opening_hours/invalid`

Should be done now :) :) :)

I'll check after the next weekly, but I think as long as we enforce it on new spiders we should be good.

Cj-Malone:
Some dirty code to do spider name -> PascalCase + upper country + "Spider" gives us an awful lot, and a bunch of false positives.

| spider_name | calculated_name | actual_name |
| ----------- | --------------- | ----------- |
[...]

PR #5244, opened 2023-05-01
download_delay of 1s is the default, remove from spiders where that is the value set

mjoe999:
A few general clean ups around download_delay. Should not be too contentious. We should think about some of the other non-default values  perhaps.


Cj-Malone:
Thanks, this was on my TODO list. https://github.com/alltheplaces/alltheplaces/issues/4878#issuecomment-1444806199

PR #5251, opened 2023-05-03
Update Waterstones spider to use alphabetical directory

rjw62:
The existing implementation crawls through a paginated list starting from https://www.waterstones.com/bookshops/viewall . However, a coding error by Waterstones means only 15 pages are linked, ending at https://www.waterstones.com/bookshops/viewall?page=15 . But in fact there's a hidden page 16: https://www.waterstones.com/bookshops/viewall?page=16 that contains the last few stores at the end of the alphabet.

This means the existing spider misses around 20 stores. This PR should fix this, by instead polling the 26 alphabetical pages.


Cj-Malone (approved):

Cj-Malone:
LGTM, thanks.

PR #5263, opened 2023-05-12
Add more EV brands

Cj-Malone:
@rjw62 hopefully you can use these too.


Cj-Malone:
Rel: https://github.com/osmlab/name-suggestion-index/pull/8126

PR #5266, opened 2023-05-13
Fix big_boy and rename to big_boy_us (59 stores)

davidhicks:


locations/google_url.py
```diff
@@ -65,8 +65,10 @@ def get_query_param(link, query_param):
             if m := re.match(r"^(-?[.\d]+),(-?[.\d]+)$", ll):
                 return float(m.group(1)), float(m.group(2))
     elif url.startswith("https://www.google.com/maps/place/"):
-        lat, lon = url.split("/")[5].split(",")
-        return float(lat.strip()), float(lon.strip())
+        slash_splits = url.split("/")
+        if len(slash_splits) > 5:
```
iandees, line 69: Can you add a test case that checks this new branch?

PR #5283, opened 2023-05-16
kohls: improve and rename to kohls_us

davidhicks:
* Extract opening hours
* Fix SitemapSpider rules to avoid unnecessary pages
* Extract store reference ID from store URL pages
* Drop common fields across all stores: Facebook, Twitter and Image
* Remove unnecessary text from store names


locations/spiders/kohls_us.py
```diff
@@ -0,0 +1,26 @@
...
+class KohlsUSSpider(SitemapSpider, StructuredDataSpider):
+    name = "kohls_us"
+    item_attributes = {"brand": "Kohl's", "brand_wikidata": "Q967265"}
+    allowed_domains = ["www.kohls.com"]
+    sitemap_urls = ["https://www.kohls.com/sitemap_store.xml"]
+    sitemap_rules = [(r"\/stores\/\w{2}\/\w+-(\d+)\.shtml$", "parse_sd")]
+    wanted_types = ["DepartmentStore"]
+
+    def post_process_item(self, item, response, ld_data):
+        item["ref"] = re.search(self.sitemap_rules[0][0], response.url).group(1)
```
Cj-Malone, line 18: FYI, this should automatically happen. I recently fixed a bug where it was only working for full url marches before, we can also override `get_ref` if needed.

PR #5292, opened 2023-05-17
Add Camping World spider

reidab:
This adds a spider for Camping World locations via https://rv.campingworld.com/locations

### DIFF
--- a/locations/spiders/campingworld.py
+++ b/locations/spiders/campingworld.py
@@ -1,0 +1,67 @@
+import json
+
+import scrapy
+from scrapy.http import JsonRequest, Request
+
+import base64
+
+from locations.items import Feature
+from locations.user_agents import BROWSER_DEFAULT
+
+
+class CampingWorldSpider(scrapy.Spider):
+    name = "campingworld"
+    item_attributes = {"brand": "Camping World", "brand_wikidata": "Q5028383"}
+
+    locator_url = "https://rv.campingworld.com/locations"
+    api_url = "https://api.rvs.com/api/geodata/getclosestdealer"
+    user_agent = BROWSER_DEFAULT
+
+    request_json = """
+        {
+            "zipCode":"66952",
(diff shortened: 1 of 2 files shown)


locations/spiders/campingworld.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def start_requests(self):
+        yield Request(url=self.locator_url, callback=self.parse_locator)
+
+    def parse_locator(self, response):
+        # The locator HTML contains a JWT in a meta tag that we need to make the API call.
+        meta = response.xpath('//meta/@content')
+        meta_content = [s.get() for s in meta]
+        tokens = [s for s in meta_content if s.startswith("eyJ") and ".eyJ" in s]
+        token = tokens[0]
```
Cj-Malone, line 41: Would it be better to do this in the xpath?
```suggestion
        token = response.xpath('//meta[starts-with(@content, "eyJ")][contains(@content, ".eyJ")]/@content').get()
```

locations/spiders/campingworld.py
```diff
@@ -0,0 +1,67 @@
...
+from scrapy.http import JsonRequest, Request
+
+import base64
+
+from locations.items import Feature
+from locations.user_agents import BROWSER_DEFAULT
+
+
+class CampingWorldSpider(scrapy.Spider):
+    name = "campingworld"
```
Cj-Malone, line 13: If this is US only, we'd prefer to flag that:
```suggestion
class CampingWorldUSSpider(scrapy.Spider):
    name = "camping_world_us"
```

And a filename of `camping_world_us.py`

locations/spiders/campingworld.py
reidab: Works for me.

locations/spiders/campingworld.py
reidab: Updated!

PR #5296, opened 2023-05-17
house_au: fix spider

davidhicks:


iandees:
There are a few points really far outside Australia. Do those actually exist or is it just bad data on their side?

davidhicks:
Good catch. The data sucks. There is a store in a suburb of Melbourne, Victoria, Australia that shows up in France. This incorrect location is also shown in their official store page at https://www.house.com.au/stores/house---malvern

davidhicks:
Fixed it with a reverse geocoder check to skip those stores with wildly incorrect coordinates :)

Cj-Malone:
Do we want to throw away the whole POI, maybe just remove the coords on bad ones?

davidhicks:
Oh I didn't realise we even kept locations without coordinates. I've done as suggested and changed it to keep the location without providing any geometry for the location.

Issue #5309, opened 2023-05-19
petsathome_gb spider unstable


rjw62:
The most recent run of the petsathome_gb spider from 2023-05-15 has returned 50 fewer stores than the previous run from 2023-04-15. I've checked a few of the missing stores, and they all appear to still exist. For example:
* https://community.petsathome.com/store/yeovil/
* https://community.petsathome.com/store/east-grinstead/
* https://community.petsathome.com/store/barrow/
* https://community.petsathome.com/store/ipswich/

I've tried running the spider locally, with similar (though not identical) results to the 2023-05-15 run. The only clue I've spotted is that the log output shows a number of HTTP 404 responses. I don't know if these are on store URLs or from the search points. Is anyone able to shed any light on this?

Cj-Malone:
This doesn't fix it either. [South East](https://www.petsathome.com/shop/en/pets/store-locations-results/southeast) links to `http://community.petsathome.com/stores/ST_ALBANS/` which redirects a couple times, then 404s. The real url is `https://community.petsathome.com/store/st-albans/`

There site is just broken at the moment... I think we just come back in a few week and see if the site is fixed.

```python
from scrapy.link import Link
from scrapy.linkextractors import LinkExtractor
from scrapy.spiders import CrawlSpider, Rule

from locations.structured_data_spider import StructuredDataSpider


class PetsAtHomeGBSpider(CrawlSpider, StructuredDataSpider):
    name = "petsathome_gb"
    item_attributes = {"brand": "Pets at Home", "brand_wikidata": "Q7179258"}
    allowed_domains = ["petsathome.com"]
    start_urls = ["https://www.petsathome.com/shop/en/pets/store-locations"]
    rules = [
        Rule(LinkExtractor(allow=["https://www.petsathome.com/shop/en/pets/store-locations-results/"])),
        Rule(
            LinkExtractor(allow=["http://community.petsathome.com/stores/"]),
            callback="parse_sd",
            process_links="clean_links",
        ),
    ]
    wanted_types = ["PetStore"]
    custom_settings = {"ROBOTSTXT_OBEY": False}

    # Skip a bunch of redirects
    @staticmethod
    def clean_links(links: [Link]):
        for link in links:
            link.url = (
                link.url.lower().replace("http://", "https://").replace(":443", "").replace("/stores/", "/store/")
            )
            yield link
```

CloCkWeRX:
Is this one still busted? Or closable since skipped?

[...]

PR #5313, opened 2023-05-19
Skip Pets at Home web crawl

Cj-Malone:
Partial fix in #5309 as the website is causing issues.

@rjw62 I think the only missing POI is [Guernsey](https://community.petsathome.com/store/guernsey/), but overall this should be better for the Chain Reaction.


Cj-Malone:
@rjw62 some have bad coords, the usual bad, wrong side of town. But I've [updated OSM](https://www.openstreetmap.org/changeset/136304025), looking forward to seeing if I missed any when you update Chain Reaction.

PR #5322, opened 2023-05-19
whsmith_gb: fix spider

davidhicks:


Cj-Malone:
I'm not in favour of doing this in every spider. **If** we decide on doing it, it should be a pipeline.

PR #5333, opened 2023-05-21
cintas: fix spider

davidhicks:


iandees:
If you look at the map, it seems like this endpoint has data for more than just the US and silently sets the max radius to something like 1000 miles.

davidhicks:
> 

Thanks @iandees I've updated the Spider to use 4 different search locations that should cover all of USA and Canada (I hope).

PR #5355, opened 2023-05-26
Manually apply categories

Cj-Malone:
@iandees could you checkout Target. Maybe you'll have heard of the store formats? NSI has no concept of them, we should probably add them as `matchNames`.


iandees:
The mappings you have for Target seem reasonable. To be honest I think most of them are old and no longer used.

PR #5361, opened 2023-05-27
chickfila: fix and rename to chick_fil_a

davidhicks:


locations/spiders/chick-fil-a.py
```diff
@@ -0,0 +1,25 @@
+from locations.categories import Extras, apply_yes_no
+from locations.storefinders.yext import YextSpider
+from locations.structured_data_spider import clean_instagram, clean_twitter
+
+
+class ChickfilASpider(YextSpider):
+    name = "chick-fil-a"
```
Cj-Malone, line 7: I'd vote for chick_fil_a

iandees: Agree. Please use underscores for the spider name.

davidhicks: Done, thanks for the review.

I wasn't sure in this case as the official name has hypens not spaces. Agreed that underscores are best to replace any punctuation or spaces in official names :)

PR #5362, opened 2023-05-27
fazolis_us: fix spider

davidhicks:


locations/spiders/fazolis_us.py
```diff
@@ -1,10 +1,21 @@
...
-    sitemap_urls = ["https://locations.fazolis.com/robots.txt"]
-    sitemap_rules = [(r"https://locations\.fazolis\.com/\w\w/[-\w]+/[-\w]+\.html$", "parse_sd")]
+    api_key = "e6cda871db03e0fc9eb2211470649126"
+
+    def parse_item(self, item, location):
+        if location.get("c_pagesURL"):
+            item["website"] = location["c_pagesURL"]
+        item.pop("twitter")
+        if "instagram" in item["extras"].keys():
+            item["extras"].pop("instagram")
```
Cj-Malone, line 15: FYI this could be done in 1 line with:
```python
item["extras"]["instagram"] = None
```
or
```python
item["extras"].pop("instagram", None)
```

Cj-Malone: Also is it not `contact:instagram`? This is the OSM tag directly because we don't have it mapped in the exporter (because it's not a field on Feature, because it's not that common.)

davidhicks: Thanks @Cj-Malone I've suggested changes to fix this problem at https://github.com/alltheplaces/alltheplaces/pull/5369

PR #5375, opened 2023-05-30
zaxbys: fix spider and rename to zaxbys_us

davidhicks:
Returns 865 locations which is lower than the 900-930 locations than commercial shop location scrapers indicate. The company doesn't appear to publish a store location count in an obvious public location, however it was published they had 905 locations in May 2020[1] and a PR release from April 2023 states "over 900 locations across 17 states"[2].

The discrepancy in numbers is probably due to the sitemap on their website being out of date. Many location pages are redirects back to the main location finder page, indicating defunct locations still exist in the sitemap. Perhaps newer locations haven't been updated into the sitemap either.

[1] https://www.fastcasual.com/articles/zaxbys-shares-covid-19-recovery-plan/

[2] https://www.verdictfoodservice.com/news/zaxbys-launches-loyalty-programme/


davidhicks:
I think the sitemap URL is returning something else (error page?) when a data centre is the source IP address? I'm not sure we have a way with `requires_proxy` to denote that a residential IP address may be required.

The spider is running fine in Australia with results of:
```
{"atp/brand/Zaxby's": 865,
 'atp/brand_wikidata/Q8067525': 865,
 'atp/category/amenity/fast_food': 865,
 'atp/field/email/missing': 865,
 'atp/field/image/missing': 865,
 'atp/field/postcode/missing': 865,
 'atp/field/twitter/missing': 865,
 'atp/nsi/perfect_match': 865,

```

davidhicks:
Perhaps we need a better system for defining that a spider is deliberately blocked.

yt-dlp is one such repository that I thought might have an existing way of dealing with geoblocking, but it looks similar to ATP in that it only allows policies to be defined by country code[1][2].

I think we need something more expressive for these blocking rules. Some ideas below:
```blocking_rules = "+residential +cellular" #only residential or cellular IP addresses
blocking_rules = "+US +CA -Tor -VPN" #US or CA IP addresses only, excluding Tor exit nodes and VPN services
blocking_rules = "-VPN -Tor -datacentre" #anything other than VPN, Tor exit nodes or data centre IP addresses
blocking_rules = "+GB-residential" #only GB residential IP addresses```

The idea wouldn't be to create an exhaustive list of blocking rules, rather, if someone cared enough about using Tor or VPN services with ATP for avoiding geoblocking rules, they could update these blocking rules as necessary for each spider.

[1] https://github.com/yt-dlp/yt-dlp/blob/18f8fba7c89a87f99cc3313a1795848867e84fff/yt_dlp/extractor/common.py#L1160
[2] https://github.com/yt-dlp/yt-dlp/blob/18f8fba7c89a87f99cc3313a1795848867e84fff/yt_dlp/utils/_utils.py#L1121

Cj-Malone:
> blocking_rules

I sort of like it, but it's a lot of complexity for something that's not even hooked up yet. (`requires_proxy` isn't used in the weekly)

PR #5381, opened 2023-05-30
sweetgreen: fix spider and rename to sweetgreen_us

davidhicks:


davidhicks:
The CI results shows no output, but that's because the first grid search for a Sweetgreen store is (at default download delay) a few minutes into the run so we get a timeout for CI.

iandees:
Why not use the existing store listing URL?

davidhicks:
> Why not use the existing store listing URL?

Do you mean a SitemapSpider/CrawlSpider over the locations listed at https://www.sweetgreen.com/locations/ ? I think with this approach we'd then need to parse the page for each location to find a location ID that is then used to dynamically load location data for one location at a time via GraphQL query. I count then 1 + 177 * 2 = 354 queries needed which is about half the number proposed via this merge request, so probably a better option. We could probably beat this figure if a 200 or 300 mile centroid list for the USA was available instead of 100 mile variant currently available. I'm just not sure how those centroid lists were developed--is there a process documented previously?

iandees:
@wessport did you ever create documentation for how you made these searchable points files?

PR #5385, opened 2023-06-01
Add more categories

Cj-Malone:


Cj-Malone:
pytest works locally. It looks like another dependency issue, this time `typing_extensions`.

iandees:
I think the issue is that the Pipfile.lock was created with python 3.10 (what my local virtualenv was using) and we're running 3.11 on CI (and 3.10 on the Docker image for the weekly). I'm working on making them consistent now.

PR #5431, opened 2023-06-07
biggby: fix and rename to biggby_coffee_us

davidhicks:


davidhicks:
The last successful run of this bot in the weekly run returned 423 results. The difference with this revised spider returning 356 results is discarding of stores that within the official store finder are labelled as "coming soon".

PR #5432, opened 2023-06-07
belk: delete as HUMAN Bot Defender is in use

davidhicks:
Reference: https://docs.perimeterx.com/docs/bd-overview


davidhicks:
It might be worth checking whether clients in the US are challenged. If not, this spider could be kept and `requires_proxy = "US"` set.

davidhicks:
`hibbett_sports` is also using HUMAN bot defender within the past week or two.

Any thoughts on how to archive/disable these spiders as broken due to bot blocking?

davidhicks:
I like the idea of placing them in a separate directory, at least until ATP has in-built support for working around bot detection systems and spiders being able to specify the bot blocking mechanism that needs to be worked around.

iandees:
Looks like we got some results last week, so I'm going to close this and keep it around.

PR #5473, opened 2023-06-15
Added picard spider (1136 locations) and fixed systeme_u spider

Jean-Huang98:


locations/spiders/picard.py
```diff
@@ -0,0 +1,39 @@
...
+
+from locations.hours import OpeningHours
+from locations.items import Feature
+
+
+class PicardSpider(scrapy.Spider):
+    name = "picard"
+    item_attributes = {"brand": "Picard Surgelés", "brand_wikidata": "Q3382454"}
+    allowed_domains = ["magasins.picard.fr"]
+    start_urls = [
```
Cj-Malone, line 13: I don't like manually listing the URLs, because once they add a sitemap4 we'll be losing POIs.

Instead of inheriting Spider, you could inherit SitemapSpider. It would mean and you could delete parse, since it does than for you. Also you could set sitemap_urls to https://magasins.picard.fr/sitemap.xml, or even robots.txt, then a sitemap_follows of ["locationsitemap"]. So it will automatically fetch all the matching sitemaps.

Please forgive any typos, I've done this on my phone.

iandees:
Closing as stale.

PR #5500, opened 2023-07-06
Add GLSC Gazetteer Spider

### LINKED ISSUE #5499: Guyana Gazetteer
### Brand name

Guyana

### Wikidata ID

_No response_

### Store finder url(s)

https://gazetteer.glsc.gov.gy/gazetteer/

Cj-Malone:
Fixes #5499

@justinelliotmeyers we try to be as close to OSM categories as possible, feel free to improve the mapping if you can.

### DIFF
--- a/locations/spiders/glsc_gazetteer_gy.py
+++ b/locations/spiders/glsc_gazetteer_gy.py
@@ -1,0 +1,48 @@
+import json
+
+from scrapy import Spider
+
+from locations.categories import apply_category
+from locations.items import Feature
+
+
+class GLSCGazetteerGYSpider(Spider):
+    name = "glsc_gazetteer_gy"
+    start_urls = ["https://gazetteer.glsc.gov.gy/gazetteer/data/NationalGazetteer_0.js"]
+    no_refs = True
+
+    cats = {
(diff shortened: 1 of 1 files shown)


iandees:
There's a lot of data of questionable quality in there, too...

Cj-Malone:
I was originally annoyed by there being only points, even for features like rivers. But I the lack of data is quite understandable, and so I can't blame them for that. Honestly, I think any nation without an established mapping agency by now should just embrace OSM, but that's not my decision to make.

I have however just found there [category](https://gazetteer.glsc.gov.gy/index.php/explanatory-notes/#103) definitions, and I would definitely blame them for having both `Fall` and `Falls` in the data when only the first is in the spec.

iandees:
I'm not happy with the quality of the data in here, so I'm going to close this out.

I think the GeoJSON produced by the CI here would be useful for someone who might want to add this to OSM, though.

justinelliotmeyers:
Agree and thank you all for looking into it 
 Appreciate your help

iandees:
Thanks for finding the data, Justin. Keep it up!

Issue #5495, opened 2023-07-05
searchable_points generation


jimbofreedman:
Hi,

Is the code that was used to generate these available somewhere? I'd like to generate some more and could save a bit of time.

Thanks!

jimbofreedman:
I found the wikidata query

iandees:
Unfortunately we don't have information about how the searchable points were generated. I can probably regenerate something like them, but I don't think we ever got @wessport's readout on how they were made.

wessport:
@iandees apologies missed the earlier tag. 

A short write up on how the points were created would be good. I think I have some notes written down somewhere. I can also generate a new set if that would be helpful. Traveling this week, but will be back in early next week and can write up how the points were generated.

jimbofreedman:
Amazing, thank you!

davidhicks:
I've been working on something similar for the `western_union` spider that needs searchable points across the entire surface of the Earth for 350km radius searches. I've proposed a documented whole-of-Earth searchable points generation method for radius search APIs at https://github.com/alltheplaces/alltheplaces/pull/5509 (using `western_union` as an example consumer).

This proposed icosahedral discrete global grid approach to generating searchable points for radius searches is a bit trickier to apply on a country-by-country level as the software ([DGGRID](https://github.com/sahrk/DGGRID)) only appears to have crude means to optimise the returned searchable points for a small area of the Earth. For example, it is possible to crop a whole-of-Earth list of carefully spaced centroids to just those centroids overlapping a supplied shapefile (shapefile outline of a country). It is also possible to shift all centroids +/- latitude and +/- longitude to find a somewhat optimal overlap of cells over a country, minimising the total number of centroids needed to completely cover a country. This could perhaps be overcome with some scripts that run DGGRID multiple times to shift +/- latitude and +/- longitude and within X attempts, picks the one that results in complete coverage of a country shapefile with the fewest number of centroids required.

jimbofreedman:
#5509 answers my question, thanks

davidhicks:
@jimbofreedman also see https://github.com/alltheplaces/alltheplaces/blob/master/locations/searchable_points/au_centroids_iseadgg_175km_radius.meta for an example of how to generate searchable points for a country code, and https://github.com/alltheplaces/alltheplaces/blob/master/locations/searchable_points/gg_gb_im_je_centroids_iseadgg_48km_radius.meta for multiple country codes.

I'm happy to generate more searchable points as needed just let me know.

PR #5509, opened 2023-07-08
western_union: fix spider to use correct geographic search

davidhicks:
The Western Union GraphQL endpoint does use a 350km radius from each supplied search coordinate to find locations to return (over multiple pages of results). The previous attempt to fix this spider incorrectly assumed a 'country' parameter to the GraphQL search function was used instead of a radius being applied.

To achieve this search across the entire surface of the Earth, a new searchable points file is provided, along with documentation for how this searchable points file was created, and how similar could be created in the future for different search radiuses.


davidhicks:
There's a small number of erroneous coordinates that will need filtering from the sample results stated above. For example, two coordinates near the North pole and one in the middle of an ocean. Can I come back to address this after this spider completes a weekly run so we have all the data needed to see whether any of these erroneous coordinates can be quickly and efficiently filtered out (e.g. ignore all coordinates near the North pole if they're all wrong).

davidhicks:
@iandees Is this waiting for some further resolution of issue #5495 for how searchable points are generated?

iandees:
I was still trying to understand the tools you used to generate the CSV and it sounded like you needed to filter some coordinates out of the file? Maybe you meant coordinates in the Western Union data.

In the future, it would be helpful to make separate PRs for different chunks of work.

davidhicks:
There are a few bad coordinates I've seen in the Western Union data, somewhat like the problem I fixed recently for moneygram per pull request #5515. These bad coordinates likely existed in the previous CrawlSpider version of western_union too. I just haven't had a chance to do a full run of the western_union spider to then figure out how many coordinates might be bad, and how best to remove those coordinates (if it's practical to do so).

The reason for putting the new searchable points generation method within this pull request was so that it was committed along with the first consumer of the method (western_union), allowing the effectiveness to be verified. I was planning to wait for issue #5495 discussion to conclude before extending this new searchable points generation method to be used country-by-country. And then, my thought was if we commit new searchable points, that a few spiders be switched over to the new method at the same time so that the effectiveness can be verified (new location count = old location count). Or is it best to keep separate PRs for the change to searchable points, and a separate PR for each spider switched to the new method?

iandees:
Thanks for the explanation. Generally I would prefer that new features (like the new searchable points) get added as a separate PR so that they're easier to find and to encapsulate discussion about them.

davidhicks:
Thanks I'll make sure I do that for planned searchable points generation per-country as proposed in #5495

PR #5518, opened 2023-07-11
rolld_au: convert to WPStoreLocatorSpider

davidhicks:


davidhicks:
Actually this is probably truncating at 100 locations. WPStoreLocatorSpider is often configured with a hard max_results limit and I don't think there is a workaround from the WPStoreLocatorSpider source code (eg no paging).

Cancelling

PR #5539, opened 2023-07-14
Add Autograph (AU/NZ) (58 stores)

davidhicks:


iandees:
Can you double-check the exception happening here?

davidhicks:
That was a really weird exception about chompjs encountering a control character, even though I couldn't find the stated control code in the ld+json blob. Fixed by disabling json.loads strict mode to allow control characters to be parsed.

iandees:
Interesting, thanks for checking!

PR #5562, opened 2023-07-17
bp: fix spider

davidhicks:


davidhicks:
Leaving this as a draft as I have to run this spider through a full pass which takes time, especially given the API has a 429 rate limiting mechanism if going too quick.

I'll report back the results but some initial results look positive :)

davidhicks:
Fixes #5568

iandees:
(FYI the "closes X" or "fixes X" have to be in the description of the PR to take action on the issue)

PR #5564, opened 2023-07-18
Revert "[pre-commit.ci] pre-commit autoupdate"

iandees:
Reverts alltheplaces/alltheplaces#5563

Seems to have caused build problems.


iandees:
The build did update. Latest run is July 17, which is yesterday.

PR #5578, opened 2023-07-24
albertsons: fix errors in spider

davidhicks:
* microdata_parser was erroring on incorrect URLs in Albertson's store webpages which contain e.g.: a href="https://www.[c_bannerURLName].com/pharmacy/covid-19.html" (example: https://local.pharmacy.acmemarkets.com/pa/west-chester/907-paoli-pike.html)

* Facebook and Twitter fields are not store specific, so ignore them.


davidhicks:
Thank you @iandees I missed that facebook and twitter fields are only conditionally set.

PR #5581, opened 2023-07-25
Add spider for Maggianos Little Italy

nbeecher:


locations/spiders/maggianos_little_italy.py
```diff
@@ -0,0 +1,49 @@
...
+        return opening_hours.as_opening_hours()
+
+    def parse(self, response):
+        data = response.json()
+
+        for place in data:
+            properties = {
+                "name": place["properties"]["business_name"],
+                "ref": place["properties"]["store_code"],
+                "addr_full": place["properties"]["slug"]["address_line_1"],
```
iandees, line 35:
```suggestion
                "street_address": place["properties"]["slug"]["address_line_1"],
```

davidhicks: The source data also has field `address_line_2` so if this field is not blank, it should be appended to the street address:

```
"street_address": ", ".join(filter(None, [place["properties"]["slug"]["address_line_1"], place["properties"]["slug"]["address_line_2"]])),
```

locations/spiders/maggianos_little_italy.py
```diff
@@ -0,0 +1,50 @@
...
+    ]
+
+    def parse_hours(self, hours):
+        opening_hours = OpeningHours()
+
+        for hour in hours:
+            day = hour["day_name"][:2]
+            opening_hours.add_range(day, open_time=hour["open_time"], close_time=hour["end_time"])
+
+        return opening_hours.as_opening_hours()
```
davidhicks, line 26: Return just `opening_hours` instead of the textual representation generated through `opening_hours.as_opening_hours()`.

locations/spiders/maggianos_little_italy.py
```diff
@@ -0,0 +1,50 @@
...
+                "state": place["properties"]["slug"]["state_abbreviation"],
+                "postcode": place["properties"]["slug"]["postal_code"],
+                "country": place["properties"]["country"],
+                "phone": place["properties"]["primary_phone"],
+                "lat": place["geometry"]["coordinates"]["latitude"],
+                "lon": place["geometry"]["coordinates"]["longitude"],
+            }
+
+            hours = self.parse_hours(place["properties"]["store_hours"])
+            if hours:
```
davidhicks, line 46: No need to check `if hours:`, you can safety return a 'blank' OpeningHours object here.

locations/spiders/maggianos_little_italy.py
davidhicks: The source data also has a `full_address` field that can be extracted to `properties["addr_full"]`. Then you have both `properties["addr_full"]` and `properties["street_address"]`

davidhicks (changes requested):

PR #5586, opened 2023-07-26
Add Żabka spider (~9506 stores)

alufers:
Polish Convenience store brand.

I have extracted the API endpoint and key from the store's app: https://play.google.com/store/apps/details?id=pl.zabka.apb2c&hl=en_US

It is better than the one that can be found on the website, because it returns all the shops in one huge JSON along with all the data for each shop. The website API would need to make a HTTP request for each of the 9000 shops separately. 

I have added opening hours for Monday-Saturday, but due to Polish law Sundays are considered non-working days, but the shops can be open if the owner or the owner's family member is working there. This means that each franchisee sets the opening hours for each specific Sunday, and it ends up in the `"nonTradingDays"` field:

```json
   "nonTradingDays": [
      {
        "date": "2023-07-16",
        "openTime": "09:00",
        "closeTime": "21:00"
      },
      {
        "date": "2023-07-23",
        "openTime": "09:00",
        "closeTime": "21:00"
      },
      {
        "date": "2023-07-30",
        "openTime": "09:00",
        "closeTime": "21:00"
      }
    ]
```

In practice almost all stores will be open each Sunday with the same hours. Unfortunately not all of them have the data filled in (even if the plaque on the door says otherwise), so I did not attempt to parse this field.

### DIFF
--- a/locations/spiders/zabka_pl.py
+++ b/locations/spiders/zabka_pl.py
@@ -1,0 +1,35 @@
+import scrapy
+from scrapy.http import JsonRequest
+
+from locations.dict_parser import DictParser
+from locations.hours import OpeningHours
+
+
+class ZabkaPLSpider(scrapy.Spider):
+    name = "zabka_pl"
+    item_attributes = {"brand": "Żabka", "brand_wikidata": "Q2589061"}
+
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Thank you

PR #5588, opened 2023-07-27
Add spider for Ameriprise

jleedev:
Don't convert {"fax":None} to the string 'None'.

### DIFF
--- a/locations/spiders/ameriprise.py
+++ b/locations/spiders/ameriprise.py
@@ -3,8 +3,8 @@ import json
 from scrapy.spiders import SitemapSpider
 
 from locations.google_url import extract_google_position
-from locations.linked_data_parser import LinkedDataParser
 from locations.items import Feature
+from locations.linked_data_parser import LinkedDataParser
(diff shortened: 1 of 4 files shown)


locations/spiders/ameriprise.py
```diff
@@ -0,0 +1,39 @@
...
+
+from scrapy.spiders import SitemapSpider
+
+from locations.google_url import extract_google_position
+from locations.items import Feature
+from locations.linked_data_parser import LinkedDataParser
+
+
+class AmeripriseSpider(SitemapSpider):
+    name = "ameriprise"
```
iandees, line 11: Should this be `ameriprise_us`? i.e. will there be entries from other countries?

iandees: Can we change this to `AmeripriseUSSpider` and `ameriprise_us` name? I think we're trying to move towards this on spiders that only produce data for a single country.

iandees:
Thanks!

PR #5595, opened 2023-07-30
Add GB car repair spider

Cj-Malone:
So a little to unpack here.

MOT is a test on cars every year. Garages that do the test register with the government.

This is a gov dataset, not first party. So I've prefixed it with `gov_`. I've not thought about this much, so there may be a better solution.

I've added `brand:wikidata` for the bigger chains

```
    127 FORMULA ONE AUTOCENTRES LIMITED
    141 ARNOLD CLARK
    189 ATS EUROMASTER LIMITED
    506 HALFORDS AUTOCENTRE
    530 KWIK FIT
```

But it's the smaller/independent ones that are quite interesting, we wouldn't usually see that data and its license is compatible with OSM.

### DIFF
--- a/locations/spiders/gov_mot_gb.py
+++ b/locations/spiders/gov_mot_gb.py
@@ -1,0 +1,61 @@
+import json
+
+from scrapy import Request
+from scrapy.spiders import CSVFeedSpider
+
+from locations.categories import Categories, apply_category
+from locations.items import Feature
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Oh and no coords. It's unfortunate.

I've been toying with a pipeline to geocode items using https://nominatim.openstreetmap.org/search. It isn't great data wise, for GB addresses at least, and it's a blocking http request per item so I'll leave it for now.

Issue #5598, opened 2023-08-01
Should address fields all be strings (rather than string|lists)?


jimbofreedman:
https://github.com/alltheplaces/alltheplaces/blob/0606485dc5d9141b6704bac5de96c6687c3d6168/locations/linked_data_parser.py#L80

There are a few spiders returning lists for address fields - so far I've found `mydentist`, `national` and `new_look_gb`.

Could/should we join these like we do for street address? 

(I'm volunteering)

Cj-Malone:
Yes, it should be a string. And, yeah we'd appreciate a PRs for fixing this, and in general for moving closer to the OSM spec. I think there are a few booleans in the output, they are on my todo list to change to yes/no.

jimbofreedman:
Great - I'm loading them into bigquery and it breaks my schema so I'll fix all the ones that are otherwise currently returning items.

Cj-Malone:
I just got a little confused, because I thought you were saying street address is a list. But you weren't, you weren't, you were saying other address fields are.

```python
{'mydentist_gb/addr:city/list': 230,
 'mydentist_gb/contact:email/list': 5,
 'national/addr:full/list': 1522,
 'new_look_gb/addr:state/list': 1}
```

It's still a bug. But I don't know if `LinkedDataParser` is the best place to fix it. It won't help with `national`, as it doesn't use that code. And the first occurrence of it in `mydentist_gb` is:
```json
{"addr:street_address": "Unit 6A", "addr:city": ["Broad Street Plaza", "Halifax"]}
```
But the desirable output would be:
```json
{"addr:street_address": "Unit 6A, Broad Street Plaza", "addr:city": "Halifax"}
```

To do that, it would have to be fixed in spider, not in `LinkedDataParser`

Cj-Malone:
And everything:
```python
{'decathlon_sk/nsi_id/int': 13,
 'verizon/store_type_note/list': 4444,
 'speedway/amenity:fuel/bool': 3284,
 'speedway/atm/bool': 2180,
 'speedway/fuel:diesel/bool': 2816,
 'speedway/fuel:octane_87/bool': 3263,
 'speedway/fuel:octane_89/bool': 3234,
 'speedway/fuel:octane_93/bool': 3233,
 'speedway/fuel:propane/bool': 1526,
 'speedway/car_wash/bool': 323,
 'speedway/fuel:e85/bool': 488,
 'speedway/fuel:HGV_diesel/bool': 277,
[...]

PR #5599, opened 2023-08-01
Add insights option to check value types

Cj-Malone:
As noted in #5598, some of our output values aren't strings. This counts them so we can detect and fix them.

### DIFF
--- a/locations/commands/insights.py
+++ b/locations/commands/insights.py
@@ -1,11 +1,13 @@
 import json
 import os
+import pprint
 import re
 from collections import Counter
 from zipfile import ZipFile
 
 import ijson
 import requests
+import scrapy.statscollectors
 from scrapy.commands import ScrapyCommand
 from scrapy.exceptions import UsageError
 
@@ -87,6 +89,12 @@ class InsightsCommand(ScrapyCommand):
             help="Do not process data for spider matching this file name fragment",
         )
         parser.add_argument(
+            "--value-types",
+            dest="value_types",
+            action="store_true",
+            help="Check property values are strings",
+        )
+        parser.add_argument(
             "--country-codes",
             dest="country_codes",
             action="store_true",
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Actually I'm going to put a error in the pipeline, so we hopefully catch these while writing spiders.

jimbofreedman:
Heavily related to this, I'm actually using Scrapy Cloud with Spidermon and I'm adding validation. Do you already have an idea of what type enforcement there should be? Particularly around fields you consider _required_. I don't want to be inventing any rules here so better to just ask :grin:, and can't see anything in any of the create-a-spider docs.

Cj-Malone:
> fields you consider required

Not even `ref` is literally required any more, just strongly recommended.
[Here](https://github.com/alltheplaces/alltheplaces/blob/225a97029966f7cc7cc5c3de389bd9d3c31827ef/locations/pipelines/check_item_properties.py#L47-L58) + `lat`, `lon`, `twitter` and `opening_hours` are the ones we complain about when missing. Although it's very rare we get all of them.

Maybe we should trim down the missing counters, it's probably been hiding the `wrong_type` counters and time of writing spiders.

> everything should be a string, except extras, geometry, and lat/lon

I don't know enough about Scrapy Cloud/Spidermon to know the which point in atp you are looking at. If you are looking at the exported files (GeoJSON), then even the `extras` should be applied to the feature properties, and there too they should really be string key, string value. This is to keep us as close as possible to [OSM](https://wiki.openstreetmap.org/wiki/Main_Page).

If you are looking before the export, then yes `extras` is a dict, but ideally it's just string key, string value for when it gets exported.

We are generally pushing to `geometry` over lat/lon, since we now support outputting "ways"/"line strings" or in theory polygons, or any GeoJSON `geometry`.

`geometry` should be a dict (or null). lat/lon should be floats by the time they are exported, but inside atp they are often strings. Only getting converted towards the end.

PR #5600, opened 2023-08-01
more consistency for brand name in insight.json

mjoe999:
The insights command produces output for consumption by: https://www.alltheplaces.xyz/wikidata.html. We are not comparing multiple language names at this point so prefer the English brand name if available for consistency. This avoids:

image

### DIFF
--- a/locations/commands/insights.py
+++ b/locations/commands/insights.py
@@ -224,6 +224,12 @@ class InsightsCommand(ScrapyCommand):
             wikidata_dict[wikidata_code] = record
             return record
 
+        def get_brand_name(item_tags: dict):
+            # Prefer English brand name for insights application (https://www.alltheplaces.xyz/wikidata.html).
+            if brand_name := item_tags.get("brand:en"):
+                return brand_name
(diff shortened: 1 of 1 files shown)


locations/commands/insights.py
```diff
@@ -224,6 +224,12 @@ def lookup_code(wikidata_code):
             wikidata_dict[wikidata_code] = record
             return record
 
+        def get_brand_name(item_tags: dict):
+            # Prefer English brand name for insights application (https://www.alltheplaces.xyz/wikidata.html).
+            if brand_name := item_tags.get("brand:en"):
+                return brand_name
+            return item_tags.get("brand")
```
iandees, line 231: Python lets us shorten this up to this. What do you think?

```suggestion
            return item_tags.get("brand:en") or item_tags.get("brand")
```

mjoe999: For sure better and done!

JoeDixon-TomTom:
@iandees your new Python tip got me out of the flake8 bind also :-)

PR #5606, opened 2023-08-02
Improve Midcounties Co-operative brand and categories

Cj-Malone:
Rel: https://github.com/osmlab/name-suggestion-index/issues/8469

@TheUKHighStreet you are very welcome to check the output on this. There will be a "map" link in the comment below.


TheUKHighStreet:
Thanks - that looks very similar to the map on their website of all their locations:
https://www.midcounties.coop/stores/

Sorry, I'm not quite sure yet how all this works or how alltheplaces ties in with the Open Street Map. I think I have a lot of questions but I'll have another look tomorrow. Thanks for getting this sorted so quickly.

Cj-Malone:
> Thanks - that looks very similar to the map on their website of all their locations:

Yeah, this is a copy of that. But accessible and in OSM format.

> Sorry, I'm not quite sure yet how all this works or how alltheplaces ties in with the Open Street Map.

So ATP collects and normalises POIs, unfortunately this data isn't directly used in OSM, although there are [QA tools](https://osm.mathmos.net/chains/).

> I think I have a lot of questions but I'll have another look tomorrow. Thanks for getting this sorted so quickly.

Happy to answer any questions.

PR #5608, opened 2023-08-02
Add spider for Carter's

jleedev:


jleedev:
Interesting, it was responding for me from AWS.

Cj-Malone (approved):
Can confirm this works GB residential

PR #5616, opened 2023-08-05
Fix FedEx.

jleedev:
Test me with `scrapy sd https://local.fedex.com/pl-pl/zyrardow/wawek`.

The microdata contains nested itemprop=name, which gets translated as an array of names. Teach LinkedDataParser to use the first name when a list is found.

When an item contains an array of names, an exception is thrown. Teach ClosePipeline to guard against this so that it does not cause items to be lost.

The `scrapy sd` command prints out the found item, then a message saying "failed to decode structured data", which does not explain anything. Change its log level from disabled to warning so that exceptions are not swallowed but it otherwise remains terse.

Sort the stats output from `scrapy sd`.

Fix isinstance() call in CheckItemProperties.

### DIFF
--- a/locations/commands/sd.py
+++ b/locations/commands/sd.py
@@ -95,4 +95,4 @@ class SdCommand(BaseRunSpiderCommand):
         if stats_dict.get("item_scraped_count", 0) == 0:
             print("failed to decode structured data")
         if opts.stats:
-            print(pprint.pformat(stats_dict))
+            pprint.pprint(stats_dict)
--- a/locations/commands/sd.py
+++ b/locations/commands/sd.py
@@ -1,6 +1,7 @@
 import json
 import os
 import pathlib
+import pprint
 
 from scrapy.commands import BaseRunSpiderCommand
 from scrapy.exceptions import UsageError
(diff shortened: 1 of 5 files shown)


jleedev:
It seems like raising an exception (which is not DropItem) in a pipeline's process_item leads to a pretty major termination in scrapy, based on the tracebacks at https://data.alltheplaces.xyz/runs/2023-07-29-13-32-45/logs/fedex.txt going right out to Twisted and not passing through any Scrapy code, which might explain why no items are produced at all despite only some pages triggering this exception. This can probably be described as a bug in Scrapy.

locations/commands/sd.py
```diff
@@ -94,5 +95,4 @@ def run(self, args, opts):
         if stats_dict.get("item_scraped_count", 0) == 0:
             print("failed to decode structured data")
         if opts.stats:
-            for k, v in stats_dict.items():
-                print(k + ": " + str(v))
+            print(pprint.pformat(stats_dict))
```
Cj-Malone, line 98:
```suggestion
            pprint.pprint(stats_dict)
```

May be more readable?

Cj-Malone (approved):

PR #5621, opened 2023-08-07
Add spider for O'Charley's

nbeecher:


locations/spiders/ocharleys.py
```diff
@@ -0,0 +1,37 @@
...
+import re
+
+import scrapy
+
+from locations.hours import OpeningHours
+from locations.items import Feature
+
+
+class OcharleysSpider(scrapy.Spider):
+    name = "ocharleys"
```
iandees, line 12: Can you please change the spider name here, class name, and file name to reference that it's US-only?

nbeecher: Yes, fixed

iandees:
Thanks!

PR #5654, opened 2023-08-11
solasalons: fix and rename to sola_salon_studios

davidhicks:
Note: geographic coordinates do not appear to be available for salon locations. Locations exist in both US and CA.


davidhicks:
Actually I found https://platform.solasalonstudios.com/v1/get-web-locations-list which has geographic coordinates and other details.

Will need to redo this fix starting from https://platform.solasalonstudios.com/v1/get-web-locations-list and then adding extra details from extracting data from the site specific location pages.

PR #5674, opened 2023-08-15
decathlon: fix spider

davidhicks:
Decathlon implements the Woosmap store finder for their larger (generally European) websites. There is a separate API key per country that Decathlon operates within. This change removes the old Decathlon spider and replaces it with country-specific WoosmapSpider spiders. No locations that would have been captured by the old "decathlon" spider would be missed with the new country-specific WoosmapSpider spiders.

A new Python dependency "unidecode" is also introduced as Decathlon's website URLs are an ASCII-equivalent transliteration of the store name, and "unidecode" provides the ability to convert European language diacritics to the closest ASCII-equivalent character.

There are other countries within which Decathlon operates however they appear to have only a few stores each, and the websites do not use a country-specific Woosmap store finder API key. For most of these other countries, it is possible that each store page needs to be parsed individually to extract store location information. This is separate work unrelated to the replacement of the old "decathlon" spider with country-specific WoosmapSpider spiders.


davidhicks:
Depends upon PR#5673

iandees:
`unidecode` shouldn't be necessary – we should be able to output UTF-8 encoded Unicode codepoints without a problem. What particular problems were you seeing there?

davidhicks:
Woosmap API doesn't return a URL or slug for each store. Instead, we have to create this URL ourselves.

For example, Woosmap API returns the store name "Pozuelo de Alarcón" and store identifier "0070070200702". The required URL is "https://www.decathlon.es/es/store-view/tienda-de-deportes-pozuelo-de-alarcon-0070070200702". The prefix "https://www.decathlon.es/es/store-view/tienda-de-deportes-" is common across all Spanish locations.

Notice the diacritic "ó" is transliterated to just "o" in the resulting URL.

I've just realised that the store name in the URL is not verified server-side, even though it is expected and can't be blank, so "https://www.decathlon.es/es/store-view/tienda-de-deportes-abc-0070070200702" redirects to "https://www.decathlon.es/es/store-view/tienda-de-deportes-pozuelo-de-alarcon-0070070200702". I'm not sure we'd want to record the website incorrectly though, and rely upon the server redirecting to the correct URL?

iandees:
Got it, that makes sense.

PR #5675, opened 2023-08-15
jardiland: replace spider as Woosmap store finder is no longer used

davidhicks:
Jardiland appear to no longer use the Woosmap store finder on their websites (FR, ES and PT) even though the old API key appears to continue to function.

This change removes the old "jardliland" WoosmapSpider and replaces it with spiders designed for newer FR and ES/PT websites.


davidhicks:
Thanks for fixing the country property... oops I missed that!

iandees:
No problem!

PR #5676, opened 2023-08-15
dunkin_us: fix spider

davidhicks:


iandees:
This doesn't seem to pull in lat/lon. Is the data not available with linkeddata?

davidhicks:
Thanks for the heads up. I missed the part where their ld+json blob doesn't contain coordinates :(

There is another JS blob on the web page which does usually, but not always, contain coordinates. I've updated the spider to extract these coordinates when they exist.

davidhicks:
@iandees

PR #5682, opened 2023-08-15
Add Baskin Robbins India

iandees:


iandees:
@davidhicks This Stockist site doesn't seem to support the /locations/all endpoint. Any ideas for how to work around that?

iandees:
(At https://baskinrobbinsindia.com/pages/store-locator)

davidhicks:
It looks like the user of Stockist has to enable a feature to allow all locations to be returned at once:

https://help.stockist.co/article/35-not-seeing-stores

With the URL of https://stockist.co/api/v1/u11410/locations/search?latitude=0&longitude=0&distance=100000000000 I am unsure how to change the maximum results beyond 100. There is a maximum results setting referred to at https://help.stockist.co/article/14-im-not-seeing-results-when-searching-by-country-or-state but I can't find any API documentation from Stockist, just user interface documentation for users.

davidhicks:
This store locator is fixed as `max_results=100` and `max_distance=50000` per https://stockist.co/api/v1/u11410/widget.js?callback=_stockistConfigCallback_u11410 so I don't think adjusting maximum results will help.

I think we'll need a new/modified Stockist store locator that handles store locator pages where locations/all is disabled.

I think I've found the answer too. You can query https://stockist.co/api/v1/u11410/locations/overview.js which returns [geohashes](https://en.wikipedia.org/wiki/Geohash) of all locations. The geohashes are 9 characters long which appear to be granular to +/- 4.77m. Thus I think our search algorithm could be:

1. Get all geohashes and store in dictionary X.
2. For one geohash picked from dictionary X, convert to a coordinate and then search the Stockist API with a `max_distance` as obtained from widget.js.
3. For all locations returned from the Stockist API, geohash the coordinates. Remove any geohashes from dictionary X that have been found.
4. Continue steps 2-3 until no geohashes remain in the dictionary.

davidhicks:
Ideally we'd also track in a separate array the coordinates which have already been searched, and then pick a new coordinate from dictionary X which has the largest minimum haversine distance to all points already searched. This approach would hopefully minimise searches by tending to avoid overlap in search radiuses, without requiring a more scientific approach such as constrained k-means clustering (e.g. https://github.com/joshlk/k-means-constrained).

I'm happy to write up a change to the Stockist store finder, but would like to check that I haven't missed an obviously better search algorithm, or a better approach to selecting next coordinates from dictionary X to search.

Cj-Malone:
There also seems to be [overview.js](https://stockist.co/api/v1/u11410/locations/overview.js) which gives you the [geohash](http://geohash.co/) of every location.

Cj-Malone:
We could also try and ask stockist.co for there data.

davidhicks:
I've implemented a fix for StockistSpider that is transparent to users of StockistSpider. Users still just need to supply a single "key" parameter for the Stockist API key, and StockistSpider will do the rest.

Cj-Malone:
I asked stockist for all there data, they would want permission from there customers, and it isn't top of the todo list. But they seemed positive about it.

davidhicks:
You can find most of the big users of Stockist with a search for `"Loading store locator from Stockist store locator..." site:.co.uk`, `"Loading store locator from Stockist store locator..." site:.au` etc. The problem I find is that Stockist is very frequently used by brands that stock their products in other brick and mortar stores, so there is a fair bit of hunting through Stockist users to find ones that actually have their own physical presence. One of the better store finder APIs though!

Cj-Malone:
> The problem I find is that Stockist is very frequently used by brands that stock their products in other brick and mortar stores,

Commonly known as [stockists](https://www.wordnik.com/words/stockist) :p

I raised this in my last email, I hope if they give us data they'd give us a flag so we'd know.

> One of the better store finder APIs though!

"there are just a couple of us" so yeah, good API and a good team by the sounds of it.

PR #5685, opened 2023-08-15
Add Baskin-Robbins (GB/IE/JE) (130 stores)

davidhicks:
All locations appear to be colocated at Cineworld or Millie's Cookies. Branding appears to be separate and distinct from each other, such that Baskin-Robbins appears as a store-within-a-store.

PR #5687, opened 2023-08-15
Add Cineworld GB/JE Spider

Cj-Malone:
Unfortunately this seems to be blocked from none GB ips.

### DIFF
--- a/locations/spiders/cineworld_gb.py
+++ b/locations/spiders/cineworld_gb.py
@@ -1,0 +1,30 @@
+import json
+import re
+
+from scrapy import Spider
+
+from locations.dict_parser import DictParser
+from locations.spiders.vapestore_gb import clean_address
+
+
+class CineworldGBSpider(Spider):
+    name = "cineworld_gb"
+    item_attributes = {"brand": "Cineworld", "brand_wikidata": "Q5120901"}
+    start_urls = ["https://www.cineworld.co.uk/"]
+    requires_proxy = True
+
+    def parse(self, response, **kwargs):
+        for location in json.loads(re.search(r"apiSitesList = (\[.+\]),", response.text).group(1)):
+            item = DictParser.parse(location)
+            item["street_address"] = clean_address(
+                [
+                    location["address"]["address1"],
+                    location["address"]["address2"],
+                    location["address"]["address3"],
+                    location["address"]["address4"],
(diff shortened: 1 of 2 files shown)


davidhicks:
Could start_urls also include `https://www.cineworld.ie/` or a different spider is needed?

Cj-Malone:
Ha I'm blocked from IE

davidhicks:
Looks like two spiders needed, with `requires_proxy = "GB"` and `requires_proxy = "IE"` respectively. Sigh

Cj-Malone:
Yeah, but it there is only 1 location in IE, so not writing a spider for that.

https://www.cineworld.ie/cinemas/dublin/0001

davidhicks:
Sounds good then! I just noticed that Dublin location come up in the PR #5685 (baskin_robbins_gb_ie) spider so wasn't sure if there were more Cineworld locations in Ireland beyond the single Dublin location.

locations/spiders/cineworld_gb.py
```diff
@@ -11,7 +11,7 @@ class CineworldGBSpider(Spider):
     name = "cineworld_gb"
     item_attributes = {"brand": "Cineworld", "brand_wikidata": "Q5120901"}
     start_urls = ["https://www.cineworld.co.uk/"]
-    requires_proxy = True
+    requires_proxy = "GB"
```
Cj-Malone, line 14: The theory is that it would default to the spider location, from the name cineworld_**gb**. So we don't need to be explicit in most cases.

davidhicks:
Ah I stuffed that up... I think with Jersey included (JE), `requires_proxy = "GB"` is probably once again needed?

Cj-Malone:
> Sounds good then! I just noticed that Dublin location come up in the PR https://github.com/alltheplaces/alltheplaces/pull/5685 (baskin_robbins_gb_ie) spider so wasn't sure if there were more Cineworld locations in Ireland beyond the single Dublin location.

And it's well [mapped](https://www.openstreetmap.org/way/269884096), my opinion is these smaller chains/independent stores should just have good data there. It's not worth the drama here of looking after them.

> Ah I stuffed that up... I think with Jersey included (JE), `requires_proxy = "GB"` is probably once again needed?

Probably

PR #5691, opened 2023-08-15
winterthur_ch: fix spider, no refs available

davidhicks:
Reference fields have been removed from the source data. There are no more unique identifiers remaining for each location.


iandees:
🤔 I really don't like including records that don't have a ref. I think a relatively stable ID is pretty important for us to generate (probably more important than lat/lon). We have precedent for `no_ref` right now, but I might want to rethink that in the future.

PR #5699, opened 2023-08-16
Add Freshop Store Locator

davidhicks:


davidhicks:
There's heaps of small supermarket chains in America that use this store locator.

Search query `"store locator" "Enter City and State, or Postal Code"` brings up pages of such store locators. Most are very small with 1-15 locations only. The time consuming part is creating Wikidata items if they don't currently exist.

PR #5703, opened 2023-08-19
Add REWE (DE)

### LINKED ISSUE #660: REWE
Grocery store in Germany

https://marktsuche.rewe.de/

davidhicks:
Note: Cloudflare is used.

Fixes #660


davidhicks:
I don't have a system setup for using Playwright. Does anyone want to continue work on this spider and see if the Playwright errors can be fixed, and also complete the `parse_location` method to extract all required information from the JSON response? There should be something like 3000-3500 locations returned.

The easiest approach to this spider would be using SitemapSpider + StructuredDataSpider however this would request HTML pages. Cloudflare would thus be able to apply more sophisticated fingerprinting that is more likely to cause this spider to fail. Therefore the most reliable approach to use would be the JSON API calls this spider proposes to use.

I noted too that `curl` is apparently an acceptable client.

davidhicks:
@iandees This one should now be ready to merge. Works with Zyte.

PR #5704, opened 2023-08-20
Change spider `winterthur_ch`

brawer:
* Update location of upstream data feed
* Import playground reference numbers for emergency calls
* Support newly added address tags

### DIFF
--- a/locations/spiders/winterthur_ch.py
+++ b/locations/spiders/winterthur_ch.py
@@ -15,14 +15,14 @@ class WinterthurCHSpider(scrapy.Spider):
         "attribution:name:en": "City of Winterthur",
         "attribution:wikidata": "Q9125",
         "license": "Creative Commons Zero",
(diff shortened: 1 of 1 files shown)


locations/spiders/winterthur_ch.py
```diff
@@ -42,6 +42,10 @@ def parse_playgrounds(self, response):
             item = {
                 "lat": lat,
                 "lon": lon,
+                "street": props.get("streetName"),
+                "housenumber": props.get("houseNo"),
+                "city": "Winterthur",
+                "country": "CH",
```
davidhicks, line 48: There is middleware that will populate `item["country"]` automatically from the `_ch` in the spider name. Thus no need to specify the country explicitly. Not a big deal, just a style preference that seems to have come into existence.

iandees: I tend to prefer it sitting explicitly in the spider, personally, so this is fine 👍🏻

davidhicks: I prefer it too :)

PR #5710, opened 2023-08-21
sainsburys: use Categories

davidhicks:


davidhicks:
Depends upon update to Categories and Extras in #5716.

davidhicks:
Thanks for the typo fix!

PR #5717, opened 2023-08-21
costacoffee_gb: fix and rename to costa_coffee_gg_gb_im_je

davidhicks:
The store locator has been switched to GraphQL.

Note that the GraphQL query used is a "closest N locations to coordinates X,Y" query where N can be any large number. If N is too large, the GraphQL endpoint times out.

We thus need to split up the GraphQL query search into smaller search areas. A density of 2500 locations per 100km radius seems to return all locations whilst avoiding some duplication. This is not a perfect way to search because the density in Northern Scotland is quite low, and the nearest 2500 locations are still returned. For London with a very high density, a density of 2500 locations per 100km radius seems to be necessary. Due to the inability for this spider to currently control for density of search areas, there is a high number of duplicate item warnings.


davidhicks:
Depends upon removal of function call to costacoffee_gb in #5710, #5711, #5712, #5713, #5714, #5715, #5716.

davidhicks:
Forgot to push #5736 previously which this pull request depends upon.

PR #5718, opened 2023-08-21
storefinders/woosmap: fix pagination search

davidhicks:
An Origin header needs to be sent on every API request, but a previous change had forgotten to send the Origin header on page 2..n requests.

Also do not send a referrer header for page 1 -> page 2 -> page 3 -> etc.


iandees:
For posterity: https://docs.scrapy.org/en/latest/topics/spider-middleware.html#acceptable-values-for-referrer-policy

PR #5737, opened 2023-08-24
Add Visa Spider

### LINKED ISSUE #4550: VISA ATM Locator
### Brand name

Visa

### Wikidata ID

[Q328840](https://www.wikidata.org/wiki/Q328840)

### Store finder url(s)

https://www.visa.com/locator/atm

Cj-Malone:
Fixes #4550

### DIFF
--- a/locations/spiders/visa.py
+++ b/locations/spiders/visa.py
@@ -1,0 +1,82 @@
+import json
+import urllib.parse
+
+from scrapy import Spider
+from scrapy.http import JsonRequest
+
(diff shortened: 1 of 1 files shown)


iandees:
This is an interesting dataset. It looks like they're geocoding (if I had to guess, with HERE) and falling back to a city pin if they can't find the address. There are lots of spots in Africa with dozens of pins with the same location. I think that's probably OK, but I bet it'll cause data consumers to scratch their heads in confusion 🙂

PR #5740, opened 2023-08-24
Fix allowed_domains for spider winterthur_ch

brawer:


brawer:
I’m surprised that #5740 worked just fine, even though I forgot to adjust `allowed_domains` to the new domain in that pull request. Does `allowed_domains` make any difference at all?

iandees:
From [doc](https://docs.scrapy.org/en/latest/topics/spiders.html#scrapy.Spider.allowed_domains):

> An optional list of strings containing domains that this spider is allowed to crawl. Requests for URLs not belonging to the domain names specified in this list (or their subdomains) won’t be followed if [OffsiteMiddleware](https://docs.scrapy.org/en/latest/topics/spider-middleware.html#scrapy.spidermiddlewares.offsite.OffsiteMiddleware) is enabled.

I bet we don't have OffsiteMiddleware enabled, so it probably doesn't matter. We probably should enable it to reduce the chances of scraping too far.

Cj-Malone:
We do have it enabled, I didn't think it triggered on `start_urls`, but maybe skipped on all `start_requests` requests.

davidhicks:
https://github.com/scrapy/scrapy/blob/HEAD/scrapy/spidermiddlewares/offsite.py only implements `from_crawler` (documentation: https://docs.scrapy.org/en/latest/topics/spider-middleware.html#scrapy.spidermiddlewares.SpiderMiddleware.from_crawler) which means `OffsiteMiddleware` only applies to requests caused by the crawling processes of CrawlSpider, SitemapSpider, etc.

If `OffsiteMiddleware` implemented `process_start_requests` then I think any `yield Request(...)` calls from the `start_requests` function would be filtered too by `OffsiteMiddleware`. I'm not sure that a `yield Request(...)` call anywhere else (e.g. in a `parse` function) would ever be filtered by `OffsiteMiddleware` unless you used some more complex Scrapy magic to trigger `OffsiteMiddleware` manually.

PR #5755, opened 2023-08-30
canadiantire: fix and rename to canadian_tire_ca

davidhicks:


davidhicks:
@iandees Should be ready for merge now. Fixed by using Zyte to overcome data centre IP blocking.

PR #5765, opened 2023-08-31
Pmtiles support in weekly run

iandees:
This adds Tippecanoe to the Docker image used in the weekly run and then calls it as part of the weekly run to generate a pmtiles file that will get uploaded as part of the artifacts generated by the build.

This can then by used so @jleedev doesn't have to regenerate the mbtiles with [alltheplaces-serve](https://github.com/jleedev/alltheplaces-serve/tree/main) and we can load them up in [Rapid editor](https://github.com/facebook/Rapid/discussions/993).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,6 +49,15 @@ ENV LANG=en_US.UTF-8 \
     PYENV_ROOT="/home/ubuntu/.pyenv" \
     PATH="/home/ubuntu/.pyenv/versions/${PYTHON_VERSION}/bin:/home/ubuntu/.pyenv/bin:/home/ubuntu/.pyenv/shims:$PATH"
 
+# install tippecanoe
+ARG TIPPECANOE_VERSION=2.29.0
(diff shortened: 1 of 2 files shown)


jleedev:
Nice. The demo map can be made to work by adding the pmtiles maplibre plugin. Also need to enable CORS from the `www` to the `data` subdomain, I think. Range requests require a preflight request, which should be [made cacheable] to avoid adding extra latency.

[made cacheable]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age

Cj-Malone:
https://data.alltheplaces.xyz/runs/latest.json gives you `pmtiles_url`, but I see the value of a `/latest.pmtiles`.

iandees:
Yep I was kinda hoping to do the whole "latest" thing as part of a general revamp away from "run the whole thing weekly" to "run spiders at different intervals and build a collection of the latest data weekly", but doing a simple "latest" in our current setup is quick and easy.

Cj-Malone:
@bhousel please could you share the url with street imagery enabled. Coordinates are the worst part of this dataset.

PR #5789, opened 2023-09-02
bremer_bank_us: Fix address, phone no. parsing

jimbofreedman:
Also: Do we prefer to consistently parse US states to abbreviations?

### DIFF
--- a/locations/spiders/bremer_bank_us.py
+++ b/locations/spiders/bremer_bank_us.py
@@ -1,3 +1,5 @@
+import re
+
 from scrapy.spiders import SitemapSpider
 
 from locations.google_url import extract_google_position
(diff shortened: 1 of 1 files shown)


iandees:
> Do we prefer to consistently parse US states to abbreviations?

If it's easy. We don't make any promises on it at this point.

jimbofreedman:
> > Do we prefer to consistently parse US states to abbreviations?
> 
> If it's easy. We don't make any promises on it at this point.

I just found `StateCodeCleanUpPipeline` - 

> > Do we prefer to consistently parse US states to abbreviations?
> 
> If it's easy. We don't make any promises on it at this point.

Setting the spider's name to `bremer_bank_us` (I didn't notice and only renamed the file triggered auto-adding the country `US` which triggered `StateCodeCleanUpPipeline` to finish the job, all good.

PR #5810, opened 2023-09-06
Update dm websites per country

### LINKED ISSUE #5752: dm - use national sites
Is it possible to use the national sites for dm stores instead of the German one? The format is `dm.[country code]` for all countries except for Bulgaria, Bosnia and Italy (which use `dm-drogeriemarkt.[country code]`) and Slovakia (`mojadm.sk`).

Dimitar5555:
Resolves #5752


Dimitar5555:
This PR is ready for merging.

PR #5811, opened 2023-09-06
Fixed the format for Fantastiko's refs

Dimitar5555:
Currently the refs look like `https://www.fantastico.bg/shops/f{number}` when they should be a simple number. This PR addresses this problem.


Dimitar5555:
It is working as expected now.

iandees:
Please note that since the ID we generate for each spidered place is based on the `ref`, so we try to keep them consistent as much as possible. I agree with you that the number by itself is the right one to use, though.

PR #5819, opened 2023-09-07
Try zyte api

iandees:
This is an experiment to use the [Zyte API](https://docs.zyte.com/zyte-api/get-started.html) to assist with the `requires_proxy=True` spiders.


iandees:
Neat, I think this will be interesting to try.

You can add `ZYTE_TRANSPARENT_MODE: True` in `custom_settings` on a particular spider to use it. I don't want to use it for all spiders because it will be cost prohibitive. I currently have a $5 spending limit on the account, and if we hit that limit their API will [return 403's](https://docs.zyte.com/zyte-api/usage/errors.html#account-suspension).

This will work in both the weekly run and the spider PR runner. Please try not to use it on a spider that makes a lot of requests 😄

jimbofreedman:
Umm we don't have Zyte API, we're still using Smart Proxy. I also think it will be interesting to try!

I looked at it and the Auto Extraction is only for specific types of data, the IDE I was curious about but obv didn't want to write spiders that break compatibility with ATP.

I'm curious to see it's effectiveness at some of the sites that still `429` a lot with Smart Proxy. I'll do some envelope maths  to see what the comparative costs are.

PR #5820, opened 2023-09-07
Add Zyte to Olive Garden

iandees:

PR #5830, opened 2023-09-07
Add middleware proxy to honour requires_proxy

Cj-Malone:
Rel: https://github.com/alltheplaces/alltheplaces/pull/5820#discussion_r1318709864

I actually don't have credentials, so I can't test this. I saw some spiders work on CI, if the secrets are in GitHub, this should work.

### DIFF
--- a/locations/middlewares/zyte_api_by_country.py
+++ b/locations/middlewares/zyte_api_by_country.py
@@ -1,0 +1,50 @@
+import re
+
+from scrapy import Request, Spider
+from scrapy.crawler import Crawler
+
+
+def get_proxy_location(requires_proxy: bool | str, spider_name) -> str | None:
+    """
+    Get the country to proxy from
+    :param requires_proxy: True or country code from the spider
+    :param spider_name:
+    :return: 2-Letter country code
+    """
+    if isinstance(requires_proxy, str):
+        return requires_proxy
+    if m := re.match(r".*_(\w{2})$", spider_name):
+        return m.group(1)
+    return None
+
+
+class ZyteApiByCountryMiddleware:
+    """
+    This middleware directs requests to the Zyte API
+    according to the spider's requires_proxy attribute.
+    """
+
+    zyte_api_automap = None
+
+    def __init__(self, crawler: Crawler):
+        self.crawler = crawler
+
+    @classmethod
+    def from_crawler(cls, crawler: Crawler):
+        return cls(crawler)
+
+    def process_request(self, request: Request, spider: Spider):
+        # Calculate zyte_api_automap on the first request
+        if self.zyte_api_automap is None:
+            if requires_proxy := getattr(spider, "requires_proxy", False):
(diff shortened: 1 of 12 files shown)


iandees:
I was holding off on this because I thought I saw somewhere in the docs that using geolocation adds some cost to the API calls, but I don't see it anywhere. Zyte's pricing info is annoyingly difficult to understand!

Cj-Malone:
I still get spammy emails from Zyte, so they aren't my favourite company.

I could disable the geolocation for now, honouring `requires_proxy` is my primary goal. Otherwise we'd have to change every spider.

Cj-Malone:
CI is now failing because there are more than 5 spiders changed, CI did work in the middle. But as mentioned, I don't have credentials so I haven't tested much locally.

locations/middlewares/zyte_api_by_country.py
```diff
@@ -0,0 +1,50 @@
...
+    def from_crawler(cls, crawler: Crawler):
+        return cls(crawler)
+
+    def process_request(self, request: Request, spider: Spider):
+        # Calculate zyte_api_automap on the first request
+        if self.zyte_api_automap is None:
+            if requires_proxy := getattr(spider, "requires_proxy", False):
+                if cc := get_proxy_location(requires_proxy, spider.name):
+                    # Use the country code set in spider
+                    self.zyte_api_automap = {"geolocation": cc.upper()}
```
Cj-Malone, line 42: If we want to (temporally) disable the geolocation part
```suggestion
                    # TODO: Use the country code set in spider
                    # self.zyte_api_automap = {"geolocation": cc.upper()}
                    self.zyte_api_automap = True
```

Cj-Malone:
Hyatt errored because https://www.hyatt.com/hyatt-centric/melct-hyatt-centric-melbourne doesn't include linked data. It should probably be changed to a StructuredDataSpider, or parse the `__NEXT_DATA__` data.

The log is interesting though, should we be dropping user-agents on these requests?

`2023-09-07 15:41:40 [scrapy_zyte_api._params] WARNING: Request <GET https://www.hyatt.com/en-US/hotel/united-arab-emirates/hyatt-place-dubai-wasl-district/dxbzm> defines header b'User-Agent', which cannot be mapped into the Zyte API customHttpRequestHeaders parameter.`

iandees:
The user-agent thing feels like a bug in the zyte-api/scrapy integration. It shouldn't try to set that header if it's going to ignore it.

iandees:
Yea, it seems like it should be skipping user-agent by default according to [their readme](https://github.com/scrapy-plugins/scrapy-zyte-api).

iandees:
I'm going to merge this as-is. We can fix spiders later. I'm curious to see what the cost will be for a weekly run.

Cj-Malone:
> I'm curious to see what the cost will be for a weekly run.

Let me know.

PR #5833, opened 2023-09-07
shell: extract additional tags

davidhicks:


davidhicks:
CI times out before any results are returned, due to the way in which this spider makes lots of queries before the first location data is obtained.

locations/spiders/shell.py
```diff
@@ -12,14 +22,28 @@ def parse_item(self, item, location):
         amenities = location["amenities"]
         fuels = location["fuels"]
         apply_category(Categories.FUEL_STATION, item)
-        if "shop" in amenities:
+        if "twenty_four_hour" in amenities:
+            item["opening_hours"] = OpeningHours()
+            item["opening_hours"].add_days_range(DAYS, "00:00", "23:59")
+        if "shop" in amenities or "selectshop" in amenities:
```
Cj-Malone, line 28: See https://github.com/alltheplaces/alltheplaces/pull/5767#issue-1875398065

I think Shell Select is big enough to be its own POI. I don't think "shop" is, so I'd prefer it mapped to `shop=yes`.

locations/spiders/shell.py
```diff
@@ -12,14 +22,28 @@ def parse_item(self, item, location):
...
             apply_category(Categories.SHOP_CONVENIENCE, item)
-        apply_yes_no("amenity:chargingstation", item, "electric_charging_other" in fuels or "shell_recharge" in fuels)
-        apply_yes_no("toilets", item, "toilet" in amenities)
-        apply_yes_no("atm", item, "atm" in amenities)
-        apply_yes_no("car_wash", item, "carwash" in amenities)
-        apply_yes_no("hgv", item, "hgv_lane" in amenities)
-        apply_yes_no("wheelchair", item, "disabled_facilities" in amenities)
+        if "restaurant" in amenities:
+            apply_category(Categories.RESTAURANT, item)
```
Cj-Malone, line 31: `food=yes` is an OSM attribute for saying that a place also serves food. It's typically used on pubs, but I think it applies here.

locations/spiders/shell.py
```diff
@@ -12,14 +22,28 @@ def parse_item(self, item, location):
...
+        if "charging" in amenities or "electric_charging_other" in fuels or "shell_recharge" in fuels:
+            apply_category(Categories.CHARGING_STATION, item)
```
Cj-Malone, line 33: `shell_recharge` is big enough to be it's own POI (and has it's own spider). I don't know the about the others.

If we want to collect them lets use `fuel:electric=yes`, it's not used in OSM, but `amenity=charging_station;fuel` breaks compatibility with OSM.

locations/spiders/shell.py
davidhicks: For a roadhouse/roadside service centre with an apron for refueling cars and building that contains a convenience store and separate restaurant area with seating, does this become 3 POIs? And air compressors, car washes, etc become separate POIs too?

Typically we'll only get one coordinate to share for all POIs, but an ATP consumer can see that there are different related POIs all in the same nearby vicinity.

Cj-Malone: From an OSM perspective, yes, a fuel station will eventually get "micro mapped" into several POIs. [eg](https://www.openstreetmap.org/node/9101318090#map=19/-37.75909/144.97475&layers=D)

From an ATP perspective, we don't add any value when splitting one `amenity=fuel` + `compressed_air=yes` into 2 POIs (`amenity=fuel` and `amenity=compressed_air`). But we do add value when we have specific details about them, like different "brands", if we know there is a Shell fuel station, a Budgens convenience store and a Subway takeaway we should output that. [eg](https://www.openstreetmap.org/way/395120804#map=19/51.36058/-1.14869&layers=D)

locations/spiders/shell.py
Dimitar5555: `fuel:electricity` is a more common tag (1.6k vs 12 uses for `fuel:electric`) but it's used for BBQs. Maybe we can create a discussion on the forum and see what other mappers think?

Dimitar5555: Should've checked Overpass before writing the previous reply. Of those 1.6k elements with `fuel:electricity`, ~1.3k are on `amenity=fuel`. I think that it should do the job.

Cj-Malone: I'm good with `fuel:electricity`. I don't do the tagging mailing list, but you are welcome to reach out.

```suggestion
        apply_yes_no("fuel:electricity", item, True)
```

iandees:
@davidhicks how would you like to resolve this? I'm happy to merge it in if we resolve the conflict. Or we can close it if it has been superseded by other PRs.

davidhicks:
We can close this I think as #7632 has already been merged and extracts most of the same additional tags.

PR #5857, opened 2023-09-12
storefinders/store_locator_plus: split into cloud vs. self-hosted

### LINKED ISSUE #5844: Ruler Foods (US) (48 locations)
### Brand name

Ruler Foods

### Wikidata ID

Q17125470

### Store finder url(s)

https://rulerfoods.com/locations/

davidhicks:
StoreLocatorPlus is both a software-as-a-service store finder, as well as an older self-hosted WordPress plugin. This change introduces a new StoreLocatorPlusSelfSpider for self-hosted, and renames the existing StoreLocatorPlusSpider to StoreLocatorPlusCloudSpider.

hemmakvall_se uses the new StoreLocatorPlusSelfSpider.

ruler_foods_us is a new spider that uses StoreLocatorPlusSelfSpider.

national_seating_mobility uses StoreLocatorPlusCloudSpider.

Fixes #5844


locations/storefinders/store_locator_plus_self.py
```diff
@@ -0,0 +1,75 @@
...
+        if url and len(self.searchable_points_files) > 0 and self.search_radius != 0 and self.max_results != 0:
+            for searchable_points_file in self.searchable_points_files:
+                for lat, lon in point_locations(searchable_points_file):
+                    formdata = {
+                        "action": "csl_ajax_onload",
+                        "lat": str(lat),
+                        "lng": str(lon),
+                        "radius": str(self.search_radius),
+                    }
+                    yield FormRequest(url=url, formdata=formdata, method="POST")
```
Cj-Malone, line 60: FYI Wordpress can do url parameters. I think it is slightly better as the logs are more helpful.
https://www.hemmakvall.se/wp-admin/admin-ajax.php?action=csl_ajax_onload

davidhicks: I just tried it out and I don't think this is true for at least StoreLocatorPlusSelf. For example:

`https://rulerfoods.com/wp-admin/admin-ajax.php?action=csl_ajax_onload&lat=42.4255736&lng=-67.8072665&radius=500`

The response to that HTTP GET query shows that StoreLocatorPlus is not able to determine the coordinates or search radius from parameters supplied in the URL.

Cj-Malone: Hmm, maybe some plugins don't do it. TIL.

PR #5891, opened 2023-09-19
Add Pepco Spider

### LINKED ISSUE #5868: Pepco Europe
### Brand name

Pepco

### Wikidata ID

Q11815580

### Store finder url(s)

https://pepco.eu/find-store/

The coordinates can be retrieved from the map's json data (but no other information). For opening hours and store names, HTML parsing is required.

Cj-Malone:
Fixes #5868
cc @Dimitar5555

### DIFF
--- a/locations/spiders/pepco.py
+++ b/locations/spiders/pepco.py
@@ -1,0 +1,25 @@
+import json
+
+from scrapy import Spider
+
+from locations.items import Feature
+
+
+class PepcoSpider(Spider):
+    name = "pepco"
+    item_attributes = {"brand": "Pepco", "brand_wikidata": "Q11815580"}
+    start_urls = ["https://pepco.eu/find-store/"]
+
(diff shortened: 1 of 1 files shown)


Dimitar5555:
Thanks @Cj-Malone! The output looks good.

> Only the first 50 load in the HTML, I can't get more to load, even with the "show more" button in a browser.

It looks like the rest are loaded using a POST request to `https://pepco.eu/wp-admin/admin-ajax.php`.The parameters are 
`action=get_more_shops&page=1&search=`. Page is the current page, search is empty. It *should* return an array of `shops`, `page` (or the returned page) and `last_page` (boolean). The shops parameter seems to be empty for some reason..

Cj-Malone:
Yeah, I'm just getting an empty string. It's a shame, but we can't get the extra data until they fix there site.

https://pepco.eu/wp-admin/admin-ajax.php?action=get_more_shops

PR #5892, opened 2023-09-20
Add GB train station spiders

Cj-Malone:


iandees:
Looks like one of the scotrail locations doesn't have a name? I'll merge these in if you want to fix in another PR.

Cj-Malone:
I've got a feeling something went weird with the proxy there. It does have a name on the [page](https://www.scotrail.co.uk/plan-your-journey/stations-and-facilities/sum) and it ran to completion here without any errors.

Issue #4772, opened 2023-02-10
Brands and store formats


Cj-Malone:
I was going to write this in the Gap PR #4765, but I forgot.

I think at some stage we need to discuss some of the rough edges with tagging differences OSM (and NSI). This doesn't really apply to the 99% of brands that are just 1 brand, and 1 store format, it doesn't even apply to brands that have one brand but crosses OSM categories, eg convenience vs supermarket, this is about the edge cases. I don't think there is any action to be taken right now, just food for thought.

ATP
`brand`, typically broken down "Gap Factory Store" vs "Gap", "Tesco Extra" vs "Tesco Express"
`name`, often brand + locality

NSI
`brand` usually generic "Gap" or "Costco"
`name` would be the store type "Gap Factory" in this example

OSM
`brand` almost exclusively comes from NSI
`name` used to be the fascia, sometimes brand + locality, but really just follows NSI now
`branch` the "new" place to put the locality/store name

Making changes to any of the above will be difficult, and painful, but it needs to be done at some stage.

- I think NSI should embrace that `brand` should be fine grain. They are often similar, and under the same company but they are different. "Costco Business Center" is different to "Costco Wholesale" they have different logos, soon enough they will have different Wikidata entries.
- I think ATP needs to embrace `branch`, our `name` is a mess.
- ATP needs to think about `operator`

![Screenshot 2023-02-10 at 11-56-01 Name Suggestion Index - brands_shop_clothes](https://user-images.githubusercontent.com/10121513/218086798-96678d76-7efa-4a9c-9228-d1cdc743f7dd.png)
(Also I see no evidence in our data the Gap Kids/Baby Gap/Gap Body are really a thing other than branded departments, but that's a NSI discussion)

rjw62:
As far as OSM is concerned, I think the name=* should normally be the primary wording on the main sign at the shop. Many chains do indeed just use their brand name for this. But a number do use other location-based or other names. (Convenience store symbol groups are a good example of the latter. Many "Premier" branded convenience stores will actually be called something quite different.) I think NSI is often a bit heavy handed in enforcing brand values in the name=* tag.

As for ATP, I'm not sure what the best plan is. It would probably be useful for data users to have a bit more consistency in the fields provided. There is usually some sort of identifying location/branch name applied to each store. But, for instance, should this include the brand name or not? Does ATP want to try to capture the "name" that would be suitable for OSM as well?

brawer:
[...]

PR #5899, opened 2023-09-21
Clean up name and branch

Cj-Malone:
I've mentioned this before. But I'd like to get closer to the OSM `name` and `branch`. Instead of our current thing.

This takes the NSI names and `strip_names` supplied by a spider and removes them from a `name` and sets the remainder to `branch`.

I still want to run through a weekly and get a reasonable set of `strip_names` before this is merged. I'm opening this now because I've been ignoring it for too long.


iandees:
Yea, this makes sense 👍🏻

Cj-Malone:
Rel: #4772

Cj-Malone:
So, I don't think doing this automatically from all the tags (`name`, `brand`, `operator`) is viable. It will work most of the time, but we'll end up with edge cases. My latest idea is just `name` from NSI, or `strip_names` is spider.

iandees:
I'm going to close this as stale. Please feel free to reopen if we want to continue this line of inquiry.

PR #5911, opened 2023-09-23
Add Orlen Paczka (PL) (9682 locations)

### LINKED ISSUE #5898: Orlen amenity=parcel_locker (9699 POIs)
### Brand name

Orlen

### Wikidata ID

Q110457879

### Store finder url(s)

https://www.orlenpaczka.pl/pokaz-punkty/

Note that Orlen fuel station spider exists already

davidhicks:
Fixes #5898


matkoniecz:
tags for parcel lockers are missing, is it intentional?

davidhicks:
Thanks, fixed. I should have checked NSI to see this brand is across multiple categories, so we need to set the category as a hint to which NSI tags to apply.

locations/spiders/orlen_paczka_pl.py
```diff
@@ -0,0 +1,34 @@
...
+
+class OrlenPaczkaPLSpider(Spider):
+    name = "orlen_paczka_pl"
+    item_attributes = {"brand": "Orlen Paczka", "brand_wikidata": "Q110457879"}
+    start_urls = ["https://ruch-osm.sysadvisors.pl/api.php"]
+    allowed_domains = ["ruch-osm.sysadvisors.pl"]
+
+    def start_requests(self):
+        for url in self.start_urls:
+            yield Request(
```
Cj-Malone, line 16: I'm on my phone, so can't confirm. But FormRequest should probably be used here. It adds the headers automatically, and converts the data and does post by default.

The following should work:
```suggestion
yield FormRequest(url=url, formdata={"a":"f","s":0})

matkoniecz:
> this brand is across multiple categories

Weird, I would expect to have it matching only to parcel lockers

locations/spiders/orlen_paczka_pl.py
davidhicks: That's what I thought too but it doesn't work because FormRequest doesn't want to convert integer 0. From vague memory, Scrapy doesn't like URL encoded form data?

davidhicks:
> > this brand is across multiple categories
> 
> Weird, I would expect to have it matching only to parcel lockers

It's also in `amenity/post_office` per `https://nsi.guide/index.html?t=brands&k=amenity&v=post_office&tt=Orlen+Paczka`

locations/spiders/orlen_paczka_pl.py
Cj-Malone: Probably wants a string, "0"

davidhicks: > Probably wants a string, "0"

The problem then is that the server ends up receiving a body of:

`{"a":"f","s":"0"}`

Instead of:

`{"a":"f","s":0}`

Treating the 0 as a string causes the server to return no results.

Cj-Malone: Ahh I wasn't paying attention, this isn't actually a Form request, the header is a lie. It's a Json POST request.

`yield JsonRequest(url=url, data={"a": "f", "s": 0})`

The special classes, `JsonRequest` and `FormRequest` do 2 main things.

1. They automatically set the headers, which I think is polite but often not required (or there are times when sites require other or even wrong herders).
2. Automatically populate the body. In the case of Json it's probably just json.dumps, but it's nicer to see and alter a dict than a string containing json.

They also automatically change them to POST requests when there is `data` or `formdata`.

PR #5921, opened 2023-09-23
Add Lilly (RS)

### LINKED ISSUE #5860: Lilly Drogeries
### Brand name

Lilly

### Wikidata ID

Q111764460

### Store finder url(s)

Greece: https://lilidrogerie.gr/wp-json/v1/get-store/?lang=el
Serbia: https://www.lilly.rs/phpsqlsearch_genjson.php?lat=44.8019&lng=20.4671&radius=600&limit=200&keyword=none
Bulgaria: https://www.lillydrogerie.bg/bg/stores (look for `var locations = ` in the page source code, requires scraping HTML data in order to get opening hours)

It may be a good idea to have a separate spider for each country since all of them use diffrent systems and we don't know when (and if) they will standartise it.

davidhicks:
Fixes #5860

### DIFF
--- a/locations/hours.py
+++ b/locations/hours.py
@@ -378,6 +378,34 @@ DAYS_RO = {
     "Duminică": "Su",
 }
 
+DAYS_SR = {
+    "Pon": "Mo",
+    "Ponedjeljak": "Mo",
+    "Ponediljak": "Mo",
+    "Понедељак": "Mo",
+    "Понеделник": "Mo",
+    "Uto": "Tu",
+    "Utorak": "Tu",
+    "Уторак": "Tu",
+    "Sri": "We",
+    "Sreda": "We",
+    "Среда": "We",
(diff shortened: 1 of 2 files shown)


locations/hours.py
```diff
@@ -378,6 +378,34 @@
     "Duminică": "Su",
 }
 
+DAYS_SR = {
```
iandees, line 381: Looks like this is Croatian? If so, can you make this `DAYS_HR`?

davidhicks: I think it's correct as `SR` because locations are in Serbia where the dominant language is ISO-639-1 language code `SR`, not Croatia with the dominant language of ISO-639-1 language code `HR`?

iandees: Yea, makes sense. I was trusting Google Translate too much.

PR #5936, opened 2023-09-25
[Rompetrol] Namespace ref

### LINKED ISSUE #5932: Rompetrol - missing stations
There are about 80 Romanian stations and about 30 Bulgarian stations which are missing. It seems that the problem is caused by duplicated `ref`s on their end (the databases seem to be independent so each site has station with `ref=1`, `ref=2` and so on). Is it possible to resolve that issue? Possibly without having 4 almost identical spiders?

Cj-Malone:
Fixes #5932

### DIFF
--- a/locations/spiders/rompetrol.py
+++ b/locations/spiders/rompetrol.py
@@ -34,6 +34,8 @@ class RompetrolSpider(Spider):
             elif "www.rompetrol.ro" in response.url:
                 item["country"] = "RO"
 
+            item["ref"] = "{}-{}".format(item["country"], item["ref"])
+
             apply_yes_no(Fuel.OCTANE_98, item, "27" in location["services"], False)
             apply_yes_no(Fuel.OCTANE_95, item, "28" in location["services"], False)
             apply_yes_no(Fuel.GTL_DIESEL, item, "30" in location["services"], False)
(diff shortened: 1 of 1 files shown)


Dimitar5555:
Thank you @Cj-Malone. Unfortunately it looks like it's a bigger mess than I first thought. The Romanian site is returning stations in Bulgaria as well which means that that they are on the right coordinates but with the wrong `addr:country` tag. I haven't checked if the data is duplicated or not. Is it possible to fix it somehow?

Cj-Malone:
It does seem like every BG location is in the RO data. Hows this look?

Dimitar5555:
I don't see any missing pins when comparing the map here and the map on the Bulgarian website.

locations/spiders/rompetrol.py
```diff
@@ -8,50 +8,38 @@ class RompetrolSpider(Spider):
...
-            if "14" in location["services"]:
-                apply_category(Categories.CHARGING_STATION, item)
-            if "33" in location["services"]:
-                apply_category(Categories.SHOP_LAUNDRY, item)
-            if "20" in location["services"]:
-                apply_category(Categories.COMPRESSED_AIR, item)
-            if "17" in location["services"]:
-                apply_category(Categories.RESTAURANT, item)
+
+            apply_yes_no("fuel:electric", item, "14" in location["services"])
```
Dimitar5555, line 40: The number for EV chargers is 34. 14 is for AdBlue.

Cj-Malone: Thanks.

PR #5937, opened 2023-09-25
[CanadianTireCA] Skip fetching html page

Cj-Malone:


Cj-Malone:
We weren't getting anything from the html, so it was just a waste. Now we have half the requests it'll be faster, and hopefully get us blocked less.

davidhicks (approved):
Thanks!

PR #5949, opened 2023-09-26
[RalphLauren] Fix spider

Cj-Malone:
So I need a proxy for this as well as slowing it down, but that's because I get redirected to the UK site. I think it will work without a proxy from US ips.


iandees:
Can you fix the merge conflict?

iandees:
Thank you

PR #5957, opened 2023-09-27
Fix mcdonalds_be:  add category fast food

jhutton1:


jhutton1:
already fixed PR 5944

PR #5966, opened 2023-09-28
Rename a1.py to a1_bg.py

Dimitar5555:


iandees:
Can you also rename the class and the spider contained in this file?

Dimitar5555:
I tought that ve missed something but I wasn't sure what. It's now fixed.

PR #5973, opened 2023-09-28
[cvs_us] add category + more brands + log failed hours.

panasonic-mafia:


locations/spiders/cvs_us.py
```diff
@@ -46,22 +62,27 @@ def post_process_item(self, item, response, ld_data, **kwargs):
         yield item
 
     def parse_hours(self, hours):
-        opening_hours = OpeningHours()
-        for row in hours:
-            if row["startTime"] == "Open 24 Hours":
-                row["startTime"] = "12:00 AM"
-            if row["endTime"] == "Open 24 Hours":
-                row["endTime"] = "11:59 PM"
+        try:
```
iandees, line 65: Instead of putting the try/except inside the function, can you put it around the call to `self.parse_hours()` up above?

panasonic-mafia: Thanks for suggestion. The code is a bit cleaner now)

PR #5978, opened 2023-09-28
Added Cateogries based on Facility Type Mercy Health

ckpiper:


locations/spiders/mercyhealth.py
```diff
@@ -41,4 +54,7 @@ def parse(self, response):
                 "website": store_data["Location"]["Link"],
             }
 
-            yield Feature(**properties)
+            item = Feature(**properties)
+            facility_type = store_data.get("Location", {}).get("FacilityType", {}).get("SchemaPlaceType", "")
+            apply_category(AMENITIES[facility_type], item)
+            yield item
```
iandees, line 60: I believe the `apply_category()` function will work on a plain dictionary like `properties`, so this can be rearranged to:

```suggestion
            facility_type = store_data.get("Location", {}).get("FacilityType", {}).get("SchemaPlaceType", "")
            apply_category(AMENITIES[facility_type], properties)
            yield Feature(**properties)
```

iandees: (For future reference. It's fine as-is.)

PR #5994, opened 2023-09-30
Add Doctors Category Tag for United Surgical Partners Brand

ckpiper:


iandees:
Are you adding these categories because the Name Suggestion Index data isn't adding them for us?

ckpiper:
> Are you adding these categories because the Name Suggestion Index data isn't adding them for us?

A lot of them are not in NSI as they are smaller brands. Should I add the brands to NSI instead, of adding the category here?

iandees:
I'm ok with adding category here. We should make sure the data in NSI is correct/exists where it makes sense though. If we think smaller brands like this don't need to be in there, that's OK.

PR #5995, opened 2023-09-30
Add McGrath (AU) (127 locations)

davidhicks:


iandees:
There's some stray locations with lat/lons up near Japan. Do you see an easy way to filter those out?

davidhicks:
Some latitudes are incorrectly signed (should be e.g. -33 not 33). Hopefully this patch fixes it now :)

iandees:
Great, thanks!

PR #5997, opened 2023-09-30
Add First National Real Estate (AU) (225 locations)

davidhicks:


iandees:
Can we add an office tag to these? Maybe https://wiki.openstreetmap.org/wiki/Tag:office%3Destate_agent

davidhicks:
NSI brand proposed per: https://github.com/osmlab/name-suggestion-index/pull/8715

Issue #5864, opened 2023-09-14
DSK Bank


Dimitar5555:
### Brand name

DSK Bank / Банка ДСК

### Wikidata ID

Q5206146

### Store finder url(s)

https://dskbank.bg/контакти/клонова-мрежа/GetOffices/

The type of office is stored in the `BranchType` variable. Some of the offices have lunch breaks which are denoted by `*Почивка`.

Exterminator11:
hey @Dimitar5555, is this issue up for grabs?

Dimitar5555:
Hi @Exterminator11, yes the issue is up for grabs.

Exterminator11:
I am a novice contributor, so if you could explain what the issue expects to be done it would be very helpful.

Dimitar5555:
The goal is to create a spider that gets the coordinates, opening hours, addresses and other information that is in the json file and convert it to All The Places data format. I think that the spider for DSK would be similar to A1 (https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/a1.py). There is more information in [README.md](https://github.com/alltheplaces/alltheplaces/blob/master/README.md) and [DATA_FORMAT.md](https://github.com/alltheplaces/alltheplaces/blob/master/DATA_FORMAT.md).

Exterminator11:
Thank you so much, I'll get to work right away!

davidhicks:
This would be a fairly simple spider, mostly just:

`
start_urls = ["https://dskbank.bg/%D0%BA%D0%BE%D0%BD%D1%82%D0%B0%D0%BA%D1%82%D0%B8/%D0%BA%D0%BB%D0%BE%D0%BD%D0%BE%D0%B2%D0%B0-%D0%BC%D1%80%D0%B5%D0%B6%D0%B0/GetOffices/"]

def parse(self, response):
  for location in response.json():
    item = DictParser.parse(location)
    yield item
`

Then some extra tweaks to extract additional fields from `location` to `item`, and to clean up any messy fields.

Exterminator11:
hey @davidhicks and @Dimitar5555, I was able to come up with a solution would you like me to open up a PR?. This is what the output looks like when I run the file.
<img width="1438" alt="Screenshot 2023-10-01 at 8 40 25 AM" src="https://github.com/alltheplaces/alltheplaces/assets/88940647/5091bba6-559b-4c55-91d4-ecd5d257ac82">

davidhicks:
[...]

PR #6006, opened 2023-10-01
Added spider for DSKBank

Exterminator11:
**Description**:
Added spider for DSKBank

**Changes Made:**
created a DskBank.py file 

**Screenshots:**
Screenshot 2023-10-01 at 11 28 58 AM
Screenshot 2023-10-01 at 11 31 19 AM

**Related issues:**
#5864 : DSK Bank

**Additional notes:**
I'm open to feedback and further improvements. Please let me know if any changes are needed.


locations/spiders/DskBank.py
```diff
@@ -0,0 +1,67 @@
...
+                apply_category(Categories.BANK, item)
+            else:
+                apply_category(Categories.ATM, item)
+
+            text = data["Text"]
+            clean_text = re.sub(r"<.*?>", "\n", text)
+            clean_text = re.sub("\n+", "\n", clean_text)
+            clean_text = clean_text.strip("\n").split("\n")
+
+            item["addr_full"] = clean_text[0]
```
Cj-Malone, line 37: I don't know anything about Bulgarian addresses, but this is probably `street_address`, not `addr_full`.

locations/spiders/DskBank.py
```diff
@@ -0,0 +1,67 @@
...
+class DSKBankBGSpider(scrapy.Spider):
+    name = "dsk_bank_bg"
+    item_attributes = {"brand": "Банка ДСК", "brand_wikidata": "Q5206146"}
+    allowed_domains = ["https://www.dskbank.bg/"]
+    start_urls = ["https://dskbank.bg/контакти/клонова-мрежа/GetOffices/"]
+
+    def parse(self, response):
+        item = Feature()
+        response_json = response.json()
+        for data in response_json:
```
Cj-Malone, line 19:
```suggestion
        for data in response.json():
            item = Feature()
```

locations/spiders/DskBank.py
Exterminator11: Thank you for pointing that out. I apologize for the confusion. I'll check this out and get back.

locations/spiders/DskBank.py
```diff
@@ -0,0 +1,66 @@
...
+    start_urls = ["https://dskbank.bg/контакти/клонова-мрежа/GetOffices/"]
+
+    def parse(self, response):
+        for data in response.json():
+            item = Feature()
+            item["ref"] = data["Id"]
+            item["name"] = data["Name"]
+            item["lat"] = data["Latitude"]
+            item["lon"] = data["Longitude"]
+            item["email"] = data["Email"]
```
davidhicks, line 23: I think you could replace all of this with `item = DictParser.parse(data)` after adding an import `from locations.dict_parser import DictParser`

locations/spiders/DskBank.py
```diff
@@ -0,0 +1,66 @@
...
+                        if day in DAYS_BG:
+                            hour_from = re.sub(r"\.", ":", hour_from)
+                            hour_to = re.sub(r"\.", ":", hour_to)
+                            hour_to += ":00"
+                            hour_from += ":00"
+                            item["opening_hours"].add_range(sanitise_day(day, DAYS_BG), hour_from, hour_to, "%H:%M:%S")
+                    else:
+                        pass
+            else:
+                item["opening_hours"] = None
```
davidhicks, line 64: I think you could replace all of this with:

```
from locations.hours import DAYS_BG, OpeningHours
item["opening_hours"].add_ranges_from_string(data["OpenHours"], days=DAYS_BG)
```

locations/spiders/DskBank.py
```diff
@@ -0,0 +1,66 @@
...
+            clean_text = re.sub(r"<.*?>", "\n", text)
+            clean_text = re.sub("\n+", "\n", clean_text)
+            clean_text = clean_text.strip("\n").split("\n")
+
+            item["addr_full"] = clean_text[0]
+
+            phone = data["Phone"]
+            if phone:
+                phone = re.sub(r"\(|\*|\|\);", "", phone)
+            item["phone"] = phone
```
davidhicks, line 41: Phone numbers get cleaned up automatically by Scrapy middleware, and are formatted according to the particular country the phone number belongs within.

I think all you need here is `item["phone"] = data["Phone"].split(";", 1)[0]` ?

locations/spiders/DskBank.py
davidhicks: I believe this could/should be replaced with: `item["addr_full"] = data["Address"]` and that `data["Text"]` (HTML data) does not need parsing. In the future, if you do need to parse HTML, use this approach:

```
from scrapy import Selector
item["addr_full"] = Selector(text=data["Text"]).xpath('//p[1]/span/text()').get()
```

davidhicks (changes requested):

Exterminator11:
As I am an open source newbie I might have overlooked better methods to solve the problem but I will make the necessary changes by the earliest.

locations/spiders/DskBank.py
Exterminator11: when I was going through the data I found out that sometimes `data["Address"]` contains the full address and sometimes the HTML text. So, would you like me to select the best one possible ? and If both of them contain only the street address would it be better to add the city name to make it a complete address?

davidhicks: Have you got an example? I can only see text (no HTML) in the "Address" and "City" fields in source JSON data.

davidhicks: If you run DictParser over the source data like this:

```
for location in response.json():
  yield location
```

You'll also probably need to add after the DictParser.parse call:
```
  item["street_address"] = item.pop("addr_full", None)
```

This is because DictParser by default will see an "Address" field and assume it is a full address with city, postcode, etc rather than just the street number and street name concatenated.

We also only extract fields available for addresses, and generally don't try to extract postcodes from full addresses, city names from full addresses, street numbers, etc. It is essentially up to ATP consumers to clean up the source address data, likely requiring access to non-free data sets such as government/commercial street address data sets.

Exterminator11: This is one piece of data I found
`{
  "Id": "d7b892b1-4d38-650e-9d59-ff0900048325",
  "ServiceId": "D0022402",
  "Name": "Албена",
  "Location": "43.3661586-28.0797264",
  "Address": "к.к. Албена",
  "CityId": "1991090806",
  "Longitude": 28.0797264,
  "Latitude": 43.3661586,
  "Text": "<p><i class=\"i loc\"></i><span>к.к. Албена, Хотел \"Laguna Garden\"</span></p>",
  "BranchType": [
    "Atm"
  ],
  "Services": [],
  "Types": null,
  "OpenHours": null,
  "VaultOpenHours": null,
  "City": "Албена, обл. Добрич",
  "RegionSfId": "59cb52b1-4d38-650e-9d59-ff0900048325",
  "CitySfId": "3bb392b1-4d38-650e-9d59-ff0900048325",
  "Signature": null,
  "Email": null,
  "Phone": null,
  "CityFriendlyName": "Албена"
}`
In this context, the `Address` field displays a portion of the complete address, specifically `к.к. Албена`. However, after applying the cleaning process using `Selector(text=data["Text"]).xpath("//p[1]/span/text()").get()`, the resulting content in the `Text` field includes the complete address, `к.к. Албена, Хотел Laguna Garden`, which, in my opinion, should be considered as the accurate address. I may be wrong because I am not familiar with how addresses look in Bulgaria.

locations/spiders/DskBank.py
```diff
@@ -0,0 +1,39 @@
...
+                item["phone"] = data["Phone"].split(";", 1)[0]
+            else:
+                item["phone"] = None
+
+            item["opening_hours"] = OpeningHours()
+
+            if data["OpenHours"]:
+                item["opening_hours"].add_ranges_from_string(data["OpenHours"], days=DAYS_BG)
+            else:
+                item["opening_hours"] = None
```
davidhicks, line 37: These two lines can be omitted. It's fine to return a blank `OpeningHours()` object.

davidhicks (changes requested):
Looking good I just have a few minor additional comments :)

locations/spiders/DskBank.py
```diff
@@ -0,0 +1,39 @@
...
+    item_attributes = {"brand": "Банка ДСК", "brand_wikidata": "Q5206146"}
+    allowed_domains = ["https://www.dskbank.bg/"]
+    start_urls = ["https://dskbank.bg/контакти/клонова-мрежа/GetOffices/"]
+
+    def parse(self, response):
+        for data in response.json():
+            item = DictParser.parse(data)
+
+            branch_type = data["BranchType"]
+            if "Branch" in branch_type:
```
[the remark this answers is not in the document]
Exterminator11: Sure

locations/spiders/DskBank.py
```diff
@@ -0,0 +1,39 @@
...
+                apply_category(Categories.BANK, item)
+            else:
+                apply_category(Categories.ATM, item)
+
+            if data["Phone"]:
+                item["phone"] = data["Phone"].split(";", 1)[0]
+            else:
+                item["phone"] = None
```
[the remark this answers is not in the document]
Exterminator11: alright!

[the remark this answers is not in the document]
davidhicks: > Alright! But this should raise an `Attribute error` because when `data["Phone"]` is None the `spilt` function cannot be applied to it.

The second example `item["phone"] = data.get("Phone", "").split(";", 1)[0]` would need to be used then if data["Phone"] is sometimes `None`. Basically `item["phone"] = ""` (empty string) and a Scrapy middleware will ignore the field because it isn't a valid phone number.

davidhicks (approved):

PR #6008, opened 2023-10-01
Add BeallsFlorida spider

### LINKED ISSUE #896: Bealls
https://stores.beallsflorida.com/search.html

tonyskapunk:
Fixes #896

### DIFF
--- a/locations/spiders/bealls_florida.py
+++ b/locations/spiders/bealls_florida.py
@@ -1,6 +1,7 @@
-from re import search
 from json import loads
+from re import search
 from urllib.parse import unquote
+
 from scrapy.spiders import SitemapSpider
 
 from locations.hours import OpeningHours
--- a/locations/spiders/bealls_florida.py
+++ b/locations/spiders/bealls_florida.py
@@ -1,0 +1,46 @@
+from re import search
+from json import loads
+from urllib.parse import unquote
+from scrapy.spiders import SitemapSpider
+
+from locations.hours import OpeningHours
+from locations.structured_data_spider import StructuredDataSpider
+
+
+class BeallsFlorida(SitemapSpider, StructuredDataSpider):
+    name = "bealls_florida"
+    item_attributes = {"brand": "Bealls", "brand_wikidata": "Q4876153"}
+    allowed_domains = ["stores.beallsflorida.com"]
+    sitemap_urls = ["https://stores.beallsflorida.com/sitemap/sitemap_index.xml"]
+    sitemap_rules = [(r"stores\.beallsflorida\.com/fl/\w+/$", "parse_sd")]
+
+    def post_process_item(self, item, response, ld_data, **kwargs):
+        item.pop("image")
+        item.pop("facebook")
+        item["email"] = unquote(item.pop("email")).strip()
+
+        if m := search(
+            r"map_list_data\s*=\s*(.*)", response.xpath('//script[@type="text/javascript"]/text()').extract_first()
+        ):
+            d = loads(m.group(1))[0]
+
+            item["city"] = d.get("city")
+            item["lat"] = d.get("lat")
+            item["lon"] = d.get("lng")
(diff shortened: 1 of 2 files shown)


iandees:
Thanks for adding this as a sitemap spider. Did you try the single API call that was suggested in the original issue? It'd be nicer if we could do it in one call.

locations/spiders/bealls_florida.py
```diff
@@ -0,0 +1,46 @@
...
+            r"map_list_data\s*=\s*(.*)", response.xpath('//script[@type="text/javascript"]/text()').extract_first()
+        ):
+            d = loads(m.group(1))[0]
+
+            item["city"] = d.get("city")
+            item["lat"] = d.get("lat")
+            item["lon"] = d.get("lng")
+            item["name"] = d.get("location_name")
+            item["phone"] = d.get("local_phone")
+            item["postcode"] = d.get("post_code")
```
davidhicks, line 33: For a future spider, the following is preferred:

```
item = DictParser.parse(d)
```

DictParser automatically searches a dictionary to find fields relevant to extract (e.g. field named "postcode").

It's not perfect and typically a handful of adjustments need to be made manually after DictParser has completed parsing.

locations/spiders/bealls_florida.py
```diff
@@ -0,0 +1,46 @@
...
+            item["street_address"] = d.get("address_1") + d.get("address_2")
+
+            oh = OpeningHours()
```
davidhicks, line 36: `item["opening_hours"] = OpeningHours()` is preferred. It's OK to return a blank/null OpeningHours() object should the for loop not call `add_range`.

locations/spiders/bealls_florida.py
```diff
@@ -0,0 +1,46 @@
...
+from json import loads
+from urllib.parse import unquote
+from scrapy.spiders import SitemapSpider
+
+from locations.hours import OpeningHours
+from locations.structured_data_spider import StructuredDataSpider
+
+
+class BeallsFlorida(SitemapSpider, StructuredDataSpider):
+    name = "bealls_florida"
```
davidhicks, line 12: If this is a US brand only, the class name should be `BeallsFloridaUSSpider` and `name = "bealls_florida_us"`

davidhicks (changes requested):

tonyskapunk:
> Thanks for adding this as a sitemap spider. Did you try the single API call that was suggested in the original issue? It'd be nicer if we could do it in one call.

Hi @iandees, sorry for the delayed response. I looked at it initially but didn't see the hours, but after reviewing it I see hours are included in an xml, I guess I could give it a try that way and see how it goes.

locations/spiders/bealls_florida.py
tonyskapunk: Thanks I'll take a look to it and adjust as needed

locations/spiders/bealls_florida.py
tonyskapunk: I looked to other spiders and see the name also includes the country code, I'll rename this one as well. thanks!

locations/spiders/bealls_florida.py
tonyskapunk: I gave it a try, and I don't feel is worth the effort to extract the info and decode multiple times the data they provide (xml in a json, with other json embedded) since the other option is not taking that long and there aren't so many stores I feel is not that bad the current way, but let me know if you feel is not the case.

tonyskapunk:
> I guess I could give it a try that way and see how it goes.

I gave it a try, and I don't feel is worth the effort to extract the info and decode multiple times the data they provide (xml in a json, with other json embedded) since the other option is not taking that long and there aren't so many stores I feel is not that bad the current way, but let me know if you feel is not the case.

davidhicks (approved):
Looks good thanks for the changes!

PR #6023, opened 2023-10-03
[PigglyWigglyUS] Collect more locations

Cj-Malone:


locations/spiders/piggly_wiggly_us.py
```diff
@@ -1,11 +1,25 @@
...
+
+class PigglyWigglyUSSpider(SitemapSpider):
     name = "piggly_wiggly_us"
     item_attributes = {"brand": "Piggly Wiggly", "brand_wikidata": "Q3388303"}
-    app_key = "piggly_wiggly_nc"
+    sitemap_urls = ["https://www.pigglywiggly.com/wp-sitemap-posts-page-1.xml"]
+    sitemap_rules = [(r"/store-locations/.+/", "parse")]
+
+    def parse(self, response, **kwargs):
+        locations = json.loads(re.search(r"locations = (\[?{.+}\]?);", response.text).group(1))
```
davidhicks, line 16: `.+` would be greedy and return beyond the end of that dictionary, possibly into later dictionaries in the same JavaScript blob?

Would the following work instead:
```
from chompjs import parse_js_object

locations = parse_js_object(response.xpath('//script[contains(text(), "var locations = ")]/text()').get().split("var locations = ", 1)[1].split("\n", 1)[0])
```

locations/spiders/piggly_wiggly_us.py
```diff
@@ -1,11 +1,25 @@
...
+        if isinstance(locations, dict):
+            locations = locations.values()
 
-    def parse_item(self, item, location):
-        item["phone"] = item["phone"].split("|", 1)[0].split("\n", 1)[0].strip()
-        yield item
+        for location in locations:
+            item = DictParser.parse(location)
+            item["ref"] = location["storeNumber"]
+            item["website"] = item.pop("email")
```
iandees, line 23: It looks like they have store-specific URLs available e.g. https://www.shopthepig.com/stores/sister-bay-95/5431. Can you try to include those?

locations/spiders/piggly_wiggly_us.py
Cj-Malone: The current regex wont go beyond the line, and if they minimised it in future it would break on the first bit

```patch
-locations = (
+locations=(
```

I try to avoid chompjs as much as possible now, it's broken too many times, it's our only option for javascript. But for json, I skip it as much as possible.

locations/spiders/piggly_wiggly_us.py
Cj-Malone: I don't think they were available, a couple had them in the name. I can check again when I get home, or tomorrow.

iandees: Looks like there are multiple companies operating stores under the "Piggly Wiggly" brand. https://www.thepig.net/store-locations/ does some, https://www.shopthepig.com/ does Wisconsin, etc. They all seem to have their own store locators, so it's probably ok to keep what we have for now.

locations/spiders/piggly_wiggly_us.py
davidhicks: I think the issue would be if a string in the JSON blob contained `];` but as you said, this would cause the JSON blob to stop being parsed half way and we'd see the error as 0 or a smaller number of locations returned for this spider than usual.

PR #6031, opened 2023-10-04
Added Category Tag for Aramark Clothes Brand

ckpiper:
Amarak Uniform Services US is not in NSI and I am going to add Amarak but it is a Food service, facilities, and uniform services. Thus the category I don't think will be shop=food. I think it might have to be shop=general.


locations/spiders/aramark_uniform_services_us.py
```diff
@@ -65,4 +66,7 @@ def parse(self, response):
             elif "Cleanroom Services" in item["name"].title():
                 item["brand"] = "Aramark Cleanroom Services"
                 item["name"] = item["brand"] + " " + item["city"]
+
+            apply_category(Categories.SHOP_CLOTHES, item)
+            apply_category({"rental": "clothes"}, item)
```
davidhicks, line 71: I think categories should be applied in the `if/elif` statements above? If it's `Uniform Services` or `Cleanroom Services`, then `shop=laundry` (with other tags to denote it's a private laundry that customers can't just walk into) may be most suitable? For `Cleanroom Services`, I see that they advertise themselves as supplying and maintaining reusable cleanroom garments for businesses, hence I think these locations are primarily where they clean garments, rather than a shopfront where customers can walk in to buy garments, or where they warehouse new garments for distribution to customers.

davidhicks (changes requested):

locations/spiders/aramark_uniform_services_us.py
ckpiper: Looking more deeply into it they do offer laundry services as well as uniform rentals. Okay so uniform services will be clothe rentals and cleanroom will be laundry. I can get behind that.

Also Looks like Aramack changed its name to Vestis last month.

davidhicks: Thanks for investigating. I see now why it's difficult to decide between `rental:clothes` and `shop:laundry` as the difference is seemingly only who owns the garments, and these type of industrial laundry services would probably allow for both (for example, a hotel that has their own branded linen, or another hotel that just uses plain linen mixed with other hotels).

We can rename this spider as needed.

ckpiper: Does that mean you are okay with what I added below? Having Cleanroom Services being laundry and the general being rental clothes?

Looking into it more. 
> today announced that Vestis™ will be the name of the new company to be created through the expected spin-off of its uniform services business.
I don't think we need to change the spider name. Seems vestis will be its own thing. But the names of the shops that are being scraped all have Vestis in it.

davidhicks: I think we should delete:
```
apply_category(Categories.SHOP_CLOTHES, item)
```

(as I think this category is more for places you can walk into to buy a bunch of goods and then walk out with those goods)

And perhaps both cleanroom and uniform services get the same categories:
```
apply_category(Categories.SHOP_LAUNDRY, item)
apply_yes_no("laundry_service", item, True)
apply_category({"rental": "clothes"}, item)
apply_category({"access": "private"}, item)
```

Which would mean it's a private laundry service that either accepts privately owned garments, or garments previously rented from the business?

Generally `brand` should match the name the location is marketed as, which sounds like `Verstis` in this case. I agree that the spider name can remain the same as current, but I think we should override `brand` and `brand_wikidata` for uniform services.

davidhicks: As `Vestis` has been spun off as a new company and new brand, a new Wikidata item is needed as there are now two separate companies and thus a need for two Wikidata items.

ckpiper: Have created a new wikidata item for Vesits Corporation (https://www.wikidata.org/wiki/Q122947676)

locations/spiders/aramark_uniform_services_us.py
```diff
@@ -59,10 +60,20 @@ def parse(self, response):
...
+
+            def add_tags(item):
+                apply_category(Categories.SHOP_LAUNDRY, item)
+                apply_yes_no("laundry_service", item, True)
+                apply_category({"rental": "clothes"}, item)
+                apply_category({"access": "private"}, item)
+
+            if "Cleanroom Services" in item["name"].title():
+                item["brand"] = "Vestis Cleanroom Services"
                 item["name"] = item["brand"] + " " + item["city"]
```
iandees, line 72: I know this was there already, but we shouldn't craft our own name for a place unless this is what the brand themselves use. Even then, I would prefer to pull it from the page rather than build it ourselves by concatenating strings.

locations/spiders/aramark_uniform_services_us.py
```diff
@@ -1,12 +1,13 @@
 from scrapy import Spider
 from scrapy.http import JsonRequest
 
+from locations.categories import Categories, apply_category, apply_yes_no
 from locations.dict_parser import DictParser
 
 
 class AramarkUniformServicesUSSpider(Spider):
     name = "aramark_uniform_services_us"
```
iandees, line 9: If Aramark is rebranding themselves, we should also update the spider name, class name, and file name too.

davidhicks: I think you're right, and the source data differentiating names Aramark vs. Verstis is just a case of "haven't had a chance to rebrand all locations yet". But in the future, https://www.aramark.com/our-services/uniform-services makes it sound like Aramark won't deliver any uniform or cleanroom services under an Aramark brand, that will be done solely by Verstis instead. Aramark then keeps food services, facilities management, etc services under their existing brand.

iandees: Yea, it does look like there will still be Aramark places. Vestis is a whole new company that spun off, so at some point we'll probably make a scraper for them and revisit this one.

(Vestis Pittsburgh should probably file a trademark infringement lawsuit... https://www.vestispgh.com/)

ckpiper: I changed the filename, classname, spidername, and urls as they vestis.com/graphql now works. Thus I think this should be good to merge.

locations/spiders/aramark_uniform_services_us.py
ckpiper: Understandable. I did look if they have anything and the location (`location["node"]["postLocation"]["name"]`) only have 2 outputs, `Vestis` and `Vestis Cleanroom Services`. We could change it to just use that.

davidhicks (approved):

PR #6033, opened 2023-10-05
`petrol_bg`: add services data

### LINKED ISSUE #5939: petrol_bg: add services data
29 - AdBlue
25 - car wash
26 - atm
22 - Gasoline A-95
21 - Gasoline A-100
23 - LPG
20 - Diesel Green Force
19 - Diesel Pro Force
31 - EV charging station
24 - CNG
27 - restaurant
28 - self service

Note on the two diesels: Green Force has some Liqui Molly additive and is more expensive, the other is standard diesel. I think that we can put both of them under `fuel:diesel`.

Dimitar5555:
Resolves #5939


Dimitar5555:
I'm not sure why it can't find `location["categories"]`. It is present in https://www.petrol.bg/wp-admin/admin-ajax.php?action=asl_load_stores&nonce=b2bcd077ec&load_all=1&layout=1.

locations/spiders/petrol_bg.py
```diff
@@ -12,4 +13,18 @@ def parse_item(self, item, location):
         if m := re.match(r"^(\d+) (.+)$", item["name"]):
             item["ref"] = m.group(1)
             item["name"] = m.group(2)
+
+        categories = location["categories"].split(",")
```
Cj-Malone, line 17:
```suggestion
        categories = location.get("categories", "").split(",")
```

Sounds like a few don't have any categories.

locations/spiders/petrol_bg.py
```diff
@@ -12,4 +13,18 @@ def parse_item(self, item, location):
         if m := re.match(r"^(\d+) (.+)$", item["name"]):
             item["ref"] = m.group(1)
             item["name"] = m.group(2)
+
+        categories = location.get("categories", "").split(",")
```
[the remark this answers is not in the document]
Dimitar5555: No worries. I've also just noticed that some of them are `null`.

locations/spiders/petrol_bg.py
```diff
@@ -12,4 +13,18 @@ def parse_item(self, item, location):
...
+
+        categories = (location["categories"] or "").split(",")
+        apply_yes_no(Fuel.DIESEL, item, ("19" in categories or "20" in categories))
+        apply_yes_no(Fuel.OCTANE_100, item, "21" in categories)
+        apply_yes_no(Fuel.OCTANE_95, item, "22" in categories)
+        apply_yes_no(Fuel.LPG, item, "23" in categories)
+        apply_yes_no(Fuel.CNG, item, "24" in categories)
+        apply_yes_no(Extras.CAR_WASH, item, "25" in categories)
+        apply_yes_no(Extras.ATM, item, "26" in categories)
+        apply_yes_no("restaurant", item, "27" in categories)
```
davidhicks, line 25: Ideally immediately after `categories=` the following may be possible:

```
if "27" in categories:
  restaurant_item = copy.deepcopy(item)
  apply_category(Categories.RESTAURANT, item)
  yield item
```

This will create a separate item with the same address, phone number, etc. Perhaps some fields need to be pop'd off (such as phone number) if they differ between fuel station and restaurant. Perhaps a different brand needs to be applied, etc.

locations/spiders/petrol_bg.py
```diff
@@ -12,4 +13,18 @@ def parse_item(self, item, location):
...
+        apply_yes_no("self_service", item, "28" in categories)
+        apply_yes_no(Fuel.ADBLUE, item, "29" in categories)
+        apply_yes_no("amenity:chargingstation", item, "31" in categories)
```
davidhicks, line 28: @Cj-Malone has some insight on desired tagging of charging stations/fuel stations at #5833 which may be helpful here.

locations/spiders/petrol_bg.py
Dimitar5555: They have two kinds of restaurants. One is the typical petrol station restaurant (sandwiches, hot dogs, cafe). The other is an actual restaurant called "Street Chefs" which is available on very few Petrol station (probably less than 10). I will check back on it in the coming days when I have a bit more time.

Dimitar5555: The stations that have `restaurant` actually have cafe which at best sells some baked snacks and has outdoor seating area. The ones that have actual restaurant ("Street Chefs") are only two. `cafe=yes` would probably be a better tag.

iandees:
Is there interest in resolving the conflict here and merging in?

Dimitar5555:
@iandees I've resolved the conflict but the changes don't show up here. Can you reopen the PR? 

https://github.com/Dimitar5555/alltheplaces/blob/patch-2/locations/spiders/petrol_bg.py

locations/spiders/petrol_bg.py
Dimitar5555: Updated the PR to use `cafe=yes` instead.

locations/spiders/petrol_bg.py
Dimitar5555: The spider now generates a new POI for the charging station.

PR #6042, opened 2023-10-05
[WilkoGB] Remove spider

Cj-Malone:
I won't be offended if you don't want this. The Wilko spider will be getting [deleted soon](https://www.timeout.com/uk/news/heres-the-full-list-of-wilko-stores-that-are-closing-for-good-080723), the company is in the process of [firing all the staff and selling the brand](https://www.theguardian.com/business/2023/sep/13/the-range-buys-wilko-brand-in-multimillion-pound-deal), the also "lost" [a little bit of pension money](https://www.theguardian.com/business/2023/sep/29/wilko-pension-fund-50m-in-deficit-after-firms-collapse), but lets not get political.

I want the closed POIs so I can [update OSM](https://www.openstreetmap.org/changeset/142207326) as each batch of stores close. However ATP can't actually output `key=0` as as `if 0:` is the same is `if False:`... I've changed the exporter and added a test. I also hardcoded it in the `osm` exporter. That exporter is still super opinionated and fails on a bunch of spiders, but that's kinda how I want it, it's not recommanded or documented for anyone else.

Finally I changed the closed checker pipeline to skip POIs that are tagged as closed.


locations/spiders/wilko_gb.py
```diff
@@ -8,10 +8,17 @@ class WilkoGBSpider(SitemapSpider, StructuredDataSpider):
...
-    ]
+    sitemap_rules = [(r"https:\/\/stores\.wilko\.com\/gb\/([\w-]+\/[-\w\/]+)$", "parse_sd")]
     wanted_types = ["DepartmentStore"]
+
+    def post_process_item(self, item, response, ld_data, **kwargs):
+        if closed := response.xpath(
+            '//span[@class="AlertBanner-text"][contains(text(), "The store will permanently close on ")]/text()'
+        ).get():
+            closed_date = closed.replace("The store will permanently close on ", "")
+            item["extras"]["confidence:feature"] = 0.0
```
davidhicks, line 19: Is `confidence:feature` an OSM tag or something we've decided upon for ATP? I can only see related `operational_status:reliability` tag for OSM per https://wiki.openstreetmap.org/wiki/Humanitarian_OSM_Tags/Humanitarian_Data_Model#Operational_status.

What does this tag intend to convey, particularly for value `0.0`? It sounds like a statement that we're 0% confident the feature exists? Which isn't _quite_ true if we're still waiting for `end_date` to occur because we're somewhat confident the feature still exists until `end_date`. Then after `end_date` we're now closer to 100% confident the feature no longer exists?

davidhicks (commented):
I support the proposal to keep track of stores which have an upcoming nominated closure date, using the `end_date` tag for this purpose.

For stores that have already closed and continue to be returned by a store locator API for a period afterwards (just marked as closed), how long would ATP keep track of such closures? Indefinitely as long as the store locator continues to know of these stores? What if the store locator keeps track of closed stores for X years to come, and in the meantime another tenant moves into a building, creating another feature on top of the closed one? Should the closed feature be exported from ATP and then shown on the map at alltheplaces.xyz, included in the feature counts, etc?

Cj-Malone:
We can't rely on `end_date`, as most sites won't provide it. Even some of these POIs don't supply it. There needs to be another tag.

`confidence:feature` isn't an OSM tag, and there isn't an equivalent. [`operational_status=closed`](https://wiki.openstreetmap.org/wiki/Tag:operational_status%3Dclosed) is "deprecated", most of the "community" never liked it or really any of the HOT tags.

In OSM when a POI closes it does:

```patch
addr:street=abc
building=retail
-name=x
-brand=y
-shop=z
+disused:shop=z
```

Then when it's reopened:

```patch
addr:street=abc
building=retail
+name=x
+brand=y
+shop=z
-disused:shop=z
```

The OSM object is the premises, not the company.

I copied `confidence:feature=0` from Overture, I agree it's counter intuitive, but I don't particularly want to invent a new system. Especially since this spider (the only spider) will be removed soon, it's not worth that much code.

locations/spiders/wilko_gb.py
Cj-Malone: These end dates are in the past (and un formatted as I wasn't using them), I hadn't considered an end_date in the future. I agree if we get that data, we should output it. I don't think I've seen a site give us that data yet.

davidhicks: > These end dates are in the past (and un formatted as I wasn't using them), I hadn't considered an end_date in the future. I agree if we get that data, we should output it. I don't think I've seen a site give us that data yet.

I've only seen future closure dates be mentioned a few times, generally in the context of temporary closures for renovations/maintenance or an upcoming relocation to a new address. I think in all cases though, it wouldn't be worth parsing because it would have been information generally provided in freeform comments fields, hence hard to reliably extract.

davidhicks:
> We can't rely on `end_date`, as most sites won't provide it. Even some of these POIs don't supply it. There needs to be another tag.
> 
> `confidence:feature` isn't an OSM tag, and there isn't an equivalent. [`operational_status=closed`](https://wiki.openstreetmap.org/wiki/Tag:operational_status%3Dclosed) is "deprecated", most of the "community" never liked it or really any of the HOT tags.
> 
> In OSM when a POI closes it does:
> 
> ```diff
> addr:street=abc
> building=retail
> -name=x
> -brand=y
> -shop=z
> +disused:shop=z
> ```
> 
> Then when it's reopened:
> 
> ```diff
> addr:street=abc
> building=retail
> +name=x
> +brand=y
> +shop=z
> -disused:shop=z
> ```
> 
> The OSM object is the premises, not the company.
> 
> I copied `confidence:feature=0` from Overture, I agree it's counter intuitive, but I don't particularly want to invent a new system. Especially since this spider (the only spider) will be removed soon, it's not worth that much code.

Should we be using the `disused:` prefix? Sounds reasonable. It would be a bit of work though in deciding whether to report disused features in weekly run statistics, output them to the weekly geojson output, etc.

My main concern at the moment is with this proposed change, we would seemingly be over-reporting brands in statistics, and in the weekly geojson output, regardless of whether `disused:` or `end_date` tags or whatever is used. For example, one location closes and another pops up due to a relocation of a franchise to a new premises a few doors away. Currently we'd just report whichever of the two locations is currently open. With this patch, seemingly we'd be counting both old and new locations.

Cj-Malone:
> Should we be using the `disused:` prefix? Sounds reasonable. It would be a bit of work though in deciding whether to report disused features in weekly run statistics, output them to the weekly geojson output, etc.

Using `disused:shop=x` would be a lot of work, to categories, NSI pipeline. It's too much code for 1 spider that will be getting deleted in a few weeks. Also I consider `brand=x` `disused:shop=y` bad tagging in OSM anyway.

> My main concern at the moment is with this proposed change, we would seemingly be over-reporting brands in statistics, and in the weekly geojson output, regardless of whether `disused:` or `end_date` tags or whatever is used.

The output stats can't currently be used for analysis. 1 spider can output multiple categories and multiple brands. To workout how many Kroger supermarkets there are, you have to look at the data.

> For example, one location closes and another pops up due to a relocation of a franchise to a new premises a few doors away. Currently we'd just report whichever of the two locations is currently open. With this patch, seemingly we'd be counting both old and new locations.

I don't think that's true. We basically always output what we're given. In this example it would be 2 locations, or 1.

I added the closed check so I could review spiders that may be outputting closed locations. TBH I can't remember if I looked at the stats from it, I don't think I edited any spiders based on it, yet.

This wasn't meant to be an official way for ATP to output closed locations. I don't know if ATP should do that, or throw them away. I see advantages for both. It was just meant to get me the data to update OSM over the next few weeks, I like archiving that. I tried similar for Paperchase when it was closing, however they never flagged the locations so ATP outputted 100% of POIs until it was 0.

Cj-Malone:
If this is about the stats we can add a counter for confirmed closed POIs easily.

davidhicks:
I have no objection to this particular PR because the brand is soon disappearing. Just curious what our future intent is for closed locations reported by brands. Do we aim to capture `end_date` where it exists (I've personally seen it quite a few times and just ignored it) and then leave it up to ATP consumers to check closure status of each feature before using each feature?

As ATP tries to mostly capture authoritative (not second hand) data, someone comparing OSM and ATP feature sets for a brand should be fairly confident in removing an OSM feature if the the last few ATP runs have completed without error and the OSM feature is not close to any from the most recent non-failed ATP runs. Even then, if `end_date` information or other confirmation of closure is available, this will raise confidence to 100% that a feature in OSM but explicitly labelled as closed in ATP should be removed.

PR #6043, opened 2023-10-06
added spider for sunglass_hut_2

### LINKED ISSUE #6003: Sunglass Hut (various countries)
### Brand name

Sunglass Hut

### Wikidata ID

Q136311

### Store finder url(s)

https://mena.sunglasshut.com/api/content/render/false/limit/9999/type/json/query/+contentType:SghStoreLocator%20+languageId:11%20+deleted:false%20+working:true/orderby/modDate%20desc

Covers countries:
* Algeria
* Bahrain
* Cyprus
* Egypt
* Jordan
* KSA
* Kuwait
* Lebanon
* Oman
* Qatar
* UAE

Netherlands: https://nl.sunglasshut.com/api/content/render/false/limit/9999/type/json/query/+contentType:SghStoreLocator%20+languageId:3%20+deleted:false%20+working:true/orderby/modDate%20desc

Portugal: https://pt.sunglasshut.com/api/content/render/false/limit/9999/type/json/query/+contentType:SghStoreLocator%20+languageId:4%20+deleted:false%20+working:true/orderby/modDate%20desc

Turkey: https://tr.sunglasshut.com/api/content/render/false/limit/9999/type/json/query/+contentType:SghStoreLocator%20+languageId:4%20+deleted:false%20+working:true/orderby/modDate%20desc

Hong Kong: https://hk.sunglasshut.com/api/content/render/false/limit/9999/type/json/query/+contentType:SghStoreLocator%20+languageId:8%20+deleted:false%20+working:true/orderby/modDate%20desc

Arshad-Ali-1a:
closes #6003 

### Description:
Added spider for Sunglass Hut (various countries)

### Related Issue
[Sunglass Hut (various countries)](https://github.com/alltheplaces/alltheplaces/issues/6003)

### Output
image


locations/spiders/sunglass_hut_2.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def parse(self, response):
+        for data in response.json()["contentlets"]:
+            item = DictParser.parse(data)
+
+            item["country"] = data["innerCountry"] if data["innerCountry"] else None
+
+            item["lat"] = data["geographicCoordinatesLatitude"]
+            item["lon"] = data["geographicCoordinatesLongitude"]
+            item["geometry"] = {"coordinates": [item["lat"], item["lon"]], "type": "Point"}
```
davidhicks, line 37: `item["geometry"]` is set automatically if you've already specified `item["lat"]` and `item["lon"]`, so no need to specify it again here. We'd set `item["geometry"]` instead of `item["lat"]`/`item["lon"]` if the source data provided something like Schema.org GeoCoordinates:

```
  "geo": {
    "@type": "GeoCoordinates",
    "latitude": "40.75",
    "longitude": "-73.98"
  },
```
(ref: https://schema.org/GeoCoordinates)

locations/spiders/sunglass_hut_2.py
```diff
@@ -0,0 +1,45 @@
...
+
+            item["opening_hours"] = OpeningHours()
+            item["opening_hours"].add_ranges_from_string(
+                "\n".join([f"{day} {data[day]}" for day in DAYS_LOWER]), delimiters=[" : "]
```
davidhicks, line 41: This is a bit of a nitpick, but instead of specifying new constant `DAYS_LOWER`, it could be a one liner to get the string which `add_ranges_from_string` will parse?

```" ".join(["{}: {}".format(day, data[day.lower()]) for day in DAYS_FULL])```

File: locations/spiders/sunglass_hut_2.py [hunk 3]
```diff
@@ -0,0 +1,45 @@
...
+            )
+            item["opening_hours"] = item["opening_hours"].as_opening_hours()
```
davidhicks, line 43: This can be omitted and the `OpeningHours` object returned directly as-is.

locations/spiders/sunglass_hut_2.py
```diff
@@ -0,0 +1,45 @@
...
+        "https://pt.sunglasshut.com/api/content/render/false/limit/9999/type/json/query/+contentType:SghStoreLocator%20+languageId:4%20+deleted:false%20+working:true/orderby/modDate%20desc",
+        "https://tr.sunglasshut.com/api/content/render/false/limit/9999/type/json/query/+contentType:SghStoreLocator%20+languageId:4%20+deleted:false%20+working:true/orderby/modDate%20desc",
+        "https://hk.sunglasshut.com/api/content/render/false/limit/9999/type/json/query/+contentType:SghStoreLocator%20+languageId:8%20+deleted:false%20+working:true/orderby/modDate%20desc",
+    ]
+
+    def parse(self, response):
+        for data in response.json()["contentlets"]:
+            item = DictParser.parse(data)
+
+            item["country"] = data["innerCountry"] if data["innerCountry"] else None
```
davidhicks, line 33: Perhaps we need a fallback country_code of `response.url.split("https://", 1)[1].split(".", 1)[0].upper()` and if that happens to be `mena` then we leave `country` as `None`. I think the middleware will then apply reverse geocoding of coordinates where `country=None` for the `mena` API call.

davidhicks (changes requested):
What an excellent 1st PR! I've suggested just a few minor tweaks

locations/spiders/sunglass_hut_2.py
Arshad-Ali-1a: 😄 I was going to do that, but I thought making a new constant would be better, as it is only 7 days, and `.lower()` would not have to be used.

Anyways, I will change it.

locations/spiders/sunglass_hut_2.py
Arshad-Ali-1a: Ya, that's better.
Is this fine: 
```
if  data["innerCountry"]:
                item["country"] = data["innerCountry"] else:
                item["country"] = response.url.split("https://", 1)[1].split(".", 1)[0].upper()
                item["country"] = None if (item["country"] == "MENA") else item["country"]
```

davidhicks (approved):

locations/spiders/sunglass_hut_2.py
```diff
@@ -0,0 +1,45 @@
...
+
+    start_urls = [
+        "https://mena.sunglasshut.com/api/content/render/false/limit/9999/type/json/query/+contentType:SghStoreLocator%20+languageId:11%20+deleted:false%20+working:true/orderby/modDate%20desc",
+        "https://nl.sunglasshut.com/api/content/render/false/limit/9999/type/json/query/+contentType:SghStoreLocator%20+languageId:3%20+deleted:false%20+working:true/orderby/modDate%20desc",
+        "https://pt.sunglasshut.com/api/content/render/false/limit/9999/type/json/query/+contentType:SghStoreLocator%20+languageId:4%20+deleted:false%20+working:true/orderby/modDate%20desc",
+        "https://tr.sunglasshut.com/api/content/render/false/limit/9999/type/json/query/+contentType:SghStoreLocator%20+languageId:4%20+deleted:false%20+working:true/orderby/modDate%20desc",
+        "https://hk.sunglasshut.com/api/content/render/false/limit/9999/type/json/query/+contentType:SghStoreLocator%20+languageId:8%20+deleted:false%20+working:true/orderby/modDate%20desc",
+    ]
+
+    def parse(self, response):
```
davidhicks, line 27: Oops I missed something in my review, and it's causing 1 result to return because there is a dupefilter that thinks all locations are the same without a distinguishing ID.

We need to set `item["ref"]` manually in this case, and there is no ID field. Instead I suggest that the `seoUrl` field be used, similar to `item["website"] = response.url.split(...) + data["seoUrl"]` and then `item["ref"] = item["website"]`.

davidhicks (changes requested):

locations/spiders/sunglass_hut_2.py
Arshad-Ali-1a: Oh, I didn't see that. The following should be better, as the websites would also be legit.
I did not add the website previously, as these are not exactly _websites_, they just provide info about the specific store.

```suggestion
    def parse(self, response):
      region=response.url.split("https://", 1)[1].split(".", 1)[0].lower()
      item["ref"]=f"{region}-{data['seoUrl']}"
      
      REGION_TO_LANGUAGE={"mena":"en" , "nl":"nl" ,...} 
      item["website"]="https://{}.sunglasshut.com/{}/store-locator/{}".format(region,REGION_TO_LANGUAGE[region],data['seoUrl'])
```

the websites would be like this:
https://mena.sunglasshut.com/en/store-locator/riyadh-panorama
https://nl.sunglasshut.com/nl/store-locator/lelystad-lelystad

davidhicks: Those websites are fine to use. Looks good!

Arshad-Ali-1a:
I will also change [sunglass_hut.py](https://github.com/alltheplaces/alltheplaces/blob/bf033b5493ef2d25b36341e4a1e46ee31e553a40/locations/spiders/sunglass_hut.py#L7) to `sunglass_hut_1.py`

davidhicks:
> I will also change [sunglass_hut.py](https://github.com/alltheplaces/alltheplaces/blob/bf033b5493ef2d25b36341e4a1e46ee31e553a40/locations/spiders/sunglass_hut.py#L7) to `sunglass_hut_1.py`

According to the failed CI, `sunglass_hut_gb` now needs to be updated to import attributes from `sunglass_hut_1`.

Arshad-Ali-1a:
> According to the failed CI, `sunglass_hut_gb` now needs to be updated to import attributes from `sunglass_hut_1`.

Yaa, Sorry, I did see the file earlier, but forgot that it was importing from `sunglass_hut`.
I will update it.

PR #6046, opened 2023-10-06
Add Putka (PL) (220 stores)

starsep:


matkoniecz:
not sure is it fixable, but I spotted that links are borked a bit (part of URL decides which submenu will be unrolled) - see https://github.com/alltheplaces/alltheplaces/issues/12048

PR #6047, opened 2023-10-06
Fix Delikatesy Centrum: parse NextJS buildId

starsep:


starsep:
After implementing I realized that [QuickBELU](https://github.com/alltheplaces/alltheplaces/blob/f2d1b0a5df46dbc567f62a2d40eb4914d202a418/locations/spiders/quick_be_lu.py) uses the same technique. Perhaps should be extracted to some function.

Two other spiders will also break after NextJS rebuild: [TeknikmagasinetSE](https://github.com/alltheplaces/alltheplaces/blob/f2d1b0a5df46dbc567f62a2d40eb4914d202a418/locations/spiders/teknikmagasinet.py) and [CrumblCookiesUS](https://github.com/alltheplaces/alltheplaces/blob/f2d1b0a5df46dbc567f62a2d40eb4914d202a418/locations/spiders/crumbl_cookies_us.py)

davidhicks (approved):

PR #6048, opened 2023-10-07
Add Stokrotka (PL) (975 stores)

starsep:


locations/spiders/stokrotka_pl.py
```diff
@@ -0,0 +1,31 @@
...
+class StokrotkaPLSpider(SitemapSpider, StructuredDataSpider):
+    name = "stokrotka_pl"
+    item_attributes = {}
+    allowed_domains = ["stokrotka.pl"]
+    sitemap_urls = [
+        "https://stokrotka.pl/data/domains/1/sitemap/c_objects_map.xml",
+    ]
+    sitemap_rules = [(r"nasze-sklepy\/[a-zA-Z0-9-]+\.html", "parse_sd")]
+
+    def post_process_item(self, item, response, ld_data, **kwargs):
```
davidhicks, line 15: Coordinates aren't stored in the ld+json blob, but are contained in a separate JavaScript blob on each page.

Could you extract the coordinates with something like:
```
data = response.xpath('//script[contains(text(), "lat:")]/text()').get()
item["lat"] = data.split("lat: ", 1)[1].split(",", 1)[0]
item["lon"] = data.split("lng: ", 1)[1].split(",", 1)[0]
```

davidhicks (changes requested):

locations/spiders/stokrotka_pl.py
starsep: Thanks! I tested that it works
I added it in https://github.com/alltheplaces/alltheplaces/pull/6048/commits/618352a4502b792a9da72419d8cf6294dea23457

davidhicks (approved):
Thank you looks good!

PR #6049, opened 2023-10-07
Add ABC (PL) (7602 stores)

starsep:


locations/spiders/abc_pl.py
```diff
@@ -0,0 +1,17 @@
...
+class ABCPLSpider(Spider):
+    name = "abc_pl"
+    item_attributes = {"brand": "abc", "brand_wikidata": "Q11683985"}
+
+    def start_requests(self):
+        yield JsonRequest(url="https://sklepyabc.pl/wp-content/themes/abc/api/js/gps.json")
+
+    def parse(self, response, **kwargs):
+        for feature in response.json()["shops"]:
+            item = DictParser.parse(feature)
```
davidhicks, line 16: We'll also need to add:
```
item["ref"] = feature["idCRM"]
item["housenumber"] = feature["number"]
```

This will ensure all 7602 unique features are extracted from the JSON file, as a unique reference is needed for each. The missing street number is needed to complement the existing extraction of the street name.

davidhicks (changes requested):

locations/spiders/abc_pl.py
starsep: Added in https://github.com/alltheplaces/alltheplaces/pull/6049/commits/e0735bcef968eb64591dfa09f7c2efd220485ee3

davidhicks (approved):
Thank you looks good!

PR #6070, opened 2023-10-09
Add Auchan (PL) (137 locations).

panasonic-mafia:


locations/spiders/auchan_pl.py
```diff
@@ -0,0 +1,40 @@
+from urllib.parse import urljoin
+
+from locations.categories import Categories, apply_category
+from locations.storefinders.woosmap import WoosmapSpider
+
+# Types stored at https://api.auchan.com/corp/cms/v4/pl/template/store-types?lang=pl (accessible only from website)
+CATEGORY_MAPPING = {
```
davidhicks, line 7: Should this mapping be:

```
"123": {"brand": "Auchan Whatever", "brand_wikidata": "Q12345", "shop": "supermarket"},
"456": ...
```

And then lower down in the spider, it becomes `item.update(CATEGORY_MAPPING[store_type]`

That way we're distinguishing between Auchan sub-brands.

davidhicks (changes requested):

locations/spiders/auchan_pl.py
panasonic-mafia: Makes sense, thanks!

davidhicks (approved):

PR #6074, opened 2023-10-09
Fixed the bug easypay_bg.py that caused countries of all items to be set to BG

### LINKED ISSUE #6032: `easypay_bg.py` - foreign offices are marked as located in Bulgaria
EasyPay seems to have some offices outside of Bulgaria (1 in Austria, 5 in Greece and 14 in Cyprus), but all of them have `addr:country=BG`.

Arshad-Ali-1a:
Fixes #6032 

## Description
So, as there is no country data in the csv, the country is set to `None` for all items. The pipeline then runs [country_code_clean_up.py](https://github.com/alltheplaces/alltheplaces/blob/master/locations/pipelines/country_code_clean_up.py#L20). And it extracts the country from the **spider name**, which in this case is `easypay_bg`. This is why all countries were set to `BG`. The reverse geocoding did not even happen.

So the sipder name is changed to `easypay`. Now reverse geocoding takes care of all the countries. The distribution is as follows:

BG-3267
RO-24
CY-14
GR-10
MK-3
AT-1

In the future, if more spiders for easypay were to be created, the spider name can be changed to `easypay_1`.
The file name, and the class name can also be changed  if required, in order to match the spider name.

### DIFF
--- a/locations/spiders/easypay.py
+++ b/locations/spiders/easypay.py
@@ -3,8 +3,8 @@ from scrapy.spiders import CSVFeedSpider
 from locations.dict_parser import DictParser
 
 
-class EasypayBGSpider(CSVFeedSpider):
-    name = "easypay"  # the name is not set to easypay_bg, because then name would be used to extract "country", and "country" in every item would be set to "BG", which is wrong.
+class EasypaySpider(CSVFeedSpider):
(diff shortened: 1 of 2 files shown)


iandees:
The spider name shouldn't include a country suffix if it produces data in multiple countries.

Thanks for making this change.

PR #6092, opened 2023-10-10
bmo_harris: fix spider and rename to bmo

davidhicks:
BMO is self-hosting Where2GetIt on their own domains. The Where2GetItSpider has been updated to allow for self-hosted instances.

Due to BMO's implementation of Where2GetIt being slow to respond (502 Gateway Timeout errors), it is also now necessary for Where2GetItSpider to allow querying on a state/province-by-state/province level. The parameters for the Where2GetItSpider have been updated to allow level 0 (try to return all locations in a single query), level 1 (return locations per country) and level 2 (return locations per state/province).

BMO ATMs are located in a large number of third-party locations. The list of located_in/located_in_wikidata fields for BMO ATMs has been vastly expanded to include many additional third-party locations that already have ATP spiders.

In order to import modules (in this case spiders), Python expects module names to start with an alphabetical character, not a number. Spiders which had names starting with a number have been renamed to allow them to be imported by BMOSpider. Some additional renaming of other spiders has occurred to ensure new module imports into BMOSpider are more consistently named.


davidhicks:
FYI failed spiders `market_basket`, `seven_eleven_dk` and `thrifty_foods_ca` are already broken, and not broken by this change. They'll need fixing in due course.

iandees:
Hmm, I don't think BMO runs their own ATMs, so I think we might want to skip them? I think they partner with Allpoint's ATM network.

davidhicks:
> Hmm, I don't think BMO runs their own ATMs, so I think we might want to skip them? I think they partner with Allpoint's ATM network.

Oh!

https://www.bmo.com/en-us/main/personal/ways-to-bank/ isn't clear on whether BMO operate their own ATM network as well as using Allpoint's ATM network. For example, at a BMO branch, this spider sets `atm=yes` if the branch has ATMs present. Would these be BMO operated or Allpoint operated? And then all external ATMs from branches would be Allpoint operated?

PR #6093, opened 2023-10-10
added spider for DSV Locker (ZA).

### LINKED ISSUE #6045: DSV Locker parcel lockers (ZA)
### Brand name

DSV Locker

### Wikidata ID

Q1155771

### Store finder url(s)

https://dsvdirect.co.za/dsv-locker/#Locator (457 lockers)

Arshad-Ali-1a:
Closes #6045 

## Description

added spider for DSV Locker(ZA).

image


locations/spiders/dsv_locker_za.py
```diff
@@ -0,0 +1,30 @@
...
+class DSVLockerZASpider(Spider):
+    name = "dsv_locker_za"
+    item_attributes = {"brand": "DSV Locker", "brand_wikidata": "Q1155771"}
+    allowed_domains = ["dsv-clientzone-map.afrigis.co.za"]
+    start_urls = ["https://dsv-clientzone-map.afrigis.co.za/proxy/api/wfs/?layer=locker"]
+
+    def parse(self, response):
+        for data in response.json()["result"]["features"]:
+            data = data["properties"]
+            item = DictParser.parse(data)
```
davidhicks, line 16: The source data has `data["location_code"]` which is perhaps a better choice for `item["ref"]` than leaving things to default where you get `data["id"]` instead. It looks like `id` is just local to the store finder, whereas `location_code` is the brand's internal site code that is used across systems.

locations/spiders/dsv_locker_za.py
```diff
@@ -0,0 +1,30 @@
...
+
+            item["name"] = data.get("location_name")
+            item["street"] = data.get("suburb")
```
davidhicks, line 19: This doesn't look correct. In one example, `data["suburb"] = "Northmead"` and `data["street_address"] = 2, Fifth Avenue, Northmead, 1501, Johannesburg, Gauteng`.

Therefore "Fifth Avenue" is the street, and "Northmead" is the "city".

I see the source data has `suburb`, `town` and `province` but ATP only allows `city` and `state`. `city` in ATP should be the smallest administrative region corresponding to a suburb (or town if the town has no suburbs). `state` in ATP should be the 2nd level of government beneath the federal government. Perhaps one level in the source data needs to be skipped?

davidhicks (changes requested):

locations/spiders/dsv_locker_za.py
Arshad-Ali-1a: So, I guess suburb should be skipped.

davidhicks: Actually I think you'd skip Johannesburg in this case, and keep Northmead ("city") and Gauteng ("state"). The ATP term of "city" isn't what the name implies sadly. It generally means "suburb, or town/city if the town/city has no suburbs".

davidhicks: Yep I think that's right, just set `item["city"] = data["suburb"]`

davidhicks (approved):

PR #6133, opened 2023-10-12
Add M-Texx (BG) (362 locations)

### LINKED ISSUE #6038: M-texx clothes and shoes containers
### Brand name

M-texx

### Wikidata ID

Q122947768

### Store finder url(s)

https://m-texx.com/локации

It requires parsing of all `li` html elements. The four digit numbers are postcodes, there is no `ref` system.

Dimitar5555:
Resolves #6038.

### DIFF
--- a/locations/spiders/mtexx_bg.py
+++ b/locations/spiders/mtexx_bg.py
@@ -24,6 +24,6 @@ class MtexxBGSpider(Spider):
                     r'((\d+\.\d+),\s(\d+\.\d+))|((\d+°\d+\'\d+\.\d+"[NS])\s(\d+°\d+\'\d+\.\d+"[EW]))', text
                 )
                 if coords:
-                    properties.lat = coords.group(1)
-                    properties.lon = coords.group(2)
+                    properties["lat"] = coords.group(1)
+                    properties["lon"] = coords.group(2)
                 yield Feature(**properties)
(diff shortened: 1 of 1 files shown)


Dimitar5555:
It seems like it's almost done. I'm not sure why it's complaining about the lack of `ref`. Also some coordinates are in `43°29'39.4"N 24°04'44.9"E` format. I'm not sure if they are automatically converted or not.

iandees:
> I'm not sure why it's complaining about the lack of ref

Every item that a spider produces needs to have a `ref` property set that is unique per place. We do this to create unique IDs and to remove duplicate places.

> some coordinates are in 43°29'39.4"N 24°04'44.9"E format. I'm not sure if they are automatically converted or not

They won't be converted, no. Can you please convert them?

Dimitar5555:
> > I'm not sure why it's complaining about the lack of ref
> 
> Every item that a spider produces needs to have a `ref` property set that is unique per place. We do this to create unique IDs and to remove duplicate places.

It seemed like they are generated automagically in other spiders. Am I missing something here? 

> > some coordinates are in 43°29'39.4"N 24°04'44.9"E format. I'm not sure if they are automatically converted or not
> 
> They won't be converted, no. Can you please convert them?

Should be done.

Dimitar5555:
There are 5 locations that are not in the ATP dataset. One of them is "Coming soon", the other 4 have no coordinates. Otherwise it should be good to go.

iandees:
> It seemed like they are generated automagically in other spiders. Am I missing something here?

Yea, true. I forgot that we moved away from requiring ref a while ago. If we can't find a stable identifier for each location we will generate one randomly. This is less ideal, but if there isn't any data there then it's the best we can do.

PR #6142, opened 2023-10-13
Improve coop GB tagging

Cj-Malone:
I'm sure there are still issues, but it's an improvement.

"Coop" in GB is a PATA, it's a mix of companies (I think like Goodwill in US), however they have had internal drama. So the don't actually share a brand any more, but do share the mindshare.

I have 2 within 5 mins of my flat. 1 is a Southern (aka green), I always thought the other was Central, but it's actually Group.

Group is the big one, the small ones stayed with it. They are a coop of coops and (mostly) share a brand (aka blue). The mid sized ones fragmented.

Thanks for reading my anecdote.

### DIFF
--- a/locations/spiders/central_england_cooperative.py
+++ b/locations/spiders/central_england_cooperative.py
@@ -1,24 +1,46 @@
 from scrapy.spiders import SitemapSpider
 
-from locations.linked_data_parser import LinkedDataParser
-from locations.microdata_parser import MicrodataParser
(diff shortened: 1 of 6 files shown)


Cj-Malone:
I'm gonna check my Qcodes.

Cj-Malone:
This will be in line with https://github.com/osmlab/name-suggestion-index/pull/8799

PR #6148, opened 2023-10-13
Add Epaka.pl spider (360 locations)

Adikso:
I also fixed typo in Polish weekday name.

### DIFF
--- a/locations/hours.py
+++ b/locations/hours.py
@@ -255,7 +255,6 @@ DAYS_PL = {
     "Pi": "Fr",
     "Pia": "Fr",
     "Piątek": "Fr",
-    "Piątek": "Fr",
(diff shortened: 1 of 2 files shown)


locations/hours.py
```diff
@@ -264,7 +263,7 @@
     "Ni": "Su",
     "Nie": "Su",
     "Niedz": "Su",
-    "Niedzela": "Su",
```
Cj-Malone, line 267: I assume this was added because a website was giving us "Niedzela", if so, we are now losing Sunday hours from that spider.

iandees: Hmm, when I looked at the diff here I thought it was a duplicate entry, not a misspelling. Must have been going too fast. Thanks for adding it back.

PR #6158, opened 2023-10-15
Add Coral Travel PL (134 offices)

starsep:


Cj-Malone:
Are the "autorized" stores real Coral Travel locations? If not we can get the "wezyr" locations with:

```python
from scrapy.linkextractors import LinkExtractor
from scrapy.spiders import CrawlSpider, Rule

from locations.structured_data_spider import StructuredDataSpider


class CoralTravelPLSpider(CrawlSpider, StructuredDataSpider):
    name = "coral_travel_pl"
    item_attributes = {"brand": "Coral Travel", "brand_wikidata": "Q58011479"}
    start_urls = ["https://www.coraltravel.pl/biura-podrozy"]
    rules = [Rule(LinkExtractor("/biura-podrozy/.+/.+/.+$"), callback="parse_sd")]
    wanted_types = ["TravelAgency"]

        item["lat"] = response.xpath('//input[@id="map_position_lat"]/@value').get()
        item["lon"] = response.xpath('//input[@id="map_position_lng"]/@value').get()

        # TODO: Opening hours, more details?

        yield item

```

Cj-Malone:
I have a strong dislike for index refs, I would actually prefer no refs over indexes. So if you want to stick with the current spider, I'd request you remove the refs.

starsep:
> Are the "autorized" stores real Coral Travel locations? If not we can get the "wezyr" locations with:

I think yet are only partner locations. 
They are available by clicking "Autoryzowany agent" filter in the bottom left corner of the map.

Coral Travel logo is visible on photos I looked at but they seem to be only part of the branding.
Usually there is another text which would fit better OpenStreetMap `name` tag.

Thanks for suggested code. I will use!

> I have a strong dislike for index refs, I would actually prefer no refs over indexes. So if you want to stick with the current spider, I'd request you remove the refs.

Understood! I agree that index refs are unstable as the order might change. Shops can get removed from the middle.
I added it because without `ref` exception is raised.
Is it expected to return `None` when it is not available?
Then only first item is taken, rest is dropped due to duplicated `ref`.

Cj-Malone:
A spider level attribute `no_refs = True` will allow POIs without a ref, obviously we want refs when possible but in cases like this we have to go without.

PR #6170, opened 2023-10-16
pep8 naming

starsep:
- Fix pep8-naming: class names
- Fix pep8-naming: function names
- Fix pep8-naming: argument name
- Fix pep8-naming: variable in function
- Fix pep8-naming: mixedCase in global scope


locations/dict_parser.py
```diff
@@ -183,10 +183,10 @@ def get_variations(key):
         flatcase = key.lower().replace("-", "")
         results.add(flatcase)
 
-        FLATCASEUPPER = flatcase.upper()
```
iandees, line 186: Can you make this file follow the casing rules and add comments with what the casing names mean instead of using noqa?

locations/hours.py
```diff
@@ -3,6 +3,8 @@
 from collections import defaultdict
 
 DAYS = ["Mo", "Tu", "We", "Th", "Fr", "Sa", "Su"]
+DAYS_3_LETTERS = ["Mon", "Tue", "Wed", "Thu", "Fri", "Sat", "Sun"]
+DAYS_3_LETTERS_SUNDAY = DAYS_3_LETTERS[-1:] + DAYS_3_LETTERS[:-1]
```
davidhicks, line 7: `_SUNDAY` may be a bit clearer as `_SUN_TO_MON` or `_FROM_SUNDAY` or `_SUNDAY_START`?

starsep: Renamed to `DAYS_3_LETTERS_FROM_SUNDAY` in https://github.com/alltheplaces/alltheplaces/pull/6170/commits/5e58dce224fd759b728f7725b628e3a08294c794

iandees:
Do the tests fail when you run them locally?

starsep:
> Do the tests fail when you run them locally?

No, they pass. I see at the end of `pipenv run pytest` output
> 97 passed, 28 warnings in 2.36s

iandees:
Thanks for making this change. I merged it with #6188 so that I could fix the tests. The tests were failing because of a change upstream in scrapy.

PR #6173, opened 2023-10-17
Add Chorten PL (1964 stores)

starsep:


locations/spiders/chorten_pl.py
```diff
@@ -0,0 +1,48 @@
...
+
+        if response.url == self.start_urls[0]:
+            last_page_number = max(
+                map(
+                    lambda x: int(x.removeprefix("https://chorten.com.pl/sklepy/lista/p").removesuffix("?")),
+                    response.xpath("//a[@class='page-link']/@href").getall(),
+                )
+            )
+            for i in range(2, last_page_number + 1):
+                yield Request(url=f"https://chorten.com.pl/sklepy/lista/p{i}?")
```
Cj-Malone, line 25: I'm not too fussed now the code has been written, but FYI this could probably have been a [CrawlSpider](https://docs.scrapy.org/en/latest/topics/spiders.html#crawlspider), with a rule for `/sklepy/lista/p\d+?` and a rule with `/sklepy/[\w%_-]+_\d+$` with the `parse_store` call back.

starsep: I migrated to CrawlSpider in https://github.com/alltheplaces/alltheplaces/pull/6173/commits/027a389c3f732d2e0992ff80b5f935658dac7c3f
I checked that 2015 shops are scraped now

Cj-Malone: Another tip you might not know, when developing these request heavy spiders scrapy can do some caching which will save a bunch of time when re running the spider.

`pipenv run scrapy crawl -s HTTPCACHE_ENABLED=True -s HTTPCACHE_EXPIRATION_SECS=604800 chorten_pl`

However it breaks on spiders that send requests to the same url but with different headers, it's not common so when it happens I always get confused.

PR #6178, opened 2023-10-18
Fixed code to run hotels

ckpiper:
It works for pages that actually have hotels listed. Some of the links will give errors. These are mainly tourist attractions or non-hotel pages. 

I do not have stats as I ran it for 24 hours straight and it still was not finished. Not sure if there is a timeout feature that could be activated.


iandees:
Can you use `sitemap_filter` to limit the number of requests it makes to pages that are location specific? An example in another spider: https://github.com/alltheplaces/alltheplaces/blob/1b1662a6706cb27866d9aa65628d962d2e03b9fc/locations/spiders/tui_gb.py#L16

PR #6190, opened 2023-10-22
Create spider for cashterminal_bg

Dimitar5555:


Dimitar5555:
It seems like this site doesn't like robots.. Otherwise, I don't see why it would fail.

PR #6198, opened 2023-10-24
Use actual ref for JYSK store

Dimitar5555:
It seems like there are no duplicated refs across the websites so there is no point in using the URL address as a ref.

### DIFF
--- a/locations/spiders/jysk.py
+++ b/locations/spiders/jysk.py
@@ -50,7 +50,8 @@ class JyskSpider(scrapy.Spider):
         store["house_number"] = store.pop("house")
         item = DictParser.parse(store)
         item["image"] = store.get("image")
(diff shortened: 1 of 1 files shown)


Dimitar5555:
I'm not sure why it timeouts at 71 stores... The previous PR on this spider returned 74 stores (https://github.com/alltheplaces/alltheplaces/actions/runs/6577701571/job/17869890069) but the regular Sunday run returned 3k+ stores. I assume that it should be fine.

Cj-Malone (approved):
That runs for about 60 seconds, POI count will vary on network latency.

Weekly runs for a max of 8 hours.

panasonic-mafia:
Nice one, thanks @Dimitar5555!

PR #6220, opened 2023-10-27
chuckecheese: fix spider

davidhicks:


locations/spiders/chuckecheese.py
```diff
@@ -1,55 +1,17 @@
...
-            "website": response.url,
-        }
-
-        hours = self.parse_hours(response.xpath('//*[@itemprop="openingHours"]/@content').extract())
-
-        if hours:
-            properties["opening_hours"] = hours
-
-        yield Feature(**properties)
+    item_attributes = {"brand": "Chuck E. Cheese", "brand_wikidata": "Q2438391"}
```
Cj-Malone, line 8: This is one of the times I think NSI is wrong (it's usually us)

https://github.com/osmlab/name-suggestion-index/pull/8871

davidhicks:
> I like this diff :)

It's like winning a lottery to find a brand that has correct structured data!

PR #6264, opened 2023-10-31
Add 'Wesola Pani' (77 locations)

Adikso:


Adikso:
Hm... it works for me locally... maybe the website errored out randomly on checker attempt

locations/spiders/wesolapani.py
```diff
@@ -0,0 +1,40 @@
...
+        storage = chompjs.parse_js_object(storage_js)
+
+        for shop in storage["shops"]:
+            opening_hours = OpeningHours()
+            opening_hours.add_ranges_from_string(ranges_string=shop["shedule"], days=DAYS_RU)
+
+            address = ADDRESS_PATTERN.search(shop["address"])
+            address = address.groupdict()
+
+            yield {
```
Cj-Malone, line 32: The pipelines will fail with a dict, this needs to be a Feature

Cj-Malone:
I've added it to NSI so we should get the category in future. https://github.com/osmlab/name-suggestion-index/commit/083c40d02956f1677707cf8a83f075c5aa377628

PR #6299, opened 2023-11-08
Add Nordsee (DE) (299 locations)

### LINKED ISSUE #654: Nordsee
Fast food seafood in Europe

http://www.international.nordsee.com/en/filialen

davidhicks:
Fixes #654

### DIFF
--- a/locations/spiders/nordsee_de.py
+++ b/locations/spiders/nordsee_de.py
@@ -1,0 +1,29 @@
+from scrapy import Selector, Spider
+from scrapy.http import JsonRequest
+
+from locations.dict_parser import DictParser
+from locations.hours import OpeningHours
+
+
+class NordseeDESpider(Spider):
+    name = "nordsee_de"
+    item_attributes = {"brand": "Nordsee", "brand_wikidata": "Q74866"}
+    allowed_domains = ["www.nordsee.com"]
+    start_urls = ["https://www.nordsee.com/en/?type=2001"]
+    custom_settings = {"ROBOTSTXT_OBEY": False} # Missing robots.txt
+
+    def start_requests(self):
+        for url in self.start_urls:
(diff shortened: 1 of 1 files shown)


locations/spiders/nordsee_de.py
```diff
@@ -0,0 +1,32 @@
...
+from locations.dict_parser import DictParser
+from locations.hours import OpeningHours
+
+
+class NordseeDESpider(Spider):
+    name = "nordsee_de"
+    item_attributes = {"brand": "Nordsee", "brand_wikidata": "Q74866"}
+    allowed_domains = ["www.nordsee.com"]
+    start_urls = ["https://www.nordsee.com/en/?type=2001"]
+    custom_settings = {"ROBOTSTXT_OBEY": False}  # Missing robots.txt
```
Cj-Malone, line 13: Since this is a missing robots.txt, not a robots.txt that blocks us/breaks us. I'd prefer to leave this out. This gets set in the output file (`spider:robots_txt=ignored`). I think brawer added it so he could make a decision on if he wanted to use the data.

```suggestion
```

davidhicks: We just end up with a few robots.txt parsing errors in output in these cases. I think that's on Scrapy which doesn't seem to know what to do when a HTML page is returned instead of a plain text robots.txt

Cj-Malone: In scrapys defence I have seen html robots.txt before, but should they waste our time trying to parse it is another question (no).

I'm fine with the extra few lines in the log, it wont even affect the weeklies as they are ran with `-L INFO`

davidhicks: All good, I've removed the robots.txt disobey setting :)

PR #6329, opened 2023-11-14
mercy_health_us: fix spider and improve specialisation categories

davidhicks:
* Mercy Health has a v2 API but it is not returning any results (broken website?)

* Mercy Health can also be scraped using a sitemap and parsing ld+json data. This approach has been selected.

* Healthcare specialisation categories have been implemented with Mercy Health the first spider to use these categories.

* Clothing specialisation categories have been updated to match the approach used for healthcare specialisation.


davidhicks:
Note that both `gap_us` and `hugoboss` spiders are broken, but it's out of scope of this PR to fix them. I can get around to fixing them later.

locations/categories.py
```diff
@@ -185,17 +210,21 @@ def apply_category(category, item):
...
+    "healthcare:speciality",
     "highway",
     "landuse",
     "leisure",
     "man_made",
     "office",
     "public_transport",
     "shop",
+    "social_facility",
+    "social_facility:for",
```
Cj-Malone, line 227: Even though these are part of the category, they aren't top level tags. They are secondary.

`clothes=x` is not a valid category in OSM. `shop=clothes` is, and `shop=clothes` + `clothes=x` is.

This list/get_categories should only be returning the minimal category (`shop=clothes`).

davidhicks: Thanks, fixed it. Could you please check again to make sure the revised approach is ok.

PR #6347, opened 2023-11-17
davids_bridal: fix spider

davidhicks:


Cj-Malone:
I'm starting to really like graphql, at lease when it is open.

You can do things like trim the request down so we only get the attributes we want. Rename attributes so DicrParser handles them. Have them twice (eg location and address)

```graphql
query StoreLocationList {
    storeLocationList(active: true) {
        name
        phone
        ref: storeId
        storeType
        hours {
            regular {
                close
                day
                open
            }
        }
        address: location {
            building
            address1
            address2
            city
            postalCode
            state
            countryCode
        }
        location {
            latitude
            longitude
        }
    }
}
```

(No code changes required, just commenting on graphql)

davidhicks:
I don't mind the concept of GraphQL for the reasons you provided (particularly trimming fields and renaming fields), but find it annoying that the GraphQL community have generally decided to disable introspection.

I also wonder whether we could tidy up our GraphQL spiders by moving GraphQL query text into a separate .graphql file of the same name as the spider, just with a different extension. Then we create a `GraphQLSpider` which takes a single `start_urls[0]` as the GraphQL API endpoint, takes an optional dictionary of headers, and takes an optional dictionary of parameters for the query, and then loads the .graphql file from disk as the query. I am probably overthinking things, but I find it a bit annoying to work with spiders with 2 screen heights worth of GraphQL query text embedded within.

Cj-Malone:
Introspection was left on here, that's what I meant when I said "at least when it is open". Yeah, it's really annoying when it's disabled or it's just a persisted query.

My IDE can minimize variables, sure `data` here annoys me, especially the indenting. But it's not annoying me enough to do something about it. If it is annoying you enough to do something about it, go for it. (I think we all trust you to make sensible PRs) Keep in mind that Jimbo recently pointed out that file paths don't work as expected in some runtime environments, but it wasn't a complicated fix.

davidhicks:
> Introspection was left on here, that's what I meant when I said "at least when it is open". Yeah, it's really annoying when it's disabled or it's just a persisted query.
> 
> My IDE can minimize variables, sure `data` here annoys me, especially the indenting. But it's not annoying me enough to do something about it. If it is annoying you enough to do something about it, go for it. (I think we all trust you to make sensible PRs) Keep in mind that Jimbo recently pointed out that file paths don't work as expected in some runtime environments, but it wasn't a complicated fix.

It doesn't bother me _that_ much, just a thought for the future on how to simplify GraphQL spiders by writing a GraphQLSpider class that has some in-built methods to:
- Open a GraphQL query file from disk to read
- Specify variables for use with the GraphQL query
- Send the GraphQL query to the server
- Parse results with DictParser automatically (where possible)
- Allow overriding a `post_process_items(..)` method to parse tricky data or clean up data

Maybe one day when things are slow :)

Issue #6328, opened 2023-11-14
Auto-detect known store locators?


iandees:
We have a number of known store locator classes that make it easier to pull down locations.

Can we add a scrapy command that gets a domain and then passes it off to each known store locator type and asks the class to figure out if it can scrape that site? The goal would be to make it easier to build these spiders, maybe even automating it.

Cj-Malone:
Related to Common Crawl, but not this issue. There is [Web Data Commons](http://webdatacommons.org/) which extracts the structured data from Common Crawl. I'd love to use that data at some point.

davidhicks:
This would be quite easy to achieve for most store finders, as there is generally a common pattern such as an `<a href="https://storelocator.example">` or `<script src="https://storelocator.example/file.js">` that gives away the presence of a storefinder from a supplied store locator page on a brand website. I've been thinking about this idea too, but haven't implemented anything yet.

Would a function like this suffice for each storefinder:
```
@staticmethod
def find_spider_parameters(url: str)
```

Where the function returns, if successful, a dictionary:
```
{"storefinder": ClassNameOfStoreFinder,
  "api_key": "",
  ...etc (other parameters such as api_key that are specific to a storefinder)
}
```

If the function is unsuccessful, it'd return `None`. Or perhaps an exception is raised instead?

Then a new `BaseRunSpiderCommand` named `sf` is created which iterates through the store finders calling `find_spider_parameters(..)` until a value other than `None` is returned, or until there is no exception raised.

Improvement suggestions/thoughts?

Cj-Malone:
@davidhicks that's how I see it working

davidhicks:
I guess we could just return a pre-populated `Spider` object (subclass for the relevant storefinder) instead of a dictionary of parameters, then the `sf` BaseRunSpiderCommand would pretty print the `Spider` object ready for someone to pipe the output directly to a new `.py` file.

davidhicks:
> Related to Common Crawl, but not this issue. There is [Web Data Commons](http://webdatacommons.org/) which extracts the structured data from Common Crawl. I'd love to use that data at some point.

[...]

PR #6350, opened 2023-11-17
Automatic storefinder spider generator proof of concept

davidhicks:
Refer to #6328


davidhicks:
Proof of concept test:

`pipenv run scrapy sf --spider-key=ultra_tune_au --spider-class-name=UltraTuneAUSpider --wikidata=Q29025649 'https://www.ultratune.com.au/store-locations/'`

davidhicks:
Some thoughts:
- Should we introduce a class such as `AutomaticSpiderGenerator` from which store finders that support automatic spider generation from a URL and some basic supplied parameters (such as Wikidata ID) inherit from?
- `StructuredDataSpider` could also automatically generate a spider.
- NSI should be queried for the given Wikidata ID and spider automatically named.
- Wikidata could be queried for an item with an official homepage containing the domain of the provided URL.
- `sf` could be changed to return the name of a detected store finder, or could try to generate real output as `sd` does. A new `autogen` command or similar name could be created instead.

Cj-Malone:
> * NSI should be queried for the given Wikidata ID and spider automatically named.
> 
> * Wikidata could be queried for an item with an official homepage containing the domain of the provided URL.

NSI has a cache of urls, it should allow us to go from domain to Wikidata most of the time. https://github.com/osmlab/name-suggestion-index/blob/main/dist/wikidata.json

davidhicks:
> * Perhaps a new Wikidata property for store locator pages could be created as the source for generating spiders automatically. Obviously customisation is often needed to clean up data.

[P11707](https://www.wikidata.org/wiki/Property:P11707) already allows us to generate `StructuredDataSpider` spiders for each brand automatically from a given domain or Wikidata ID for a brand.

What I think is missing is a property to get to the landing page of a store locator search form / map, and/or a property for getting to the sitemap or HTML page that contains links matching P11707 for each location/store.

davidhicks:
> > * NSI should be queried for the given Wikidata ID and spider automatically named.
> > * Wikidata could be queried for an item with an official homepage containing the domain of the provided URL.
> 
> NSI has a cache of urls, it should allow us to go from domain to Wikidata most of the time. https://github.com/osmlab/name-suggestion-index/blob/main/dist/wikidata.json

OK I've implemented this successfully (I believe):

Test command:
```pipenv run scrapy sf 'https://www.ultratune.com.au/store-locations/'```

Output:
```from locations.storefinders.agile_store_locator import AgileStoreLocatorSpider


class UltraTuneSpider(AgileStoreLocatorSpider):
	name = "ultra_tune_au"
	item_attributes = {"brand": "Ultra Tune", "brand_wikidata": "Q29025649"}
	allowed_domains = ["www.ultratune.com.au"]
```

Cj-Malone:
I'll do a proper review at some point, but this is looking great. Thank you.

davidhicks:
Additional thoughts:

- Store finder classes could inherit from a new `AutomaticSpiderGenerator` class which contains common code for class naming and handling `name` and `item_attributes`. Some store finders may require additional parameters such as `api_key`. Perhaps each store finder overrides a method of `AutomaticSpiderGenerator` to return a dictionary of additional attributes, and `AutomaticSpiderGenerator` is then smart enough to pretty print those attributes whether it is a string, array of strings, dictionary of strings, etc.
- Command of `genspider` (what this POC branch currently names `sf`) would try to generate spider source code automatically from a provided URL. Output can be piped directly to a new .py file which is then ready to execute (noting a post_process_item method or similar may still need overwriting to clean and modify data).
- Command of `scanurl` would try to output information about what type of store finder exists at a URL, or whether the URL is a store page that uses Structured Data, etc. If source data is JSON formatted, one sample location of raw source data is returned. Additionally one sample location of parsed/cleaned data is also returned. The user can then check raw source data against automatically parsed/cleaned data to see how much manual cleaning/overriding may be required.

Cj-Malone:
> Store finder classes could inherit from a new `AutomaticSpiderGenerator` class which contains common code for class naming and handling `name` and `item_attributes`. Some store finders may require additional parameters such as `api_key`. Perhaps each store finder overrides a method of `AutomaticSpiderGenerator` to return a dictionary of additional attributes, and `AutomaticSpiderGenerator` is then smart enough to pretty print those attributes whether it is a string, array of strings, dictionary of strings, etc.

I've thought about having a base spider class for all spiders, instead of scrapy.spiders.Spider. How nice would it be to have our IDEs autocomplete `item_attributes`, `skip_auto_cc_spider_name` or whatever else? (I'm bad at spelling)

Just wanted to write it down, it's not for this PR.

panasonic-mafia:
Nice POC, thank you for current implementation!

> Command of genspider (what this POC branch currently names sf) would try to generate spider source code automatically from a provided URL. Output can be piped directly to a new .py file which is then ready to execute (noting a post_process_item method or similar may still need overwriting to clean and modify data).

Just a few thoughts on this.

Current [`genspider`](https://docs.scrapy.org/en/latest/topics/commands.html#std-command-genspider) command use templates, custom templates can be defined for each store finder, e.g.

```shell
$ mkdir templates/spiders
$ touch templates/spiders/agile_store_locator.tmpl
``` 
```python
from locations.storefinders.agile_store_locator import AgileStoreLocatorSpider


class $classname(AgileStoreLocatorSpider):
    name = "$name"
    allowed_domains = ["$domain"]


```
Then modify `settings.py` with:
```python
TEMPLATES_DIR=os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))), "templates")
```
`genspider` command can be inherited and modified to work with `item_attributes`. It possibly could be invoked from `SfCommand`.

davidhicks:
> Nice POC, thank you for current implementation!
> 
> > Command of genspider (what this POC branch currently names sf) would try to generate spider source code automatically from a provided URL. Output can be piped directly to a new .py file which is then ready to execute (noting a post_process_item method or similar may still need overwriting to clean and modify data).
> 
> Just a few thoughts on this.
> 
> Current [`genspider`](https://docs.scrapy.org/en/latest/topics/commands.html#std-command-genspider) command use templates, custom templates can be defined for each store finder, e.g.
> 
> ```shell
> $ mkdir templates/spiders
> $ touch templates/spiders/agile_store_locator.tmpl
> ```
> 
> ```python
> from locations.storefinders.agile_store_locator import AgileStoreLocatorSpider
> 
> 
> class $classname(AgileStoreLocatorSpider):
>     name = "$name"
>     allowed_domains = ["$domain"]
> ```
> 
> Then modify `settings.py` with:
> 
> ```python
> TEMPLATES_DIR=os.path.join(os.path.dirname(os.path.dirname(os.path.abspath(__file__))), "templates")
> ```
> 
> Then run:
> 
> ```shell
> $ pipenv run scrapy genspider -t agile_store_locator UltraTuneAU https://www.ultratune.com.au/store-locations/
> ```
> 
> `genspider` command can be inherited and modified to work with `item_attributes`. It possibly could be invoked from `SfCommand`.

Thanks for the excellent information and ideas. I hadn't thought that `genspider` was already a command name.

I've had a look at Scrapy's `genspider` and think we should copy a lot from it, but I also think we'd need to re-implement it rather than subclass from it. In particular, I think we should copy the following:
- Using the existing template notations.
- Setting template variables similar to what is done at `https://github.com/scrapy/scrapy/blob/master/scrapy/commands/genspider.py#L122` by asking the store finder class to amend standard template variables (e.g. spider name, Wikidata item, etc).
- Providing an option to save the newly created spider to disk.
- Providing an option to edit the newly created spider after creating it.
- Providing an option to execute the newly created spider after saving it to disk.

Some proposed changes:
- Users should be able to supply either a URL to the storefinder, or a Wikidata Q code.
- If a Wikidata Q code is supplied, our new command should:
  - Retrieve from a new to-be-created Wikidata property a link to the store/location finder page. Then we can check against the store finder classes to see if any of them recognise the store/location finder page being `YextSpider`, `StorepointSpider`, etc. If all of these fail, then see if there are links on the store/location finder page which match with the regex [`P11707`](https://www.wikidata.org/wiki/Property:P11707) to create a `CrawlSpider + StructuredDataSpider`. I'm not sure yet how this would work with store/location finder pages that split out into per-country then per-state pages, which then finally have links to individual store/location pages that we want to parse.
  - Try finding through `robots.txt` / `sitemap.xml` / `.well-known` etc any URLs which match the pattern provided at the Wikidata property [`P11707`](https://www.wikidata.org/wiki/Property:P11707) and check if they contain structured data.
- If a URL to a storefinder page is supplied instead of a Wikidata Q code, use NSI data to try and find the relevant Wikidata Q code, then perform the actions above as if a Wikidata Q code had been supplied.
  - If the NSI data does not allow matching of a URL to Wikidata Q code, then the supplied URL is checked similar to if a Wikidata Q code had been supplied. I'm not sure if it'd be best to ask the user to manually provide the Wikidata Q code, or just use a randomly generated placeholder spider name that the user can edit after the spider has been created.

locations/name_suggestion_index.py
```diff
@@ -35,7 +37,23 @@ def _ensure_loaded(self):
...
+        Attempt to return a single wikidata code corresponding to
+        the brand of the supplied URL.
+        :param url: URL to find the corresponding wikidata code for
+        :return: wikidata code, or None if no match found
+        """
+        self._ensure_loaded()
+        for wikidata_code, brand_parameters in self.wikidata_json.items():
+            for official_website in brand_parameters.get("officialWebsites", []):
+                official_website_domain = urlparse(official_website).netloc
+                supplied_url_domain = urlparse(url).netloc
```
Cj-Malone, line 51: This should probably be outside the loop

Cj-Malone:
I would have done operator in a separate PR. It's a big change, in terms of spiders and pipelines. It needs to be reviewed properly.

locations/pipelines/apply_nsi_categories.py
```diff
@@ -84,6 +87,8 @@ def apply_tags(self, match, item):
         for key, value in match["tags"].items():
             if key == "brand:wikidata":
                 key = "brand_wikidata"
+            if key == "operator:wikidata":
```
Cj-Malone, line 90:
```suggestion
            elif key == "operator:wikidata":
```

Cj-Malone:
You seem to have added a bunch of spiders in here too. I do trust that your spiders are good enough to be added without a real review, but if you're like me, my local uncommitted spiders are because I didn't finish them for various reasons.

There is also a test spider in here. I have the exact same file, we should add test*.py to git ignore.

davidhicks:
> I would have done operator in a separate PR. It's a big change, in terms of spiders and pipelines. It needs to be reviewed properly.

Yep I agree. Will rewrite it as a separate branch and PR.

davidhicks:
Superseded by revised attempt: #6526

PR #6357, opened 2023-11-20
Fix yettel_bg spider

### LINKED ISSUE #6323: yettel_bg is broken
The spider does not return anything. It seems like they've moved the page for getting their store data. I wasn't able to find the new address or how the data reaches their store locator page.

Store locator: https://www.yettel.bg/faq/digital-customer-service/store-locator

Spider: https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/yettel_bg.py

Dimitar5555:
Resolves #6323


Pipfile
```diff
@@ -39,3 +39,4 @@ unidecode = "*"
 pygeohash = "*"
 scrapy-zyte-api = "*"
 cryptography = "*"
+openpyxl = "*"
```
davidhicks, line 42: You'll need to run `pip lock` to update `Pipfile.lock` and then add to this PR.

davidhicks (changes requested):

locations/spiders/yettel_bg.py
```diff
@@ -10,20 +14,45 @@ class YettelBGSpider(Spider):
         "brand_wikidata": "Q14915070",
         "country": "BG",
     }
-    start_urls = ["https://www.yettel.bg/store-locator/json"]
+    start_urls = ["https://www.yettel.bg/dA/1d9e589cca"]
```
Cj-Malone, line 17: The spreadsheet has already changed to ``.

I recommend starting with the store finder and extracting the url from there:
```suggestion
    start_urls = ["https://www.yettel.bg/faq/digital-customer-service/store-locator"]
```

CloCkWeRX:
Conflicts at the moment.

If you need a quick instance, maybe start a code space for this branch. Will be clunky but let you check everything is happy

Dimitar5555:
It seems to work as expected now.

iandees:
It looks like you might have been running something other than Python 3.11 when you updated the Pipfile. Can you double-check that?

Dimitar5555:
> It looks like you might have been running something other than Python 3.11 when you updated the Pipfile. Can you double-check that?

It was 3.10. Should be fixed now.

PR #6361, opened 2023-11-20
Fix Coral Travel PL after website changes

starsep:


iandees:
Looks like the address field is causing an error: https://data.alltheplaces.xyz/ci/6935803642/coral_travel_pl/log.txt

locations/spiders/coral_travel_pl.py
```diff
@@ -2,24 +2,28 @@
...
 
 
-class CoralTravelPLSpider(CrawlSpider, StructuredDataSpider):
+class CoralTravelPLSpider(CrawlSpider):
     name = "coral_travel_pl"
     item_attributes = {"brand": "Coral Travel", "brand_wikidata": "Q58011479"}
-    start_urls = ["https://www.coraltravel.pl/biura-podrozy"]
-    rules = [Rule(LinkExtractor("/biura-podrozy/.+/.+/.+$"), callback="parse_sd")]
+    start_urls = ["https://www.coraltravel.pl/travel-offices"]
+    rules = [Rule(LinkExtractor("/travel-offices/.+/.+$"), callback="parse")]
```
davidhicks, line 12: As this is a regular expression string, it should be prefixed r""?

davidhicks (changes requested):

davidhicks:
> Looks like the address field is causing an error: https://data.alltheplaces.xyz/ci/6935803642/coral_travel_pl/log.txt

Quite an annoying error too--no easy way to fix it other than perhaps check the length of the returned array from `.split()` is as expected, and only attempt to extract fields if the length is as expected. I can't see the address data in the location page in a format that doesn't contain the broken HTML tags, so it looks like the error is in source data in a backend database.

locations/spiders/coral_travel_pl.py
iandees: Fixed in fc895d2.

PR #6362, opened 2023-11-20
Fix Odido PL

starsep:


iandees:
It looks like there's a map with the locations for the stores but they aren't getting saved to the output of the spider. Can you try to extract them?

PR #6364, opened 2023-11-20
Apply autoflake8

starsep:


iandees:
What settings are you running with flake8? pre-commit runs it already, so there shouldn't be any changes.

Cj-Malone:
It might have been isort

iandees:
> It might have been isort

Yea, I thought that might be, but isort should be running/making changes as part of pre-commit too. Maybe the config was different for this run?

Cj-Malone:
I always thought it was isort that removed unused imports... Nope, looks like it's actually autoflake8.

I have IntelliJ set up to run autoflake8, black and isort, and I have noticed a few unused imports coming from Davids spiders.

precommit runs flake8, not autoflake8. I don't know much about the linters, but apparently [autoflake is bad](https://inventwithpython.com/blog/2022/11/19/python-linter-comparison-2022-pylint-vs-pyflakes-vs-flake8-vs-autopep8-vs-bandit-vs-prospector-vs-pylama-vs-pyroma-vs-black-vs-mypy-vs-radon-vs-mccabe/)/unnecessary. But I would like to see these unused imports disappear.

iandees:
I added autoflake to remove unused imports. Let's see what happens!

PR #6405, opened 2023-11-23
Fix category for all Decathlon spiders

panasonic-mafia:
For almost all Decathlon spiders NSI match failed due [to multiple categories for the same wikidata](https://nsi.guide/?t=brands&tt=Q509349). This PR is adding category to `item_attributes` of `decathlon_fr` which all Decathlon spiders use.

### DIFF
--- a/locations/spiders/decathlon_fr.py
+++ b/locations/spiders/decathlon_fr.py
@@ -2,12 +2,13 @@ import re
 
 from unidecode import unidecode
 
+from locations.categories import Categories
 from locations.storefinders.woosmap import WoosmapSpider
 
 
 class DecathlonFRSpider(WoosmapSpider):
     name = "decathlon_fr"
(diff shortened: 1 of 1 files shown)


panasonic-mafia:
Stats: https://data.alltheplaces.xyz/ci/6970220219/decathlon_fr/log.txt

panasonic-mafia:
I debugged `apply_nsi_category` and found that there [is an NSI operator with Decathlon wikidata code](https://nsi.guide/index.html?t=operators&k=leisure&v=pitch&tt=q509349), and because of that NSI mach was failing. Maybe we should handle operators differently...

Cj-Malone:
> Maybe we should handle operators differently...

I agree, some spiders will need to be changed. NSI matching will need to be changed. But it needs to be done.

PR #6408, opened 2023-11-23
[Lifetime] Switch to StructuredDataSpider

Cj-Malone:
To some degree this is just noise. No benefit other than maintainability.


Cj-Malone:
Correcting the brand makes it worth it.

PR #6440, opened 2023-11-30
Add CMA Fuel spider

Cj-Malone:
https://www.gov.uk/guidance/access-fuel-price-data

This is "temporary" dataset for fuel prices, but it is nice for locations too. As it's big brands, it's 100% branded.

### DIFF
--- a/locations/spiders/gov_cma_fuel_gb.py
+++ b/locations/spiders/gov_cma_fuel_gb.py
@@ -1,0 +1,89 @@
+from scrapy import Request, Spider
+
+from locations.categories import Categories, Fuel, apply_category, apply_yes_no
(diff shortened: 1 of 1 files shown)


Cj-Malone:
I wrote it bulk of this [2 months ago](https://gist.github.com/Cj-Malone/044dffc32b87000d0f3c751c8e9f38f2), but didn't do a PR for some reason. I had issues with the data (incorrect fields), and access (Cloudflare) but I contacted CMA, they fixed there stuff straight away and informed the companies about the other issues. Seems fixed now, other than needing to fake a user agent.

iandees:
Can you prefix the pricing tags with something to make it clear that those are prices?

PR #6449, opened 2023-12-02
Implement top-level operator and operator_wikidata fields

davidhicks:
Many locations of interest to AllThePlaces are known primarily by their operator. Examples include fire stations, police stations and hospitals which may be operated by the same public or private organisation on behalf of a government agency. Many other locations are known primarily by their brand, but due to their franchise arrangements, are also known by their franchise owner (operator). For numerous franchises, hundreds of locations may be operated by the same franchise owner.

Treating `operator` and `operator_wikidata` as top-level fields of the `Feature` class makes sense because these fields often carry equal weight ot he existing `brand` and `brand_wikidata` top-level fields. Benefits include better statistics collection on operators, better name-suggestion-index interoperability (where operators and brands are equally weighted) and simplifed spider code.

### DIFF
--- a/locations/exporters/geojson.py
+++ b/locations/exporters/geojson.py
@@ -33,6 +33,8 @@ mapping = (
     ("image", "image"),
     ("brand", "brand"),
     ("brand_wikidata", "brand:wikidata"),
+    ("operator", "operator"),
+    ("operator_wikidata", "operator:wikidata"),
     ("located_in", "located_in"),
     ("located_in_wikidata", "located_in:wikidata"),
     ("nsi_id", "nsi_id"),
--- a/locations/pipelines/apply_nsi_categories.py
+++ b/locations/pipelines/apply_nsi_categories.py
@@ -11,7 +11,7 @@ class ApplyNSICategoriesPipeline:
         if item.get("nsi_id"):
             return item
 
-        code = item.get("brand_wikidata")
+        code = item.get("brand_wikidata", item.get("operator_wikidata"))
         if not code:
             return item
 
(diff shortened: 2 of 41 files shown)


locations/exporters/geojson.py
```diff
@@ -33,6 +33,8 @@
     ("image", "image"),
     ("brand", "brand"),
     ("brand_wikidata", "brand:wikidata"),
+    ("operator", "operator"),
+    ("operator_wikidata", "operator:wikidata"),
```
Cj-Malone, line 37: Should we consider just doing `operator:wikidata`? `brand:wikidata` is still one of my biggest typos when writing spiders.

Cj-Malone: Oh I guess we can't since it's a variable.

locations/pipelines/apply_nsi_categories.py
```diff
@@ -11,7 +11,7 @@ def process_item(self, item, spider):
         if item.get("nsi_id"):
             return item
 
-        code = item.get("brand_wikidata")
+        code = item.get("brand_wikidata", item.get("operator_wikidata"))
```
Cj-Malone, line 14: This is not how I want to do this long term, but I want to re write this code anyway. So fine for now.

Cj-Malone (approved):
I'm sure there will be something we've forgotten, but seems OK. Thank you.

One bit of drama that we will encounter overtime is that we have spiders that output brand data, but they should be outputting operator data. This will continue to "work" as they currently do, they get matched to the operator but still have brand:wikidata. Something like:
```python
{'brand_wikidata': 'Q4829193',
 'extras': {'amenity': 'fire_station'},
 'nsi_id': 'avonfireandrescueservice-ab1fb5',
 'operator': 'Avon Fire and Rescue Service',
 'operator_wikidata': 'Q4829193'}
```

I can't remember any off the top of my head, but we'll need to keep an eye out. They will probably fail to match when I rewrite the NSI matching. (I'm definitely procrastinating)

Cj-Malone:
Thank you David :)

PR #6451, opened 2023-12-03
abercrombie_and_fitch: fix spider

davidhicks:
Try using the worldwide ("wd") API endpoint as this appears to be where the UK endpoint is redirecting towards, at least for most of the world. If worldwide ("wd") redirects back to the UK endpoint in the UK, perhaps this spider needs to allow following the first redirect.


davidhicks:
Requires a proxy too it seems? (data centre IPs blocked?)

PR #6455, opened 2023-12-03
the_keg_steakhouse: fix spider

davidhicks:


davidhicks:
@Cj-Malone As the author of the Playwright middleware, I thought you may be interested in reviewing this more complex usage of Playwright. For this spider, we have to click a "View All" button to trigger some obfuscated/minimised JavaScript (including location data that is obfuscated/minimised) to add more DOM nodes containing links to individual location pages which can then be easily parsed.

locations/settings.py
```diff
@@ -149,6 +149,19 @@
     "DOWNLOADER_MIDDLEWARES": {"locations.middlewares.playwright_middleware.PlaywrightMiddleware": 543},
 }
 
+DEFAULT_PLAYWRIGHT_SETTINGS_WITH_EXT_JS = {
```
Cj-Malone, line 152: Can probably be something like:
```python
DEFAULT_PLAYWRIGHT_SETTINGS_WITH_EXT_JS = DEFAULT_PLAYWRIGHT_SETTINGS | { "PLAYWRIGHT_ABORT_REQUEST": lambda request: not request.resource_type == "document"
    and not request.resource_type == "script"}
```

davidhicks: Thanks for the review! I had tried `DEFAULT_PLAYWRIGHT_SETTINGS.update({..})` without success and totally forgot about `|` which is working fine with `from location.settings import DEFAULT_PLAYWRIGHT_SETTINGS_WITH_EXT_JS`.

PR #6473, opened 2023-12-05
[bridgestone_select] add category

panasonic-mafia:
TBH I would be happy to just use `brand:wikidata=Q179433` and let pipeline to apply category from NSI, but I'm not sure if it's right to use a brand from NSI here. 

@davidhicks what do you think? [It looks like OSM has most of these POIs with `brand:wikidata=Q179433`.](https://overpass-turbo.eu/s/1Elt)

### DIFF
--- a/locations/categories.py
+++ b/locations/categories.py
@@ -100,6 +100,7 @@ class Categories(Enum):
     SHOP_TOYS = {"shop": "toys"}
     SHOP_TRADE = {"shop": "trade"}
     SHOP_TRAVEL_AGENCY = {"shop": "travel_agency"}
+    SHOP_TYRES = {"shop": "tyres"}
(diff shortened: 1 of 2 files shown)


davidhicks (approved):
I think the existing Wikidata item is more correct than what NSI uses, as `Bridgestone` is a brand of tyres, and possibly also fitment/service centres in some countries. In Australia, there is effectively a sub-brand/distinct brand `Bridgestone Select` or `Bridgestone Select Tyre & Auto Service` (if including the additional suffix from signage) that copies from the `Bridgestone` global brand, but adds the `Select Tyre & Auto Service`.

As a result, I'm happy with the changes you've suggested and think it is NSI not ATP that needs changing.

panasonic-mafia:
Thanks for your comment @davidhicks!

PR #6506, opened 2023-12-12
[peugeot_nl] add category

DivyaB1:
Hi,

Adding category to peugeot_nl

Thanks,

### DIFF
--- a/locations/spiders/peugeot_nl.py
+++ b/locations/spiders/peugeot_nl.py
@@ -1,7 +1,7 @@
 import scrapy
 
 from locations.items import Feature
-
+from locations.categories import Categories, apply_category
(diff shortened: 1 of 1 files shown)


panasonic-mafia:
May I kindly ask to do more detailed tagging?

The data has `services` array from where more detailed tagging can be done, e.g. if `shop=car` also provides a repair service it would be nice to include `service:vehicle:car_repair` or if a POI provides only car repair and does not sell cars `shop=car_repar` should be applied.

iandees:
Argh, I meant to leave a comment, not merge. I'm sorry.

I agree with @panasonic-mafia. Please add more detailed tagging.

PR #6508, opened 2023-12-12
[pet_paradise] add category

DivyaB1:
Adding category and wikidata to pet paradise

### DIFF
--- a/locations/spiders/pet_paradise.py
+++ b/locations/spiders/pet_paradise.py
@@ -3,11 +3,12 @@ import json
 import scrapy
 
 from locations.items import Feature
+from locations.categories import Categories, apply_category
 
 
 class PetParadiseSpider(scrapy.Spider):
     name = "petparadise"
-    item_attributes = {"brand": "Pet Paradise"}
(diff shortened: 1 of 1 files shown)


locations/spiders/pet_paradise.py
```diff
@@ -43,4 +44,5 @@ def parse(self, response):
                 "website": store_data["interestpointMoreInfoLink"],
             }
 
+            apply_category({"shop": "pet_grooming"}, properties)
```
panasonic-mafia, line 47: Could you please add this category into `Categories` enum in [categories.py](https://github.com/alltheplaces/alltheplaces/blob/master/locations/categories.py) and then use it from Categories enum?

Cj-Malone: The website looks like it's more focused on https://wiki.openstreetmap.org/wiki/Tag:amenity=animal_boarding. This is one of the reasons I prefer the category in NSI, I have no idea about this brand, but someone in OSM will eventually fix it if it's in NSI, they wont fix it here.

PR #6511, opened 2023-12-12
[lidl_lt][lidl_cy][lidl_ee][lidl_lv] Added new lidl country spiders

ckpiper:
Added some more lidl country spiders.

### DIFF
--- a/locations/spiders/lidl_cy.py
+++ b/locations/spiders/lidl_cy.py
@@ -1,0 +1,27 @@
+import re
+
+from locations.hours import DAYS_ES, OpeningHours, sanitise_day
+from locations.spiders.lidl_gb import LidlGBSpider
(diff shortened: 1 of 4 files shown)


panasonic-mafia:
Since these spiders look exactly the same it makes sense to reuse class of one of the spiders in others.

iandees:
I agree on merging the spiders but merged too quickly. Please add a PR for that.

ckpiper:
@iandees, When you say merge, does that mean merge most countries into one spider? Or as @panasonic-mafia said should I just use one spider as a parent spider and the rest as child spiders. This would be in a separate PR.

panasonic-mafia:
For this case I prefer to have 4 spiders with common class, if one of the spiders errored it's higher probability that we notice that and fix it.

PR #6512, opened 2023-12-12
[ribcrib_us] add category

DivyaB1:
Hi,

Adding category to ribcrib_us.

Thanks,


Cj-Malone:
OSM has these as restaurants.

Cj-Malone:
While adding the category here is helpful, the real fix is to add it to [NSI](https://github.com/osmlab/name-suggestion-index/commit/b812846ea438966f0287c0348169e5bee150dfb9). ATP will then match to it, and the OSM community will keep the category and Wikidata accurate.

iandees:
My rule of thumb for restaurant vs. fast food in OSM is if there's table service. Table service = restaurant, No table service = fast food.

PR #6518, opened 2023-12-14
[lidl] Cleaned up spiders.

ckpiper:
I made lidl_at and lidl_be parent spiders as many others were the almost the exact same except for days, key, dataset_id, and dataset_name.


ckpiper:
I did test all of them. And the log info was all the same. But I guess the Action refuses to test more then 15 spiders at a time.

locations/spiders/lidl_be.py
Cj-Malone, line 1: I think you forgot to inherit LidlATSpider on this one?

Cj-Malone: Oh I see, different format, but you still moved to self.days

panasonic-mafia:
Thanks @ckpiper !

PR #6625, opened 2024-01-09
Make spider count check work

iandees:
As pointed out in #6624, when a PR only deletes spiders we fail the spider-pr build. The change in that PR didn't work because of a technicality with echo + wc commands (it counts the trailing newline as a line so you can never get to zero).

I moved that change to this PR and made the count a separate command that pipes them all together.


.github/workflows/spider-pr.yaml
```diff
@@ -39,21 +39,28 @@ jobs:
...
-        SPIDERS=$(echo $pr_file_changes | jq -r '.[] | select(.status != "removed") | .filename' | grep "^locations/spiders/")
+        SPIDERS=$(echo "${pr_file_changes}" | jq -r '.[] | select(.status != "removed") | .filename' | grep "^locations/spiders/")
         retval=$?
         if [ ! $retval -eq 0 ]; then
             (>&2 echo "checking file changes failed. response was ${pr_file_changes}")
             exit 1
         fi
 
-        if [ $(wc -l <<< "$SPIDERS") -gt 15 ]; then
+        spider_count=$(echo "${pr_file_changes}" | jq -r '.[] | select(.status != "removed") | .filename' | grep "^locations/spiders/" | wc -l | tr -s " ")
```
Cj-Malone, line 53:
```suggestion
        spider_count=$(echo "${pr_file_changes}" | jq -r '[.[] | select(.status != "removed") | select(.filename | startswith("locations/spiders/"))] | length')
```

We should be able to do this all in jq, then we can avoid this drama.

iandees: I think the problem is that something is modifying the response from the `curl` because it detects an authorization header in the diff. This breaks the `jq` and causes the whole thing to exit.

Cj-Malone: If so, it should work with a different test spider?

iandees: Thanks for the jq help. I'm not entirely sure what the problem was, but using the startswith filter seems to have helped.

PR #6674, opened 2024-01-12
fiveguys: expand spider with more countries and fix European locations

davidhicks:
* YextSpider is now used for some European locations: DE, ES, FR and GB.

* Add missing countries to five_guys.

* Rename fiveguys to five_guys.


Cj-Malone:
Did you forget to add fiveguys.py to the commit?

Even though these are the same brand and have the same code, separating into different spiders mean they can be ran in parallel. I don't know how long this takes to run, but I think I recommend splitting out a couple of the big ones.

davidhicks:
Thanks for the feedback, I agree, the spiders are better split. I've split them all up as we do for other large global brands like KFC and McDonalds.

PR #6689, opened 2024-01-12
Add Marathon Petroleum (US) to replace marathon and arco

davidhicks:
Marathon Petroleum owns the Marathon and Arco brands, and the same store finder is used for both brands. Currently the `marathon` and `arco` spiders are broken. These old spiders are now replaced by the single Marathon Petroleum spider this commit introduces.

Fuel types are very confusing and there are 100's of different codes applying to different types of fuels. The matching between codes and OSM fuel tag values is best-effort.


locations/spiders/marathon_petroleum_us.py
```diff
@@ -0,0 +1,85 @@
...
+                        fuel_types_matched.append(Fuel.OCTANE_87)
+                    case "UNLD88":
+                        fuel_types_matched.append(Fuel.OCTANE_88)
```
iandees, line 50: Looks like `Fuel.OCTANE_88` might need to get added

davidhicks: Thanks, I lost that change to categories.py with an accidental git command. Added 88 and also 96 that was missing (completing the full list of 87 through 100 inclusive)

iandees: Thanks for going through all the fuel types. That's a lot of distinct values!

iandees:
Wonder why it's only getting one on Marathon. I see many more in the start url.

davidhicks:
> Wonder why it's only getting one on Marathon. I see many more in the start url.

Doh I forgot to add back `yield item`! I removed it during testing

davidhicks:
> Wonder why it's only getting one on Marathon. I see many more in the start url.

I'm missing the `ref` field... fixing it

davidhicks:
Fixed. I had also accidentally lost that change to DictParser during testing. Hopefully I've got everything now included :)

PR #6692, opened 2024-01-12
oxxo: rename to oxxo_mx as all locations in MX

davidhicks:


davidhicks:
Sorry about this change (better earlier than later). I incorrectly saw some US locations in the geocoded country names, but this was due to inaccurate/incorrect geocoding. All locations show on the map as being in MX only.

iandees:
No problem!

PR #6693, opened 2024-01-12
Add autoflake to remove unused imports

iandees:
Ref #6364


iandees:
pre-commit.ci autofix

PR #6706, opened 2024-01-14
burger_king_ru: fix spider

davidhicks:


davidhicks:
This brand is using Qrator anti-bot blocking per:

```
<!DOCTYPE html>
<html>
<head>
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
 
    <meta name="referrer" content="no-referrer" />
    <script src="/__qrator/qauth_utm_v2_v2dad.js"></script>
 
</head>
<body>
 
</body>
</html>
```

It looks like Zyte is not able to overcome this?

PR #6739, opened 2024-01-16
[Poundland] Drop extra locations

### LINKED ISSUE #6673: Poundland spider returning Pep&Co only stores as both Poundland and Pep&Co
Poundland has a clothing brand called "Pep&Co", and typically their larger stores have a "Pep&Co" area inside, looking a bit like a concession in a department store. In their storefinder data, these stores are indicated by a "Pep Shop" flag. The ATP poundland.py spider looks for this, and when it's found generates an additional "Pep&Co" store item in addition to the "Poundland" one.

So far so good. However, it appears that there are also a few (about 14) "Pep&Co" stores that aren't also regular Poundland stores. For example, these can occur in shopping centres, where Poundland is renting two nearby units and using one as a regular Poundland store (without Pep&Co) and the other as a Pep&Co only store. Currently ATP is returning a "Poundland" entry for the  "Pep&Co" only store. Since there's a separate entry in the storefinder for the actual Poundland store, this leads to duplicates.

Here's an example mapped in OSM:
Poundland - https://www.openstreetmap.org/node/11293224534 - CV21 2JT
Pep&Co - https://www.openstreetmap.org/node/11293224520 - CV21 2JS

Another (unmapped example) is two stores with postcode TS8 0TJ.

I think these  "Pep&Co" only stores can be detected by the branch name (stored in item["branch"]) starting with "Pep & Co". I guess we should test for this, and then not yield the final item if it's found. (In case there's any inconsistency in their data, it would probably be good to add the same test to the` if "Pep Shop" in ` line as an alternative. This would ensure that we don't completely drop a branch who's name starts with "Pep&Co" but doesn't have the "Pep shop" flag set.

Cj-Malone:
Possibly fixes #6673, I want @rjw62 to review.

The drops the numbers:
```patch
 'atp/brand/Pep&Co': 769,
- 'atp/brand/Poundland': 807,
+ 'atp/brand/Poundland': 793,
```

But it leaves us with some ones that I'm questioning. They end up with 3 pois in proximity, a Poundland with it's Pep and a near by Pep (Which used to have a Poundland). They may be bad source data, or they may be valid.

1. B25 8UJ
2. PE21 6EH/PE21 6QX
3. L20 4XX
4. N9 0TZ

### DIFF
--- a/locations/spiders/poundland.py
+++ b/locations/spiders/poundland.py
@@ -12,7 +12,7 @@ class PoundlandSpider(WoosmapSpider):
     def parse_item(self, item: Feature, feature: dict, **kwargs):
         item["branch"] = item.pop("name")
 
-        if "Pep Shop" in feature["properties"]["tags"]:
+        if "Pep Shop" in feature["properties"]["tags"] or item["branch"].startswith("Pep & Co "):
             pep = item.deepcopy()
 
             pep["ref"] = pep["ref"] + "_pep"
@@ -24,6 +24,9 @@ class PoundlandSpider(WoosmapSpider):
             pep["located_in_wikidata"] = self.item_attributes["brand_wikidata"]
 
             yield pep
+
+        if item["branch"].startswith("Pep & Co "):
+            return
 
         apply_yes_no(Extras.ATM, item, "ATM" in feature["properties"]["tags"])
         item["extras"]["icestore"] = "yes" if "Ice Store" in feature["properties"]["tags"] else "no"


rjw62:
I expect the stand-alone Poundland and Pep&Co stores on those will all be correct. I guess the question is whether or not the Poundland stores have a Pep&Co 'concession' within them, given that there's a stand-alone Pep&Co nearby. Maybe they do, maybe the data is wrong, or maybe someone's taken the conscious decision to add "pep shop"=yes to the stores where there's a stand-alone Pep&Co nearby. I don't have any actual information about these stores, so I guess we should take the data at face value leave them in ATP.

PR #6757, opened 2024-01-16
ollies_bargain_outlet: requires proxy due to geo blocking

davidhicks:


davidhicks:
Closed as spider needs more than just a proxy--it also needs rewriting, and this is best done by someone not impacted by geoblocking.

PR #6761, opened 2024-01-16
[WesternUnion] Remove top level tag

Cj-Malone:
Fixes #5889

Issue #5881, opened 2023-09-18
@spider=fedex has point without main tag for convenience shop providing package services


matkoniecz:
Ideally it would be marked that it is not separate POI, but extra property of some existing POI to be attached to. In such case many of tags listed here should be dropped. (or should it be separate POI? in such case main tag is missing, not entirely sure about tagging)

https://www.alltheplaces.xyz/map/#16.82/50.072221/20.041573

![screen01](https://github.com/alltheplaces/alltheplaces/assets/899988/46772969-5a13-4a19-9b09-3eb5bde7a36f)

panasonic-mafia:
I believe it should be a separate POI. Category tag was missing due to failed NSI match because [Fedex NSI](https://nsi.guide/index.html?t=brands&tt=fedex) has multiple categories.

matkoniecz:
If that applies `amenity=post_office` tag then it is still wrong. For reference, right now many small shops in Poland allow picking up packages from one or more delivery services. Adding `amenity=post_office` near them is definitely not a valid tagging.

panasonic-mafia:
> For reference, right now many small shops in Poland allow picking up packages from one or more delivery services.

Ah I see what you mean. 

Ref: 

- https://local.fedex.com/en-pl/warsaw/waw6e
- and POI data for above POI at https://validator.schema.org/#url=https%3A%2F%2Flocal.fedex.com%2Fen-pl%2Fwarsaw%2Fwaw6e

The name of such POIs has 'OnSite' ([more info](https://www.fedex.com/en-pr/shipping/onsite.html)) which may help to identify them from standalone locations. I'm not sure about tagging though. Perhaps `description` tag can be added, something like e.g. `<service name> inside <name of OnSite location>`

Cj-Malone:
In OSM this is probably `post_office=post_partner` as an attribute on another POI. https://wiki.openstreetmap.org/wiki/Key:post_office for these locations.

We can't merge this into an attribute of another POI in ATP. Anyone consuming this data would have to.

Cj-Malone:
> rapid is showing this data to import

Not really, Rapid can show pmtiles now, in the same way JOSM can open GeoJSON.

Personally I hope ATP to becomes a supported dataset, but I envision that as a separate server project. Where we would have rules like you are talking about. I'd exclude any POI without a top level category. I'd exclude spiders like visa, western union and gbfs. I'd exclude loads of tags that don't make sense in an OSM context.

[...]

PR #6763, opened 2024-01-17
[WesternUnion] Remove top level tag

### LINKED ISSUE #5889: @spider=western_union in Poland list `amenity=money_transfer` POIs not actually existing as separate objects
very similar to #5881

It would be better to drop main tag over showing it like this. And in this case it seems dubious to me is it mappable as all on https://www.openstreetmap.org/node/5873034793 bank note.

https://www.alltheplaces.xyz/map/#16.47/50.076332/20.032325

https://location.westernunion.com/pl/malopolskie/krakow/e6d7165e8f86df94dacd8de6f1bfc780

I can visit that place and check in which form Western Union appears there.

matkoniecz:
Fixes #5889

Closes #6761

### DIFF
--- a/locations/spiders/western_union.py
+++ b/locations/spiders/western_union.py
@@ -4,7 +4,6 @@ from scrapy import Spider
 from scrapy.downloadermiddlewares.retry import get_retry_request
 from scrapy.http import JsonRequest
 
-from locations.categories import Categories
 from locations.dict_parser import DictParser
 from locations.geo import point_locations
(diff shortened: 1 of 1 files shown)


matkoniecz:
note to self: if this will be accepted, then look at very similar https://github.com/alltheplaces/alltheplaces/issues/5881

Cj-Malone:
Thank you and welcome @matkoniecz

PR #6821, opened 2024-01-21
[SizzlerUS] Fix spider

Cj-Malone:
Rel: https://github.com/osmlab/name-suggestion-index/pull/9178


davidhicks:
Should this delete/replace the existing `sizzler` spider that returns fewer locations?

PR #6867, opened 2024-01-23
[lidl_at] Added code to delete address information from the name of the POI

ckpiper:
Currently the item name of the POI is grabbed from ShownStoreName key. Now I noticed that all of the values for name from that are the street_address and city of the POI. Thus I think we could either not add a name at all or delete the city and street address from the name if it is in it. Then later the name (Lidl) will be added from NSI as the QCode is added. Would not mind the thoughts. But the value from the ShownStoreName is not the way to go. When I look at a map I want Lidl as the name. Not the address.

What are peoples thoughts on this?

<details><summary>Updated Stats</summary>

```python
{'atp/brand/Lidl': 887,
 'atp/brand_wikidata/Q151954': 887,
 'atp/category/shop/supermarket': 887,
 'atp/field/email/missing': 887,
 'atp/field/image/missing': 887,
 'atp/field/name/missing': 615,
 'atp/field/operator/missing': 887,
 'atp/field/operator_wikidata/missing': 887,
 'atp/field/phone/missing': 887,
 'atp/field/state/missing': 887,
 'atp/field/twitter/missing': 887,
 'atp/field/website/missing': 887,
 'atp/nsi/category_match': 887,
 'downloader/request_bytes': 4056,
 'downloader/request_count': 6,
 'downloader/request_method_count/GET': 6,
 'downloader/response_bytes': 81899,
 'downloader/response_count': 6,
 'downloader/response_status_count/200': 5,
 'downloader/response_status_count/404': 1,
 'elapsed_time_seconds': 6.119464,
 'finish_reason': 'finished',
 'finish_time': datetime.datetime(2024, 1, 23, 17, 0, 8, 451804, tzinfo=datetime.timezone.utc),
 'httpcompression/response_bytes': 1493745,
 'httpcompression/response_count': 5,
 'item_scraped_count': 887,
 'log_count/INFO': 9,
 'memusage/max': 138354688,
 'memusage/startup': 138354688,
 'request_depth_max': 1,
 'response_received_count': 6,
 'robotstxt/request_count': 1,
 'robotstxt/response_count': 1,
 'robotstxt/response_status_count/404': 1,
 'scheduler/dequeued': 5,
 'scheduler/dequeued/memory': 5,
 'scheduler/enqueued': 5,
 'scheduler/enqueued/memory': 5,
 'start_time': datetime.datetime(2024, 1, 23, 17, 0, 2, 332340, tzinfo=datetime.timezone.utc)}
```
</details>


iandees:
Yea, the name value that is produced right now doesn't look right. I would prefer we not set name if it's ambiguous what the it should be.

ckpiper:
@iandees Should we leave it as is? Taking out the address from the name field leaves it empty, meaning apply_nsi_categories pipelines should add the name later on, correct? They all seem to be Lidl supermarkets so the name Lidl would be more correct then just an address I think. 

I agree that setting the name if it is ambiguous. I am curious on your thoughts.

locations/spiders/lidl_pl.py
```diff
@@ -9,3 +11,38 @@ class LidlPLSpider(LidlATSpider):
...
+                item["name"] = item["name"].replace(item["street_address"], "").strip()
+
+        # Takes care of the case where city is in name
+        if item["city"] + "," in item["name"]:
+            item["name"] = item["name"].replace(item["city"] + ",", "").strip()
+        elif item["city"] in item["name"]:
+            item["name"] = item["name"].replace(item["city"], "").strip()
+
+        item["name"] = item["name"].strip()
+
```
Cj-Malone, line 39: IMO, it's not worth it. There is no reasonable branch name, so drop it/don't collect it.

```suggestion
```

Cj-Malone: This may be the same for the other Lidl spiders, I've not looked.

ckpiper:
I checked the other spiders that use lidl_at as a parent and all of them also have the name being some combination of the street_address, district, city, and store_number/reference number. Thus I think it is safe to say that the ShownStoreName is more for internal use as a reference then the name of the store. 

Thus I think it is safe to delete grabbing the value of it. Thus changing results from the following spiders.

- lidl_ch
- lidl_cy
- liidl_cz
- lidl_dk
- lidl_ee
- lidl_es
- lidl_fi
- lidl_gr
- lidl_hu
- lidl_it
- lidl_lt
- lidl_lv
- lidl_nl
- lidl_pl
- lidl_pt
- lidl_sk

iandees:
Thanks!

PR #6880, opened 2024-01-24
Add NSW National Parks and Wildlife Service (AU) (452 locations)

davidhicks:
'atp/category/tourism/camp_site': 360,
 'atp/category/tourism/yes': 92,

tourism=yes sites are not readily taggable, but are various types of accommodation ranging from lodging, chatels, remote huts, hotels, ensuite cabins, etc. See justification in source file for lack of detailed tagging.


davidhicks:
Depends on Categories update in #6878

PR #6913, opened 2024-01-27
Add Little Free Library (72173 locations)

davidhicks:
There is no robots.txt, but there is a rate limit. The default 1 second delay appears to not trigger this rate limit.


iandees:
Can we remove the email contact? It seems like those are private people's emails. While they're already posted on the Little Free Library's website I don't think they provide value to us and I don't want to be a part of making it easier to spam them 😄

locations/spiders/little_free_library.py
```diff
@@ -0,0 +1,54 @@
...
+            "lat": location.get("Library_Geolocation__Latitude__s"),
+            "lon": location.get("Library_Geolocation__Longitude__s"),
+            "street_address": location.get("Street__c"),
+            "city": location.get("City__c"),
+            "state": location.get("State_Province_Region__c"),
+            "postcode": location.get("Postal_Zip_Code__c"),
+            "country": location.get("Country__c"),
+            "operator": location.get("Primary_Steward_s_Name__c"),
+            "phone": location.get("Primary_Steward_s_Phone__c"),
+            "email": location.get("Primary_Steward_s_Email__c"),
```
CloCkWeRX, line 45:
```suggestion
```
As per PR feedback

davidhicks:
Should we just remove the `operator` point of contact entirely including name and phone number?

It looks like their own mobile app is used to report problems with each location, rather than having people email/call individuals who look after each location. See https://littlefreelibrary.org/docs/how-do-i-report-a-little-free-library-that-is-missing-or-damaged/

iandees:
Yea, that sounds good.

davidhicks:
Fixed at last :)

PR #6915, opened 2024-01-28
Add basic devcontainer, so it's easier for new/casual contributors to start

CloCkWeRX:
As you have a dockerfile, this more or less just works - though it's a bit slow.
An alternative would be to base off one of the debian python 3.11 microsoft containers


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for more information:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+# https://containers.dev/guide/dependabot
+
+version: 2
+updates:
+ - package-ecosystem: "devcontainers"
```
iandees, line 9: Does this block existing Dependabot updates?

CloCkWeRX: Security ones run independently - so those *shouldn't* be affected.
If there is an existing dependabot config for general deps, happy to merge this into that.

matkoniecz:
Maybe linking https://containers.dev/supporting somehow or something similar would be nice idea? It took me a bit to figure out how to use devcontainers without using proprietary program like Visual Studio Code.

(to be more specific, I just located Dev Container CLI that seems to be a solution and I am working on getting it up - hopefully it will work)

Issue #6562, opened 2023-12-30
bupa_gb spider is including closed branches


rjw62:
The bupa_gb.py spider is returning a number of closed practices, as these still have web pages and are still listed in the sitemap. Current examples include:

https://www.bupa.co.uk/dental/dental-care/practices/harleston
https://www.bupa.co.uk/dental/dental-care/practices/leckhampton

I think these can be reliably detected by checking if the name ends (case-insensitively) with "closed". So I'd suggest we drop any whose name matches /closed$/i. There are about 12 of these in total out of 388 branches.

(Sorry, I can no longer run the code on my computer to implement and test this myself, since I'm unable to install a recent enough version of Python.)

Cj-Malone:
I've flagged them as closed in #6564 in the method that was discussed for Wilko in #6042 (`end_date=yes`). TT has a strong preference for flagging these, not removing them so they can remove the closed POIs from other datasets.

PR #6925, opened 2024-01-29
mention end_date=yes

matkoniecz:
per https://github.com/alltheplaces/alltheplaces/issues/6562#issuecomment-1872906888 it is kept due to TT request

document this so this will be less surprising (I would not expect nonexisting places to be listed)

### DIFF
--- a/DATA_FORMAT.md
+++ b/DATA_FORMAT.md
@@ -51,7 +51,7 @@ Each GeoJSON feature will have a `properties` object with as many of the followi
 | `located_in`          | The name of the feature that this feature is located in.
 | `located_in:wikidata` | The [Wikidata](https://www.wikidata.org/wiki/Wikidata:Main_Page) [item ID](https://www.wikidata.org/wiki/Help:Items) for the brand or chain of the feature that this feature is located in. This is a machine-readable identifier counterpart for the human-readable `located_in` above.
 | `nsi_id`              | The [Name Suggestion Index](https://nsi.guide/) (NSI) ID for the feature. NSI IDs aren't stable, so you may require [old NSI data](https://github.com/osmlab/name-suggestion-index/tree/main/dist) if you are working with old ATP data.
-| `end_date`              | `end_date=yes` is applied when given location is closed and not operating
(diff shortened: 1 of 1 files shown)


DATA_FORMAT.md
```diff
@@ -51,6 +51,7 @@ Each GeoJSON feature will have a `properties` object with as many of the followi
 | `located_in`          | The name of the feature that this feature is located in.
 | `located_in:wikidata` | The [Wikidata](https://www.wikidata.org/wiki/Wikidata:Main_Page) [item ID](https://www.wikidata.org/wiki/Help:Items) for the brand or chain of the feature that this feature is located in. This is a machine-readable identifier counterpart for the human-readable `located_in` above.
 | `nsi_id`              | The [Name Suggestion Index](https://nsi.guide/) (NSI) ID for the feature. NSI IDs aren't stable, so you may require [old NSI data](https://github.com/osmlab/name-suggestion-index/tree/main/dist) if you are working with old ATP data.
+| `end_date`              | `end_date=yes` is applied when given location is closed and not operating
```
Cj-Malone, line 54: It's actually year-month-day or `yes` if unknown. Spec was decided in https://github.com/alltheplaces/alltheplaces/pull/6042

I think we also talked about things closing in the future, although I don't think there is any data yet.

https://github.com/alltheplaces/alltheplaces/blob/156ec96c13d8dd040b617bf73791b625bec126dd/locations/items.py#L85

matkoniecz: > `end_date=yes` is applied when given location is closed at unknown date and can be assumed to not operate right now, `end_date` may also have values in year-month-day format, including future dates for planned closures.

Though if that is not just for one spider, temporarily, I would consider lifecycle prefix or something else not involving "and this extra field inverts meaning".

Cj-Malone: > > `end_date=yes` is applied when given location is closed at unknown date and can be assumed to not operate right now, `end_date` may also have values in year-month-day format, including future dates for planned closures.
> 
> ?

LGTM, thanks.

matkoniecz: done now

Cj-Malone (approved):

Issue #5888, opened 2023-09-18
@spider=zabka_pl opening hours are missing Sunday data, proposed `opening_hours` are broken


matkoniecz:
https://www.alltheplaces.xyz/map/#16.47/50.073227/20.037421

![screen](https://github.com/alltheplaces/alltheplaces/assets/899988/a19656c7-65ad-4db2-8907-b5e0f2186f26)

https://www.openstreetmap.org/node/4271289403

Opening hours in OSM (`Mo-Sa 06:00-22:00; Su 11:00-20:00`) are correct.

At https://www.zabka.pl/znajdz-sklep it refuses to show Sunday opening hours.

Maybe it is caused by their internal structure? This shops are franchises and maybe they are obligated to be open `Mo-Sa 06:00-22:00` and Sundays can be decided by operator?

Overall it seems that Monday to Saturday data is likely correct.

Cj-Malone:
> Maybe it is caused by their internal structure? This shops are franchises and maybe they are obligated to be open `Mo-Sa 06:00-22:00` and Sundays can be decided by operator?

https://github.com/alltheplaces/alltheplaces/blob/a0bc9ae0f99026e6235a4e71a1b70d83231c3f5c/locations/spiders/zabka_pl.py#L30-L35

https://github.com/alltheplaces/alltheplaces/pull/5586

Sounds like you and @alufers agree, and it sounds like at least some of them have Sunday hours available. It would be nice to collect them, especially when all the Sundays in `nonTradingDays` agree.

matkoniecz:
Not entirely sure what should be done?

Ideally parsing `nonTradingDays` would be nice, but not sure what should be done with shops missing it. Are they closed on Sundays? Missing data and their opening hours should be skipped?

For now I would propose to not emit any opening hours rather than known incomplete ones.

Cj-Malone:
> For now I would propose to not emit any opening hours rather than known incomplete ones.

Interesting, we don't actually output `Su off`. But I acknowledge, in reality, OSM doesn't differentiate. We also don't output any PH data, anyone conflating with OSM should take that into account.

For this specific spider, currently 262 are missing (length 0) `nonTradingDays`, 9,365 have it. Ideally someone with knowledge of the chain could work out if the 262 are closed on Sunday, or missing data.

More than 9,000 currently have the same `nonTradingDays` opening hours, we can collect them. But what happens near Christmas? I imagine that number will be a lot lower, I wouldn't want you to update OSM then.

PR #6933, opened 2024-01-30
add opening_hours info

matkoniecz:
based on #5888 - how it can differ from OSM even if syntax is the same


DATA_FORMAT.md
```diff
@@ -46,7 +46,7 @@ Each GeoJSON feature will have a `properties` object with as many of the followi
 | `contact:twitter`     | The twitter account for the venue. We try to make this specific to the venue and not generic for the brand that is operating the venue.
 | `contact:facebook`    | The facebook account for the venue. We try to make this specific to the venue and not generic for the brand that is operating the venue.
 | **Other**             | _Other information about the venue_
-| `opening_hours`       | The opening hours for the venue. When we can, the format for this field follows [OpenStreetMap's `opening_hours` format](https://wiki.openstreetmap.org/wiki/Key:opening_hours#Examples).
+| `opening_hours`       | The opening hours for the venue. When we can, the format for this field follows [OpenStreetMap's `opening_hours` format](https://wiki.openstreetmap.org/wiki/Key:opening_hours#Examples). Note that in some cases only some days of week are parsed.
```
davidhicks, line 49: I think we should be more specific about the quality of `opening_hours` data and how it might differ from user expectations. For example, the documentation could state that:
- Opening hour strings don't exactly match OSM syntax when time ranges extend across midnight (reference: discussion #4959).
- If a location is not open on a given day, that day is omitted from the opening hours string rather than included with a time range of "off".
- Opening hour information accuracy is only as good as the source data, which is sometimes out-of-date or containing errors that can't be parsed unambiguously.

davidhicks (changes requested):

DATA_FORMAT.md
matkoniecz: > Opening hour information accuracy is only as good as the source data, which is sometimes out-of-date or containing errors that can't be parsed unambiguously.

This probably should go into some general disclaimer/warning section as it applies to every single part of data.

> Opening hour strings don't exactly match OSM syntax when time ranges extend across midnight 

From looking at https://wiki.openstreetmap.org/wiki/Key:opening_hours#Closing_after_midnight it is actually matching it (or wiki is wrong)

> It is fairly common for source data to change hours on a week-by-week basis depending on whether any public holidays exist within the week. In these situations, the source data provides no means to ascertain whether the hours are normal/regular or public holiday hours. Thus it is not always possible to rely upon some brand's opening hours information unless data captured from previous weeks is checked and compared to find the most common (regular) opening hours for a location.

Ooooo, that is a nice trap.

Though I guess that it is getting a bit long - maybe separate more in-depth page would be warranted at this point? Also, it sounds like something that would be ideally done by ATP itself rather by each data consumer separately, but warning is definitely worth adding for now at least.

davidhicks: > This probably should go into some general disclaimer/warning section as it applies to every single part of data.

I agree, although the difference for opening hours is that we should state that if hours for a particular day cannot be unambiguously interpreted from source data, hours for that day will be skipped. Hours for the remainder of the week will be returned as usual if they can be unambiguously interpreted. Thus a day without hours specified could be due to inability to interpret source data _or_ because the source data tells us the location is closed that day.

> From looking at https://wiki.openstreetmap.org/wiki/Key:opening_hours#Closing_after_midnight it is actually matching it (or wiki is wrong)

I think the issue is that OSM users are instructed by the specification on the OSM wiki to treat anything after a semicolon as overwriting hours information supplied previously in the string. Made up example:
- "Mo 22:00-02:00; Tu 06:00-14:00" technically means "Mo 22:00-23:59; Tu 06:00-14:00", but to some OSM consumers not exactly following the specification means "Mo 22:00-23:59; Tu 00:00-02:00, Tu 06-14:00"
- "Mo 22:00-02:00, Tu 06:00-14:00" means "Mo 22:00-23:59; Tu 00:00-02:00, Tu 06-14:00"

> Though I guess that it is getting a bit long - maybe separate more in-depth page would be warranted at this point? Also, it sounds like something that would be ideally done by ATP itself rather by each data consumer separately, but warning is definitely worth adding for now at least.

I agree, there are enough gotchas with opening hours information to warrant a section of the specification to explain in much more detail. Sometimes it is as silly as a plaintext opening hours string on one location's web page containing a typo such as "Weeknds: 06:00-14:00" instead of "Weekends: 06:00-14:00". All other 1000 locations of the brand might not have this typo.

For the regular vs. public holidays ambiguity many brands present, I don't think it is something we can solve during a crawl. It'd have to be post-processed as part of the weekly run which then has the opportunity to compare this week's hours against, for example, the last 4 weekly runs. Then we could add another field during post-processing to differentiate "current_week_hours" (extracted from latest run) from "mode_of_last_5_weeks_hours" (comparison of 5 weekly runs).

matkoniecz: > I think the issue is that OSM users are instructed by the specification on the OSM wiki to treat anything after a semicolon as overwriting hours information supplied previously in the string

Oh yes, semicolon overwrite is a big trap full of unexpected behaviour. Ideal solution is to use comma separation which allows saner formatting.

davidhicks: > Maybe in such cases adding something like `fixme=yes fixme:opening_hours=parsing_failed` would be a good idea?

I wonder whether `Mo-Tu 08:00-16:00; We open; Th-Fr 08:00-16:00; ...` would be better as we'd likely know the location is not closed on the day, but we don't know the hours it is open. Reference: https://wiki.openstreetmap.org/wiki/Key:opening_hours#Summary_syntax

Additionally, `fixme` might indicate to someone there is a possibility that a human could fix the broken source data. I think more often than not this is not the case, and neither a human nor spider could determine the true opening times. Many spiders pick up basic problems like typos and will fix those automatically and effortlessly without a human needing to apply serious thought and consideration.

matkoniecz: > might indicate to someone there is a possibility that a human could fix the broken source data.

I would expect issue tracker to be used for cases where ATP code should be improved.

I thought about `fixme` due to such OSM tag, more in meaning that data users would need to somehow magically fix it (in place survey or something) but should be aware of known defects. Something else like `suspect_data=yes` or `low_quality=yes` or even `inscription=caveat emptor` would still work :)

Just to have some way to communicate to data users - as otherwise there is no way to distinguish data with known failed parsing from data that is likely of higher quality and more trustworthy (but still with no guarantee).

matkoniecz: > I wonder whether Mo-Tu 08:00-16:00; We open; Th-Fr 08:00-16:00; ... would be better as we'd likely know the location is not closed on the day, but we don't know the hours it is open.

For such cases? Yes. Maybe even `Mo-Tu 08:00-16:00; We "parsing failed" Th-Fr 08:00-16:00; ...` if it is unclear is it actually open or not.

matkoniecz: Are you fine with just copying your comments into that new opening hours section (in other words, licensing it on repo license) ?

matkoniecz: Also, I opened "Distinguish parsed opening hours and partially parsed opening hours" #6943

matkoniecz: @davidhicks 

Are you fine with just copying your comments into that new opening hours section (in other words, licensing it on repo license) ?

matkoniecz: @davidhicks

Are you fine with just copying your comments into that new opening hours section (in other words, licensing it on repo license) ?

davidhicks: @matkoniecz Sorry I missed your messages earlier. All my above writing can be treated with a CC-0 license.

matkoniecz: @davidhicks added in https://github.com/matkoniecz/alltheplaces/commit/2954dfc08afd95d9ace585b1014805659d16c86a

matkoniecz:
Moving to draft as I apparently broke the table.

matkoniecz:
I would consider keeping note that

> Day is omitted from the opening hours string when object is closed on that day

is normal for OSM opening_hours syntax and not ATP-specific wrinkle

davidhicks:
Thanks @matkoniecz and @iandees for editing. Easy to read and understand :)

PR #6935, opened 2024-01-30
[LLoydsPharmacy] Delete as no longer has stores

karen-kerr:
Lloyds Pharmacy in the UK has gone online only. Their stores have been sold to a number of other high street pharmacies rather than all becoming another brand.

LLoyds Pharmacy also had a large number of pharmacies within Sainsburys supermarkets. Our sainsburys.py checks for pharmacy but creates nothing, i.e.
`        elif store["store_type"] == "pharmacy":
            return None  # LloydsPharmacyGBSpider`


karen-kerr:
https://www.pharmacy.biz/news/lloydspharmacy-confirms-exit-from-the-high-street/

PR #6924, opened 2024-01-28
Add ruby_tuesday_us spider (216 locations)

sterojos:

PR #6942, opened 2024-01-30
Remove manually added NSI ID from spiders

sterojos:


sterojos:
I manually specified NSI ID for the spiders I created, this is actually counterproductive - see the conversation for #6924.

PR #6946, opened 2024-01-31
[HM] Add hardcoded country list to avoid blocking

sterojos:


sterojos:
This spider seems to get 403'd in weekly runs, but I just ran it locally without problems. There's also another spider (`hm_nl`) which uses the same API and seems to have no issues.

Alternatively, the countries could be hardcoded into the spider.

iandees:
Let's try it with the hardcoded country list?

sterojos:
This works, but there are so many countries that it might be difficult to spot mistakes in the list in the future...

I'm thinking about iterating over all countries from `GeonamesCache` instead. This would waste some requests, but also ensure that no country is left out. 

The API returns an empty `stores` array if an invalid country is requested.

iandees:
If it's just one query per country then I think I'm ok with iterating the whole list.

PR #6957, opened 2024-02-01
[NissanEurope] Add spider (5799 locations)

michaelgrahamevans:


michaelgrahamevans:
As well as South Africa, this also finds dealers in Botswana and Namibia. Is `NissanZA` an acceptable name? All of these dealerships are shown in the [ZA website](https://www.nissan.co.za).

davidhicks:
> As well as South Africa, this also finds dealers in Botswana and Namibia. Is `NissanZA` an acceptable name? All of these dealerships are shown in the [ZA website](https://www.nissan.co.za).

"nissan_za" wouldn't be a good choice if there are other countries mixed in. I prefer "nissan_bw_na_za" as the spider name and file name, and "NissanBWNAZASpider" as the class name, but others may differ in opinion. The other choice I guess is "nissan_africa", "nissan_1", or some other non-country-code suffix.

locations/spiders/nissan_za.py
```diff
@@ -0,0 +1,46 @@
+import scrapy
+
+from locations.dict_parser import DictParser
+from locations.hours import DAYS, OpeningHours
+
+
+class NissanZASpider(scrapy.Spider):
```
davidhicks, line 7: Suggest changing to `class NissanEuropeSpider(scrapy.Spider):`

It appears all the countries listed at "https://www.nissan-europe.com/" are using the same API and thus this spider can obtain dealer locations across dozens of countries.

locations/spiders/nissan_za.py
```diff
@@ -0,0 +1,46 @@
...
+    name = "nissan_za"
+    item_attributes = {"brand": "Nissan", "brand_wikidata": "Q20165"}
```
davidhicks, line 9: Suggestion to add:

```
from locations.categories import Categories

item_attributes = {"brand": "Nissan", "brand_wikidata": "Q20165", "extras": Categories.SHOP_CAR.value}
```

locations/spiders/nissan_za.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def start_requests(self):
+        yield scrapy.Request(
+            url="https://eu.nissan-api.net/v2/publicAccessToken?locale=en_ZA&scope=READ&proxy=*&brand=nissan&environment=prod",
+            headers={"origin": "https://www.nissan.co.za"},
```
davidhicks, line 14: This should iterate across all the combinations of locales and origins for "Find My Dealer" pages linked from each country webpage at https://www.nissan-europe.com/

They all appear to be using the same eu.nissan-api.net API and thus this spider can be readily expanded beyond 3 countries to 20 or whatever it ends up being.

davidhicks (changes requested):
Looks good. My suggested changes are mostly a realisation this spider can be easily expanded to capture Nissan dealers in many more countries.

michaelgrahamevans:
> Looks good. My suggested changes are mostly a realisation this spider can be easily expanded to capture Nissan dealers in many more countries.

Thanks for the review. Let me see what other countries I can add :slightly_smiling_face:

locations/spiders/nissan_za.py
```diff
@@ -0,0 +1,46 @@
+import scrapy
+
+from locations.dict_parser import DictParser
+from locations.hours import DAYS, OpeningHours
+
+
+class NissanZASpider(scrapy.Spider):
+    name = "nissan_za"
```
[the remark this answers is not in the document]
michaelgrahamevans: I am now finding dealers on nearly all continents. Would it be fair to call this spider just `nissan`?

locations/spiders/nissan_za.py
michaelgrahamevans: It turns out that the origin is not too important, only the locale seems to determine what dealers are returned.

I'm now looping over a list of locales I found in the `locale` package. A lot of them get 400 responses and no dealers so the requests feel wasted. Is this a good approach or would it be better to maintain a list of known working locales in the source that we update over time?

locations/spiders/nissan_europe.py
```diff
@@ -0,0 +1,65 @@
...
+            callback=self.parse_dealers,
+            dont_filter=True,
+        )
+
+    def parse_dealers(self, response, **kwargs):
+        for dealer in response.json()["dealers"]:
+            dealer["id"] = dealer.pop("dealerId")
+            item = DictParser.parse(dealer)
+            if item["state"] in ("Botswana", "Namibia"):
+                item["country"] = item.pop("state")
```
michaelgrahamevans, line 40: This is my first ATP spider. Is it right to fixup data like this when I see a mistake or should it rather be mapped as is?

I'll need to find a way to do this for all the new countries too.

davidhicks: Is the problem that the source data `state` field could be a South African 2nd level division OR another country name? They've mashed two data fields `country` and `state` together in one and there is no way to get a state/province in Botswana or Namibia etc?

michaelgrahamevans: Yes that's right

davidhicks (approved):

locations/spiders/nissan_europe.py
```diff
@@ -0,0 +1,69 @@
...
+            item = DictParser.parse(dealer)
+            if item["state"] in self._countries:
+                item["country"] = item.pop("state")
+            for w in dealer["contact"].get("websites", []):
+                item["website"] = w.get("url")
+                break
+            oh = OpeningHours()
+            for day in dealer.get("openingHours", {}).get("regularOpeningHours", []):
+                for interval in day.get("openIntervals", []):
+                    oh.add_range(
```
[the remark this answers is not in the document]
davidhicks: (this is not a blocker to merging, just a comment)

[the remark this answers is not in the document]
michaelgrahamevans: Ooo thanks, I will use this next time

PR #6960, opened 2024-02-01
gamestop: fix spider and split into US, CA, DE and IT spiders

davidhicks:
- Split gamestop into gamestop_us, gamestop_ca, gamestop_de and gamestop_it.
- Update US postcode data.
- Introduce ability to filter postcodes by population (where the source data allows this).


locations/spiders/gamestop_us.py
```diff
@@ -0,0 +1,97 @@
...
+            # yet implemented in ATP and is a complex problem
+            # requiring significant experimentation and
+            # documentation to explain the methodology which others
+            # can reproduce.
+            #
+            # The 2022 Annual Report (https://news.gamestop.com/static-files/fe325562-c087-4fad-809c-efd183364196)
+            # states that there were 2949 open stores at 28 Jan 2023
+            # operating in the US.
+            #
+            # One year later, the following search results were
```
Cj-Malone, line 56: The API looks to be limited by size, 3MB and 1 million chars, not the radius. Pushing the radius up high quickly hits the limit, however 200 radius and 50000 population gives me:

```
 'downloader/response_count': 1017,
 'downloader/response_status_count/200': 1017,
 'item_scraped_count': 2824,
```

davidhicks: Thanks for finding that the radius can be more than 100! And for confirming a population filter of 50000 gives good results :)

I went on a wild hunt for algorithms but found nothing suitable for preparing a minimal set of postcode centroids buffered by a radius, where the resulting circles fully cover the USA (or at least populated areas of the USA). "Point thinning", "Point decimation", and similar approaches were giving me nonsensical results where it'd reduce the number of points, but still keep clusters of 10 or so points all within the same close proximity (no need to keep all 10).

I also implemented a "consolidate_cities" feature in this latest patch that gets close to halving the number of API requests required without impacting results:

30000 population at 200 radius with city consolidation
```
 'downloader/response_count': 2013,
 'downloader/response_status_count/200': 2013,
 'item_scraped_count': 2843,
```

50000 population at 200 radius with city consolidation
```
 'downloader/response_count': 629,
 'downloader/response_status_count/200': 629,
 'item_scraped_count': 2824,
```

iandees:
Nice fix and cleanup!

PR #7072, opened 2024-02-08
[Engen] Add spider (1244 locations)

michaelgrahamevans:


iandees:
https://alltheplaces-data.openaddresses.io/map.html?show=https://alltheplaces-data.openaddresses.io/ci/7836056626/engen/output.geojson

It looks like some of the postcode values are "0". Can you filter those out?

PR #7073, opened 2024-02-08
[SparZA] Add spider (1106 locations)

michaelgrahamevans:
For #346


iandees:
https://alltheplaces-data.openaddresses.io/map.html?show=https://alltheplaces-data.openaddresses.io/ci/7836052799/spar_za/output.geojson

Looks like several of these have really bad lat/lon's, but we can tackle that later.

PR #7077, opened 2024-02-09
Add sutter_health_us spider (574 locations)

impiaaa:


iandees:
Looks good: https://alltheplaces-data.openaddresses.io/map.html?show=https://alltheplaces-data.openaddresses.io/ci/7838753072/sutter_health_us/output.geojson

Thanks!

Issue #7126, opened 2024-02-11
Brow Art 23 (US) (shop/beauty) (AgileStoreLocator) (help wanted)


CloCkWeRX:
### Brand name

Brow Art 23

American beauty parlor chain

### Wikidata ID

Q115675881
https://www.wikidata.org/wiki/Q115675881
https://www.wikidata.org/wiki/Special:EntityData/Q115675881.json

### Store finder url(s)

https://browart23.com/location/

Official Url(s): https://browart23.com/
- [x] I have run the below and if anything in generated, added it.
- [x] I have manually checked for a storefinder
- [x] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q115675881 https://browart23.com/`
```
from locations.storefinders.agile_store_locator import AgileStoreLocatorSpider


class BrowArt23USSpider(AgileStoreLocatorSpider):
        name = "brow_art_23_us"
        item_attributes = {
                "brand_wikidata": "Q115675881",
                "brand": "Brow Art 23",
        }
        allowed_domains = [
                "browart23.com",
        ]
```

CloCkWeRX:
Associated PR does work, but only from certain locations. I can run the spider locally. I can curl locally. From github's actions; vs code code spaces... Timeouts.

This needs some kind of proxy or similar to advance further, but the ones in the project I tried did not work

PR #7134, opened 2024-02-11
Brow Art 23 (US) (shop/beauty) (AgileStoreLocator)

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/7126

For some reason by browser can happily load:
https://browart23.com/wp-admin/admin-ajax.php?action=asl_load_stores&load_all=1
But executing the scraper locally is a timeout


CloCkWeRX:
image
Utterly confused.
From my local machine, happy.
From github codespaces, timeout.

davidhicks:
I think they might be using TLS fingerprinting to block TLS clients that don't look like web browsers?

You could try `is_playwright_spider = True` or if that doesn't work, `requires_proxy = True`, to see if one of those approaches gets around the blocking.

CloCkWeRX:
Tried playwright to no avail, and requires_proxy = True also which is what confuses me further.
It seems odd they would block all of the GitHub ranges

CloCkWeRX:
Going to close for now, as playwright didn't solve this.

iandees:
The `requires_proxy=True` setting won't work unless you set a Zyte API token. The CI running from this repo has it set using donated funds. Github won't share that with PRs running in forks though.

Does the spider make a bunch of requests? If not, we can merge it as is and let it run on the weekly.

davidhicks:
I think you can merge @iandees 

https://alltheplaces-data.openaddresses.io/map.html?show=https://alltheplaces-data.openaddresses.io/ci/7883076133/brow_art_23_us/output.geojson shows that `requires_proxy=True` works. Very few Zyte requests needed

davidhicks (approved):

CloCkWeRX:
Oh, thanks for checking this

PR #7207, opened 2024-02-14
checkers: replace with checkers_rallys_us (798 locations)

davidhicks:
YextSpider is used by both Checkers and Rally's, two fast food chains in America with the same owner. Both brands share the same YextSpider API key.

 'atp/brand/Checkers': 513,
 "atp/brand/Rally's": 285,
 'atp/brand_wikidata/Q63919315': 513,
 'atp/brand_wikidata/Q63919323': 285,


davidhicks:
Checkers tags will only be pulled in correctly after this fix is merged into NSI for the correct Checkers Wikidata ID: https://github.com/osmlab/name-suggestion-index/pull/9244

Issue #7048, opened 2024-02-08
Ultra Liquors (ZA) (shop/alcohol) (28 locations)


CloCkWeRX:
### Brand name

Ultra Liquors

### Wikidata ID

Q116620602

### Store finder url(s)

https://greenpoint.ultraliquors.co.za/store-locator
https://greenpoint.ultraliquors.co.za/sitemap.xml

PR #7220, opened 2024-02-15
[UltraLiquors] Add spider (23 locations)

michaelgrahamevans:
For #7048

### DIFF
--- a/locations/spiders/ultra_liquors_za.py
+++ b/locations/spiders/ultra_liquors_za.py
@@ -1,0 +1,30 @@
+import scrapy
+
+from locations.categories import Extras, apply_yes_no
+from locations.dict_parser import DictParser
+from locations.user_agents import BROWSER_DEFAULT
+
+
+class UltraLiquorsZASpider(scrapy.Spider):
+    name = "ultra_liquors_za"
+    item_attributes = {"brand": "Ultra Liquors", "brand_wikidata": "Q116620602"}
+    allowed_domains = ["greenpoint.ultraliquors.co.za"]
+    user_agent = BROWSER_DEFAULT
+    custom_settings = {"ROBOTSTXT_OBEY": False}
+
+    def start_requests(self):
+        yield scrapy.FormRequest(
+            url="https://greenpoint.ultraliquors.co.za/UltraCityStoreSelector/ListData",
+            formdata={"length": "1000"},
+        )
+
+    def parse(self, response, **kwargs):
+        for data in response.json()["Data"]:
+            data["name"] = data.pop("CompanyName")
+            data["address"] = data.pop("CompanyAddress")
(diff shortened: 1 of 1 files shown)


locations/spiders/ultra_liquors.py
```diff
@@ -0,0 +1,30 @@
+import scrapy
+
+from locations.categories import Extras, apply_yes_no
+from locations.dict_parser import DictParser
+from locations.user_agents import BROWSER_DEFAULT
+
+
+class UltraLiquorsSpider(scrapy.Spider):
```
davidhicks, line 8: All results returned are in South Africa, hence this spider should be renamed to have the `ZA`/`_za` suffix.

locations/spiders/ultra_liquors.py
```diff
@@ -0,0 +1,30 @@
...
+    name = "ultra_liquors"
+    item_attributes = {"brand": "Ultra Liquors", "brand_wikidata": "Q116620602"}
+    allowed_domains = ["greenpoint.ultraliquors.co.za"]
+    user_agent = BROWSER_DEFAULT
+    custom_settings = {"ROBOTSTXT_OBEY": False}
+
+    def start_requests(self):
```
davidhicks, line 15: Possibly a FormRequest is not needed and Request would suffice? If there is a cut-off of results below 1000 unless `length` is specified, I think this FormRequest may need a `method="POST"` parameter too?

Currently all 23 results are returned with a straight no-parameter Request to the supplied URL, thus `start_requests` may not need to be overridden, and `start_urls` could be set instead?

locations/spiders/ultra_liquors.py
```diff
@@ -0,0 +1,30 @@
...
+        yield scrapy.FormRequest(
+            url="https://greenpoint.ultraliquors.co.za/UltraCityStoreSelector/ListData",
+            formdata={"length": "1000"},
+        )
+
+    def parse(self, response, **kwargs):
+        for data in response.json()["Data"]:
+            data["name"] = data.pop("CompanyName")
+            data["street_address"] = data.pop("CompanyAddress")
```
davidhicks, line 24: The `CompanyAddress` field includes suburb and postcode, hence should be mapped to `addr_full` rather than `street_address`.

davidhicks (changes requested):
Looks good. Just a few minor changes I've suggested.

locations/spiders/ultra_liquors.py
michaelgrahamevans: I am seeing that only 10 results being returned when no `length` parameter is sent but the response is misleading as it says there are 23 results.

`"POST"` is the default method set by `FormRequest` if one is not supplied. Should we set it anyway?

davidhicks (approved):

PR #7232, opened 2024-02-15
[TotalEnergies] Fix flipped coordinates

sterojos:
Agip gas stations in Germany have incorrect location:

Agip gas stations

Flipping the coordinates fixes this, other gas stations in Germany are located correctly.

### DIFF
--- a/locations/spiders/total_energies.py
+++ b/locations/spiders/total_energies.py
@@ -1,3 +1,5 @@
+import reverse_geocoder
+
 from locations.categories import (
     Access,
     Categories,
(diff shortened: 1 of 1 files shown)


sterojos:
There seems to be the same issue with BP stations in Poland. I'll first check the data if there aren't any more cases of this.

davidhicks:
Plus a few more:

Cepsa branded locations from Spain (ES) are showing off the coast of Tanzania.

Avia branded locations from Germany (DE) are showing off the coast of Somalia. 


Thanks for the change, good catch on fixing these flipped coordinates :)

PR #7252, opened 2024-02-17
Add The Good Feet Store (worldwide) (253 stores)

### LINKED ISSUE #1649: The Good Feet Store
https://www.goodfeet.com/locations

davidhicks:
Fixes #1649


CloCkWeRX:
If this is getting more stores than locations/spiders/goodfeet.py ; worth removing that one?

davidhicks:
> If this is getting more stores than locations/spiders/goodfeet.py ; worth removing that one?

Thank you, I missed that spider. I tried the same StructuredDataSpider approach but quickly found only some of the 253 location pages have structured data, with the remainder requiring HTML extraction.

PR #7267, opened 2024-02-18
[SparHR] Add spider (136 locations)

sterojos:


sterojos:
https://alltheplaces-data.openaddresses.io/map.html?show=https://alltheplaces-data.openaddresses.io/ci/7950388179/spar_hr/output.geojson

PR #7268, opened 2024-02-18
[Tier] Add spider

Cj-Malone:
I just want to have another think about the pipeline change, make sure I'm not breaking anything.

Brands that aren't in NSI end up with `wikidata_cache["q123"] = []`. This fails the existing `not wikidata_cache.get`, so we loop NSI on every POI, causing quite a slow down on this spider.

### DIFF
--- a/locations/pipelines/apply_nsi_categories.py
+++ b/locations/pipelines/apply_nsi_categories.py
@@ -15,7 +15,7 @@ class ApplyNSICategoriesPipeline:
         if not code:
             return item
 
-        if not self.wikidata_cache.get(code):
+        if code not in self.wikidata_cache:
             # wikidata_cache will usually only hold one thing, but can contain more with more complex spiders
             # The key thing is that we don't have to call nsi.iter_nsi on every process_item
             self.wikidata_cache[code] = list(self.nsi.iter_nsi(code))
--- a/locations/spiders/tier.py
+++ b/locations/spiders/tier.py
@@ -8,7 +8,7 @@ from locations.items import Feature
 
 class TierSpider(Spider):
     name = "tier"
-    item_attributes = {"name": "Tier", "brand": "Tier", "brand_wikidata": "Q63386916"}
+    item_attributes = {"name": "Tier", "operator": "Tier", "operator_wikidata": "Q63386916"}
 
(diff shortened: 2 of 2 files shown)


Cj-Malone:
This doesn't affect most spiders as the cache is populated straight away. I think we are good. @panasonic-mafia have I missed anything?

Cj-Malone:
Tier isn't included in the GBFS spider, hopefully they implement the standard and we can remove this one day.

panasonic-mafia:
> This doesn't affect most spiders as the cache is populated straight away. I think we are good. @panasonic-mafia have I missed anything?

Seems fine to me, thanks!

CloCkWeRX:
@iandees another one that seems ready

iandees:
What is Tier? I see locations present in US cities and they don't line up with any transit stops that I know about.

Cj-Malone:
`amenity=bicycle_rental` or `amenity=kick-scooter_rental`, but it's not possible to tell from the data.

Cj-Malone:
https://www.tier.app/en/

iandees:
Their website doesn't list any locations in the US, but the spider picks up several hundred. Is this aggregating some other kind of data from somewhere?

Cj-Malone:
They seem to have [bought Spin a year ago](https://www.tier.app/en/press/tier-enters-north-america-by-acquiring-micromobility-operator-spin). I'll move it from brand to operator.

iandees:
Got it, thanks for looking.

davidhicks:
@Cj-Malone it looks as if the API key is no longer valid? Their website doesn't appear to have a location finder using a guest API key, so I assume it's necessary to create an account / use the mobile app and then intercept requests relating to a "find my location" feature in order to intercept and extract an API key tied to the specific account?

There additionally is a v2 beta API documented at https://api-documentation.tier-services.io/docs/tier-api-v2-beta/Zones.html#get-all-rootzones which may be replacing the older v1 API this spider was using?

Cj-Malone:
Sorry I missed this.

I have no love for Tier, in fact a bit of disdain because the refuse to use GBFS. I have no objections to deleting this spider and forgetting about them.

PR #7307, opened 2024-02-19
[OsservaprezziCarburantiIT] Add spider (21.4k locations)

sterojos:


sterojos:
There might be some more brands and fuel types to recognize - I'll look into it.

Translated from the website:

> As provided for in the Development Law (Article 51 of Law No. 99 of July 23, 2009), it is, in fact, mandatory for anyone engaged in the business of selling motor fuel to the public for civil use to report to the Ministry the prices charged for each type of fuel (gasoline, diesel, LPG and methane, including that from L-GNC regasification and that dispensed in liquid form LNG) for motor vehicles marketed, for the purpose of their publication on this website.

locations/spiders/osservaprezzi_carburanti_it.py
```diff
@@ -0,0 +1,63 @@
...
+    BRANDS = {
+        "Api-Ip": {"brand": "IP", "brand_wikidata": "Q646807"},
+        "Esso": {"brand": "Esso", "brand_wikidata": "Q867662"},
+        "Q8": {"brand": "Q8", "brand_wikidata": "Q1634762"},
+        "Tamoil": {"brand": "Tamoil", "brand_wikidata": "Q706793"},
+        "AgipEni": {"brand": "Eni", "brand_wikidata": "Q565594"},
+        "Giap": {"brand": "Giap", "brand_wikidata": None, "nsi_id": "giap-e8f712"},
+        "Energas": {"brand": "Energas", "brand_wikidata": None, "nsi_id": "energas-e8f712"},
+        "Keropetrol": {"brand": "Keropetrol", "brand_wikidata": None, "nsi_id": "keropetrol-e8f712"},
+        "San Marco Petroli": {"brand": "San Marco Petroli", "brand_wikidata": None, "nsi_id": "sanmarcopetroli-e8f712"},
```
Cj-Malone, line 30:
```suggestion
        "Giap": {"brand": "Giap", "brand_wikidata": None},
        "Energas": {"brand": "Energas", "brand_wikidata": None},
        "Keropetrol": {"brand": "Keropetrol", "brand_wikidata": None},
        "San Marco Petroli": {"brand": "San Marco Petroli", "brand_wikidata": None},
```

NSI ids aren't stable, we don't hard code them. The real solution to them being missing from the output, it to add wikidata to NSI (and here).

Cj-Malone:
I have a minor preference that we start grouping the government stuff, or the none first party stuff. I've prefixed a couple spiders with `gov_`, we may want a better solution eventually.

Related, if this is published with a specific license, or usage restrictions, please could they be added to `dataset_attributes`.

locations/spiders/osservaprezzi_carburanti_it.py
sterojos: At least I'll learn how to edit NSI 😄

Thanks for the feedback.

locations/spiders/osservaprezzi_carburanti_it.py
```diff
@@ -0,0 +1,63 @@
...
+                {
+                    "points": [{"lat": lat, "lng": lon}],
+                    "radius": 10,
+                }
+            )
+            yield JsonRequest(
+                "https://carburanti.mise.gov.it/ospzApi/search/zone",
+                method="POST",
+                body=payload,
+            )
```
Cj-Malone, line 45:
```suggestion
            yield JsonRequest(
                "https://carburanti.mise.gov.it/ospzApi/search/zone",
                data={
                    "points": [{"lat": lat, "lng": lon}],
                    "radius": 10,
                },
            )
```

`JsonRequest` can automatically encode the data for us, and when it does, it sets the method as well.

locations/spiders/osservaprezzi_carburanti_it.py
sterojos: From what I

Cj-Malone: I'm happy to do it if you want, but I expect the reason they don't have wikidata in NSI is because they don't have wikidata... So you'd be learning how to edit wikidata as well :)

I'll can do them tomorrow if you want.

sterojos: I can do it, if you don't mind it taking a bit longer. There's going to be more brands in the data that are not yet in the spider.

sterojos:
> I have a minor preference that we start grouping the government stuff, or the none first party stuff. I've prefixed a couple spiders with `gov_`, we may want a better solution eventually.
> 
> Related, if this is published with a specific license, or usage restrictions, please could they be added to `dataset_attributes`.

The spider is renamed to follow the convention. Regarding the license, I found [this document](https://download.mise.gov.it/osservaprezzicarburanti/documenti/NoteLegali.pdf), but it seems to concern some "texts", so I'm not sure if it applies.

locations/spiders/osservaprezzi_carburanti_it.py
Cj-Malone: No rush, take as long as you like. I'd probably only do the brands with more than 100 or 200 entries, there are a lot of brands with less than 50 pois in this dataset.

```json
  "atp/gov_osservaprezzi_carburanti_it/unmapped_brand/Ego": 210,
  "atp/gov_osservaprezzi_carburanti_it/unmapped_brand/SanMarcoPetroli": 215,
  "atp/gov_osservaprezzi_carburanti_it/unmapped_brand/Costantin": 232,
  "atp/gov_osservaprezzi_carburanti_it/unmapped_brand/Retitalia": 233,
  "atp/gov_osservaprezzi_carburanti_it/unmapped_brand/Beyfin": 251,
  "atp/gov_osservaprezzi_carburanti_it/unmapped_brand/Shell": 311,
  "atp/gov_osservaprezzi_carburanti_it/unmapped_brand/KEROPETROL": 312,
  "atp/gov_osservaprezzi_carburanti_it/unmapped_brand/Europam": 432,
```
`$ cat osservaprezzi_carburanti_it.json  | jq '[to_entries | .[] | select(.key | startswith("atp/gov_osservaprezzi_carburanti_it/unmapped_brand"))] | sort_by(.value) | from_entries'`

locations/spiders/gov_osservaprezzi_carburanti_it.py
```diff
@@ -0,0 +1,57 @@
...
+            )
+
+    def parse(self, response: Response):
+        for result in response.json()["results"]:
+            item = DictParser.parse(result)
+
+            apply_category(Categories.FUEL_STATION, item)
+
+            for fuel_name, fuel in self.FUELS.items():
+                apply_yes_no(fuel, item, any(f["name"] == fuel_name for f in result["fuels"]))
```
Cj-Malone, line 48: If we flip the loop, we can log failures to improve the mapping. We can also collect the fuel prices, like the cma spider. Something like:

```python
            for fuel in result["fuels"]:
                if fuel["name"] not in self.FUELS:
                    self.crawler.stats.inc_value(
                        "atp/gov_osservaprezzi_carburanti_it/unmapped_fuel/{}".format(fuel["name"])
                    )
                    continue
                key = self.FUELS[fuel["name"]]
                item["extras"]["charge:{}".format(key)] = "{} EUR/1 litre".format(round(fuel["price"], 2))
                apply_yes_no("fuel:{}".format(key), item, True)
```

```python
    FUELS = {
        "Benzina": "petrol",
        "Gasolio": "diesel",
        "GPL": "lpg",
        "Metano": "cng",
        # "Blue Diesel": Fuel.ADBLUE, ?
    }
```

locations/spiders/gov_osservaprezzi_carburanti_it.py
```diff
@@ -0,0 +1,57 @@
...
+
+            if (brand := result["brand"]) == "PompeBianche":
+                # non-branded station
+                yield item
+            if brand in self.BRANDS:
+                item.update(self.BRANDS[brand])
+            else:
+                item["brand"] = brand
```
Cj-Malone, line 56: `yield` doesn't actually stop the function like `return`, so this specific item continues on to get a brand on line 56. If you want the code to stay like this, you'll need to put a `continue` after the `yield`.

Another approach would be to add `PompeBianche` to `BRANDS`, but without a brand. I find a single "return" less confusing.

```suggestion
            brand = result["brand"]
            if brand in self.BRANDS:
                item.update(self.BRANDS[brand])
            else:
                item["brand"] = brand
                self.crawler.stats.inc_value("atp/gov_osservaprezzi_carburanti_it/unmapped_brand/{}".format(brand))

```

```python
    BRANDS = {
        "Api-Ip": {"brand": "IP", "brand_wikidata": "Q646807"},
        ...
        "PompeBianche": {},  # non-branded station
    }
```

Cj-Malone: Or it could end up in the `if`, something like:

```python
            brand = result["brand"]
            if brand in self.BRANDS:
                item.update(self.BRANDS[brand])
            elif brand == "PompeBianche":
                pass  # non-branded station
            else:
                item["brand"] = brand
                self.crawler.stats.inc_value("atp/gov_osservaprezzi_carburanti_it/unmapped_brand/{}".format(brand))
```

Cj-Malone:
This is looking good, thank you for finding this source!

FYI, scrapy can cache requests, I find it vital for developing spiders like this. I have an bash alias, but it ends up something like this: `pipenv run scrapy crawl -s HTTPCACHE_ENABLED=True -s HTTPCACHE_EXPIRATION_SECS=604800 osservaprezzi_carburanti_it`.

locations/spiders/gov_osservaprezzi_carburanti_it.py
sterojos: I believe that this actually worked somehow, but it was probably because the second "yield" got filtered out as a duplicate. Thanks for the correction.

I think the option with the "empty brand" is the most neat. I'll add the stats as well.

locations/spiders/gov_osservaprezzi_carburanti_it.py
sterojos: Neat, thank you!

For some stations there are 2 different prices for the same fuel depending on self service, is there an established way to distinguish between these? 
image

sterojos:
Hopefully adding the missing wikidata IDs in osmlab/name-suggestion-index#9258.

locations/spiders/gov_osservaprezzi_carburanti_it.py
Cj-Malone: Technically OSM can do [conditionals](https://wiki.openstreetmap.org/wiki/Conditional_restrictions), but I doubt anything supports using them, and they can be a pain to write. I wouldn't worry about it too much right now, maybe add a TODO comment.

PR #7322, opened 2024-02-20
[TegutDE] fix link detection

gorgobacka:


sterojos (approved):

Cj-Malone:
For future reference, this probably could have been a CrawlSpider. They end up being a bit more stable for these kind of minor changes.

gorgobacka:
Ok. Thanks for the feedback. I will keep that in mind.

PR #7326, opened 2024-02-21
try better phrasing

matkoniecz:
disclaimer: I am not a native speaker

but it seems much better to me, repeated forward sounded a bit weird

(I hope that it actually improves thing, sorry if it does not)

### DIFF
--- a/docs/SITEMAP.md
+++ b/docs/SITEMAP.md
@@ -69,8 +69,8 @@ https://smashburger.com/locations/us/co/denver/7305-e-35th-avenue/
 ................. and many more!
 ```
 
-That has got us a fair way forward quite quickly. One of the next steps
-forward could be to analyze one of the store URLs for SEO content
+That is a good start! One of the next steps
+could be to analyze one of the store URLs for SEO content
 using our [structured data tooling](./STRUCTURED_DATA.md), or give it to a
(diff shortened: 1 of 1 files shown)


matkoniecz:
BTW

> The ATP tooling described here and on related pages can go a long way to giving you confidence in your end spider before writing any code!

in https://github.com/alltheplaces/alltheplaces/blob/master/docs/STRUCTURED_DATA.md also seems weird (why I would have confidence in not-yet-existing spider?) but I am less sure hidden and have no obvious fixes

CloCkWeRX:
You are right, it's a bit weirdly phrased.

Native speakers would probably rewrite it as "... A further step" or something to avoid the double use of forward, if they focused on it.

But then it depends how far you go on the "vernacular Australia English" (bloody hell we've gotten a long way, and quick smart too, but that's job's not over yet! So buckle up and point your looking gear at this...) vs other dialects.

Note: this is why no one lets me near the en_au.yml i18n files at work though.

matkoniecz:
Looking at what I ended writing it may be also (over)reaction to how often "forward" appears in various corporate/bureaucracy babble (as seen from non-English language country).

iandees:
Thanks for calling these out. I did some more extensive copy editing on this file to make it clearer.

PR #7327, opened 2024-02-21
try to get rid of red marker

matkoniecz:
change from

screen

to

screen01


matkoniecz:
maybe there is a way to shut up Github about long lines

(or maybe requesting to stop linting site-wide is an option, but not sure where it would be done)

PR #7339, opened 2024-02-21
bing_lee_au: uses DataDome so attempt requires_proxy=True

davidhicks:


Cj-Malone:
The base `StructuredDataSpider.wanted_types` doesn't need to hold every possibility, just the most common ones. Although since I fixed the biggest performance issue, I'm not sure how much affect it has.

davidhicks:
> The base `StructuredDataSpider.wanted_types` doesn't need to hold every possibility, just the most common ones. Although since I fixed the biggest performance issue, I'm not sure how much affect it has.

Oh, I've been adding types as I come across them. I wouldn't have thought there is a performance penalty because `LinkedDataParser.iter_linked_data` is seemingly getting a copy of *all* JSON-LD blobs and then `StructuredDataParser.iter_linked_data` is ignoring any JSON-LD blobs that aren't in the default or overridden `wanted_types` list? If there would be a performance penalty, I would have expected the XPath query for `LinkedDataParser.iter_linked_data` to look more like: `script[contains(@type, "ElectronicsStore") or contains(@type, "BikeStore")]`, where only certain JSON-LD blobs are copied.

Cj-Malone:
We used to parse the JSON for every item in the list. I switched the loop semi recently.

I guess there is still the possibility of breaking existing spiders. If we now match new structured data, and output it first it's likely the original data will be dropped as a duplicate. I'm not sure if it's a real issue. But I do wish I never added Place, it's too generic, but I'm scared to remove it because it'll break spiders. And we should never add more common things (PostalAddress)

iandees:
I think I would prefer to keep the list in the spider and not expand the overall list, but also I think we can tackle that later.

Issue #7312, opened 2024-02-20
pipenv install is not sufficient


rkiddy:
I followed the directions. I am on a Ubuntu 22.04.3 LTS system. I ran:
```
$ git clone git@github.com:alltheplaces/alltheplaces.git
$ (installed pipenv)
$ cd alltheplaces
$ pipenv install
$pipenv run scrapy
```
And I got an error.
```
$ pipenv run scrapy
Traceback (most recent call last):
  File "/home/ray/.local/share/virtualenvs/alltheplaces-3iXQ9kuR/bin/scrapy", line 8, in <module>
    sys.exit(execute())
  File "/home/ray/.local/share/virtualenvs/alltheplaces-3iXQ9kuR/lib/python3.10/site-packages/scrapy/cmdline.py", line 128, in execute
    settings = get_project_settings()
  File "/home/ray/.local/share/virtualenvs/alltheplaces-3iXQ9kuR/lib/python3.10/site-packages/scrapy/utils/project.py", line 71, in get_project_settings
    settings.setmodule(settings_module_path, priority="project")
  File "/home/ray/.local/share/virtualenvs/alltheplaces-3iXQ9kuR/lib/python3.10/site-packages/scrapy/settings/__init__.py", line 385, in setmodule
    module = import_module(module)
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 992, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/home/ray/Projects/alltheplaces/locations/__init__.py", line 5, in <module>
    import requests_cache
[...]

PR #7341, opened 2024-02-22
Update readme

iandees:
This makes the readme's install instructions more specific (for Ubuntu and macOS) and adds information about using pyenv.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -3,6 +3,9 @@ url = "https://pypi.python.org/simple"
 verify_ssl = true
 name = "pypi"
 
+[requires]
+python_version = "3.11"
+
 [dev-packages]
 awscli = "*"
 boto3 = "*"
@@ -14,7 +17,6 @@ flake8 = "*"
 pep8-naming = "*"
 isort = "*"
 pre-commit = "*"
-exceptiongroup = "*"
 
 [packages]
 scrapy = "*"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,10 +1,12 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "786c47c535a50257752d58db0ffb1282c22096ed02d0b4c6cffdc9546294a7b0"
+            "sha256": "61b1e0e0d093e37d605ddf39c57369e793baa81483a56f1851a37299721373f9"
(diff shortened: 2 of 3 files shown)


iandees:
Relates to https://github.com/alltheplaces/alltheplaces/issues/7312

matkoniecz:
+1 - though I guess that it introduces expectation of updating this readme for each LTS version

And not sure is it reasonable to add also this, but hint about devcontainer may be a good idea (I have not figured out what Pipenv is changing on my system so I like version with container keeping such changes inside)

iandees:
> it introduces expectation of updating this readme for each LTS version

I think it's a good idea to revisit the install instructions periodically to make sure they still work, so I'm OK with this.

> hint about devcontainer may be a good idea

I'm not sure how that system works. Maybe @CloCkWeRX can suggest a change to the readme that mentions how to use it?

> I have not figured out what Pipenv is changing on my system so I like version with container keeping such changes inside

Python has the concept of [virtual environments](https://docs.python.org/3/tutorial/venv.html), with the goal of keeping your Python dependencies out of the system's python altogether. Pipenv and Pyenv work together to manage virtual environments and the dependencies in them. If you follow the instructions in the readme as-is, your global system should not be affected at all.

CloCkWeRX:
>I'm not sure how that system works. Maybe @CloCkWeRX can suggest a change to the readme that mentions how to use it?

So, codespaces _should_ be a part of the github UI for pretty much everyone.

It's simply a checkout option:
image


Which then launches a web based version of VS Code (you can also have VS Code installed locally and connect to the cloud hosted container)

image

It also executes the content of the Dockerfile; and adds any extras in the devcontainer.json - for example, bash completion is a handy quality of life improvement.

This gives you a cloud development environment which turns itself off after a while of inactivity, so it can be really handy for casual contribution or executing big runs of scraper testing from the cloud; rather than your local machine.

Biggest downsides are accidentally closing your tab when you use ctrl+w in the terminal emulator and after ~3 weeks it forgets your auth tokens

matkoniecz:
Oh  that is this one. (Though I have not used this and do not plan to use it as part of avoiding features specific to proprietary products)

PR #7349, opened 2024-02-22
Add Anti-Bot Detection middleware

davidhicks:
This new middleware is in two parts:

1. Middleware to detect use of anti-bot methods such as Cloudflare Captchas, and various other Web Application Firewalls (WAF) implementing the likes of TLS fingerprinting, browser JS challenges, etc.

2. Middleware to stop a spider from generating more requests, leading to quick (but still deferred) stopping of a crawl once anti-bot methods have been detected.

Spiders can implement a new attribute "anti_bot_methods" which is a list of AntiBotMethod as specified in
locations/middlewares/anti_bot_detection.py

AntiBotDetectionMiddleware will automatically set the attribute if anti-bot method(s) are discovered.

If a particular anti-bot method is unexpected or if Zyte API is not available to use for bypassing the anti-bot method, AntiBotStopCrawl will automatically shut down the crawl in a graceful way, generally after all requests already sent have been received and parsed.

### DIFF
--- a/locations/middlewares/anti_bot_detection.py
+++ b/locations/middlewares/anti_bot_detection.py
@@ -1,0 +1,84 @@
+from enum import Enum
+
+from scrapy import Request, Spider
+from scrapy.downloadermiddlewares.httpcompression import HttpCompressionMiddleware
+from scrapy.http import HtmlResponse, Response
+
+
+# As a temporary transition to use of AntiBotDetectionMiddleware,
+# assume that Zyte API is capable of bypassing all anti-bot
+# methods. This list will need to be amended after testing of Zyte
+# API ability to bypass various anti-bot methods. This list will
+# need to be frequently updated depending on the effectiveness of
+# Zyte API's ability to rapidly bypass new/revised anti-bot methods.
+class AntiBotMethods(Enum):
+    AZURE_WAF = {"name": "Azure WAF", "zyte_bypassable": True}
+    CLOUDFLARE = {"name": "Cloudflare", "zyte_bypassable": True}
+    DATADOME = {"name": "DataDome", "zyte_bypassable": True}
+    HUMAN = {"name": "HUMAN", "zyte_bypassable": True}
+    IMPERVA = {"name": "Imperva", "zyte_bypassable": True}
+    QRATOR = {"name": "Qrator", "zyte_bypassable": True}
+
+
+class AntiBotDetectionMiddleware:
+    def __init__(self, crawler):
+        self.crawler = crawler
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        return cls(crawler)
+
+    def process_response(self, request: Request, response: Response, spider: Spider):
+        if response.headers.get("cf-mitigated"):
+            self.add_anti_bot_method(AntiBotMethods.CLOUDFLARE, spider)
+
+        if response.headers.get("x-datadome"):
+            self.add_anti_bot_method(AntiBotMethods.DATADOME, spider)
+
+        if server := response.headers.get("server"):
+            server_utf8 = self.decode_http_header_value(server)
+            if server_utf8.upper() == "QRATOR":
+                self.add_anti_bot_method(AntiBotMethods.QRATOR, spider)
+            if response.status == 403 and server_utf8.upper() == "MICROSOFT-AZURE-APPLICATION-GATEWAY/V2":
+                self.add_anti_bot_method(AntiBotMethods.AZURE_WAF, spider)
+
+        if cookies := response.headers.getlist("set-cookie"):
+            for cookie in cookies:
+                cookie_utf8 = self.decode_http_header_value(cookie)
+                if cookie_utf8.startswith("visid_incap_") or cookie_utf8.startswith("incap_ses_"):
+                    self.add_anti_bot_method(AntiBotMethods.IMPERVA, spider)
+
+        if response.status == 403:
+            # We may need to decompress the response body before it
+            # can be parsed to detect anti-bot methods. This is
+            # necessary because HttpErrorMiddleware is ordered
+            # before HttpCompressionMiddleware per:
+            # 1. https://docs.scrapy.org/en/latest/topics/settings.html#spider-middlewares-base
+            # 2. https://docs.scrapy.org/en/latest/topics/settings.html#downloader-middlewares-base
+            # It is probably not a good idea to reorder these
+            # default middleware orders.
+            httpcm = HttpCompressionMiddleware.from_crawler(spider.crawler)
+            decompressed_response = httpcm.process_response(request, response, spider)
+            html_response = HtmlResponse(decompressed_response.url, body=decompressed_response.body)
+            if html_response.xpath('//meta[@name="description" and @content="px-captcha"]'):
+                self.add_anti_bot_method(AntiBotMethods.HUMAN, spider)
+
+        return response
+
+    @staticmethod
+    def decode_http_header_value(raw_header_value: bytes) -> str:
+        # It's not quite so simple to decode HTTP header values.
+        # The below approach is simple and will most likely work due
+        # to the encoding of HTTP header values being expected to be
+        # set by anti-bot systems. It is not expected that anti-bot
+        # systems would use different encodings for different sites.
+        return raw_header_value.decode("utf-8")
+
+    @staticmethod
+    def add_anti_bot_method(anti_bot_method: AntiBotMethods, spider: Spider):
+        if not getattr(spider, "anti_bot_methods", None):
+            setattr(spider, "anti_bot_methods", [])
+        if anti_bot_method not in spider.anti_bot_methods:
+            anti_bot_name = anti_bot_method.value["name"]
+            spider.logger.info(f"{anti_bot_name} anti-bot protection has been detected when executing spider {spider.name}.")
+            spider.anti_bot_methods.append(anti_bot_method)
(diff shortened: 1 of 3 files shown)


davidhicks:
Do we need to add to stats as well as raising an INFO log?

For example:

```
...
atp/antibot/azure_waf: True
atp/antibot/human: True
...
```

CloCkWeRX:
This seems sensible. Thinking about your auto detection branch, could this annotate spiders to automatically use a proxy? And I guess second question, should it?

davidhicks:
> This seems sensible. Thinking about your auto detection branch, could this annotate spiders to automatically use a proxy? And I guess second question, should it?

I think this anti-bot detection branch should be integrated with the automatic spider generation branch `sf` command, the existing `sd` command, etc so that these commands can be seamlessly executed via Zyte API if anti-bot methods are detected and the user has a Zyte account configured.

CloCkWeRX:
@iandees or @Cj-Malone any thoughts?

Seems like there's a plan for the next phase of this which I like a lot.

The annotation of this pipeline changing as the zyte API plays whackamole is a bit of a maintenance papercut, where keeping it in sync with the code is tricky or requires humans...
But on the other hand we get indicators for blocking methods without having a human involved. So the maintenance burden overall is reduced.

locations/middlewares/anti_bot_detection.py
```diff
@@ -0,0 +1,86 @@
...
+# API ability to bypass various anti-bot methods. This list will
+# need to be frequently updated depending on the effectiveness of
+# Zyte API's ability to rapidly bypass new/revised anti-bot methods.
+class AntiBotMethods(Enum):
+    AZURE_WAF = {"name": "Azure WAF", "zyte_bypassable": True}
+    CLOUDFLARE = {"name": "Cloudflare", "zyte_bypassable": True}
+    DATADOME = {"name": "DataDome", "zyte_bypassable": True}
+    HUMAN = {"name": "HUMAN", "zyte_bypassable": True}
+    IMPERVA = {"name": "Imperva", "zyte_bypassable": True}
+    QRATOR = {"name": "Qrator", "zyte_bypassable": True}
```
iandees, line 20: If it's easy, can you add comments with links to documentation about these systems?

locations/middlewares/anti_bot_detection.py
```diff
@@ -0,0 +1,86 @@
...
+            decompressed_response = httpcm.process_response(request, response, spider)
+            html_response = HtmlResponse(decompressed_response.url, body=decompressed_response.body)
+            if html_response.xpath('//meta[@name="description" and @content="px-captcha"]'):
+                self.add_anti_bot_method(AntiBotMethods.HUMAN, spider)
+
+        return response
+
+    @staticmethod
+    def decode_http_header_value(raw_header_value: bytes) -> str:
+        # It's not quite so simple to decode HTTP header values.
```
iandees, line 70: Why not? Have you seen non-unicode values in the headers that you're inspecting?

locations/middlewares/anti_bot_detection.py
davidhicks: Can do, it's easy :)

locations/middlewares/anti_bot_detection.py
davidhicks: I haven't looked (just relied on clients automatically decoding header values) but imagine UTF-8 is almost universal, and other encodings are esoteric enough to ignore. I had hoped there was some Scrapy method that handles decoding correctly, but haven't found it yet.

locations/middlewares/anti_bot_detection.py
iandees: Friendly nudge on this.

locations/middlewares/anti_bot_detection.py
```diff
@@ -0,0 +1,103 @@
...
+        if response.status == 403:
+            # We may need to decompress the response body before it
+            # can be parsed to detect anti-bot methods. This is
+            # necessary because HttpErrorMiddleware is ordered
+            # before HttpCompressionMiddleware per:
+            # 1. https://docs.scrapy.org/en/latest/topics/settings.html#spider-middlewares-base
+            # 2. https://docs.scrapy.org/en/latest/topics/settings.html#downloader-middlewares-base
+            # It is probably not a good idea to reorder these
+            # default middleware orders.
+            httpcm = HttpCompressionMiddleware.from_crawler(spider.crawler)
```
Cj-Malone, line 77: This probably shouldn't be initiated on emery request

locations/middlewares/anti_bot_detection.py
```diff
@@ -0,0 +1,103 @@
...
+# need to be frequently updated depending on the effectiveness of
+# Zyte API's ability to rapidly bypass new/revised anti-bot methods.
+class AntiBotMethods(Enum):
+    # Azure Web Applicatigon Firewall bot protection documentation:
+    # https://learn.microsoft.com/en-us/azure/web-application-firewall/ag/bot-protection-overview
+    AZURE_WAF = {"name": "Azure WAF", "zyte_bypassable": True}
+
+    # Cloudflare bot protection documentation:
+    # https://developers.cloudflare.com/bots/
+    CLOUDFLARE = {"name": "Cloudflare", "zyte_bypassable": True}
```
Cj-Malone, line 21: How true is this? We hit a lot of CF sites already with no issues, we log the stats too iirc. I don't know the CF terminology, but there are different levels, we handle some fine, playwright handles some, but does Zyte work for them all? Or are there limits there too? Does it matter?

locations/middlewares/anti_bot_detection.py
```diff
@@ -0,0 +1,103 @@
...
+            self.add_anti_bot_method(AntiBotMethods.DATADOME, spider)
+
+        if server := response.headers.get("server"):
+            server_utf8 = self.decode_http_header_value(server)
+            if server_utf8.upper() == "QRATOR":
+                self.add_anti_bot_method(AntiBotMethods.QRATOR, spider)
+            if response.status == 403 and server_utf8.upper() == "MICROSOFT-AZURE-APPLICATION-GATEWAY/V2":
+                self.add_anti_bot_method(AntiBotMethods.AZURE_WAF, spider)
+
+        if cookies := response.headers.getlist("set-cookie"):
```
Cj-Malone, line 62: I think scrapy is having a discussion about making cookies more accessible, I'll try and find the link tomorrow. We may want to be part of that conversation

locations/middlewares/anti_bot_detection.py
```diff
@@ -0,0 +1,103 @@
...
+            # can be parsed to detect anti-bot methods. This is
+            # necessary because HttpErrorMiddleware is ordered
+            # before HttpCompressionMiddleware per:
+            # 1. https://docs.scrapy.org/en/latest/topics/settings.html#spider-middlewares-base
+            # 2. https://docs.scrapy.org/en/latest/topics/settings.html#downloader-middlewares-base
+            # It is probably not a good idea to reorder these
+            # default middleware orders.
+            httpcm = HttpCompressionMiddleware.from_crawler(spider.crawler)
+            decompressed_response = httpcm.process_response(request, response, spider)
+            html_response = HtmlResponse(decompressed_response.url, body=decompressed_response.body)
```
Cj-Malone, line 79: Not important because html_response goes nowhere, but response.replace would be  better if it was used

locations/middlewares/anti_bot_detection.py
davidhicks: Generally I've only deemed something to be "anti-bot" if there is a CAPTCHA or JavaScript challenge required to be completed by the client. Cloudflare and many other services have confusing marketing, but some will provide the company with a choice of implementing IP address geographic restrictions, data centre IP address blocking, CAPTCHA challenges, JavaScript challenges, rate limiting, etc.

I don't believe Zyte have a status page available where they advertise their current/live status of where they are in the arms race with anti-bot services.  It therefore seems like we'd have to trial and error approach (perhaps even before each crawl commences, if Scrapy allows this) to populate the `zyte_bypassable` values.

locations/middlewares/anti_bot_detection.py
davidhicks: This only triggers on a HTTP 403 error though, which would typically result in a spider terminating its crawl. I don't think we'll see a performance impact as a result, because we wouldn't expect more than a few HTTP 403 errors (>1 because multiple requests could have been fired at once, and perhaps 10 HTTP 403 errors are still being sent back to Scrapy for handling).

AFAIK Azure WAF sometimes only triggers after X number of requests in Y seconds. So the first 20 requests could be fine, then request 21 triggers a HTTP 403 error and CAPTCHA challenge from Azure WAF.

locations/middlewares/anti_bot_detection.py
davidhicks: This is a good suggestion because some spiders may (unlikely though) gracefully handle HTTP 403 errors, for example, to pause and queue a retry to avoid rate limiting mechanisms. In such situation, we'd be decompressing the body of the response two times.

I'd have to look into this deeper because I suspect we may need to replace not just the body (compressed -> uncompressed data) but also modify the headers to omit any mention of the body being compressed.

locations/middlewares/anti_bot_detection.py
davidhicks: Is it https://github.com/scrapy/scrapy/issues/5431 ?

Cj-Malone: Or https://github.com/scrapy/scrapy/pull/6218

CloCkWeRX:
https://github.com/alltheplaces/alltheplaces/issues/10479 lists the spiders that most probably need this

iandees:
Closing as stale — this hasn't had activity in a couple of years. The idea is solid but the scope is large. Feel free to reopen or start a fresh PR if you'd like to pick this up again.

PR #7354, opened 2024-02-23
DictParser: fix reference keys

davidhicks:
Commit e3ecda5778fdff7ba8261ca4fd3a2cc678abfde7 tried to standardise case of reference keys. The idea was that the variation function would try all possibilities of case for each key (camel case, title case, uppercase, lowercase, etc).

Unfortunately it was forgotten that the "ID" acronym in reference keys cannot be handled well by the variations function. It is common for a reference key to always capitalise acronyms, particularly "ID", whilst using some other case for everything else.

This change should fix a number of spiders that were returning 0 or 1 result due to lack of ability to set unique references for each location scraped.

### DIFF
--- a/locations/dict_parser.py
+++ b/locations/dict_parser.py
@@ -2,18 +2,27 @@ from locations.items import Feature
 
 
 class DictParser:
+    # Variations can't handle capitalised acronyms such as "ID" so
+    # the common variants of case including such acronyms need to
+    # all be listed below.
     ref_keys = [
         "ref",
         "id",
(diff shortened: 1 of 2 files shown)


Cj-Malone (commented):
Would you mind adding a test so they aren't removed again. Either in `test_get_variations`, or a new test.

CloCkWeRX:
Tests added, is this now ready?

Cj-Malone:
Thank you

Issue #7360, opened 2024-02-24
Auto detection of hours - can multiple languages be attempted by default?


CloCkWeRX:
Variation of #7242.

OpeningHours has an assumption that DAYS_EN is the set you want to use.

However, with the generated WPStoreLocatorSpider and probably others, I'm noticing that I have to do an extra touch to say "Oh, and this one is in german/french/etc" all of the time:
```
from locations.hours import DAYS_DE
from locations.storefinders.wp_store_locator import WPStoreLocatorSpider


class BarbarossaBaeckereiDESpider(WPStoreLocatorSpider):
    name = "barbarossa_baeckerei_de"
    item_attributes = {
        "brand_wikidata": "Q807766",
        "brand": "Barbarossa Bäckerei",
    }
    allowed_domains = [
        "www.barbarossa-baeckerei.de",
    ]
    days = DAYS_DE
```

Generally I'm just inferring this from:

- The domain name (90% of the time)
- Chrome translation suggestions

Even if we didn't iterate over all, we could attempt the top 10:
https://www.statista.com/statistics/262946/most-common-languages-on-the-internet/
OR
Guess from the domain on the first requests the array of most probable.

The performance impact of this should be fairly low

iandees:
Generally I prefer to make the spider more explicit than to try multiple things implicitly across all spiders. It might mean more typing or one-offs, but that also means the difference is obvious.

CloCkWeRX:
Okay - fair point, but we have a lot of assumptions with the defaulting of says, DAYS_EN in the current codebase.

Yes, English is the most common language on the web (~50%), but it means that at least half the time that assumption is likely to be wrong the first time you try to write a spider.

Would you be comfortable with swapping this to defaulting to None; and having the core spiders raise an exception along the lines of "Default language(s) not defined - choose one of [DAYS_EN, DAYS_FR]" ?

It'd be a pain to refactor the existing spiders; but it would mean for the auto detection use cases; the looser "guess at language" could be run; while for the production execution / by the time its ready for a PR, the spider author has to have locked it in.

PR #7361, opened 2024-02-24
WPStoreLocator: Attempt to detect days by frequency

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/7360

I feel this is slightly better for wordpress instances as:

- English is still tried first
- As soon as a match is detected the assumption is made spider wide that everything else will be in that language, so the iteration cost is only for the first few matches
- Existing spiders that have set the days to a particular value remain unchanged/that is the way to optimise this.

I've only done it to the wpstorelocator spider for now as it tends to hit mainly one url / it's an API.
But this approach should be applicable more widely.

### DIFF
--- a/locations/storefinders/wp_store_locator.py
+++ b/locations/storefinders/wp_store_locator.py
@@ -1,3 +1,4 @@
+import logging
 from scrapy import Selector, Spider
 from scrapy.http import JsonRequest
 
@@ -85,7 +86,7 @@ class WPStoreLocatorSpider(Spider):
                 item["opening_hours"] = self.parse_opening_hours(location, self.days)
             else:
                 # Otherwise, iterate over the possibilities until we get a first match
-                logging.warn(
+                logging.warning(
                     "Attempting to detect opening hours - specify self.days = DAYS_EN or the appropriate language code to suppress this warning"
                 )
                 for days in self.possible_days:
--- a/locations/storefinders/wp_store_locator.py
+++ b/locations/storefinders/wp_store_locator.py
@@ -85,6 +85,7 @@ class WPStoreLocatorSpider(Spider):
                 item["opening_hours"] = self.parse_opening_hours(location, self.days)
             else:
                 # Otherwise, iterate over the possibilities until we get a first match
+                logging.warning(f"Attempting to detect opening hours - specify self.days = DAYS_EN or the appropriate language code to suppress this warning")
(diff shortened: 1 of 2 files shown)


davidhicks:
Automatic detection of language is a good idea. I don't disagree with this proposed change as a first step, but have some thoughts for additional work into the future:

- Implement a new middleware for detecting the HTTP `Content-Language` header, `<html lang="xx">` tag or perhaps even presence of language tags in the URL e.g. `r"^Wen-US^W"` as a regular expression for detecting English (US). This detected language (or languages) would be the first to try/assume for extraction of opening hours day names. Perhaps detected languages could be stored in a new `language` attribute for each spider, and this can be used for other purposes such as determining keys to lookup with `DictParser`.
- Also take a language hint from the name of the spider. If it has a suffix of "_ca_us" then check English day names first. If it has a suffix of "_sg", try English, Malay and Chinese languages first.
- Add automatic language detection to `OpeningHours.add_ranges_from_string()`. If the supplied string has German day names, treat the string as being in the German language unless it has been manually overridden.

iandees:
As I mention in the other issue, I would prefer to make spiders more explicit than to add assumptions and automatic detection. I could see the days-by-frequency and automatic detection used for helping build a spider during an automated detection process, but I'm not excited about every spider that uses this class do all this work every time.

locations/storefinders/wp_store_locator.py
```diff
@@ -79,19 +80,28 @@ def parse(self, response, **kwargs):
             item = DictParser.parse(location)
             item["street_address"] = clean_address([location.get("address"), location.get("address2")])
             item["name"] = location["store"]
-            item["opening_hours"] = self.parse_opening_hours(location)
+            # If we have preconfigured the exact days to use, start there
+            if self.days is not None:
+                item["opening_hours"] = self.parse_opening_hours(location, self.days)
+            else:
+                # Otherwise, iterate over the possibilities until we get a first match
```
CloCkWeRX, line 87:
```suggestion
                # Otherwise, iterate over the possibilities until we get a first match
                logging.warning(f"Attempting to detect opening hours - specify self.days = DAYS_EN or the appropriate language code to suppress this warning")
```
Possible middle ground approach - auto detect but nag the developer?

iandees: Sure, that works.

PR #6526, opened 2023-12-16
Implement automatic storefinder spider generator

davidhicks:

PR #7530, opened 2024-02-25
Yext - Add detection

CloCkWeRX:
Tested on: 
* https://locations.citizensbank.com/index.html
* BWS AU - no longer uses Yext API?
* https://locations.cariboucoffee.com/
* https://www.fnb-online.com/atm-branch-locator

Unfortunately as this library has about 50 different ways to deliver itself and ultimately make an API request to the back end, I'd like to find multiple examples of each technique.
I'm not 100% on the locator-google one.

### DIFF
--- a/locations/storefinders/yext.py
+++ b/locations/storefinders/yext.py
@@ -9,7 +9,6 @@ from locations.dict_parser import DictParser
 from locations.hours import OpeningHours
 from locations.structured_data_spider import clean_facebook
 
-
 # Documentation for the Yext API is available at:
 # 1. https://hitchhikers.yext.com/docs/contentdeliveryapis/introduction/overview-policies-and-conventions/
 # 2. https://hitchhikers.yext.com/docs/contentdeliveryapis/legacy/entities
--- a/locations/storefinders/yext.py
+++ b/locations/storefinders/yext.py
@@ -1,11 +1,14 @@
 import datetime
+from urllib.parse import urlparse
 
-from scrapy import Spider
-from scrapy.http import JsonRequest
+from scrapy import Selector, Spider
+from scrapy.http import JsonRequest, Request, Response
 
+from locations.automatic_spider_generator import AutomaticSpiderGenerator
 from locations.dict_parser import DictParser
 from locations.hours import OpeningHours
 from locations.structured_data_spider import clean_facebook
+
 
 # Documentation for the Yext API is available at:
 # 1. https://hitchhikers.yext.com/docs/contentdeliveryapis/introduction/overview-policies-and-conventions/
@@ -18,7 +21,7 @@ from locations.structured_data_spider import clean_facebook
 # to extract additional location data and to make corrections to automatically extracted location data.
 
 
-class YextSpider(Spider):
+class YextSpider(Spider, AutomaticSpiderGenerator):
     api_key = ""
     api_version = ""
     search_filter = "{}"
@@ -84,3 +87,14 @@ class YextSpider(Spider):
 
     def parse_item(self, item, location, **kwargs):
         yield item
+
+    def storefinder_exists(response: Response) -> bool | Request:
+        # "https://cdn.yextapis.com/v2/accounts/me/entities?api_key=
+        if response.xpath('//script[contains(text(), "window.Yext")]').get():
+            return True
+        return False
+
+    def extract_spider_attributes(response: Response) -> dict | Request:
+        return {
+            "allowed_domains": [urlparse(response.url).netloc],
+        }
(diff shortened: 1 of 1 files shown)


locations/storefinders/yext.py
```diff
@@ -84,3 +86,24 @@ def parse(self, response):
...
+        # https://www.fnb-online.com/atm-branch-locator
+        # As per https://hitchhikers.yext.com/guides/add-searchbar-guide/01-intializethelibrary/
+        if response.xpath('//script[contains(@src, "answers-search-bar")]').get():
+            return True
+
+        return False
+
+    def extract_spider_attributes(response: Response) -> dict | Request:
+        return {
+            "allowed_domains": [urlparse(response.url).netloc],
```
davidhicks, line 108: I see Yext referenced a lot on various brand webpages, but not always for a store locator. In many cases there is reference to Yext libraries, but the store finder doesn't make a call to the expected Yext store locator API endpoint of:

`https://cdn.yextapis.com/v2/accounts/me/entities?api_key={self.api_key}&v={self.api_version}&limit={self.page_limit}&offset={next_offset}&filter={self.search_filter}`

We should only detect Yext if there is a possibility of extracting the required `api_key` value, otherwise the use of Yext is probably unrelated to finding stores, or is possibly in use behind the scenes but the brand website is using an API proxy to hide the key from the public.

davidhicks (changes requested):

locations/storefinders/yext.py
CloCkWeRX: So, I'd be okay dropping the locator-google one.


The "answers search bar" detection I feel has a very strong correlation with being able to find API keys.
IE: https://www.fnb-online.com/atm-branch-locator + others
```
https://liveapi.yext.com/v2/accounts/me/answers/vertical/query?experienceKey=fnb-answers&api_key=10f82fdf7a37ee369b154241c59dade1&v=20190101&version=PRODUCTION&locale=en&input=&verticalKey=locations&limit=20&offset=0&retrieveFacets=true&facetFilters=%7B%7D&session_id=cb9a3c9e-9b5f-4ba7-b09a-f6b0480bdb16&sessionTrackingEnabled=true&sortBys=%5B%5D&referrerPageUrl=&source=STANDARD&jsLibVersion=v1.12.4
```
While this API endpoint is slightly, it does seem to present structured data that should be very similar to the entities endpoint.


window.Yext I have 2x examples of; and both were examples where an API call was subsequently made. However, that was existing storefinders I think.

Happy to make this `window.Yext` && some_further_pattern if you have examples where this gets a hit but doesn't give us anything useful

CloCkWeRX:
TODO: https://github.com/alltheplaces/alltheplaces/issues/7538

davidhicks:
Aaron's (#7547) has an API key at `https://locations.aarons.com/search` which looks like:

```
<link rel="stylesheet" type="text/css" href="[https://assets.sitescdn.net/answers/v1.7/answers.css](https://assets.sitescdn.net/answers/v1.7/answers.css)" /><script>function initAnswers() {ANSWERS.init({apiKey: "62158b834ca95fa3ef64cf0e8327c9c5", experienceKey: "answers", experienceVersion: "PRODUCTION", locale: "en",businessId: "3089660", onReady: function() {ANSWERS.addComponent("SearchBar", {container: ".search_form", name: "search-bar",redirectUrl: "/answers", placeholderText: "Ask us something",});},});}</script><script src="[https://assets.sitescdn.net/answers/v1.7/answers.min.js](https://assets.sitescdn.net/answers/v1.7/answers.min.js)" onload="ANSWERS.domReady(initAnswers)" async defer></script>
```

Example API call returning store locations: `https://cdn.yextapis.com/v2/accounts/me/entities?api_key=62158b834ca95fa3ef64cf0e8327c9c5&v=20240225&limit=50&offset=0&filter={}`

It appears the "Answers" product needs to know of store locations, hence an "Answers" API key can also be used to return store locations.

locations/storefinders/yext.py
davidhicks: From https://www.fnb-online.com/atm-branch-locator:

`<div class="desktop-search" id="search-box-19eafc2b374c463fa09cabc7b6739274" data-component="yext-search" data-apiKey="10f82fdf7a37ee369b154241c59dade1"`

davidhicks:
https://locations.cariboucoffee.com/ is not easy:

First you'd need to load some random minimised JavaScript blob from https://locations.cariboucoffee.com/assets/server/locator-google.0429c4c4.js and then find the Yext API key from `Ve=ke({apiKey:"c328ae6d84635fc2bd9c91497cdeedc0",`

Another brand will have their own obfuscated JavaScript mess within which the API key for Yext is randomly buried.

I'm thinking we need a new generic approach using Playwright where we use https://github.com/scrapy-plugins/scrapy-playwright?tab=readme-ov-file#handling-page-events and https://playwright.dev/python/docs/api/class-page#page-event-request to load a store finder page with Playwright, then wait for subsequent requests to known API endpoint URLs and capture API keys from these requests. That way we don't care about the 100 different ways a brand may implement JavaScript blobs to initiate calls to Yext APIs.

davidhicks:

PR #7532, opened 2024-02-25
Woosmap - Automatic detection

CloCkWeRX:
This one, I got stuck on some of the Vue or React frameworks that automatically load and insert content into the DOM.

Here's where I was starting to look at having a Playwright response, and then looking for the //script tags that had been inserted after page load.

I could use advice on where to next with this one @davidhicks

### DIFF
--- a/locations/storefinders/woosmap.py
+++ b/locations/storefinders/woosmap.py
@@ -1,6 +1,8 @@
 from scrapy import Spider
-from scrapy.http import JsonRequest
+from scrapy.http import JsonRequest, Request, Response
+from urllib.parse import urlparse
 
+from locations.automatic_spider_generator import AutomaticSpiderGenerator
 from locations.dict_parser import DictParser
 from locations.hours import DAYS, OpeningHours
 
@@ -12,7 +14,7 @@ from locations.hours import DAYS, OpeningHours
 # 'https://www.brandname.example'.
 
 
-class WoosmapSpider(Spider):
+class WoosmapSpider(Spider, AutomaticSpiderGenerator):
     dataset_attributes = {"source": "api", "api": "woosmap.com"}
     key = ""
     origin = ""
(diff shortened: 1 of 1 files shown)


locations/storefinders/woosmap.py
```diff
@@ -12,11 +15,25 @@
...
+        # Example: https://www.auchan.pl/pl/znajdz-sklep
+        DetectionRequestRule(url=r"https:\/\/webapp-conf\.woosmap\.com\/(?P<key>[\w-]+)\/webapp-conf\.json"),
+        # Example: https://www.decathlon.fr/store-locator
+        DetectionRequestRule(url=r"https:\/\/api\.woosmap\.com\/stores\?key=(?P<key>[\w-]+)"),
+        DetectionRequestRule(url=r"https:\/\/api\.woosmap\.com\/stores\/search\?key=(?P<key>[\w-]+)"),
+        DetectionRequestRule(url=r"https:\/\/api\.woosmap\.com\/project\/config\?key=(?P<key>[\w-]+)"),
+        DetectionResponseRule(js_objects={"key": "window.woosmap.public_key"}),
+        # detect from https://www.carrefour.fr/magasin/liste
+        # DetectionResponseRule(js_objects={"key": '.. | .woosmapApiKey'}),
+        DetectionResponseRule(js_objects={"key": "window.__INITIAL_STATE__.links.woosmapApiKey"}),
```
CloCkWeRX, line 33: @davidhicks I cannot get this one to work for some reason.

I check it in the browser console, and it should definitely be populated at page load.
But it does not succeed here.

CloCkWeRX:
Ok, neat, some upstream changes. Before I close this, could you take a bit of a look at the variations of this one?

locations/storefinders/woosmap.py
davidhicks: The `__INITIAL_STATE__` thing is non-standard for this store finder and just another example of the horrible mess that front-end JavaScript developers impose on Internet users. In this case, `__INITIAL_STATE__` just means the brand has used React for developing their website, and one of the React patterns is to create a global object containing all the dynamic information that is then mashed into the page once the DOM has loaded. Carrefour FR has customised the way in which the store finder JavaScript classes for Woosmap are loaded and it's unlikely any other brand will have a similar key in `__INITIAL_STATE__`. If another brand was using React and `__INITIAL_STATE__`, they might pick another key such as `window.__INITIAL_STATE__.apis.keys.woosmap`.

For this brand, I think you've just got the wrong store finder page being a list that is not dynamically loaded from Woosmap API, therefore there is no HTTP request to intercept. But you can do the following with the latest `autospidergen` branch...

```
pipenv run scrapy sf https://www.carrefour.fr/magasin
from locations.storefinders.woosmap import WoosmapSpider


class BonAppFRSpider(WoosmapSpider):
    name = "bon_app_fr"
    item_attributes = {
        "brand_wikidata": "Q90153100",
        "brand": "Bon App!",
    }
    key = "woos-26fe76aa-ff24-3255-b25b-e1bde7b7a683"
    origin = "https://www.carrefour.fr"
```

davidhicks: Which reminders me that I need to take another look at how we're guessing the correct NSI entry to extract the brand details from. In this case there are many NSI entries all for `carrefour.fr` so this automatic spider detection code usually just refuses to pick one, leaving it up to the user to decide which NSI entry is correct. I'm not sure why `Bon App!` was selected over other NSI entries in this case... I'll add it to my TODO list

PR #7533, opened 2024-02-25
Uberall - Autodetect

CloCkWeRX:
Only 1x of the methods here works without JS.
However, on the off chance someone manually loads these bundles in to the DOM, it will work.

### DIFF
--- a/locations/storefinders/uberall.py
+++ b/locations/storefinders/uberall.py
@@ -1,13 +1,16 @@
 import logging
 
-from scrapy import Spider
-from scrapy.http import JsonRequest
+from urllib.parse import urlparse
 
+from scrapy import Spider
+from scrapy.http import JsonRequest, Request, Response
+
+from locations.automatic_spider_generator import AutomaticSpiderGenerator
 from locations.dict_parser import DictParser
 from locations.hours import DAYS, OpeningHours
 
 
-class UberallSpider(Spider):
+class UberallSpider(Spider, AutomaticSpiderGenerator):
     dataset_attributes = {"source": "api", "api": "uberall.com"}
 
     key = ""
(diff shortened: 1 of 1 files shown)


locations/storefinders/uberall.py
```diff
@@ -50,3 +52,26 @@ def parse(self, response, **kwargs):
...
+        # <script type="module" src="https://locator.uberall.com/locator-assets/store-finder-widget-bundle-v2-modern.js?b=My4xODQuMjU="></script>
+        if response.xpath(
+            '//script[contains(@src, "https://locator.uberall.com/locator-assets/store-finder-widget-bundle-v2-modern.js")]'
+        ).get():
+            return True
+
+        return False
+
+    def extract_spider_attributes(response: Response) -> dict | Request:
+        return {
```
davidhicks, line 75: We should be returning `key` only.

Some examples:

https://www.yves-rocher.it/trova-il-tuo-negozio#! has the key provided as:

```<div id="store-finder-widget" data-key="HLGRPp968JZaR0D235dXJa5fMRPHuA"```

https://www.aldi.pl/informacje-dla-klienta/wyszukiwarka-sklepu.html has the key provided as:

```<div data-showbacklink="false" id="store-finder-widget" data-key="ALDINORDPL_3Lul0K0H0Vy47Jz0wGZlN30mchboaO"```

File: locations/storefinders/uberall.py [hunk 2]
```diff
@@ -50,3 +52,26 @@ def parse(self, response, **kwargs):
 
     def parse_item(self, item, feature, **kwargs):
         yield item
+
+    def storefinder_exists(response: Response) -> bool | Request:
```
davidhicks, line 56: I also observe some other reliable clues that Uberall is being used:

`<div id="store-finder-widget" data-key=`

`<div class="mod-uberall-store-locator" data-t-name="UberallStoreLocator"`

davidhicks (changes requested):

CloCkWeRX:
Okay. I'd prefer to still keep the other indicators in there, but have put the div + key detection as 'primary'.

If/when we have playwright around for this, I'd like to revisit and detect things being added to the DOM / run JS to discover keys.

davidhicks (approved):

PR #7534, opened 2024-02-25
SweetIQ - Autodetection

CloCkWeRX:
Finally, one that seems straight forward


locations/storefinders/sweetiq.py
```diff
@@ -73,3 +76,18 @@ def parse(self, response):
 
     def parse_item(self, item, location, **kwargs):
         yield item
+
+    def storefinder_exists(response: Response) -> bool | Request:
+        # Example: https://locations.thepaperstore.com/
+        if response.xpath('//script[contains(text(), "__SLS_REDUX_STATE__")]').get():
+            return True
+
+        if response.xpath('//script[contains(@src, "sls-cdn.sweetiq.com"]').get():
```
CloCkWeRX, line 85:
```suggestion
        if response.xpath('//script[contains(@src, "sls-cdn.sweetiq.com")]').get():
```

locations/storefinders/sweetiq.py
```diff
@@ -73,3 +76,18 @@ def parse(self, response):
...
+            return True
+
+        if response.xpath('//script[contains(@src, "sls-cdn.sweetiq.com")]').get():
+            return True
+
+        return False
+
+    def extract_spider_attributes(response: Response) -> dict | Request:
+        return {
+            "allowed_domains": [urlparse(response.url).netloc],
```
davidhicks, line 92: The SweetIQ spider needs `start_urls` defined as the full URL to the location page where `response.xpath('//script[contains(text(), "__SLS_REDUX_STATE__")]').get()` is not `None`.

I think the extracted attribute then is simply just `"start_urls": [response.url]`?

davidhicks (changes requested):

locations/storefinders/sweetiq.py
CloCkWeRX: Fixed/pushed

davidhicks: I think setting `allowed_domains` here to just the response domain name could break things, because in `SweetIQ.parse_locations_list` the API endpoint URL is extracted from the store locator page. I think we also need to copy `SweetIQ.parse_locations_list` in extracting the relevant API endpoint URL (domain name) here too.

davidhicks (changes requested):

CloCkWeRX:
@davidhicks got rid of the allowed domains in favour of start_urls

CloCkWeRX:
oops, missed ". I think we also need to copy SweetIQSpider.parse_locations_list in extracting the relevant API endpoint URL (domain name) for extract_spider_attributes too."

davidhicks:
Don't worry about further work on this one as I plan to replace the detection/extraction method with a Playwright approach where the API key (for example `u2C58HFygZ11EkWE4Hvk06TqBzuGOl`) is extracted from requests made from the store locator pages to the likes of `https://sls-api-service.sweetiq-sls-production-east.sweetiq.com/u2C58HFygZ11EkWE4Hvk06TqBzuGOl/locations-details?locale=en_US&ids=2772318%2C2772368%2C2772327%2C2772322%2C2772336%2C2772356%2C2772330%2C2772310%2C2772358%2C2772311%2C2772353%2C2772320%2C2772292%2C2772294%2C2772348%2C2772299%2C2772333%2C2772343%2C2772286%2C2772349%2C3085046&clientId=59d2489ea95c7f5d2eb7385a&cname=locations.thepaperstore.com`

This will be much more robust than trying to find API keys in HTML attributes or JavaScript blobs.

davidhicks:

PR #7536, opened 2024-02-25
StoreRocket.io - Add Autodetection

CloCkWeRX:


locations/storefinders/storerocket.py
```diff
@@ -41,3 +44,25 @@ def parse(self, response, **kwargs):
 
     def parse_item(self, item: Feature, location: dict, **kwargs):
         yield item
+
+    def storefinder_exists(response: Response) -> bool | Request:
+        # Example: # https://barre3.com/studio-locations
```
davidhicks, line 49: For this one, I think we'd have to inject JavaScript into a Playwright page for this URL to extract the key from object `Window.StoreRocket.configs.projectId` which in this case is `jN49m3n4Gy`.

locations/storefinders/storerocket.py
```diff
@@ -41,3 +44,25 @@ def parse(self, response, **kwargs):
...
+        # Example: https://soapynoble.com/
+        # Example: https://ribcrib.com/locations/
+        if response.xpath('//div[@id="storerocket-widget"]').get():
+            return True
+
+        return False
+
+    def extract_spider_attributes(response: Response) -> dict | Request:
+        return {
+            "allowed_domains": [urlparse(response.url).netloc],
```
davidhicks, line 67: We should be returning `storerocket_id` here (or rename this attribute to `api_key`).

For example https://soapynoble.com/
`<div id='storerocket-widget' style='width:100%;' data-storerocket-env='p' data-storerocket-id='vo8xZwy4gn'>`

For example https://ribcrib.com/locations/
`<div id='storerocket-widget' style='width:100%;' data-storerocket-env='p' data-storerocket-id='6wgpr528XB'>`

davidhicks (changes requested):

CloCkWeRX:
Addressed feedback. As I dont have a playwright detection spider (yet), I added the further work as a TODO

PR #7549, opened 2024-02-25
Add two quick notes about getting started quickly

CloCkWeRX:
Aimed at casual contributors


iandees:
I think I'd like for these to be more specific (i.e. "git clone ...", "docker build ...", "docker run ..." for the Docker section) and maybe to have text that helps decide between the options.

I'm happy to make those changes a bit later this afternoon.

PR #7566, opened 2024-02-25
Amasty - Autodetect

CloCkWeRX:


CloCkWeRX:
I'm missing something obvious here - for both my test urls, that xpath doesn't work. But it _should_ be right.

locations/storefinders/amasty_store_locator.py
```diff
@@ -52,3 +56,16 @@ def parse_items(self, items: [dict]):
 
     def parse_item(self, item, location, popup_html):
         yield item
+
+    def storefinder_exists(response: Response) -> bool | Request:
+        # Example: https://www.aussiedisposals.com.au/store-locator/
+        # Example: https://www.lcbo.com/en/stores/
+        if response.xpath('//script[contains(@text, "/amlocator/index/ajax/")]/@text').get():
```
davidhicks, line 63: `@text` is probably meant to be `text()`?

davidhicks (changes requested):

locations/storefinders/amasty_store_locator.py
```diff
@@ -52,3 +56,16 @@ def parse_items(self, items: [dict]):
...
+        # Example: https://www.aussiedisposals.com.au/store-locator/
+        # Example: https://www.lcbo.com/en/stores/
+        if response.xpath('//script[contains(text(), "/amlocator/index/ajax/")]/text()').get():
+            return True
+
+        return False
+
+    def extract_spider_attributes(response: Response) -> dict | Request:
+        return {
+            "allowed_domains": [urlparse(response.url).netloc],
```
davidhicks, line 70: Are you able to check with the existing `mephisto` spider which uses `start_urls` instead of `allowed_domains`. Is it possible to get a custom URL such as `https://www.mephisto.com/be-fr/amlocator/index/ajax/` from `response.xpath('//script[contains(text(), "/amlocator/index/ajax/")]/text()').get()`?

CloCkWeRX: Most appear to do a POST as the first request exactly to the endpoint
ie:
```
    def start_requests(self):
        # The request won't work without the headers supplied below.
        headers = {
            "X-Requested-With": "XMLHttpRequest",
        }
        for domain in self.allowed_domains:
            yield Request(url=f"https://{domain}/amlocator/index/ajax/", method="POST", headers=headers)
```

davidhicks: The issue in the case of `mephisto` is the API URL has `/be-fr/` in the path hence `f"https://{domain}/amlocator/index/ajax/"` is not correct for `mephisto`. I wonder whether the API URL can be detected from the HTML/JS content of the store finder page, rather than just assuming it is always `f"https://{domain}/amlocator/index/ajax/"`?

davidhicks: Perhaps a new spider attribute would be helpful to define whether the instance needs the `POST` method (that also adds the `X-Requested-With` header) or whether it needs a plain `GET` method instead?

CloCkWeRX: (I will have to look at this tonight/later, no dev environment at the moment)

CloCkWeRX: https://www.mephisto.com/be-fr/points-de-vente/ - 404

CloCkWeRX: I guess we can rip out:
            ajaxCallUrl: "https://www.mephisto.com/be-fr/amlocator/index/ajax/",

davidhicks: I just tried https://github.com/alltheplaces/alltheplaces/tree/autospidergen with AmastyStoreLocatorSpider modified to subclass additionally from AutomaticSpiderGenerator, and to have a `storefinder_exists` method always return True. The example of `pipenv run scrapy sf https://www.aussiedisposals.com.au/store-locator/` then works for me in returning a template for a new spider.

davidhicks:
Just to give you an update @CloCkWeRX on where I'm at with the Playwright approach to storefinder detection, here is an example of how AmastyStoreLocatorSpider would be updated for automatic detection:

```
class AmastyStoreLocatorSpider(Spider, AutomaticSpiderGenerator):
    detection_url_rules = [
        r"^https?:\/\/(?P<allowed_domains>[A-Za-z0-9\-.]+)\/amlocator\/index\/ajax\/?",
        r"^(?P<start_urls>https?:\/\/[A-Za-z0-9\-.]+(?:\/[^\/]+)+\/amlocator\/index\/ajax\/?)$",
    ]
```

The named matching groups `<allowed_domains>` and `<start_urls>` would become attributes of the generated spider. I need to figure out a way to denote the type of an attribute, perhaps one of:
1. `<identifier_name__list>` for a list type, `<identifier_name__int>` for an integer type, and `<identifier_name>` for a string type.
2. Storefinder classes need to override a method to take extracted named groups and perform whatever type conversions may be necessary to apply attributes to the generated spider.

I've also got a bit more work to do with:
1. Allowing rules to be specified for matching/extracting POST data and headers of all pages requested in the Playwright session.
2. Allowing Scrapy Selector (XPath or CSS) rules to be specified for matching/extracting HTML content of the storefinder page.
3. Allowing a storefinder class to interact with a storefinder page of a brand in the Playwright session. For example, before an interceptable API request is generated exposing an API key, a user may need to enter a value in a location search text field and click a "Find nearest stores" button.

locations/storefinders/amasty_store_locator.py
CloCkWeRX: Weird. This branch is definitely wrong on my local. Let me start over.

CloCkWeRX:
kk. Well, I think this is "working" using the older approach for now; so if it's useful great, if it's replaced by the playwright version, can just close.

davidhicks:

PR #7568, opened 2024-02-25
[Pipelines] Add pipeline to clean addresses

Cj-Malone:
Rel: https://github.com/alltheplaces/alltheplaces/pull/7355#discussion_r1501312149

TODO:
- [x] Check this actually improves the addresses (Will check against Weekly that is currently running)
- [x] Migrate spiders from `vapestore_gb.clean_address`


davidhicks:
I haven't been using `vapestore_gb.clean_address` but have bee using a lot of similar code throughout spiders that can also be cleaned up.

I wonder whether we should allow `Item.street_address` and `Item.addr_full` to both be ordered lists, as well as strings (current situation). A spider could populate `Item.street_address = [ address1, address2]` which this new pipeline can then join together into a single string? Maybe this just makes things way more messy because the order of this new pipeline could get quite tricky.

locations/pipelines/address_clean_up.py
```diff
@@ -0,0 +1,36 @@
...
+            if line:
+                return_addr.append(line)
+
+    return ", ".join(return_addr)
+
+
+class AddressCleanUpPipeline:
+    def process_item(self, item: Feature, spider: Spider):
+        if street_address := item.get("street_address"):
+            if isinstance(street_address, str):
```
davidhicks, line 31: This is what I read first, not realising that `clean_address` has been designed to accept lists of strings. Should an `isinstance(street_address, list)` be added here too as an OR condition?

locations/pipelines/address_clean_up.py
```diff
@@ -0,0 +1,36 @@
...
+                item["street_address"] = clean_address(street_address)
+        if addr_full := item.get("addr_full"):
+            if isinstance(addr_full, str):
```
davidhicks, line 34: Should an isinstance(street_address, list) be added here too as an OR condition?

davidhicks (changes requested):

Cj-Malone:
I wish we had done things like `phone` and `email` as lists, then `;` joined on export. It would have made things nicer. However right now, I think it's too much work to ensure everything works. Maybe in the future.

Cj-Malone:
I think it improves the output, 79,677 changes to `addr_full` and 148,346 `street_address`.
The full list is [here](https://gist.github.com/Cj-Malone/3e98587c302ce4241fb230e8d9d511ac), but it's as you'd expect, eg:

```patch
-IOMEGA (M) SDN BHD , PLOT 44, BAYAN LEPAS INDUSTRIAL PARK 4,
+IOMEGA (M) SDN BHD, PLOT 44, BAYAN LEPAS INDUSTRIAL PARK 4
```

Cj-Malone:
I specifically made `clean_address` compatible, taking the lists so we wouldn't have to change existing spider code. But then instead of doing the automatic refactor, I spend my Friday evening manually changing every spider, dropping it when I could, moving directly to `merge_address_lines` when we have an array. :/

locations/spiders/auchan_hu.py
```diff
@@ -24,7 +24,7 @@ def parse_store(self, response, **kwargs):
         item["addr_full"], item["phone"] = (
             address_info.split("Központi szám:") if "Központi szám:" in address_info else (address_info, None)
         )
-        item["addr_full"] = clean_address(item["addr_full"])
+        item["addr_full"] = item["addr_full"]
```
CloCkWeRX, line 27: Is this redundant now?

Cj-Malone: Yep, thanks.

Cj-Malone:
The longer this goes the more annoying this will be to merge.

I have no objection to closing it in favour of:

1. Status quo, leaving the function inside the spider, it's not ideal, but it's not hurting us.
2. "Dumb refactor", just move the function to shared code (`utils.py`?). The pipeline could come in a separate PR. 
3. 2 but leave the original function in but "deprecated", this will be a real small diff.

iandees:
FYI, marking a PR as ready for review (from a draft) doesn't trigger an email, so I don't know when that happens. Feel free to ping or request a review.

PR #7570, opened 2024-02-25
[LinkedData] Extract Tripadvisor URLs

Cj-Malone:
This is both, 1. good for users to be able to click through, and 2. helpful for conflation when using multiple POI sources.

### DIFF
--- a/locations/linked_data_parser.py
+++ b/locations/linked_data_parser.py
@@ -4,7 +4,7 @@ import chompjs
 import json5
 
 from locations.hours import OpeningHours
-from locations.items import Feature
+from locations.items import Feature, add_social_media
 
 
 class LinkedDataParser:
@@ -143,6 +143,8 @@ class LinkedDataParser:
         for t in types:
             LinkedDataParser.parse_enhanced(t, ld, item)
 
+        LinkedDataParser.parse_same_as(ld, item)
+
         return item
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


locations/linked_data_parser.py
```diff
@@ -215,3 +217,14 @@ def parse_enhanced_hotel(ld: dict, item: Feature):
...
                 item["extras"]["stars"] = stars.get("ratingValue")
+
+    @staticmethod
+    def parse_same_as(ld: dict, item: Feature):
+        if same_as := LinkedDataParser.get_clean(ld, "sameAs"):
+            if isinstance(same_as, str):
+                same_as = [same_as]
+            for link in same_as:
+                if "facebook.com" in link:
+                    add_social_media(item, "facebook", link)
```
davidhicks, line 228: Can we eventually deprecate `Item.facebook` and `Item.twitter` in favour of `add_social_media`?

davidhicks, line 228: For keys such as `facebook` and `tripadvisor` used with `add_social_media`, could these keys be mapped out in a `SocialMedia` (or similar name) Enum?

e.g. before:
`add_social_media(item, "tripadvisor", link)`

after:
`add_social_media(item, SocialMedia.TRIP_ADVISOR, link)`

Cj-Malone: `add_social_media` is a wrapper so if we add a new "social media" eg whatsapp as a top level item, we don't need to edit every spider. It was specifically for that direction, not removing things. Generally I vote for things a reasonable number of spiders output and/or we do "validation" on, being documented. We don't really check things in `extras`.

Tripadvisor may eventually meet my above requirements, in which case we should add it, I'm interested to see just how many POIs we get in the next weekly. I don't envision us deprecating `facebook` or `twitter`.

Cj-Malone: I have no objection, but I would request that anything using it takes both the enum or the string. Like how category stuff can take an enum or dict.

iandees:
`contact:tripadvisor` seems like a weird tag to use, but there's a couple thousand in OSM so...

PR #7572, opened 2024-02-25
remove amenity=post_office from general GLS entries

matkoniecz:
not all GLS POIs listed in ATP are dedicated post offices many are just a shelf in some shop where packages are waiting for pickup

note that there are many ways to solve this:

- declare that shelf with waiting parcels in say electronics/honey/ticket shop and dedicated post office will not be distnguished
- prefer to have false positive with nonexisting post offices reported as full post offices rather than not listing some real post offices
- more advanced processing to distinguish this caes somehow
- expect data consumers to spot such bad data on their own (with either curated list of valid spiders or by treating all post office data from ATP as likely false)
- add additional field indicating likely quality issues such as data_quality:poi_category=likely_false
- ????

### DIFF
--- a/locations/spiders/general_logistics_systems.py
+++ b/locations/spiders/general_logistics_systems.py
@@ -18,7 +18,6 @@ class GeneralLogisticsSystemsSpider(scrapy.Spider):
         "brand": "General Logistics Systems",
         "brand_wikidata": "Q46495823",
         "country": "DE",
-        "extras": Categories.POST_OFFICE.value,
     }
 
     def start_requests(self):


matkoniecz:
note that similar issues applies to also other package pickup locations, this one is kind of proposal/test (maybe title should be changed? maybe it should be discussed somehow somewhere?)

iandees:
What information did you use to know that some of these are shelves and some are post offices? Can we extrapolate that knowledge to make an informed guess on which of these POI are which?

matkoniecz:
> What information did you use to know that some of these are shelves and some are post offices? 

ops, that would be nice to include that 

1) General knowledge how parcel pickup works in Europe (dedicated post offices run by delivery companies sometimes exist but vast majority of pickup points is an extra service in some other POI and fulfils tertiary function - I picked up parcels in convenience shops, kiosks, repair shops etc)

I would not assume that pickup location is a dedicated post office until there is a very clear signal of that.

2) visiting specific location at https://www.alltheplaces.xyz/map/#21.31/50.9730136/14.2711012 (it was some electric/white goods store with GLS as secondary function - note `name=Hanno Schmidt Elektroanlagenbau` which was accurate)

IMG_20240222_160308

> Can we extrapolate that knowledge to make an informed guess on which of these POI are which?

No idea. I would try to find a dedicated GLS post office, if any exist, and check is https://api.gls-pakete.de/parcelshops?version=4&coordinates=50.973,14.271&distance=40 or similar gives some useful fields.

I guess I can ask on German forum are they aware of any?

matkoniecz:
> What information did you use to know that some of these are shelves and some are post offices? 

ops, that would be nice to include that 

1) General knowledge how parcel pickup works in Europe (dedicated post offices run by delivery companies sometimes exist but vast majority of pickup points is an extra service in some other POI and fulfils tertiary function - I picked up parcels in convenience shops, kiosks, repair shops etc)

I would not assume that pickup location is a dedicated post office until there is a very clear signal of that.

2) visiting specific location at https://www.alltheplaces.xyz/map/#21.31/50.9730136/14.2711012 (it was some electric/white goods store with GLS as secondary function - note `name=Hanno Schmidt Elektroanlagenbau` which was accurate)

IMG_20240222_160308

> Can we extrapolate that knowledge to make an informed guess on which of these POI are which?

No idea. I would try to find a dedicated GLS post office, if any exist, and check is https://api.gls-pakete.de/parcelshops?version=4&coordinates=50.973,14.271&distance=40 or similar gives some useful fields.

I guess I can ask on German forum are they aware of any?

Cj-Malone:
If these were to be mapped in OSM, the would be something like:
```
shop=convenience
name=7-Eleven
brand=7-Eleven
brand:wikidata=Q259340
post_office=post_partner
post_office:brand=General Logistics Systems
post_office:brand:wikidata=Q46495823
```

The output of this spider isn't stand alone POIs, but an attribute of another POI. (I think the spider should output `post_office=post_partner`)

matkoniecz:
Though maybe few of them may actually be a stand alone offices - is outputting post partner fine in such cases? Or at least better then outputting no category?

(for my specific purpose it would be - and for ones I can imagine, but probably for some purposes it would work poorly...)

iandees:
Thanks for the extra info.

PR #7597, opened 2024-02-26
Use Bunny.net API to purge its cache after uploading new data

iandees:
See https://github.com/alltheplaces/alltheplaces.xyz/issues/52


iandees:
For future Ian: I also updated the ECS task definition to include the API key from Bunny as an environment variable.

PR #7601, opened 2024-02-26
provide more info about data content

### LINKED ISSUE #5967: addr:state for Orlen in Poland are (1) in wrong language (2) not needed
`addr:state=Lesser Poland Voivodeship` in Poland is (1) not needed, we are not using it as part of address in general (2) if at all present, it should be in Polish

screen03

matkoniecz:
fixes #5967


iandees:
I think the "Data Quality" section that is a bit farther down in the document already covers this.

Issue #6907, opened 2024-01-26
vending=public_transport_tickets in Kraków


matkoniecz:
### Brand name

KKM

### Wikidata ID

Q57515549

### Store finder url(s)

https://kkm.krakow.pl/pl/punkty-sprzedazy-biletow/

see blue one mapped already https://www.openstreetmap.org/node/1907768673

not in ATP: https://www.alltheplaces.xyz/map/#19.29/50.0816335/19.8819302

(I can trying adding spider)

Cj-Malone:
This one is a bit of a weird one, well a pain. The data is only in javascript. We have chompjs that can parse it, but it's not ideal, in this case it turns `Latitude :  50.08325` into `"Latitude ":  50.08325` (trailing space). We can work with it, or replace it, but it may cause us issues if there are `:`s in the addresses, so it may be better for us to tolerate it... Your call.

Anyway, here is where I got to. Hope you can finish it off.

```python
import re
from typing import Any

import chompjs
from scrapy import Spider
from scrapy.http import Response

from locations.items import Feature


class TestSpider(Spider):
    name = "test"
    item_attributes = {"brand": "KKM", "brand_wikidata": "Q57515549"}
    start_urls = ["https://kkm.krakow.pl/pl/punkty-sprzedazy-biletow/"]

    def parse(self, response: Response, **kwargs: Any) -> Any:
        for location_raw in re.findall(r"items\.push\(\s+({.+?})\)", response.text, re.DOTALL):
            location = chompjs.parse_js_object(re.sub(r"\s:", ":", location_raw))
            item = Feature()
            item["ref"] = location["Id"]
            item["lat"] = location["Latitude"]
            item["lon"] = location["Longitude"]
            print(location)  # TODO: address? TypeId/TypeName
            yield item
```

Nykakin:
Hello, chompjs author here.

Fixed this issue (https://github.com/Nykakin/chompjs/pull/57), released in 1.2.3 version.
```python
>>> import chompjs
>>> 
>>> chompjs.parse_js_object("{Latitude :  50.08325}")
{'Latitude': 50.08325}
```
Please don't hesitate to open issues if you encounter any problems with the library.

matkoniecz:
we are using old version for now, see https://github.com/alltheplaces/alltheplaces/blob/master/Pipfile.lock#L403

[...]

PR #7610, opened 2024-02-27
Ticket sellers in Kraków

matkoniecz:
I have some pending TODOs that I will look into but maybe these are obviously fixable so if anyone knows how to solve them...

How can I stop NSI matcher from running on some cases? 

Not everything is vending machine (or should I report to NSI that they should have also offices listed?)

Maybe I should skip offices and process them in a separate spider? But it will pointlessly query twice.


matkoniecz:
"In the cases where it fails for some reason then the automation can be disabled for the spider" in https://github.com/alltheplaces/alltheplaces/blob/master/docs/WIKIDATA.mdoffers tantalizing view that it can be done.

Still no idea how to do this.

CloCkWeRX:
Which bit is failing/you want to ignore the automation on?

```
class HipermaxiBOSpider(AgileStoreLocatorSpider):
    name = "hipermaxi_bo"
    item_attributes = {
        "brand_wikidata": "Q81968262", # Override wikidata match spider wide
        "brand": "Hipermaxi", 
    }
```

or
```
def parse_item(self, item, location):
```
Is how you can generally do the specific decoration of an item from whatever you are parsing, then "yield" which is _kind of_ like a 

```return variable_name```

but it's actually kind of like

```return super(variable_name)```

(Apologies if you already know that bit, just assuming you are more used to java influenced languages... though looks like kotlin has the concept in an iteration context)

File: locations/spiders/kkm_vending_krk_pl.py [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+
+            TICKET_OFFICE_VALUES = ["2"]
+            TICKET_MACHINE_WITH_INFO_KIOSK_VALUES = ["13"]
+            TICKET_MACHINE_VALUES = ["1", "9"]
+            TICKET_MACHINE_VALUES_WITH_NO_CACH_SUPPORT = ["41"]
+            if location["TypeId"] in TICKET_OFFICE_VALUES:
+                apply_category(Categories.SHOP_TICKET, item)
+                item["extras"]["ticket"] = "public_transport"
+                item["name"] = "Punkt Obsługi Pasażerów KMK"
+                # TODO: stop matching on NSI here - these are offices
```
CloCkWeRX, line 50: https://github.com/alltheplaces/alltheplaces/blob/master/locations/pipelines/apply_nsi_categories.py#L10

matkoniecz:
https://github.com/alltheplaces/alltheplaces/pull/7610/files#diff-ebabc5f645c0760030ed6540d6f5f71b7ef7cc4d527da503cd10fc1593d76286R46 branch also gets tags derived from https://github.com/alltheplaces/alltheplaces/pull/7610/files#diff-ebabc5f645c0760030ed6540d6f5f71b7ef7cc4d527da503cd10fc1593d76286R15

I would want to disable NSI pipeline within `if location["TypeId"] in TICKET_OFFICE_VALUES:` if

locations/spiders/kkm_vending_krk_pl.py
matkoniecz: If I look at it right then presence of `nsi_id` with any value will terminate matching. I will try that.

CloCkWeRX: Option 1:
https://github.com/alltheplaces/alltheplaces/blob/4d5453f71f9d57db4969b9c77998ce054470878c/locations/spiders/argos.py#L18

post process_item, then undo whatever it did when you want to.

Cj-Malone:
> or should I report to NSI that they should have also offices listed

If there is a decent amount, yes it should be in NSI. The "hard" min is 10 POIs for NSI, however my personal min is a bit higher.

locations/spiders/kkm_vending_krk_pl.py
matkoniecz: https://github.com/alltheplaces/alltheplaces/issues/6907#issuecomment-1966174916 solves this as far as I see

thanks for all help!

matkoniecz:
> If there is a decent amount, yes it should be in NSI. The "hard" min is 10 POIs for NSI, however my personal min is a bit higher.

There is less than then that - three locations, and it is city specific so number unlikely to grow much

matkoniecz:
So what remains now is either implementing opening hours parsing for their syntax or giving up on opening hours.

(and then looking at data again)

locations/spiders/kkm_vending_krk_pl.py
```diff
@@ -0,0 +1,50 @@
...
+            item["ref"] = location["Id"]
+            item["lat"] = location["Latitude"]
+            item["lon"] = location["Longitude"]
+            item["extras"]["location_description"] = location["Location"]
+            if location["OpeningHours"] == "Całą dobę":
+                item["extras"]["opening_hours"] = "24/7"
+            else:
+                entry = OpeningHours()
+                entry.add_ranges_from_string(location["OpeningHours"], DAYS_PL)
+                item["extras"]["opening_hours"] = entry.as_opening_hours()
```
Cj-Malone, line 31:
```suggestion
                item["opening_hours"] = "24/7"
            else:
                item["opening_hours"] = OpeningHours()
                item["opening_hours"].add_ranges_from_string(location["OpeningHours"], DAYS_PL)
```

`opening_hours` isn't an extra, it's a direct property. Also if we output the `OpeningHours` object, we skip validation if it's the string (`as_opening_hours`) we do some regex on it.

locations/spiders/kkm_vending_krk_pl.py
```diff
@@ -0,0 +1,49 @@
...
+                item["nsi_id"] = "N/A"
+                yield item
+            if location["TypeId"] in TICKET_MACHINE_WITH_INFO_KIOSK_VALUES + TICKET_MACHINE_VALUES:
+                item["extras"]["payment:cash"] = "yes"
+                item["extras"]["payment:cards"] = "yes"
+                yield item
+            if location["TypeId"] in TICKET_MACHINE_VALUES_WITH_NO_CACH_SUPPORT:
+                item["extras"]["payment:cash"] = "no"
+                item["extras"]["payment:cards"] = "yes"
+                yield item
```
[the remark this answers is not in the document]
Cj-Malone: (Un tested)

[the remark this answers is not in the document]
matkoniecz: thanks, I wanted to log missing types and forgot to add this!

Cj-Malone (approved):
LGTM, thanks.
https://alltheplaces-data.openaddresses.io/map.html?show=https://alltheplaces-data.openaddresses.io/ci/8063707830/kkm_vending_krk_pl/output.geojson

PR #7620, opened 2024-02-27
Add spider for Mighty Flame

jleedev:
5500 ish propane exchange locations (gas stations, hardware stores) in US and CA.


CloCkWeRX:
Does this one have a NSI / wikidata entry? I don't think they came up in the shop/gas missing scrapers stuff recently

PR #7631, opened 2024-02-29
[CreditUnion US] Fix spider

Cj-Malone:
Coords and country codes were missing. This does lose website, but it was a homepage anyway, not POI specific.

TBH I'm not in love with this spider, it feels like secondary data, but I don't know how the US "Credit Union" works.


iandees:
Co-Op looks to be a network of credit unions (which are banks) that share some resources. I agree that it's likely secondary or aggregated data. We should probably be pulling data from the individual credit union pages, but this one's already here...

Issue #3113, opened 2021-10-01
Spider simply_self_storage is broken


scraperbot:
During the global build at 2021-09-29-14-42-48, spider **simply_self_storage** failed with **0 features** and **0 errors**.

Here's [the log](https://data.alltheplaces.xyz/runs/2021-09-29-14-42-48/logs/simply_self_storage.txt) and [the output](https://data.alltheplaces.xyz/runs/2021-09-29-14-42-48/output/simply_self_storage.geojson) ([on a map](https://data.alltheplaces.xyz/map.html?show=https://data.alltheplaces.xyz/runs/2021-09-29-14-42-48/output/simply_self_storage.geojson))

CloCkWeRX:
Just rebranded.

```
    item_attributes = {"brand": "Public Storage"}
    allowed_domains = ["www.publicstorage.com"]
    sitemap_urls = ["https://www.publicstorage.com/robots.txt"]

    sitemap_rules = [
        (r"-self-storage/$", "parse"),
    ]
```
Should still work.
But sitemap no longer has the storage URLs:
https://validator.schema.org/#url=https%3A%2F%2Fwww.publicstorage.com%2Fself-storage-ny-new-york%2F5307.html%235307%7C1%7Cnew%2520york%7C40.7127753%7C-74.0059728%7C0%7C0

PR #7774, opened 2024-03-10
Fix simply self storage (Now "Public Storage")

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/3113


Cj-Malone:
Sounds more like Public Storage bought Simply Self Storage. Spider should probably be renamed.

https://www.reuters.com/markets/deals/public-storage-acquire-simply-self-storage-22-bln-2023-07-24/

CloCkWeRX:
Ah, interesting. The structure of the site was basically the same and I had checked the wiki data entries, found nothing for public storage...

CloCkWeRX:
Turns out we have a public storage is Spider so this can be removed

Issue #7856, opened 2024-03-13
Mercator


CloCkWeRX:
----
### Brand name

Mercator

retail company in Slovenia

### Wikidata ID

Q738412
https://www.wikidata.org/wiki/Q738412
https://www.wikidata.org/wiki/Special:EntityData/Q738412.json

### Store finder url(s)
https://www.mercator.si/prodajna-mesta/
Official Url(s): http://www.mercator.si
- [x] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q738412 http://www.mercator.si`
```
```
Official Url(s): https://www.mercator.si/
- [x] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q738412 https://www.mercator.si/`
```
```

CloCkWeRX:
```
<html>
<body>
<!--StartFragment-->
<li class="store">
--
  | <div class="cols-wrapper">
  |  
  | <div class="col col-35 helvetica-typo">
  | <h4><a href="/prodajna-mesta/hipermarket-ljubljana-rudnik/" class="title-link">Hipermarket Ljubljana Rudnik</a></h4>
  | </div>
  |  
  | <div class="col col-35 helvetica-typo">
  | <p>Jurčkova cesta 223 <br> 1000 Ljubljana</p>
  |  
  | <p class="small">
  | <a href="//www.google.com/maps/place/Jurčkova+cesta+223+1000+Ljubljana" class="show-gmap-pin">Prikaži na zemljevidu</a>
  | </p>
  |  
  | </div>
  |  
  | <div class="col col-30 helvetica-typo">
  | <p class="open-time">
  |  
  | <strong>Zaprto</strong>
  |  
  | <span class="time-circle "></span>
  | <span class="distance">&nbsp;</span>
  | </p>
  |  
  |  
  | </div>
  | </div>
  | </li>

<!--EndFragment-->
</body>
</html><li class="store">
				<div class="cols-wrapper">

					<div class="col col-35 helvetica-typo">
[...]

PR #7859, opened 2024-03-13
Add Mercator (SL)

CloCkWeRX:
Fixes https://github.com/alltheplaces/alltheplaces/issues/7856

### DIFF
--- a/locations/spiders/mercator_sl.py
+++ b/locations/spiders/mercator_sl.py
@@ -1,0 +1,14 @@
+from scrapy.spiders import SitemapSpider
+
+from locations.structured_data_spider import StructuredDataSpider
+
+
+class MercatorSLSpider(SitemapSpider, StructuredDataSpider):
+    name = "mercator_sl"
+    item_attributes = {"brand": "Mercator", "brand_wikidata": "Q738412"}
+    allowed_domains = ["www.mercator.si"]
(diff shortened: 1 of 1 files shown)


locations/spiders/mercator_sl.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class MercatorSLSpider(SitemapSpider, StructuredDataSpider):
+    name = "mercator_sl"
+    item_attributes = {"brand": "Mercator", "brand_wikidata": "Q738412"}
+    allowed_domains = ["www.mercator.si"]
+    sitemap_urls = [
+        "https://www.mercator.si/sitemap.xml/sitemap/Store/1",
+        "https://www.mercator.si/sitemap.xml/sitemap/Store/2",
+    ]
```
Cj-Malone, line 13: When they add `/3`, we'll be losing POIs.

Instead use all sitemaps with `/Store/` in them with:
```suggestion
    sitemap_urls = ["https://www.mercator.si/sitemap.xml"]
    sitemap_follow = ["/Store/"]
```

I'll do a PR because I'm gonna change the country code at the same time.

CloCkWeRX: Thanks. I wavered on it and ran out of time when prepping this to retest with changes

Cj-Malone: FYI scrapy can cache requests, I use it basically all the time, but it sometimes causes issues when doing requests to the same url but with different headers.

`pipenv run scrapy crawl -s HTTPCACHE_ENABLED=True -s HTTPCACHE_EXPIRATION_SECS=604800`

PR #7861, opened 2024-03-13
fix mcdonalds_cz name attribute

vfosnar:
The `name` tag contained the address instead of the actual name. It seems that this is the case for other McDonald's scrapers as well but there are exceptions, for example `mcdonalds_hr`.


Cj-Malone:
ATPs `name` was originally closer to OSMs `branch` than `name`. It's going to be a slow process, but ATP now has `name` and `branch`.

PR #7863, opened 2024-03-14
fix `billa` `name`, `brand` tags and `phone` tag for the Czech Republic

vfosnar:


vfosnar:
for some reason it uses all caps `BILLA` for the brand

vfosnar:
Hm, why does every almost every spider contain the brand name inside `item_attributes`? I thought this should be handled by the pipeline. For example in this case it just contained an invalid name but on NSI/wikidata the name is correct.

iandees:
The `brand` is coming from NSI after you removed it from `item_attributes`. We still have brand in item_attributes mostly because of momentum, I think. Also, there are plenty of brands that don't have Wikidata + NSI entries.

PR #7867, opened 2024-03-14
[LaVieClaireFR] Add category

karen-kerr:


CloCkWeRX:
With these ones, a lot of the time I am skipping adding the category as it gets inferred by the brand Wikidata, or at least I thought it was.
@karen-kerr does it show up differently somewhere? Or are you more focused on annotating spiders from code inspection? (Don't want to make more work for you if it's a legit mistake on my part/the data is wrong!)

PR #7871, opened 2024-03-14
add coop_cz spider (2353 locations)

vfosnar:


locations/spiders/coop_cz.py
```diff
@@ -0,0 +1,47 @@
+import re
+
+from scrapy import Request, Spider
+
+from locations.items import Feature
+
+
+class CoopCZSpider(Spider):
+    name = "coop_cz"
+    item_attributes = {"brand": "COOP", "brand_wikidata": "Q52851660"}
```
Cj-Malone, line 10: Is this the same as brand=COOP Jednota brand:wikidata=Q41629254 in NSI?

vfosnar: afaik it shouldn't

Cj-Malone: OK, we either need a category applied in the spider, or this brand added to NSI so we get the category from there.

PR #7875, opened 2024-03-14
fix dm name

vfosnar:
prev:
`name=dm drogerie markt s.r.o. - prod.313`
now:
`name=dm`


Cj-Malone (approved):
LGTM

https://alltheplaces-data.openaddresses.io/map.html?show=https://alltheplaces-data.openaddresses.io/ci/8286132294/dm/output.geojson

PR #7883, opened 2024-03-15
move `name` into `branch` for `tesco_eu`

vfosnar:
## Before
name:
```
Hypermarket Komárno
EXTRA Bratislava Zlaté piesky
Supermarket Poltár
Supermarket Sečovce
TESCO Szeged Szupermarket
TESCO expressz Sopron - Lackner K. utca
TESCO Expressz Bp. - Ady Endre utca
TESCO Gyöngyös Hipermarket
Hypermarket Uherské Hradiště
Expres Praha Mendelova
Supermarket Odolena Voda
Expres Nový Bor
```

## After
name:
```
Tesco
Tesco Extra
Tesco
Tesco
Tesco
Tesco Express
Tesco Express
Tesco
Tesco
Tesco Express
Tesco
Tesco Express
```

branch:
```
Hypermarket Komárno
EXTRA Bratislava Zlaté piesky
Supermarket Poltár
Supermarket Sečovce
Szeged Szupermarket
expressz Sopron - Lackner K. utca
Expressz Bp. - Ady Endre utca
Gyöngyös Hipermarket
Hypermarket Uherské Hradiště
Expres Praha Mendelova
Supermarket Odolena Voda
Expres Nový Bor
```

### DIFF
--- a/locations/spiders/tesco_eu.py
+++ b/locations/spiders/tesco_eu.py
@@ -31,6 +31,7 @@ class TescoSpider(scrapy.Spider):
         for store in response.json()["stores"]:
             store["street-address"] = store.pop("address", "")
             item = DictParser.parse(store)
+            item.pop("name")
             item["ref"] = store["goldid"]
             item["lat"] = store.get("gpslat")
             item["lon"] = store.get("gpslng")
--- a/locations/spiders/tesco_eu.py
+++ b/locations/spiders/tesco_eu.py
@@ -19,6 +19,7 @@ class TescoSpider(scrapy.Spider):
         "hu": "https://tesco.hu/aruhazak/aruhaz/",
         "sk": "https://tesco.sk/obchody/detail-obchodu/",
     }
+    BRANDING_WORDS = ["tesco", "expres", "extra"]  # lowercase
 
     def start_requests(self):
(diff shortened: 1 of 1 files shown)


Cj-Malone:
I don't love these branches, IMO the format (eg "Hypermarket", "Expressz") shouldn't be in them. Right now, it looks like it'd be a mess to clean them, I'd be OK with removing the names completely so we get NSI names.

vfosnar:
The `name`s are being applied from the NSI. Tesco has a different NSI/wikidata entry for each of these 3 cases. In case of `branch`es I think it should include Supermarket etc. as that's part of their branding.

Cj-Malone:
It's already part of the `name` and `brand`.

vfosnar:
That makes sense, fixed it. Now it only leaves `Supermarket` and `Hypermarket` in `branch`. These are not included in `name` and shouldn't be, but they are good to know.

Cj-Malone (approved):
I still don't think store formats belong in `branch`. However it's not worth the argument, this is an improvement. Thank you.

vfosnar:
I'd personally prefer a tag to describe hypermarket, but [it's abondoned](https://wiki.openstreetmap.org/wiki/Proposal:Hypermarket). I find this information useful and I'm not sure where else to store it so it's user visible.

Issue #7886, opened 2024-03-15
NaturaSi (StructuredData, Sitemap)


CloCkWeRX:
----
### Brand name

NaturaSì

### Wikidata ID

Q60840755
https://www.wikidata.org/wiki/Q60840755
https://www.wikidata.org/wiki/Special:EntityData/Q60840755.json

### Store finder url(s)
https://negozi.naturasi.it/
Official Url(s): https://www.naturasi.it
- [x] I have run the below and if anything in generated, added it.
- [x] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q60840755 https://www.naturasi.it

CloCkWeRX:
https://negozi.naturasi.it/sitemap/sitemap_1.xml

PR #7923, opened 2024-03-17
Add NaturaSì

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/7886
Note: Some store pages don't have valid structured data, but I'm not making an effort to go after these.

### DIFF
--- a/locations/spiders/naturasi_it.py
+++ b/locations/spiders/naturasi_it.py
@@ -1,0 +1,15 @@
+from scrapy.spiders import SitemapSpider
+
+from locations.structured_data_spider import StructuredDataSpider
+
+
+class NaturasiITSpider(SitemapSpider, StructuredDataSpider):
+    name = "naturasi_it"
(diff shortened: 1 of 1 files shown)


locations/spiders/naturasi_it.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class NaturasiITSpider(SitemapSpider, StructuredDataSpider):
+    name = "naturasi_it"
+    item_attributes = {"brand": "NaturaSì", "brand_wikidata": "Q60840755"}
+    sitemap_urls = ["https://negozi.naturasi.it/sitemap/sitemap_1.xml"]
+    sitemap_rules = [(r"https://negozi.naturasi.it/.*", "parse_sd")]
+
+    def post_process_item(self, item, response, ld_data, **kwargs):
+        if item["city"]:
```
CloCkWeRX, line 13:
```suggestion
        if "city" in item:
```

PR #7956, opened 2024-03-18
[Rio SEO][Storefinder] Clean up

Cj-Malone:
Includes adding `dataset_attributes` and fixing `street`/`street_address`. But one pattern I'd like to reiterate is `post_process_feature`/`parse_item` functions, using `yield from` does make us have an extra line of code in every spider, but it also gives the spider the control over throwing away POIs.


CloCkWeRX:
Thanks! That's a nice bit of tidying.

PR #7725, opened 2024-03-09
Sitemap detection: Suggest regexp

CloCkWeRX:
Proof of concept for https://github.com/alltheplaces/alltheplaces/issues/7723

Some of these may be overly broad, or redundant as a simple regexp covers the situation.
I've gone with generating from the codebase and making a few judgement calls to consolidate.

# Example: Sitemap with Store pages
```
$ pipenv run scrapy sitemap https://locations.baskinrobbins.com/ --pages

...
https://locations.baskinrobbins.com/tx/mission/201-south-shary-road-suite-250-364911-br
https://locations.baskinrobbins.com/mi/lansing/619-north-marketplace-boulevard-364736-br
https://locations.baskinrobbins.com/googled44f8f19fc4a8f0d.html
https://locations.baskinrobbins.com/googleb3b80496355c277a.html

Possible patterns
/[^/]+/[^/]+/[^/]+$: 3685
/[-\w]+/[-\w]+/[-\w]+: 2244
/\w\w/[-\w]+/[-\w]+: 2244
https://locations(.*).com/[-\w]+/[-\w]+/[-\w]+: 2244
https://locations(.*)/.*/.*/.*: 2244
https://locations(.*)/.+/.+/.+: 2244
https://locations(.*)/[^/]+/[^/]+/[^/]+$: 2244
locations\.(.*)(?:\/[a-z]{2}){1,2}(?:\/[\w\-.]{3,}){2}: 2244
locations\.(.*)/.*/.*/.*$: 2244
/\w\w/[^/]+/[^/]+$: 2239
...
```

Actual regex used by the existing spider:
```
    sitemap_rules = [(r"/\w\w/[-\w]+/[-\w]+", "parse")]
```

# Example: A sitemap with no useful pages

Potentially, some of the regex that got hits here should be dropped in favour of more specific ones. However it lets you see at a glance there's not much useful content here.

If this were turned into automatic spider generation, would probably want to drop the broadest ones.

```
$ pipenv run scrapy sitemap https://roddapaint.com --pages


https://www.roddapaint.com/press-releases/rodda-paint-introduces-the-cascadia-color-collection/
https://www.roddapaint.com/in-the-news/rodda-paint-joins-the-paint-paul-campaign/
https://www.roddapaint.com/in-the-news/rodda-paint-has-a-new-president-and-chief-operating-officer/
https://www.roddapaint.com/press-releases/rodda-paint-company-announces-the-purchase-of-valley-paint-manufacturing-in-woods-cross-ut/
[...]

PR #7963, opened 2024-03-18
Structured data detector

CloCkWeRX:
@davidhicks I would add this one to my most wanted list.

It.. _works_... sort of.

```
$ pipenv run scrapy sf https://www.poiesz-supermarkten.nl/onze-winkels/supermarkt-appelscha/vaart-zuidzijde
from locations.structured_data_spider import StructuredDataSpider


class PoieszNLSpider(StructuredDataSpider):
    name = "poiesz_nl"
    item_attributes = {
        "brand_wikidata": "Q2521700",
        "brand": "Poiesz",
    }
    wanted_types = "http://schema.org/GroceryStore"
```

But:
- I don't seem to be able to return an arbitrary string from my xpath match or substring. I'm not sure if it's a parcel limitation or I'm doing something wrong
- There's an assumption it is only a string being returned

I'd be curious about the best way to flag "I want to give you an array of possible matches"

Maybe this doesn't belong in the structuredataspider, but as a combination of sitemap+structuredata spider.

### DIFF
--- a/locations/storefinder_detector.py
+++ b/locations/storefinder_detector.py
@@ -36,7 +36,6 @@ from locations.storefinders.storelocatorwidgets import StoreLocatorWidgetsSpider
 from locations.storefinders.storemapper import StoremapperSpider
 from locations.storefinders.storepoint import StorepointSpider
 from locations.storefinders.storerocket import StoreRocketSpider
-from locations.structured_data_spider import StructuredDataSpider
 from locations.storefinders.super_store_finder import SuperStoreFinderSpider
 from locations.storefinders.sweetiq import SweetIQSpider
 from locations.storefinders.uberall import UberallSpider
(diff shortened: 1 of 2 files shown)


davidhicks:
I like automation but feel this may need to be paired with #7725 and/or a CrawlSpider approach where we try and crawl through a selection of obvious-looking store pages with command patterns such as `/stores/abc/hollow-hill` or `/restaurants/happy-valley-1234`, etc (either from sitemap or provided feature list page) and look for Structured Data within?

CloCkWeRX:
I agree, this was somewhat my first pass of "does a URL contain structured data that we'd like"
I'm not particuarly surely the best way to make this a subclass of the current detection rules model(s) though; as it's kind of a hybrid - yes we want structured data; on the assumption it's more than one instance

CloCkWeRX:
Closing for the moment (still going to poke at this when I have time; but don't know how successful I'd be)

PR #7995, opened 2024-03-19
Add the codespec to build with codebuild

iandees:
Since Github won't let us build forks with secrets, I'm trying to do it with AWS CodeBuild instead. The goal is to get the build comment on all PRs, even the ones from forks.


iandees:
We'll see how this goes...

Cj-Malone:
Seems to be working, thanks for this.

iandees:
It looks like Github doesn't trigger the webhook on builds of forks, which kind of defeats the purpose of this. I have an alternative approach that I'll try next.

PR #8009, opened 2024-03-20
[SpaWorldAUNZ] Fix postcode type

karen-kerr:


iandees:
Maybe this is something we should do in a pipeline step?

PR #8036, opened 2024-03-23
Cache Python dependencies to S3 for CodeBuild to run faster

iandees:
This adds a step to cache Python dependencies to S3 so that install of dependencies doesn't take ~60 seconds every time.


iandees:
This should speed up builds. Let's see if it breaks anything.

Issue #7586, opened 2024-02-26
Revolution Laundry (Custom API)


CloCkWeRX:
Fetched 8 brands/shop/laundry from NSI
Missing by wikidata: 1
### Brand name

Revolution Laundry

Worldwide active operator of outdoor laundromats

### Wikidata ID

Q113516904
https://www.wikidata.org/wiki/Q113516904
https://www.wikidata.org/wiki/Special:EntityData/Q113516904.json

### Store finder url(s)
https://stores.revolution-laundry.com/de-de
Official Url(s): https://www.revolution-laundry.de/
- [x] I have run the below and if anything in generated, added it.
- [x] I have manually checked for a storefinder
- [x] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q113516904 https://www.revolution-laundry.de/`
```
```
https://stores.revolution-laundry.com/en-uk
Official Url(s): https://www.revolution-laundry.com/
- [x] I have run the below and if anything in generated, added it.
- [x] I have manually checked for a storefinder
- [x] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q113516904 https://www.revolution-laundry.com/`
```
```

----

_Originally posted by @CloCkWeRX in https://github.com/alltheplaces/alltheplaces/discussions/7450_

CloCkWeRX:
curl 'https://stores.revolution-laundry.com/Ajax/searchByCoordinates' \
  -H 'Accept: application/json, text/plain, */*' \
  -H 'Accept-Language: en-GB,en;q=0.9,en-US;q=0.8,de;q=0.7,es;q=0.6,fr;q=0.5' \
  -H 'Connection: keep-alive' \
  -H 'Content-Type: application/json' \
[...]

PR #8042, opened 2024-03-24
Add Revolution Laundry

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/7586


CloCkWeRX:
I feel like there's a few other company ids to explore here, but for now this is what they list

PR #8061, opened 2024-03-26
[MallMaverick] Add spider

Cj-Malone:


Cj-Malone:
I'll sort out the exception tomorrow

matkoniecz:
> However coords is relative to a none north facing svg

what does it mean?

matkoniecz:
see also #8978 (`contact:twitter=www.twitter.com` and similar should not be included) that seems to apply in part/mostly/only to this spider

Issue #7890, opened 2024-03-15
New Seasons Market (Sitemap, Structured Data)


CloCkWeRX:
### Brand name

New Seasons Market

Grocery store chain based in Portland, Oregon, U.S.

### Wikidata ID

Q7011463
https://www.wikidata.org/wiki/Q7011463
https://www.wikidata.org/wiki/Special:EntityData/Q7011463.json

### Store finder url(s)
https://www.newseasonsmarket.com/find-a-store
Official Url(s): http://newseasonsmarket.com/
- [x] I have run the below and if anything in generated, added it.
- [x] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q7011463 http://newseasonsmarket.com/`
```
```

CloCkWeRX:
HTML Parsing
```
<html>
<body>
<!--StartFragment-->
<div class="sr-list" data-latitude="45.5481269199285600" data-longitude="-122.6673199714278200" data-state-name="OR" data-store-name="Williams">
--
  | <div class="store-heading-container">
  | <div class="column">
  | <h4 class="store-name">Williams</h4>
  | </div>
  | <div class="column store-distance-info"></div>
  | </div>
  | <div class="open-time">
  | <p class="pull-left">
  | <span class="success">Open</span>
  | &bullet; Today
  | 7:00 AM - 10:00 PM
  | </p>
  | </div>
  | <div class="location-direction">
  | <p class="pull-left">
  | <span class="store-addressdetails">3445 N Williams Ave</span><br />
  | <span class="store-city">Portland</span>, <span class="store-state">OR</span> &nbsp; <span class="store-zipcode">97227</span>
  | </p>
  | <p class="pull-right">
  | <a href="https://www.google.com/maps/search/?api=1&query=3445 N Williams Ave, Portland, OR, 97227&query_place_id=ChIJnyFQvECnlVQR7G_VonSAO9Y"
  | class="text-cta direction"
  | target="_blank">
  | get directions</a>
  | </p>
  | </div>
  | <div class="store-contact">
  | <p class="pull-left"><a href="tel:5035282888">(503) 528-2888</a></p>
  | </div>
  | <div class="store-service">
  | <ul class="services highlighted depart-cl">
  | </ul>
  | <ul class="services highlighted amenities-cl">
  | <li data-id="Catering">
[...]

PR #8108, opened 2024-03-31
Add New Seasons Market (US) and fix LinkedData parsing of OpeningHoursSpecification when only dealing with a dict

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/7890

### DIFF
--- a/locations/hours.py
+++ b/locations/hours.py
@@ -640,7 +640,6 @@ class OpeningHours:
 
         if day not in DAYS:
             raise ValueError(f"day must be one of {DAYS}, not {day!r}")
-
         if not open_time or not close_time:
             return
         if isinstance(open_time, str):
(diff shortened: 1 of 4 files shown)


CloCkWeRX:
Hmm, 8 results but 21 listed on the storefinder page.

locations/hours.py
```diff
@@ -640,7 +640,6 @@ def add_range(self, day, open_time, close_time, time_format="%H:%M"):
 
         if day not in DAYS:
             raise ValueError(f"day must be one of {DAYS}, not {day!r}")
```
CloCkWeRX, line 642:
```suggestion
            raise ValueError(f"day must be one of {DAYS}, not {day!r}")

```

CloCkWeRX:
@Cj-Malone I feel like you've done a bit on the linked data parser, so pinging you for review specifically.

Let me know if we'd like to split this into two PRs - fixing linked data parsing vs new spider.

Cj-Malone:
Yeah I wrote from_linked_data before I wrote LinkedDataParser, I guess I'd like to move it in there, but it'll end up as a big refactor. I'll take a proper look at this tomorrow.

I'm not even sure a straight dict is valid in the spec. Yeah Google's validator seems to work, but that's just one implementation. With this, if it's fairly common, but against spec, I'd have no issue implementing it in shared code. If it's rare, I usually prefer it fixed in pre parse.

CloCkWeRX:
https://www.w3.org/TR/json-ld11/#representing-singular-values-as-arrays I think is the bit of the JSON LD spec that is relevant.

> Generally, when compacting, properties having only one value are represented as strings or [maps](https://infra.spec.whatwg.org/#ordered-map), while properties having multiple values are represented as an array of strings or [maps](https://infra.spec.whatwg.org/#ordered-map). This means that applications accessing such properties need to be prepared to accept either representation.

Cj-Malone (approved):
Seems fine thanks. As I've added in the code, I'd like to deprecate oh.from_linked_data in favour of ld.parse_opening_hours, it's where the code belongs. I see no reason to rush this though.

PR #8129, opened 2024-04-02
Add SocialMedia enum

Cj-Malone:
As @davidhicks said in https://github.com/alltheplaces/alltheplaces/pull/7570#discussion_r1501833283, this should be an enum.

### DIFF
--- a/tests/test_social_media.py
+++ b/tests/test_social_media.py
@@ -1,0 +1,36 @@
+from locations.exporters.geojson import item_to_properties
+from locations.items import Feature, SocialMedia, get_social_media, set_social_media
+
+
+def test_assigning():
+    item = Feature()
+
+    for social_media in SocialMedia:
+        set_social_media(item, social_media, social_media.value)
+
+    for social_media in SocialMedia:
+        assert get_social_media(item, social_media) == social_media.value
+
+    props = item_to_properties(item)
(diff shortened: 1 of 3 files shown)


davidhicks (approved):

tests/test_social_media.py
```diff
@@ -0,0 +1,36 @@
+from locations.exporters.geojson import item_to_properties
+from locations.items import Feature, SocialMedia, get_social_media, set_social_media
+
+
+def test_assigning():
```
davidhicks, line 5: I couldn't figure out from OSM keys for contact:* whether multiple values are allowed. However, for how we use these fields (only for location-specific social media contacts, not brand-specific) I can't see why we'd ever want to allow multiple values to be specified. Perhaps this test could double-assign a social media contact for an item and confirm that the previous value is overwritten rather than appended?

panasonic-mafia (approved):

tests/test_social_media.py
Cj-Malone: Technically OSM will allow these to contain multiple values, the semicolon rule applies to basically everything. However, in reality nothing supports it, and it should be `brand:facebook` for a generic profile and `contact:facebook` for a POI specific profile.

I'll add a test anyway.

PR #8142, opened 2024-04-05
[ccf_fr] add spider (249 locations)

panasonic-mafia:


panasonic-mafia:
rel: https://github.com/osmlab/name-suggestion-index/pull/9351

PR #8155, opened 2024-04-10
[FiveGuys AE AT AU BE BH CH KW NL] Fix coords and urls

Cj-Malone:
All these have the same backend that has recently got slightly broken json linked data. Most the other Five Guys sites have microdata, I can't remember, and haven't checked, what they all used to be, but presumably they are migrating in one direction.


Cj-Malone:
CI is objecting because of number of spiders changed. I've ran them all locally (I cancelled the big ones), they are good to go.

PR #8175, opened 2024-04-11
[champs_sports_us] add spider (410 locations)

mjoe999:
A US sports shop brand:

'atp/brand/Champs Sports': 410,
 'atp/brand_wikidata/Q2955924': 410,
 'atp/category/shop/sports': 410,
 'atp/field/country/from_spider_name': 410,
 'atp/field/email/missing': 410,
 'atp/field/image/dropped': 410,
 'atp/field/image/missing': 410,
 'atp/field/lat/missing': 1,
 'atp/field/lon/missing': 1,
 'atp/field/opening_hours/missing': 25,
 'atp/field/operator/missing': 410,
 'atp/field/operator_wikidata/missing': 410,
 'atp/field/twitter/missing': 410,
 'atp/nsi/perfect_match': 410

PR #8187, opened 2024-04-12
[Core] Drop attributes pipeline

Cj-Malone:
Rel: https://github.com/alltheplaces/alltheplaces/pull/8175#issuecomment-2051921316

I don't know how useful this will be, but it feels like we are writing the same code again and again.

If we want to go this way, I'll add some tests. If not, no hard feelings we can continue overriding functions.

### DIFF
--- a/docs/PIPELINES.md
+++ b/docs/PIPELINES.md
@@ -6,7 +6,7 @@ See [`fedex.py`](../locations/spiders/fedex.py) for an example.
 
 ### `DropAttributesPipeline`
 
-When a spider has `drop_attributes`, we remove those attributes from `Item`s, this happens early, so it may be re added by other pipelines later on.
+When a spider has `drop_attributes`, we remove those attributes from `Feature`s, this happens early, so it may be re added by other pipelines later on.
 
 ### `ApplySpiderLevelAttributesPipeline`
 
--- a/docs/PIPELINES.md
+++ b/docs/PIPELINES.md
@@ -4,6 +4,10 @@ After an `Item` is emitted from a spider, it goes through several pipelines befo
 If a spider needs to disable these pipelines, they can by overriding `ITEM_PIPELINES` in `custom_settings`.
 See [`fedex.py`](../locations/spiders/fedex.py) for an example.
 
+### `DropAttributesPipeline`
+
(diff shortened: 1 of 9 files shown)


iandees:
This seems useful. Two thoughts that might be considered "devil's advocate":

* I kind of like seeing everything related to manipulating the data in the Python functions of each spider. Having special attributes that hide business logic might get confusing if someone is trying to figure out where the "name" parameter went.
* Instead of doing this in a pipeline, why not have a helper function that does it?

Cj-Malone:
So, if this was helper functions, we still have to override functions in spiders, and at that point we may as well just None the attribute. Actually I think if we are overriding those functions we should None the attributes in spider, not switch to this. As you said, it's easier to follow.

But in the same way item_attributes is helpful most of the time, but occasionally we have to manually set the brand details. I think this will be helpful in most cases, but the edge cases have to be handled in spider.

locations/pipelines/drop_attributes.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class DropAttributesPipeline:
+    def process_item(self, item: Feature, spider: Spider):
+        if not hasattr(spider, "drop_attributes"):
+            return item
+
+        for attribute in getattr(spider, "drop_attributes"):
+            if attribute in item.fields:
+                del item[attribute]
```
Cj-Malone, line 13: I think del throws an exception thi the attribute isn't present. Add an if or switch to None

docs/PIPELINES.md
```diff
@@ -4,6 +4,10 @@ After an `Item` is emitted from a spider, it goes through several pipelines befo
 If a spider needs to disable these pipelines, they can by overriding `ITEM_PIPELINES` in `custom_settings`.
 See [`fedex.py`](../locations/spiders/fedex.py) for an example.
 
+### `DropAttributesPipeline`
+
+When a spider has `drop_attributes`, we remove those attributes from `Item`s, this happens early, so it may be re added by other pipelines later on.
```
Cj-Malone, line 9: It's actually Feature now, not Item.

locations/pipelines/drop_attributes.py
iandees: Use `item.pop(attribute, None)` instead?

iandees:
Let's add tests and merge this :+1:

PR #8203, opened 2024-04-15
[choice_hotels] upgrade latest chrome user agent

panasonic-mafia:


iandees:
When `requires_proxy` is set, it ignores the user agent and Zyte will use whatever it needs to do.

Try removing the requires_proxy?

iandees:
Hm, they're probably blocking non-residential IPs. Let's see what happens without proxy on the next run.

PR #8211, opened 2024-04-16
[BreakAndWash PL] Add spider

Cj-Malone:
I think this is ready to go, but I'm gonna wait for some opinions on categories: https://github.com/osmlab/name-suggestion-index/issues/9366


Cj-Malone:
Yeah this is good to go.

PR #8243, opened 2024-04-24
[gbfs] increase delay, map physical presence

panasonic-mafia:
Some [502's popped up in weeklys](https://alltheplaces-data.openaddresses.io/runs/2024-04-20-13-31-46/logs/gbfs.txt).


iandees:
pre-commit.ci run

PR #8245, opened 2024-04-24
Fix issues that showed up from pre-commit

iandees:
pre-commit started flagging these as errors some time in the last 24 hours.


iandees:
pre-commit.ci run

PR #8264, opened 2024-04-27
Add road safety cameras for Victoria, Australia (304 locations)

davidhicks:
Reference/source data: https://www.vic.gov.au/find-road-safety-camera#searchable-map-of-camera-locations


Cj-Malone:
I have concerns about the category. We will never be able to match OSM here, we can't output relations.

I suspect the best middle ground would be to output these as `highway=speed_camera` with `enforcement` as an extra.

davidhicks:
> I have concerns about the category. We will never be able to match OSM here, we can't output relations.
> 
> I suspect the best middle ground would be to output these as `highway=speed_camera` with `enforcement` as an extra.

Unfortunately `highway=speed_camera` is meant to be used for each speed measuring device and/or camera as far as I understand. I don't believe there is any tag for traffic signal enforcement devices.

Some of these Australian intersections that come as a single feature in the source data will have up to 6 devices on separate poles that are used for measuring speed in one or two directions, as well as measuring whether a vehicle has entered an intersection after a red light. As far as I know, `highway=speed_camera` always implies `enforcement=maxspeed` whereas most of these cameras in Australia are at intersections and also enforce traffic signals.

I don't think we can follow OSM tagging guidance for road enforcement devices/locations. OSM's tagging arrangement is quite complex with multiple devices linked together in a relation including directionality of devices and their detection distances, where we generally just have very basic single POI data to roughly indicate "Nearby here is a type of road enforcement system".

PR #8272, opened 2024-04-28
Add ACT Government road safety cameras (AU) (1225 cameras)

davidhicks:
Many of these cameras are mobile (permanent=no for lack of better OSM tagging standard). The mobile camera locations are fixed points in the ACT where mobile speed cameras may be temporarily used, sometimes for just a few hours at a time.


locations/spiders/act_government_road_safety_cameras_au.py
```diff
@@ -0,0 +1,57 @@
...
+            }
+
+            # Unfortunately camera/location codes are not unique, so we need
+            # to generate our own unique references that include the location
+            # description and lat/lon fields.
+            properties["ref"] = sha1(
+                "{} {} {} {}".format(
+                    properties["name"], properties["addr_full"], properties["lat"], properties["lon"]
+                ).encode("UTF-8")
+            ).hexdigest()
```
iandees, line 42: Can you set `no_refs` instead?

davidhicks: Done, I'll update the other PRs too which used a similar "fake refs" approach.

davidhicks:
@iandees Note the count difference of 6 is due to duplicates no longer being dropped automatically with use of `no_refs=True`. Perhaps we should update the dupe filter to not just match on `ref` but also across an entire `Item` (all fields), if `no_ref=True`?

iandees:
Hmm, I'm partially inclined to let the spider do the work of deduping if `no_refs` is set. I could see cases where we _want_ to allow duplicates.

Issue #4295, opened 2022-12-09
Test for polygon containment in NSI locationSet matching


brawer:
The OpenStreetMap [Name Suggestion Index (NSI)](https://nsi.guide/) has matching rules whose `locationSet` may refer to GeoJSON polygons. This is typically used to restrict a matching rule to a set of states/provinces, but there’s also a couple of non-state polygons.

To make the AllThePlaces NSI matching logic more accurate, consider using the [Shapely python library](https://pypi.org/project/shapely/). Internally, Shapely uses the same math library as PostGIS and other mature systems, so its geometry functions are reasonably good and fast. Shapely’s `shape` function accepts an arbitrary GeoJSON geometry, builds some internal data structures from it, and returns an object whose `contains` method can be called on another shape.

```python
from shapely.geometry import shape
nsi_shape = shape(load_geojson_from_nsi()["geometry"])
place_shape = shape(scraped_atp_feature["geometry"])
if nsi_shape.contains(place_shape):
    pass  # scraped ATP feature is contained within nsi_shape
```

Not sure when exactly ATP would want to load the GeoJSON files from NSI. The implementation would probably be simpler if all GeoJSONs were loaded from NSI in bulk. However, this would slow down the startup time for all spiders, even when the loaded polygons never get used. In production, this makes no difference because all NSI rules will eventually get executed. But for development, which is usually done on a single spider, it might be better if GeoJSON files were loaded from NSI only on demand.

I’ll gladly leave the implementation to others who are more familiar with ATP internals, but perhaps you’ll still find the pointer to Shapely useful, should you want to improve the NSI matching. (I came across this because of commit 92f5664f4b22e9fcd11823915e2f6c6ea923b99a by @Cj-Malone).

Cj-Malone:
I would like to see this happen, although I probably won't be working on it myself anytime soon. But if anyone does work on it, could you please took into caching NSI at the same time, it's currently downloaded every time a spider is ran. We know it's released every Sunday so we could burn the cache every Sunday evening or every couple days if that's easier.

CloCkWeRX:
#8683 #7647 #6909 potentially would be cleaned up by rejecting points outside of the boundaries, or warning at least

iandees:
[...]

PR #8273, opened 2024-04-28
Add City of Darwin Free Wi-Fi (42 hotspots)

davidhicks:
The Shapely library is imported to find the centroid of circles (51 vertice polygons) that are provided in the source data as the estimated range of each access point.


davidhicks:
@iandees I've pulled in Shapely with this spider for more advanced geometry processing. I realise this is a fairly heavy dependency, but it is also suggested for import within #4295 and would additionally be useful for automatically raising errors about features in the middle of oceans, features not corresponding to country codes provided, etc.

iandees:
Hmmm. I'm not excited about this for two reasons:

* The centroid of the areas here are not where the antennas are. Using centroids is similar to generating a name for a record where the source data doesn't specify its own name. Is there not a data source for this that shows the points?
* Adding shapely makes installation of the All the Places code significantly more difficult for people trying to do development. If you're lucky there's a pre-compiled wheel that includes gdal, but if you have to install from source then the install with gdal becomes more complicated.

davidhicks:
> Hmmm. I'm not excited about this for two reasons:
> 
>     * The centroid of the areas here are not where the antennas are. Using centroids is similar to generating a name for a record where the source data doesn't specify its own name. Is there not a data source for this that shows the points?
> 
>     * Adding shapely makes installation of the All the Places code significantly more difficult for people trying to do development. If you're lucky there's a pre-compiled wheel that includes gdal, but if you have to install from source then the install with gdal becomes more complicated.

Unfortunately their source data is just a 51 sided polygon approximating a circle around each Wi-Fi access point. All their Polygons have the same radius, and therefore look like a simple and fairly useless buffer of 300-and-a-bit-metres around each coordinate in their source data for where the access points are located. There are other layers with point data for Wi-Fi access points, but they all look woefully out of date.

Shapely depends on GEOS which is a C/C++ library (homepage: https://libgeos.org/) and per https://libgeos.org/usage/download/#build-requirements is stated to have no dependencies beyond CMake and GCC/Clang/VisualC++ for building the library. As far as I'm aware, Shapely doesn't depend upon GDAL, even though GDAL itself depends on GEOS too. Shapely is in official package repositories on Debian, Ubuntu, Fedora, Arch, Gentoo and more per https://repology.org/project/python:shapely/versions   Does this help the situation?

davidhicks:
I could hack together a custom algorithm for this spider to take the 51 vertices and attempt to find a centroid, avoiding the need for `Shapely` as a dependency. But if #4295 is suggesting `Shapely` (I'd be happy to implement this feature request by the way), and there are other benefits to be had with `Shapely` for warning on invalid locations in oceans, and invalid country codes, etc, I'd rather not hack a centroid algorithm if `Shapely` (or similar library) may be imported in the future.

Issue #7852, opened 2024-03-13
Chick-fil-A spider uses new Yext storefinder API


davidhicks:
Chick-fil-A is now using what appears to be a different/new Yext API which allows geographic search around a point, with paginated results. The pagination limit looks like it could max out at 50, but at least the search radius appears to be large enough to cover the whole world in one query (3058 locations listed as the total in the first page of 50 results returned).

`curl 'https://locator.chick-fil-a.com.yext-cdn.com/search?r=250000&per=50&offset=50' -H 'Accept: application/json'`

Unlike the existing YextSpider, no tokens or API keys are needed.

Let's see if any other brands are using this alternative API method, and if yes, we could adjust YextSpider or create an alternate YextSpider to handle this new Yext API.

PR #8276, opened 2024-04-29
[TGI Fridays US] Fix spider

Cj-Malone:


davidhicks:
I think you've accidentally written the replacement YextSpider that #7852 also requires :) It appears Yext are in the process of changing their customers to a new API.

davidhicks:
> I think you've accidentally written the replacement YextSpider that #7852 also requires :) It appears Yext are in the process of changing their customers to a new API.

Hmm actually maybe not, #7852 looks different again.

Cj-Malone:
I specifically wrote this so it could be easily split out into shared code. Do you think Yext is dropping the `entities` endpoint for all customers, I assumed only a few had it turned off.

davidhicks:
Perhaps not a complete replacement? As this spider shows, the brand has disabled the `entities` endpoint and are instead using a different `answers/vertical` endpoint. I've seen hints at this `answers` endpoint elsewhere when writing spiders, but perhaps we've just been noticing it, then trying the `entities` endpoint instead, and it just happens to work as most brands enable the `entities` endpoint regardless of if they use the `answers` endpoint.

If we see more uses of `answers/vertical` only (no `entities` endpoint supported) I guess we could bother to create a new `YextAnswersSpider` store finder, but seemingly no benefit just yet.

locations/storefinders/yext_answers.py
```diff
@@ -0,0 +1,74 @@
...
+    api_version: str = "20220511"
+    page_limit: int = 50
+
+    def make_request(self, offset: int) -> JsonRequest:
+        return JsonRequest(
+            url="{}?{}".format(
+                self.endpoint,
+                urlencode(
+                    {
+                        "experienceKey": "tgi-fridays-search",
```
davidhicks, line 28: `experienceKey` should be another API parameter?

davidhicks (changes requested):

locations/storefinders/yext_answers.py
Cj-Malone: Done, thanks.

davidhicks (approved):

PR #8279, opened 2024-04-29
Add TGI Fridays (AU) (20 locations)

davidhicks:


locations/spiders/tgi_fridays_au.py
```diff
@@ -0,0 +1,43 @@
...
+            facebook_account = clean_facebook(location["facebook_url"])
+            set_social_media(item, SocialMedia.FACEBOOK, facebook_account)
+
+            yield Request(url=location["map_address"], meta={"item": item}, callback=self.parse_maps_url)
+
+    # Follow shortened URL to find full Google Maps URL containing coordinates
+    def parse_maps_url(self, response):
+        item = response.meta["item"]
+        item["lat"], item["lon"] = url_to_coords(response.url)
+        yield item
```
Cj-Malone, line 43: I'm not comfortable with this, this turns `https://goo.gl/maps/DQ489S2NsYwWFw5NA` into coords. Unless `DQ489S2NsYwWFw5NA` encodes the coordinates, they are Google coordinates, not TGIs. This is very different to taking them from a url where a source is effectively doing `"https://www.google.com/maps/place/@{},{},20z/".format(lat,lon)`.

Cj-Malone: 20 POIs are probably not a legal risk, but I don't want to scrap Google.

PR #8299, opened 2024-05-01
Unify instagram tagging

Cj-Malone:
5 spiders were outputting `instagram` not `contact:instagram`. https://taginfo.openstreetmap.org/search?q=instagram#keys

### DIFF
--- a/locations/spiders/crunch_fitness.py
+++ b/locations/spiders/crunch_fitness.py
@@ -1,6 +1,6 @@
 import scrapy
 
-from locations.items import Feature
+from locations.items import Feature, SocialMedia, set_social_media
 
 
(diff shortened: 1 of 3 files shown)


iandees:
PS: It would be neat to try and think of a way to trigger re-running "child" spiders when a "parent" spider changes. In this case, a generic store finder spider was improved. Another case I'm thinking of is when a spider uses another spider's attributes, and those attributes get updated.

Maybe not that critical from a CI/CD perspective, but it'd be interesting to think about.

Cj-Malone:
(I ran the 2 `StoreRocketSpider`s locally)

Yeah it's be interesting. I could imagine a extended `list` commend that would output spiders, but it could take filters such as inherited classes.

Issue #8309, opened 2024-05-03
Use clean_address function to join multiple free text lines together


davidhicks:
The `clean_address` method added in #7568 allows a standardised approach to taking messy ordered multiple line address strings (of any type of composition) and joining them together into a single string.

We can now use `clean_address` to replace the many variants throughout spiders of attempting to join these multi-line address strings. An added benefit is being able to quickly find where multi-line address strings are parsed (via searching for `clean_address` instances), making it easier to change address handling in the future.

Related to #5598

PR #8324, opened 2024-05-09
Use clean_address to join and clean addresses (5)

davidhicks:
Fixes #8309


davidhicks:
dan_murphys_au is unfixable at the moment as the source website returns 404 errors for the store locator links on their main page, as well as individual store pages.

PR #8331, opened 2024-05-09
Use clean_address to join and clean addresses (12)

davidhicks:
Fixes #8309


davidhicks:
@iandees All spiders in this batch are now fixed and working.

Primark/Penneys spiders needed a complete rewrite and the crawl of this brand in each country can be conducted in parallel, hence the high number of new spiders resulting from this rewrite.

PR #8332, opened 2024-05-09
Use clean_address to join and clean addresses (13)

davidhicks:
Fixes #8309


davidhicks:
`sallybeauty` has some form of bot protection preventing access to the store finder page and API. I think it may be PerimeterX because `PX` appears in the URL of the "are you a real user" challenge page.

PR #8320, opened 2024-05-09
Use clean_address to join and clean addresses (1)

davidhicks:
Fixes #8309

PR #8344, opened 2024-05-13
[Core] Fix clean_address with new lines

Cj-Malone:
Apparently `\s` includes `\n`, so when #8320 replaced them with ` `, and the tests started failing.

### DIFF
--- a/locations/pipelines/address_clean_up.py
+++ b/locations/pipelines/address_clean_up.py
@@ -14,6 +14,8 @@ _multiple_spaces = re.compile(r" +")
 
 
 def clean_address(address: list[str] | str) -> str:
+    if not address:
+        return ""
     if isinstance(address, list):
         address = merge_address_lines(address)
     address_list = (
--- a/locations/pipelines/address_clean_up.py
+++ b/locations/pipelines/address_clean_up.py
@@ -10,12 +10,14 @@ def merge_address_lines(address_lines: [str]) -> str:
     return ", ".join(filter(None, [line.strip() if line else None for line in address_lines]))
 
 
+_multiple_spaces = re.compile(r" +")
+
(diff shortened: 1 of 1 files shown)


davidhicks:
Thanks @Cj-Malone I hadn't realised `\n` was meant to be used in those fields. I thought use of newlines in tag values was discouraged?

In either case, perhaps we could make it a bit stricter still:

* `_multiple_and_rare_spaces = re.compile(r"\p{Z}]+")` -- replace with single ASCII space ` ` (0x20)
* `_omitted_characters = re.compile(r"\r+")` -- replace with nothing so that line endings are LF rather than CRLF type for consistency reasons, other `\p{Cc}` excluding `\n` (LF) should probably also be omitted?
* `_multiple_newlines = re.compile(r"\n+")` -- replace with single `\n`

And whilst I think of this slightly related issue:

* Make `item['website'] fields safe automatically by default (unless disabled on a spider-by-spider basis) by converting IRIs to URLs (eg encode weird Unicode characters). Scrapy already uses [`w3lib.url.safe_url_string](https://w3lib.readthedocs.io/en/latest/w3lib.html#w3lib.url.safe_url_string) per https://docs.scrapy.org/en/latest/_modules/scrapy/http/request.html#Request so I'd suggest this library could be used without introducing any further dependencies.

`\p{Z}` is summarised at `https://en.wikipedia.org/wiki/Unicode_character_property#General_Category` and `https://www.compart.com/en/unicode/category/Zs` (for `Zs`) as including all the weird types of Unicode spaces available but excluding new line control characters which are instead specified as `\p{Cc}`.

Cj-Malone:
New lines are discouraged in address strings. This replaces them with commas.

PR #8364, opened 2024-05-15
tropical_smoothie_cafe_us: fix spider

### LINKED ISSUE #8340: Problems with Spiders built on YextSpider
I've noticed a few of the spiders whose results I was using in my OSM tool have recently stopped returning any results, including five_guys_de_es_fr_gb.py , matalan_gb.py , and three_gb.py .

A common feature of these spiders is that they use the YextSpider class. Checking some other spiders that are also based on this class, reveals they've also stopped working. The spider stats suggest 404 and 403 errors are preventing the data being retrieved.

If this is a general problem affecting multiple spiders, would someone be able to take a look and see if it can be fixed?

davidhicks:
Partially fixes #8340


davidhicks:
This needs merging after #8355 as #8355 fixes a bug with extraction of opening hours in the store finder class.

PR #8382, opened 2024-05-20
[opendata_mos_hotels_ru] add spider (1.8k locations)

panasonic-mafia:


panasonic-mafia:
The spider is proxy so it fails in spider-sample workflow.

PR #8542, opened 2024-05-29
[pearle_vision] Fixed Spider (393 locations)

jhutton1:
fixed spider - sitemap rules


iandees:
Looks like they might be filtering connections by IP or something. Hopefully it runs OK in the weekly though.

PR #8575, opened 2024-06-05
Explicitly add ssh as a dependency for git in Dockerfile

CloCkWeRX:
Turns out, its not automatically a dependency. When using this dockerfile for codespaces/dev work, it's handy to be able to push to your own repo.


iandees:
Hmm, this Dockerfile isn't meant for codespace/dev work, it's used for running the spiders weekly. Is it possible to tell Codespace to use a different Dockerfile and add the ssh dependency in that one?

CloCkWeRX:
It.... is apparently. I'll look at it a bit later though.

PR #8608, opened 2024-06-10
fix typo and add note about osm and atp mismatch for name key

matkoniecz:


matkoniecz:
See also #8658 - not sure how much it is intentional "we will record what they put on website listing as `name` even where it is not an actual name" and how much it is "a clear bug, name field is for actual names only"

PR #8625, opened 2024-06-13
add trailing slashes on website tags for Ziko

matkoniecz:
yes, they are missing from published data

yes, website redirects there


matkoniecz:
See say https://zikoapteka.pl/apteki/krakow/piastow-60/ https://zikoapteka.pl/apteki/krakow/piastow-60

iandees:
Nice find, thanks.

PR #8650, opened 2024-06-19
[Hours] Add set_closed function

Cj-Malone:
I have the same worries about `closed` as I did about `off` in the `apply_yes_no` discussion. I don't think we should set negative attributes when we aren't certain. If we incorrectly say someone is closed on Sunday, or doesn't have a drive through it's a lot more drama than just leaving it null.

That being said, `apply_yes_no` has a safer way to handle this, I'm OK with opening hours having it too, when we are careful.

(Bad)
```python
item["extras"]["drive_through"] = "yes" if services.get("Drive Through") else "no"
```
If they changed `Drive Through` to `drive-through`, or anything, we'd be outputting `drive_through=no` on everything.

(Safe)
```python
apply_yes_no(Extras.DT, item, services.get("Drive Through") is True)
```
We'd be outputting nothing.

(Safe negative)
```python
apply_yes_no(Extras.DT, item, services.get("Drive Through") is True, "Drive Through" not in services)
```
We output `yes/no` or null.

Anyway, opening hours. I don't think any of us want write `oh.add_range(day, "closed", "closed")`. `oh.set_closed("Su")` or `oh.set_closed(["Mo", "Tu"])` seems better.

### DIFF
--- a/locations/hours.py
+++ b/locations/hours.py
@@ -716,6 +716,15 @@ class OpeningHours:
         for day in days:
             self.add_range(day, open_time, close_time, time_format=time_format)
 
+    def set_closed(self, days: str | list[str]):
+        for day in [days] if isinstance(days, str) else days:
+            day = sanitise_day(day)
+
+            if day not in DAYS:
+                raise ValueError(f"day must be one of {DAYS}, not {day!r}")
+            self.day_hours.pop(day, None)
+            self.days_closed.add(day)
+
     def add_range(self, day, open_time, close_time, time_format="%H:%M"):
         day = sanitise_day(day)
 
@@ -732,6 +741,7 @@ class OpeningHours:
         ):
             self.day_hours.pop(day, None)
             self.days_closed.add(day)
+            return
         if isinstance(open_time, str):
             if open_time.lower() == "closed":
                 return
--- a/locations/spiders/aldi_sud_au.py
+++ b/locations/spiders/aldi_sud_au.py
@@ -21,7 +21,7 @@ class AldiSudAUSpider(SitemapSpider, StructuredDataSpider):
         if data := response.xpath("//@data-days").get():
             for rule in json.loads(data):
                 if rule["isClosed"]:
-                    continue
+                    item["opening_hours"].set_closed(rule["day"])
                 for time in rule["intervals"]:
                     item["opening_hours"].add_range(
                         rule["day"], str(time["start"]).zfill(4), str(time["end"]).zfill(4), "%H%M"
--- a/tests/test_opening_hours.py
+++ b/tests/test_opening_hours.py
@@ -333,6 +333,14 @@ def test_ld_parse_opening_hours_array_with_commas():
     assert o.as_opening_hours() == "Mo-Su 00:00-01:00,04:00-24:00"
 
 
+def test_opening_hours_closed():
+    oh = OpeningHours()
+    oh.set_closed("Su")
+    assert oh.as_opening_hours() == "Su closed"
+    oh.set_closed(DAYS)
+    assert oh.as_opening_hours() == "Mo-Su closed"
+
+
 def test_ld_parse_opening_hours_closed():
     o = OpeningHours()
     o.from_linked_data(


Cj-Malone:
My test [POI](https://store.aldi.com.au/wa/albany/70-80-albany-highway) gives us `opening_hours=Mo-We 08:30-18:00; Th 08:30-21:00; Fr 08:30-18:00; Sa 08:30-17:00; Su closed`

davidhicks:
Annoyingly in your example `opening_hours=Mo-We 08:30-18:00; Th 08:30-21:00; Fr 08:30-18:00; Sa 08:30-17:00; Su closed` any time range not listed explicitly such as `Sa 17:00-24:00` is still defined by the OSM `opening_hours` specification as `unknown` rather than `closed`. Thus to fully implement an unambiguous `opening_hours`, all our output strings would be very long and hard for a human to read. I wish the specification either defaulted to `closed` or allowed the default to be specified in each string. I think this approach of only unambiguously addressing closure of entire days is a good enough compromise.

davidhicks:
We'll also need to think about the example of `Mo 08:00-02:00; Tu closed` which ATM will now happily output. I think per the OSM `opening_hours` specification, a computer would be asked to interpret this input string as if it were `Mo 08:00-24:00; Tu closed`.

iandees:
I think it's fair to point out that we don't try to match OSM's `opening_hours` spec _precisely_ and document where it differs slightly. To me, all of these things are common-sense edge cases and the way I tag things when mapping in person.

jleedev:
I note that https://schema.org/openingHours doesn't say that "closed" is a valid time range but people seem to be using it anyway.

If these edge cases turn out to be problematic, we could conservatively produce an opening_hours containing `closed` _only_ if no time ranges exist, purely to say "the data said something and not nothing". Or remove the code that causes `closed` to clobber existing time ranges for the day, to better garbage out when garbage in.

jleedev:
> set negative attributes when we aren't certain. If we incorrectly say someone is closed on Sunday

To be clear, this

https://github.com/alltheplaces/alltheplaces/blob/b6322b61a2c496d7d1b388d49e4cba99bd7a950b/locations/hours.py#L834-L835

seems pretty certain to me; it will output `closed` only for the days which actually say so and nothing otherwise, which seems to align with your drive through example. Are there some parsing ambiguities you have found?

Cj-Malone:
> > set negative attributes when we aren't certain. If we incorrectly say someone is closed on Sunday
> 
> To be clear, this
> 
> https://github.com/alltheplaces/alltheplaces/blob/b6322b61a2c496d7d1b388d49e4cba99bd7a950b/locations/hours.py#L834-L835
> 
> seems pretty certain to me; it will output `closed` only for the days which actually say so and nothing otherwise, which seems to align with your drive through example. Are there some parsing ambiguities you have found?

I think the current code is good. I'm saying lets not write bad spider code, eg:


```python
for rule in ["Mo 09:00-17:00", "Tu 09.00-17:00", "We 09.00- 17:00"]:
    if m := re.match("(\w\w) (\d:\d:\d\d)-(\d:\d:\d\d)", rule):
        oh.add_range(*m.groups())
    else:
        oh.set_closed()
```

Would become `Mo 09:00-17:00, Tu,We closed`.

matkoniecz:
>  any time range not listed explicitly such as Sa 17:00-24:00 is still defined by the OSM opening_hours specification as unknown rather than closed.

not really? `opening_hours` specification specifies which times something is open (with potential open ended or descriptive coverage!), remaining parts are closed - not unknown.

This is also overwhelming in practical use, rare cases of `off` are applied for things like `;PH off` overriding regular times (and explicitly specifying PH is fairly rare),

Constructs like `Sa 00:00-17:00 off, Sa 17:00-24:00` are in basically no use and normal recording such info is `Sa 17:00-24:00`

davidhicks:
> > any time range not listed explicitly such as Sa 17:00-24:00 is still defined by the OSM opening_hours specification as unknown rather than closed.
> 
> not really? `opening_hours` specification specifies which times something is open (with potential open ended or descriptive coverage!), remaining parts are closed - not unknown.
> 
> This is also overwhelming in practical use, rare cases of `off` are applied for things like `;PH off` overriding regular times (and explicitly specifying PH is fairly rare),
> 
> Constructs like `Sa 00:00-17:00 off, Sa 17:00-24:00` are in basically no use and normal recording such info is `Sa 17:00-24:00`

* https://wiki.openstreetmap.org/wiki/Key:opening_hours/specification doesn't appear to state whether the absence of a rule for a day indicates closure on that day, or indicates a state of uncertainty about opening hours
* https://wiki.openstreetmap.org/wiki/Key:opening_hours#Summary_syntax states "unknown: initial open or closed state" (e.g. parser would prepend "Mo-Su unknown;" to all opening hours strings, then override that default state with whatever opening hours are actually then specified)
* https://wiki.openstreetmap.org/wiki/Key:opening_hours/specification#explain:rule_modifier:empty states that a rule without a rule modifier (such as "closed") is assumed to have a rule modifier of "open", so "Tu 10:00-16:00" is open without having to specify "Tu 10:00-16:00 open"
* https://wiki.openstreetmap.org/wiki/Key:opening_hours/specification#explain:rule_modifier:closed states that a rule modifier of "closed" will "not overwrite anything else as rules with other rule modifiers would do", so perhaps my example is wrong that `Mo 08:00-02:00; Tu closed` would only be treated as `Tu 02:00-24:00` being closed, as the `Tu closed` rule doesn't override the earlier rule as would typically occur had `closed` not been included

Cj-Malone:
I don't like to repeat these OSM proverbs but "the wiki isn't authoritative".

matkoniecz:
OK, maybe it is not formally defined, but in actual practice it is done this way (I guess that is question whether ATP follows `opening_hours` formal specification or `opening_hours` as used in OSM)

davidhicks:
> I don't like to repeat these OSM proverbs but "the wiki isn't authoritative".

This is OK, as most of the nuance of OSM tag usage seems to come down to GitHub issues/pull requests for OSM related software libraries where someone has just gone ahead with a particular interpretation of an unclear tag, and others follow in the footsteps for compatibility reasons. No doubt for something as complex as the `opening_hours` tag, machine interpretation of the tag and interoperability between software using this tag is quite a challenge and there are many different ways software (and users) are interpreting the tags.

I suggest we further develop the already partly existent ATP opinionated specification of `opening_hours`. ATP doesn't currently care about a lot of the features of `opening_hours` that other OSM projects care for (such as public holiday hours) so the ATP opinionated specification isn't needing to be very lengthy or complex.

I would like to however have a way to distinguish between hours provided by a source for only the current week or fortnight (including public holiday closures but this reason not stated by the source), versus the more common pattern of hours being provided by a source for a typical week, with public holiday closures listed separately from the typical week hours. Either as a gold plated `OpeningHours.add_specific_day_range(year: int|str, month: int|str, day: int|str, start_time: str, end_time: str)` or as `OpeningHours.add_comment(note: OpeningHoursComment)` where, for example, specifying `OpeningHoursComment.TEMPORARY_HOURS` results in an output such as `Mo-Fr 08:00-17:00 "temporary hours";`. Another option the OSM specification appears to support is `"temporary hours": Mo-Fr 08:00-17:00; Sa-Su 10:00-14:00` if specific days of the year are unknown.

davidhicks:
Let's move the discussion elsewhere as it's off-topic. I see this PR as ready to merge

davidhicks (approved):

PR #8659, opened 2024-06-20
also handle "closed" when whitespace-splitting a microdata openingHours

jleedev:


iandees:
Thanks for adding tests.

PR #8681, opened 2024-06-24
remove ace_safes_gb.py spider due to low quality

matkoniecz:
There is no indication that the places are actually existing. 

Let alone that these are shops. I wanted to start discussion thread is shop=safe a good tag and document it as in use at OSM Wiki, maybe even get it added to iD presets. So I went to find example of how their shop looks like. This failed, and it turned into checking are these places where you can walk-in without prearranged meeting (and qualifying for shop=). 

Later it turned into checking are these POIs existing at all and failed to find confirmation. Then I started asking UK mappers and they commented that supposed location data is a garbage data. Some locations are placed where Google API locates towns/cities (Bournemouth), York locations is heavily misplaced at best (if existing at all) with at best 300m offset where first plausible location for such POI is located.

I thought also about alternatives to just removal and it included emitting `fixme=likely to be not actually existing or at least misplaced` and dropping shop=safe and keeping spider. Or using more structured `fixme:location=highly imprecise` `fixme:existence=likely to be not actually existing`, with or without dropping `shop` tag.

Either way I think that it should be dropped or formatted in way that marks this spider as extremely dubious to make possible for data consumers to skip low-quality data.

--------------------------

And to be clear I appreciate work that went into adding it and I think it looked plausible, at least without looking closer into this dataset.


Cj-Malone (approved):

matkoniecz:
quoting SomeoneElse from osm IRC: 

> If you want, you can quote me saying that that Ace Safes location is implausible based on local knowledge.

davidhicks (approved):
Agreed, bin it. Git blame shows I added this spider but obviously I didn't check whether any of the features are real. It looks like they may just be dropping a pin on cities that they like to perform work in, and they come to customer residences/offices. I don't think customers ever attend any premises of Ace Safes.

PR #8688, opened 2024-06-24
this wikidata does not exist, was deleted as not notable

matkoniecz:
https://www.wikidata.org/wiki/Special:Log?page=Q123251317

### DIFF
--- a/locations/spiders/pijalniewedel.py
+++ b/locations/spiders/pijalniewedel.py
@@ -15,7 +15,6 @@ class PijalnieWedel(Spider):
     name = "pijalniewedel_pl"
     item_attributes = {
         "brand": "Pijalnia Czekolady E.Wedel",
-        "brand_wikidata": "Q123251317",
         "extras": Categories.CAFE.value,
     }
     start_urls = ["https://wedelpijalnie.pl/lokale"]


davidhicks:
Is [Q1273613](https://www.wikidata.org/wiki/Q1273613) the same brand?

matkoniecz:
Yes, it is the parent brand ("Pijalnia Wedel" is subbrand referring to this specific chain, note that I have no idea about ownership and financial situation but it is definitely subbrand)

matkoniecz:
replaced by #8705

davidhicks:
Thanks! Wikidata is no where near as deletionist as Wikipedia, but they do tend to eventually delete obscure items with no statements. If the sub-brand has it's own website, or signage or logo (I looked up NSI and it wasn't listed either) then it's worth having a separate Wikidata item with statements identifying it is a sub-brand, what the particular website is, etc. That way it generally won't get deleted

matkoniecz:
in this case I really do not see how sepaarte subbbranded itesms would be beneficial or needed so using Wedel brand as in https://github.com/alltheplaces/alltheplaces/pull/8705 seems good enough (and in this case benefits `brand:wikidata` are especially small, but it seems standard to include this extra tag in ATP and throwing it away by data consumers is much easier than adding it so it can stay I guess)

PR #8689, opened 2024-06-24
drop clearly broken leroy_merlin_pl locations

matkoniecz:
some locations are in Yemen/ocean near Oman, and as Poland has no colonies this is clearly broken. Address data present there confirms that coordinates are garbage (general area looks like switched lat-lons, but not tried confirming this theory)

I am not entirely sure how to best do it, other spiders are likely to be also affected (look at ATP global map, note supposed POIs in the middle of oceans). Maybe it should be separate pipeline, however these are defined?

It would be also possible for data consumers to do nothing and keep this claims as stated. And each data consumer skipping these data manually, I guess, but that seems overall vastly significant effort and keeping known trap for data users.

### DIFF
--- a/locations/spiders/leroy_merlin_pl.py
+++ b/locations/spiders/leroy_merlin_pl.py
@@ -21,8 +21,8 @@ class LeroyMerlinPLSpider(Spider):
             item["phone"] = poi["informationPointPhone"]
             item["website"] = urljoin("https://www.leroymerlin.pl", poi.get("wwwUrl"))
             # TODO: figure out opening hours
-            if item['lat'] < 48.99 or item['lat'] > 54.87 or item['lon'] < 14.07 or item['lon'] > 24.13:
+            if item["lat"] < 48.99 or item["lat"] > 54.87 or item["lon"] < 14.07 or item["lon"] > 24.13:
                 # clearly outside Poland, some locations were claimed to be in the Middle East
-                del item['lat']
-                del item['lon']
+                del item["lat"]
+                del item["lon"]
             yield item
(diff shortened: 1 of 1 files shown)


davidhicks:
@matkoniecz These two coordinates are flipped. Can you change the spider so these two coordinates have their lat/lon flipped instead of coordinates deleted?

```
{"type": "Feature", "id": "ztjjdQ6qI9q394E-ayTfR9y6jwc=", "properties": {"ref": "108", "@spider": "leroy_merlin_pl", "shop": "doityourself", "addr:street_address": "Koszali\u0144ska 1", "addr:city": "Koszalin", "addr:postcode": "75-900", "addr:country": "PL", "name": "Koszalin", "phone": "+48 94 372 88 00", "website": "https://www.leroymerlin.pl/sklepy/koszalin,dh108.html", "brand": "Leroy Merlin", "brand:wikidata": "Q889624", "nsi_id": "leroymerlin-092aaf"}, "geometry": {"type": "Point", "coordinates": [54.231566, 16.177192]}},

{"type": "Feature", "id": "4L1H_6ZlxQR_3ws5HhJQgtmAv3g=", "properties": {"ref": "109", "@spider": "leroy_merlin_pl", "shop": "doityourself", "addr:street_address": "Serbska 7", "addr:city": "Pozna\u0144", "addr:postcode": "61-696", "addr:country": "PL", "name": "Pozna\u0144 Serbska", "phone": "+48 61 679 58 00", "website": "https://www.leroymerlin.pl/sklepy/poznan-serbska,dh109.html", "brand": "Leroy Merlin", "brand:wikidata": "Q889624", "nsi_id": "leroymerlin-092aaf"}, "geometry": {"type": "Point", "coordinates": [52.432892, 16.950744]}},
```

matkoniecz:
I guess it would be good thing, recovering some - at risk of having some false positives in the future...

added to my TODO list

PR #8695, opened 2024-06-25
Create metrorower_pl.py

ireun:
Bike system in GZM Metropolis, Poland https://metrorower.transportgzm.pl/

By 01.08.2024 the system should have 1000 stations in total


iandees:
Looks like you might be able to extract the capacity of each station, too? Could you add that to `item["extras"]["capacity"]` please?

ireun:
I could, but I don't think I should. 

See this is 4th gen system, which does not require you to attach the bike to anything. That means that stations can hold much more bikes than capacity indicates.

Issue #6433, opened 2023-11-29
Zyte API account suspended


panasonic-mafia:
I noticed that spiders that require proxy failed due to Zyte API account suspension: 
https://data.alltheplaces.xyz/runs/2023-11-25-13-32-05/logs/vkusno_i_tochka_ru.txt

iandees:
Yea, they have a weird way of implementing spending limits and just raised their rates on some of the URLs we request.

I'll look at bumping up the limit.

iandees:
I don't think it's possible to do this directly. The Zyte API doesn't know what spider is originating the request and I don't see a way to get total API costs per domain, either. It would be nice if the spider got a cost in the API response so that it could sum it up as part of gathering metrics.

I submitted a support ticket to Zyte to see if we can learn more.

iandees:
Thanks for pulling together the counts.

Their pricing calculator (https://app.zyte.com/o/zyte-api/pricing-calculator) gives me the impression that 4 runs per month shouldn't be going above $30/month, so I'll keep working with them via support ticket to understand where the discrepancy is.

davidhicks:
The 20231118 run ran out of Zyte credit mid way through the run as best as I can tell.

The 20231111 run appears to have had Zyte credit the entire duration of the run:

```
./fetch_all_stats.sh 2023-11-11-13-32-03
./check_stats.sh 2023-11-11-13-32-03 scrapy-zyte-api/attempts
```

```
2023-11-11-13-32-03/american_freight.json: 378
2023-11-11-13-32-03/anthropologie.json: 243
2023-11-11-13-32-03/aussie_disposals_au.json: 1
2023-11-11-13-32-03/auto_nation.json: 2
2023-11-11-13-32-03/big_w_au.json: 2
2023-11-11-13-32-03/blains_farm_and_fleet_us.json: 47
2023-11-11-13-32-03/canadian_tire_ca.json: 988
2023-11-11-13-32-03/caseys_general_store.json: 2643
2023-11-11-13-32-03/chemist_warehouse_au.json: 2
2023-11-11-13-32-03/choice_hotels.json: 5695
2023-11-11-13-32-03/cineworld_gb_je.json: 2
2023-11-11-13-32-03/crackerbarrel.json: 664
2023-11-11-13-32-03/cube_smart_us.json: 1496
2023-11-11-13-32-03/discount_tire.json: 2
2023-11-11-13-32-03/extended_stay_america.json: 5
[...]

PR #8707, opened 2024-06-26
[OlliesBargainOutlet] Add requires proxy

karen-kerr:


matkoniecz:
Maybe run each proxied spider once three months? Or once a year? And until rerun happen republish data from last run? (fiddly, may be confusing)

Is there any large company using this dataset? Maybe they can throw 200$/month at the problem? (easier to propose than do it)

Maybe distribute running proxied parts to trusted users? If there is some PL-specific proxy (or without geoblock) maybe I can run it and upload results somewhere? And USA-specific one done by someone else (that would be nightmare to coordinate)

Cj-Malone:
What about the OSMF? If a director wants the data closer to OSM, maybe they should take it to a meeting.

davidhicks:
@matkoniecz See https://github.com/alltheplaces/alltheplaces/issues/6433 for some more background

At the time that report was generated, 80 spiders used Zyte API. The cheapest 67 spiders using Zyte API cost about $2/run for all 67 spiders. The remaining 13 cost $28/run for all 13 spiders, of which 7 spiders cost about $7/run in total and the 5 most expensive spiders cost $21/run in total.

If the report for Zyte expenditure was to be regenerated (as many more spiders now rely upon it), we could add a new spider attribute estimating the Zyte cost to run each spider. Then we could just skip spiders where the Zyte cost is not worth it, at least not worth the cost once per week. A $30/month budget back then would have allowed 73 of the 80 Zyte spiders to run every single week without overrunning the $30/month budget. Even if the budget was $8/month we could still run 67 of the 80 Zyte spiders every single week.

iandees:
Let's move this conversation over to https://github.com/alltheplaces/alltheplaces/issues/8726

PR #8713, opened 2024-06-26
[dino_pl] Fixed issue with parsing decryption parameters

ckpiper:
Found an issue with regex string for parsing the decryption parameters. I fixed it by adding some white space and different quoting characters. 

The next issue is with getting the encrypted GeoJSON from the api. Both the api and the robots.txt for `api-dino.appchance.shop` say that the site is under maintenance mode. With that I am not sure why it seems to work with `https://marketdino.pl/map`.

### DIFF
--- a/locations/spiders/dino_pl.py
+++ b/locations/spiders/dino_pl.py
@@ -11,7 +11,8 @@ from locations.hours import OpeningHours
 class DinoPLSpider(Spider):
     name = "dino_pl"
     item_attributes = {"brand": "Dino", "brand_wikidata": "Q11694239"}
-    allowed_domains = ["marketdino.pl", "api-dino.appchance.shop"]
+    allowed_domains = ["marketdino.pl"]
+    custom_settings = {"ROBOTSTXT_OBEY": False}
 
     def start_requests(self):
         yield Request(
@@ -25,7 +26,7 @@ class DinoPLSpider(Spider):
         if m := re.search(r"""\.from\([\n ]*['"]([0-9a-f]{32})['"],[\n ]*['"]hex['"][\n ]*\)""", response.text):
             iv = m.group(1)
         yield Request(
-            url="https://api-dino.appchance.shop/api/v1/dino_content/geofile/",
+            url="https://api.marketdino.pl/api/v1/dino_content/geofile/",
             meta={"key": key, "iv": iv},
             callback=self.parse_encrypted_geojson,
         )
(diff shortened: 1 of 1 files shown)


davidhicks:
Nice one. I investigated the HTTP 503 problem and found that requesting `robots.txt` causes subsequent requests to fail too. They might be detecting bots by whether they ask for `robots.txt`.

With you regular expression change and inclusion of `custom_settings = {"ROBOTSTXT_OBEY": False}` under `allowed_domains=...` the spider works for me.

ckpiper:
@davidhicks can you change it to what works for you? As adding customs_settings still does not get it to work for me. I am still getting 503 errors.

davidhicks:
Sorry I missed something--I also updated the URL of the encrypted blob of store data as it has changed. Can you see if this diff works for you:

```
diff --git a/locations/spiders/dino_pl.py b/locations/spiders/dino_pl.py
index 0fce0052..f5600624 100644
--- a/locations/spiders/dino_pl.py
+++ b/locations/spiders/dino_pl.py
@@ -11,7 +11,8 @@ from locations.hours import OpeningHours
 class DinoPLSpider(Spider):
     name = "dino_pl"
     item_attributes = {"brand": "Dino", "brand_wikidata": "Q11694239"}
-    allowed_domains = ["marketdino.pl", "api-dino.appchance.shop"]
+    allowed_domains = ["marketdino.pl"]
+    custom_settings = {"ROBOTSTXT_OBEY": False}
 
     def start_requests(self):
         yield Request(
@@ -20,12 +21,12 @@ class DinoPLSpider(Spider):
         )
 
     def parse_decryption_params(self, response):
-        if m := re.search(r'\.from\("([0-9a-f]{64})","hex"\)', response.text):
+        if m := re.search(r"""\.from\([\n ]*['"]([0-9a-f]{64})['"],[\n ]*['"]hex['"][\n ]*\)""", response.text):
             key = m.group(1)
-        if m := re.search(r'\.from\("([0-9a-f]{32})","hex"\)', response.text):
+        if m := re.search(r"""\.from\([\n ]*['"]([0-9a-f]{32})['"],[\n ]*['"]hex['"][\n ]*\)""", response.text):
             iv = m.group(1)
         yield Request(
-            url="https://api-dino.appchance.shop/api/v1/dino_content/geofile/",
+            url="https://api.marketdino.pl/api/v1/dino_content/geofile/",
             meta={"key": key, "iv": iv},
             callback=self.parse_encrypted_geojson,
         )
```

ckpiper:
Yes. That does work for me!

PR #8723, opened 2024-06-27
[OmniHotels] Change non running spider to use sitemap

karen-kerr:


iandees:
@Cj-Malone do you know where `addr:city` comes from on this? I don't see it in the spider code.

Cj-Malone:
Presumably `CountryCodeCleanUpPipeline`, with the geocoder.

Cj-Malone:
Looks like they do have broken down addresses in `https://www.omnihotels.com/hotels/amelia-island` `//script[contains(text(), "propertyTelephone")]` if we need them.

iandees:
Ah yea, I forgot that it does admins other than country:

https://github.com/alltheplaces/alltheplaces/blob/6c999bd462a5099b8c7cdecbd046b2ba5564528b/locations/pipelines/country_code_clean_up.py#L43-L44

At least in the US, `admin2` is "county", not "city". We might want to fix that.

iandees:
The property web pages didn't seem to be consistent with what data they offered, so I didn't try to parse anything else out.

PR #8727, opened 2024-06-27
Add ATP ingest code

Cj-Malone:
I've written this code several times for when I want to format shift a file from the weekly into .osm (because JOSM struggles with geojson files). It may as well go in master in case anyone else wants to use it. It also has the massive benefit of if an organisation wanted to publish their data in the ATP format, we could ingest it in basically no code, see the attached spider.

### DIFF
--- a/locations/spiders/tomtom.py
+++ b/locations/spiders/tomtom.py
@@ -1,0 +1,6 @@
+from locations.storefinders.alltheplaces import AllThePlacesSpider
+
+
+class TomTomSpider(AllThePlacesSpider):
+    name = "tomtom"
+    start_urls = ["https://download.tomtom.com/open/fcd/pois.json"]
--- a/locations/storefinders/alltheplaces.py
+++ b/locations/storefinders/alltheplaces.py
@@ -8,6 +8,10 @@ from locations.items import Feature
 
 
 class AllThePlacesSpider(Spider):
+    """
+    This class can help parse ATP style geojson files (from https://www.alltheplaces.xyz/spiders.html, or other sources)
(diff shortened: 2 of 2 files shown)


iandees:
Can you add a quick comment at the top of the `AllThePlacesSpider` to describe what's going on? The self-referential name might be confusing to people.

iandees:
Oh, so you would use `scrapy crawl spider_name --output spider_name.osm` or something to have it do the output conversion?

Cj-Malone:
Yeah, I haven't pushed an empty spider to do it because it'll show up with the list command, and that will show up in the weekly. We could do it with a command. But I just have a bunch of "test_[0-9].py` spiders full of junk. So a command, or a spider like:

```python
from locations.storefinders.alltheplaces import AllThePlacesSpider


class AtpReformatSpider(AllThePlacesSpider):
    name = "atp_reformat"

    def __init__(self, url: str = "", **kwargs):
        self.start_urls = [url]
        super().__init__(**kwargs)
```
With `pipenv run scrapy crawl atp_reformat -a url="https://alltheplaces-data.openaddresses.io/runs/2024-06-15-13-31-46/output/repsol_pe.geojson" -O repsol_pe.osm` and it'll use the osm exporter.

The last successful run of Little Caesars US has `website` tags for the API, not the actual website, so I did the equivalent of overriding `post_process_feature` and fixing the link, eg `https://api.cloud.littlecaesars.com/bff/api/stores/11413/` to `https://littlecaesars.com/en-us/store/11413/` and then using that data to update OSM.

iandees:
Nice, good idea.

PR #8734, opened 2024-06-28
Add contacts attribute on spider

Cj-Malone:
I've been vocal in that some brands should be looking after their own spiders, when they are, we should document a contact method for them. In theory it could be plumed in to email them automatically when there is an issue with a spider.


iandees:
Nice, good idea.

PR #8762, opened 2024-07-03
not all abs shops are abc branded or having abc name

matkoniecz:
see say  https://www.openstreetmap.org/?mlat=50.042003&mlon=20.060863#map=19/50.04192/20.06127

IMG_20240702_173635
screen01

### DIFF
--- a/locations/spiders/abc_pl.py
+++ b/locations/spiders/abc_pl.py
@@ -6,7 +6,7 @@ from locations.dict_parser import DictParser
 
 class ABCPLSpider(Spider):
     name = "abc_pl"
-    item_attributes = {"brand": "abc", "brand_wikidata": "Q11683985"}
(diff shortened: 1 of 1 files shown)


matkoniecz:
from data I see no way to recover it

there is occasional `"format": "Supermarket"` in data but I do not see evidence of it being right

matkoniecz:
>  raise KeyError(f"{self.__class__.__name__} does not support field: {key}")
> KeyError: 'Feature does not support field: shop'

Cj-Malone:
> raise KeyError(f"{self.**class**.**name**} does not support field: {key}")
> KeyError: 'Feature does not support field: shop'

Same as chorten_pl

matkoniecz:
> Same as chorten_pl

ops. See #8765 for unbreak attempt

PR #8764, opened 2024-07-03
[five_guys_cn] Fixed issue

ckpiper:


iandees:
There appear to be lat/lons in a uriencoded JSON blob on each store page. Might not be worth it for 5 stores though.

PR #8767, opened 2024-07-04
fix #8692 - drop service areas listed as shops

### LINKED ISSUE #8692: tapi_gb has many fake entries with fictional nonexisting shops (for SEO/ad purposes)
They created many fake entries for "home visit"

It seems possible that ones with "shop" in name are actually real

But say https://www.tapi.co.uk/stores/hoddesdon is an example of bogus entry

> Book a Free Home Visit and we’ll bring our store to you!

(report based on IRC discussion on OSM server, thanks for review)

matkoniecz:


locations/spiders/tapi_gb.py
```diff
@@ -17,4 +17,5 @@ def parse_store(self, response):
         item = LinkedDataParser.parse(response, "HomeGoodsStore")
         if item:
             item["ref"] = response.url
-            yield item
+            if item.get("image") != "https://www.tapi.co.uk/":  # drop fake CEO "locations" pretending to be shops
```
CloCkWeRX, line 20:
```suggestion
            if item.get("image") != "https://www.tapi.co.uk/":  # drop fake SEO "locations" pretending to be shops
```
?

PR #8782, opened 2024-07-08
[Dino PL] Improve name

matkoniecz:
and I have not confirmed is it fitting `branch` - and these are unusual in Poland

see also #8739

(I expect that many like this will be needed before I will be able to trust ATP data)

### DIFF
--- a/locations/spiders/dino_pl.py
+++ b/locations/spiders/dino_pl.py
@@ -40,16 +40,11 @@ class DinoPLSpider(Spider):
             if location["properties"]["status"] != "MARKET OTWARTY":  # "MARKET OPEN"
                 continue
             item = DictParser.parse(location["properties"])
+            item["branch"] = item.pop("name", None)
             item["geometry"] = location["geometry"]
             item["opening_hours"] = OpeningHours()
             if week_hours := location["properties"].get("weekHours"):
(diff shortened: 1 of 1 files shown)


matkoniecz:
note that it is almost certainly not actually a `branch` code, at least not in https://taginfo.openstreetmap.org/keys/branch#overview meaning

Cj-Malone:
https://github.com/alltheplaces/alltheplaces/pull/7883#issuecomment-2000272067

matkoniecz:
So should `branch` be used for any identifier in web listing, no matter whether it is actually used within shops?

matkoniecz:
(the same should be done with https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/empik_pl.py - I will try doing it after this PR is merged, personally I would prefer to emit `branch` where this distinction is seriously used, but moving such web listing code to `branch` field from `name` field already significantly helps)

Cj-Malone (approved):

matkoniecz:
BTW, with name being popped I am a bit confused from where name comes now. NSI?

iandees:
The name comes from NSI, yes. The [`ApplyNSICategoriesPipeline`](https://github.com/alltheplaces/alltheplaces/blob/master/locations/pipelines/apply_nsi_categories.py) code uses the wikidata ID to find an entry in Name Suggestion Index and then applies the tags from NSI to the outputted item.

PR #8796, opened 2024-07-09
[Hours] Stop outputting 24/7

### LINKED ISSUE #6537: 24/7 opening hours
Technically 24/7 means Mo-Su,PH 00:00-24:00, but we mean Mo-Su 00:00-24:00. We should probably change to output the latter.

Cj-Malone:
Fixes #6537

As mentioned before `opening_hours=24/7` actually means `opening_hours=Mo-Su,PH 00:00-24:00`, but we don't (and never will) collect `PH`. We probably will collect specific days in the future.

### DIFF
--- a/locations/hours.py
+++ b/locations/hours.py
@@ -786,22 +786,17 @@ class OpeningHours:
         day_groups.append(this_day_group)
 
         opening_hours = ""
-        if len(day_groups) == 1 and day_groups[0]["hours"] in (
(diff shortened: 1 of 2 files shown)


iandees:
This is one of those times where I think OSM's tagging is wrong (public holidays should not be included in `24/7`), but I'm ok with removing 24/7 in order to be more specific.

PR #8798, opened 2024-07-10
Add parquet output alongside GeoJSON

### LINKED ISSUE #8752: Add GeoParquet format
How about adding GeoParquet format to the current output format GeoJSON?

Consuming GeoParquet is much more efficient than GeoJSON - even for "full-blown, row-oriented" data exchange needs.

This would improve the usability of ATP even more - together with issue #4255.

iandees:
Fixes #8752 

This adds a script that converts the GeoJSON files outputted by scrapy to (Geo)Parquet format and then uses that script to generate parquet in the pull request and weekly "run all spiders" steps. Those scripts then upload the parquet files so that they can be downloaded.

Once this is confirmed working, I'll add it as a download option on the main page.


iandees:
I am using a script to convert GeoJSON to Parquet here because I initially wrote it for OpenAddresses. I think I'll probably ditch the script and write an item exporter to write to Parquet format and use that instead before merging this.

jleedev:
According to [gpq], the output is missing some metadata.

```
 ¶ gpq describe output.parquet
 ⚠️ Not a valid GeoParquet file (missing the "geo" metadata key).
 ¶ gpq convert output.parquet output.geoparquet
 ¶ duckdb
D select * from parquet_kv_metadata('output.parquet');
┌────────────────┬──────────────┬──────────────────────────────────────────────┐
│   file_name    │     key      │                    value                     │
│    varchar     │     blob     │                     blob                     │
├────────────────┼──────────────┼──────────────────────────────────────────────┤
│ output.parquet │ pandas       │ {\x22index_columns\x22: [\x22__index_level…  │
│ output.parquet │ ARROW:schema │ /////xg4AAAQAAAAAAAKAA4ABgAFAAgACgAAAAABBA…  │
└────────────────┴──────────────┴──────────────────────────────────────────────┘
D select * from parquet_kv_metadata('output.geoparquet');
┌───────────────────┬────────┬─────────────────────────────────────────────────┐
│     file_name     │  key   │                      value                      │
│      varchar      │  blob  │                      blob                       │
├───────────────────┼────────┼─────────────────────────────────────────────────┤
│ output.geoparquet │ pandas │ {\x22index_columns\x22: [\x22__index_level_0_…  │
│ output.geoparquet │ geo    │ {\x22version\x22:\x221.0.0\x22,\x22primary_co…  │
└───────────────────┴────────┴─────────────────────────────────────────────────┘
D
```

[gpq]: https://pkg.go.dev/github.com/planetlabs/gpq/cmd/gpq

Also, don't use processes=32, let it use the number of CPUs or add a -j flag, or both.

iandees:
Ok, this now uses a Parquet ItemExporter based on Geopandas instead of a script that (incorrectly) did the conversion from GeoJSON.

The conversion script now is a "concatenate" script that (hopefully) correctly concatenates multiple parquets into one big one during the weekly run. `gpq validate` (thanks for the pointer, @jleedev) seems to be happy with both the output of the ItemExporter and the concatenated output.

jleedev:
Some obervations:
- Geopandas peaked at 39 GiB converting the latest run, but only used one CPU >_>
- Geopandas produces a 290 MB file using SNAPPY. Gpq converts this to ZSTD at 202 MB.
- The file contains 1028 columns. This maximizes selectivity but means that `select *` is very slow. Perhaps some of the keys could be aggregated at a `:` character into a record or array outside of a well-known list of useful keys. For instance, there are 495 `ref:*`, 62 `payment:*`, 46 `fuel:*`, 30 `name:*`, 26 `brand:*`.
- I'm pretty new at (Geo)Parquet, but I think it would require a `bbox` column in order to support spatial selectivity, otherwise you can only query the binary WKB column.
- It would in principle be possible to query across many runs by using Hive Partitioning, if A) it were possible to List Objects in the object store, and B) the files were named like `/run_id=2024-07-06-13-31-59/filename1.parquet`.

iandees:
> Geopandas peaked at 39 GiB converting the latest run, but only used one CPU

That's annoying. Do you know of any other tooling that would concatenate these files together fairly easily?

jleedev:
Also, I get this error:

> ImportError: Missing optional dependency 'pyarrow.parquet'. pyarrow is required for Parquet support.  "
        "Use pip or conda to install pyarrow.parquet.

jleedev:
> > Geopandas peaked at 39 GiB converting the latest run, but only used one CPU
> 
> That's annoying. Do you know of any other tooling that would concatenate these files together fairly easily?

This is inherently a somewhat expensive operation since the schemata need to expanded to contain all the columns (many of which will be null). Geopandas really didn't take too long, it was at most 10-15 minutes on a GCP t2a-standard-32 (32 vCPU, 128 GB memory).

Anyway, maybe give duckdb a try:

```sql
copy (
  select * from read_parquet('./output/*/output.parquet', union_by_name=True)
) to 'output.parquet' (format 'parquet', codec 'zstd');
```

although it does not preserve the `geo` metadata, needing another pass over with `gpq`.

iandees:
I'm going to merge this and see how it goes. The feedback so far is great, but not a blocker I don't think. We can definitely improve output later.

iandees:
For the record, the instance that runs the weekly uses a Fargate instance with 8GB memory and 2 vCPUs via Elastic Container Service. The whole dataset might not fit in memory there.

jleedev:
Oh I bet Geopandas is trying to store the _entire_ uncompressed, mostly null dataset in memory.

2855967 * 1028 * 8 = 22 Gi

Yeah that tracks.

locations/exporters/geoparquet.py
```diff
@@ -0,0 +1,27 @@
...
+        if self.first_item:
+            # TODO Figure out a way to attach dataset attributes to the parquet file.
+            # self.write_dataset_attributes_table(item)
+            self.first_item = False
+
+        self.features.append(item_to_geojson_feature(item))
+
+    def finish_exporting(self):
+        # Create a GeoDataFrame from the features
+        gdf = geopandas.GeoDataFrame.from_features(self.features)
```
Cj-Malone, line 24: This will probably need to be written in batches. I used an extreme example (lots of POIs, big json in memory too), but I got close to 6GB of ram used for the single spider.

```python
from locations.storefinders.alltheplaces import AllThePlacesSpider


class TestSpider(AllThePlacesSpider):
    name = "test_3"
    start_urls = ["https://alltheplaces-data.openaddresses.io/runs/2024-06-08-13-32-01/output/visa.geojson"]
    custom_settings = {"DOWNLOAD_TIMEOUT": 60}
    no_refs = True
```

jleedev:
For what it's worth:

- Python needs 3.3 GiB just to deserialize that visa.geojson into memory. Jq needs 4.2 GiB. Gpq converts it to geoparquet in 1.0 GiB.
- A small rust program using the apache parquet crate can compute the union of all 2,000 intermediate parquet files in about a minute using a few hundred MiB of memory. (It has a special type to represent "a column of _n_ nulls", which I imagine is absolutely essential for conjuring up sparse columnar data.)

So Geopandas is probably fine for smaller files, perhaps hitting some limits earlier than other tools given the overhead of Python, but it shouldn't be used to combine the output. Pyarrow directly ought to be reasonable.

PR #8094, opened 2024-03-29
Add Shoprite (ZA)

davidhicks:
Fixes #8069

PR #8813, opened 2024-07-11
Trying to do a better job installing playwright dependencies

iandees:


davidhicks (commented):
Another suggestion: `pipenv run playwright install firefox` and if dependencies are needed, and actually work with the runner operating system: `sudo pipenv run playwright install-deps firefox` ?

Looks like perhaps these binary browser builds are against dynamic libraries that can't be installed on the runner operating system?

iandees:
The problem is that the CodeBuild configuration uses an Amazon Linux image that uses an older glibc. The build fails with:

```
/tmp/opt/python/site-packages/playwright/driver/node: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by /tmp/opt/python/site-packages/playwright/driver/node)
/tmp/opt/python/site-packages/playwright/driver/node: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by /tmp/opt/python/site-packages/playwright/driver/node)
/tmp/opt/python/site-packages/playwright/driver/node: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by /tmp/opt/python/site-packages/playwright/driver/node)
/tmp/opt/python/site-packages/playwright/driver/node: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by /tmp/opt/python/site-packages/playwright/driver/node)
```

This StackOverflow mentions installing a newer glibc (which we can't do because we don't get root on these lambda runners), install an older node version (might be an option if the `playwright install-deps` will use the existing installed older node), and install node from source (which is where I think I was heading next).

davidhicks:
Can the CodeBuild image be changed from Amazon Linux to Ubuntu 22.04 per https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html

Not because I recommend Ubuntu, but because Amazon only provides Ubuntu docker images if one doesn't want to use Amazon Linux, and Ubuntu 22.04 comes with glibc-2.35 and Playwright browser builds fall back to Ubuntu builds of Firefox per "BEWARE: your OS is not officially supported by Playwright; installing dependencies for ubuntu20.04-x64 as a fallback." so I assume Playwright developers like Ubuntu. Probably easier than finding another Docker image not supported by Amazon CodeBuild out of the box?

davidhicks:
> Can the CodeBuild image be changed from Amazon Linux to Ubuntu 22.04 per https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-available.html
> 
> Not because I recommend Ubuntu, but because Amazon only provides Ubuntu docker images if one doesn't want to use Amazon Linux, and Ubuntu 22.04 comes with glibc-2.35 and Playwright browser builds fall back to Ubuntu builds of Firefox per "BEWARE: your OS is not officially supported by Playwright; installing dependencies for ubuntu20.04-x64 as a fallback." so I assume Playwright developers like Ubuntu. Probably easier than finding another Docker image not supported by Amazon CodeBuild out of the box?

Cancel this, I didn't read the part where you said AWS Lambda is being used, hence Ubuntu docker images are out of the question and we're stuck with Amazon Linux :(

iandees:
It looks like I can specify a custom image from a local ECR, so I specified the alltheplaces image that gets used for the weekly run. But the Lambda runner uses a custom user rather than the default in the image, so everything gets "Permission denied".

iandees:
While playing with the Docker image for ATP earlier this morning I noticed that Playwright is already installed as part of building the image. The script doesn't need to do it at CI time.

iandees:
Closing this in favor of what I'm trying in #8094

PR #8817, opened 2024-07-11
[Visa] Remove broken spider (Fixes #8636)

### LINKED ISSUE #7647: Visa Spider: some transposed coordinates
https://www.alltheplaces.xyz/map/#14.13/-75.62091/45.38981

Unclear if just data error upstream or whole spider.

Recommended fix: bounding box for validity

#5737

### LINKED ISSUE #8636: visa: server now requires geographic radius search of <60mi/<100km and maximum of 100 results per page
The `visa` spider is broken as the server now prevents an unlimited geographic radius search and instead enforces a <60mi or <100km limit. Additionally, the current page size of 1000 is rejected and seemingly only 100 results per page are now allowed to be returned.

To fix this spider it appears we need to add a `earth_centroids_iseadgg_100km_radius.csv` searchable points file (or as close to 100km as practical) and then work through returned results 100 per page.

At the very minimum, 100 per page for all 1.5m approx features will require 15000 queries. ISEA DGG does result necessarily in having overlapping areas be queried, but this overlap is minimised. The spider will have to drop locations duplicated across overlapping areas and the total expected number of queries would be somewhat higher than 15000 queries.

Cj-Malone:
Also fixes #7647


iandees:
We'll miss you, Visa.

PR #8822, opened 2024-07-11
[dorevitch_pathology_au] Fix Spider by new url (462 items)

ckpiper:
I was looking at the website and looks like there are many other brands (which we have spiders for) under Healius name.  Such as Abbott, Laverty, QML, TML, and Western_diagnostic_pathology. What is the standard for these. Should we merge all these together under just a HealiusPathologyAU spider?


davidhicks:
Thanks!

I prefer spiders being split usually because `allowed_domains` are different and by splitting, spiders can be executed in parallel. But in this case that parallel execution doesn't apply. So I just like splitting because it generates a separate .geojson output for each brand, and the `item_attributes` property is simple and there is no need for a dictionary of brand information or as much if/else for brand-specific data that may be able to be captured.

PR #8761, opened 2024-07-03
[kate_spade_au_nz] Fix Spider (10 locations)

jhutton1:
<details><summary>stats</summary>

```python
{'atp/brand/Kate Spade New York': 10,
 'atp/brand_wikidata/Q6375797': 10,
 'atp/category/shop/clothes': 10,
 'atp/cdn/cloudflare/response_count': 1,
 'atp/cdn/cloudflare/response_status_count/200': 1,
 'atp/field/email/missing': 10,
 'atp/field/image/missing': 10,
 'atp/field/operator/missing': 10,
 'atp/field/operator_wikidata/missing': 10,
 'atp/field/twitter/missing': 10,
 'atp/field/website/missing': 10,
 'atp/nsi/perfect_match': 10,
 'downloader/request_bytes': 488,
 'downloader/request_count': 1,
 'downloader/request_method_count/POST': 1,
 'downloader/response_bytes': 3457,
 'elapsed_time_seconds': 1.558215,
 'finish_reason': 'finished',
 'httpcompression/response_bytes': 20012,
 'httpcompression/response_count': 1,
 'item_scraped_count': 10,
 'log_count/DEBUG': 22,
 'log_count/INFO': 9,
 'response_received_count': 1,
 'scheduler/dequeued': 1,
 'scheduler/dequeued/memory': 1,
 'scheduler/enqueued': 1,
 'scheduler/enqueued/memory': 1,
 'start_time': datetime.datetime(2024, 7, 3, 14, 11, 48, 312257, tzinfo=datetime.timezone.utc)}
```
</details>

PR #8823, opened 2024-07-11
Fix spiders by disobeying the robotstxt

ckpiper:
I tested 5 random spiders that used StockInStoreSpider and they all did not work before and do know. I assume this will propogate to the other 35 that use it.


davidhicks (approved):
Identified as well in #8761 but thanks for fixing with this PR :)

I got the impression that some of these brands using StockInStore have implemented other store finder methods. Generally StockInStore is used for "Find available stock" widgets within online stores, and brands then use another product for "Find your closest store" pages.

Issue #8819, opened 2024-07-11
Inconsistencies in Spider Naming Conventions: File Names vs. Class Names


Ebadm:
## Issue Description

During a recent audit of the spider files in the AllThePlaces repository, we've identified several inconsistencies between the names of the Python files and the name attribute defined within the spider classes. According to the project's naming conventions, these names should be similar or identical to maintain consistency and ease of maintenance.

### Current Status
Our audit revealed:

Total Python files: 2,573
Files with matching names: 2,484
Files with non-matching names: 89

This means that approximately 3.46% of the spider files have inconsistent naming.

## Detailed Findings
Here's a list of the mismatched spiders, showing the file name and the corresponding spider name:

99_bikes.py: File name: 99_bikes, Spider name: 99_bikes_au
albert_and_walter.py: File name: albert_and_walter, Spider name: albert_walter
allegro_one_box.py: File name: allegro_one_box, Spider name: allegro_one_box_pl
autonation.py: File name: autonation, Spider name: auto_nation
bangkok_bank_th.py: File name: bangkok_bank_th, Spider name: bangkok_bank
bank_of_cyprus_cp.py: File name: bank_of_cyprus_cp, Spider name: bankofcyprus_cy
baptist_health_arkansas_us.py: File name: baptist_health_arkansas_us, Spider name: bha_us
barnesandnoble_college.py: File name: barnesandnoble_college, Spider name: barnesnoblecollege
bass_pro_shops.py: File name: bass_pro_shops, Spider name: bassproshops
baylorscottwhitehealth.py: File name: baylorscottwhitehealth, Spider name: baylorscottwhite
bc_pizza.py: File name: bc_pizza, Spider name: bcpizza
bjrestaurants.py: File name: bjrestaurants, Spider name: bj_restaurant
bobs_burgers.py: File name: bobs_burgers, Spider name: bobs
bravoitalian_us.py: File name: bravoitalian_us, Spider name: bravo_italian_us
bremer_bank_us.py: File name: bremer_bank_us, Spider name: bremer_bank
burger_king.py: File name: burger_king, Spider name: burgerking
burlington.py: File name: burlington, Spider name: burlington_us
california_pizza_kitchen.py: File name: california_pizza_kitchen, Spider name: cpk
canada_post.py: File name: canada_post, Spider name: canadapost
cell_only.py: File name: cell_only, Spider name: cellonly
[...]

PR #8829, opened 2024-07-15
Add pre-commit check for spider naming consistency

iandees:
For #8819

This adds a pre-commit hook that checks naming consistency with spider file names, class names, and spider names.


iandees:
This is pretty noisy. I'm not 100% sure I want to do this (and fix all the issues) right now.

matkoniecz:
https://results.pre-commit.ci/run/github/61166935/1721015102.P1zuVnt0Qjuy3qlsdeiTvg - if anyone wants to fix some...

I assume that PRs are welcome?

iandees:
I think the right way to tackle this is to make changes on this branch so that when merged the build is still clear of issues. If you'd like to make a pull request, please set the target to `check-naming` instead of `master`.

iandees:
Ok, this is ready for input. There are a LOT of spider name, class name, and file name changes in here. I like that we end up with a consistent naming strategy going forward, even if it means consumers that currently check spider names might get confused in the short term.

@matkoniecz I imagine the work you're doing on OSM POI merging might be affected here as you have manually checked viability for many spiders by name. Those names might change with this PR.

@Cj-Malone I'm curious how this might affect your workflows.

I'd like to come to a decision relatively quickly so that I don't have to keep updating this branch as new spider changes come in elsewhere.

Cj-Malone:
> @Cj-Malone I'm curious how this might affect your workflows.

Class names change literally nothing for us. Spider names change the ids, this is a little bit more of a headache, but we can live with it. It's worse when ids and/or category and/or brand changes at the same time.

iandees:
I can go back through and undo spider name changes, but the new names make a lot more sense in most cases. I think the question is how annoying it is. Is this change causing a huge amount of work/rematching?

matkoniecz:
> I imagine the work you're doing on OSM POI merging might be affected here as you have manually checked viability for many spiders by name. Those names might change with this PR.

Should not cause too much fiddling to me, definitely consistency moving forward is worth more.

Cj-Malone:
> I can go back through and undo spider name changes, but the new names make a lot more sense in most cases. I think the question is how annoying it is. Is this change causing a huge amount of work/rematching?

Rip the bandaid off. We'll deal with it.

ci/check_spider_naming_consistency.py
```diff
@@ -0,0 +1,263 @@
...
+    "SA",
+    "SE",
+    "SG",
+    "SI",
+    "SK",
+    "TH",
+    "TR",
+    "TW",
+    "UA",
+    "UK",
```
Cj-Malone, line 90: UK isn't a real country code. I guess this comes from "HSBC UK". Slight preference for `HsbcUkGBSpider`, since "HSBC UK" is the brand for GB.

iandees: Thanks, I thought it was for something else but that was the only one that broke when I removed it so away it goes!

ci/check_spider_naming_consistency.py
```diff
@@ -0,0 +1,262 @@
+import ast
+import os
+import re
+import sys
+from concurrent.futures import ThreadPoolExecutor
+from pathlib import Path
+from typing import List, Tuple
+
+COUNTRYCODE_COMPONENTS = {
```
davidhicks, line 9: Can we just use `pycountry` ISO 3166-1 alpha-2 codes instead?

iandees: Yea I thought about that but wanted to be pretty restrictive about what got to be capitalized at the end of a spider. I guess that's not a great reason...

PR #8831, opened 2024-07-15
handle spaces if any are present

matkoniecz:
no idea are there any there now and on which stage of pipeline numbers are formatted...

But in the worst case it provides a bit of futureproofing


iandees:
The telephone number formatting happens in the `PhoneCleanUpPipeline` class _after_ the spider yields a scraped item. In the spider code, you would want to check for what's in the content of the page.

PR #8838, opened 2024-07-15
[Family Dollar US] Improve name

matkoniecz:
at least I assume so, based also on OSM (but I never ever visited USA!)


locations/spiders/family_dollar_us.py
```diff
@@ -47,4 +47,5 @@ def parse_item(self, item, location):
             hours_string = hours_string + f" {day}: {open_time} - {close_time}"
         item["opening_hours"] = OpeningHours()
         item["opening_hours"].add_ranges_from_string(hours_string)
+        item["branch"] = item.pop("name", None)
```
Cj-Malone, line 50: Don't destroy `branch`.

```suggestion
        item["extras"]["ref:branch"] = item.pop("name", "").removeprefix("FAMILY DOLLAR #")
```

PR #8841, opened 2024-07-16
[Cinemark] Add download delay to prevent HTTP 429 errors

impiaaa:
Also limit parsing to theater pages, and don't include logo image


iandees:
I think it's using IP range or SSL fingerprinting to block connections.

PR #8849, opened 2024-07-16
[Big Table Group GB] Add and fix spiders

TheUKHighStreet:
Updating spiders for https://www.bigtablegroup.com
Please let me know if I'm doing something wrong as I've only just started using github.


Cj-Malone:
I'll fix the website

TheUKHighStreet:
Thanks - fast learning curve as I'm used to programming in perl rather than python.
I think I need that fix for the website for Cafe Rouge as well.

Cj-Malone:
I think this is better

TheUKHighStreet:
Theres also Banana Tree - can you see that or do I need to do another pull request?

TheUKHighStreet:
Oh - and chiquito - which is an update of an existing spider.

Cj-Malone:
I don't usually do ones that small (Banana Tree and Café Rouge), but I will since it's almost no code, and to welcome you.

I'll sort out Chiquito too.

TheUKHighStreet:
Thanks :)

PR #8854, opened 2024-07-16
[American Golf GB] Create american_golf_gb.py

TheUKHighStreet:
Deleted previous branch. Trying again.


TheUKHighStreet:
Problem here seems to be that scrapy is blocked by robots.txt so I'm deleting this.

CloCkWeRX:
There is a ROBOTS_OBEY or similar flag, btw

TheUKHighStreet:
Thanks - I had found that but wasn't sure about the ethics of using it on a public project.

PR #8869, opened 2024-07-17
[Hays Travel GB] add spider

TheUKHighStreet:


TheUKHighStreet:
I think I've fixed the mess of the opening hours in multiple formats. However, I'm not sure what I'm supposed to do with the closed days - can they just be left out?

I've added the links to the branches - but some of them don't work and I can't see how they built the links from the available data. For example Hull Northpoint. I'm not sure whether the best thing is to miss out all the links or leave some (may be only one) broken?

iandees:
It looks like the content of the "Store Locator" side pane get loaded with a separate API call to https://branches.haystravel.co.uk/api/branches/getbranchlocatorcontent. That comes back with 404 for me, but maybe they're checking for referer or something.

If we can't find the website, we should remove the website from the output items.

TheUKHighStreet:
Right - thanks for that. I have a database of spiders that I've been using and thought I'd gradually add them here. But I clearly need to check more thoroughly that the json files I've been using are still being used on the sites. This is the url to the api page that seems to work: https://www.haystravel.co.uk/umbraco/api/branches/getbranchlocatorcontent
I'll move this back to a draft.

davidhicks:
Can you use https://www.haystravel.co.uk/umbraco/api/branches/getbranchlocatorcontent instead? Seems to work fine and is a single HTML response with all locations including name, coordinates, address, phone number, website URL and opening hours.

TheUKHighStreet:
I think this is fixed now. The opening hours aren't available on that page though.

PR #8882, opened 2024-07-19
New storefinder: Algolia; and new spider BlueBottleCoffee that uses it (85 locations)

impiaaa:
Also migrate BaptistHealthArkansasUS, DunelmGB, IowaUniversityUS, Jbhifi, MarinemaxUS, OrangePL, and Warhammer to the new storefinder. In doing so, IowaUniversityUS gained 20 new locations; Jbhifi, 46; OrangePL, 227; and Warhammer, 19.

Side note: The storefinder class uses the "Search multiple indices" API endpoint, which is a bit overkill since we're not searching, and all spiders only use one index. But not all sites give access to the "browse" endpoint, and this one works with all of them.


iandees:
Love it when we find a new storefinder. Thanks for taking the work to add it!

PR #8886, opened 2024-07-19
[paczkomat_inpost_pl] cleanup POI

ireun:
1. Some POI have `streetname` starting with `ul.` (ulica - street) those should be removed
2. Remove `addr:housenumber=n/n` (no number?)


matkoniecz:
This street name transformation could be potentially applicable also in other Polish spiders, not sure is there some good way to make it reusable

matkoniecz:
(and thanks for reviewing this data!)

PR #8889, opened 2024-07-19
[paczkomat_inpost_pl] Cleanup

ireun:
1. Remove `-go` ex. `1-go Maja` should be `1 Maja`
2. Replace `oś.` -> `Osiedle`


matkoniecz:
> Set skip_auto_cc_domain = True as addr:country is not used for POI in Poland

It is still useful for running location checking, I would not remove it.

PR #8893, opened 2024-07-19
[DRAFT] Add scraper for Ben & Jerry's ice cream

riordan:
Scraper for Ben & Jerry's Ice Cream globally. This is my first contribution, so please excuse the job that won't run to completion & over the top attempts to parse opening hours.

I'm coming to folks looking for feedback on better ways to handle hours. What can I do to take better advantage of your normalizing code and handle a very wide variety of variation in global opening hours?


davidhicks:
It looks like this spider will need to have a Where2GetIt `api_filter` specified to only return branded Ben & Jerry's locations, not just retailers/stockists that sell their product.

davidhicks:
If you copy an example of the raw opening hours data from the Where2GetIt response I can let you know the easiest way to extract the hours information. As far as I'm aware Where2GetIt doesn't specify an opening hours format and each brand has their own format(s) they use.

PR #8897, opened 2024-07-20
does not qualify even as a branch code

matkoniecz:


iandees:
@matkoniecz can you please use a more descriptive commit message/pull request title? It's easy to read these as condescending or critical of the previous author of the code. Something like "Remove name from `empik_pl`" or something would be more helpful.

matkoniecz:
I will try to do this

PR #8913, opened 2024-07-23
[NordeaSE] Add website

karen-kerr:


iandees:
I think I would prefer to only have location-specific websites in the data. Do you see a way to get that?

Cj-Malone:
It doesn't look like they have urls per location.

Issue #7903, opened 2024-03-15
Fibank Bulgaria


Dimitar5555:
### Brand name

Fibank

### Wikidata ID

Q3367065

### Store finder url(s)

https://www.fibank.bg/bg/branch_network_xhr?method=get_locations with `X-Requested-With=XMLHttpRequest`. There is extra data for ATMs (atm_filter_deposit, atm_filter_contactless and atm_filter_blind).

PR #8920, opened 2024-07-23
Add Fibank spider

Dimitar5555:
Resolves https://github.com/alltheplaces/alltheplaces/issues/7903

It looks like it's not caching the opening hours and the phone correctly. Also I'm unsure on how to tag "Blind access" for the ATMs, which is why it's commented.


Dimitar5555:
It's working as expected now. I've skipped the complex opening hours, will fix them in a subsequent PR.

PR #8940, opened 2024-07-25
remove name from grycan_pl

matkoniecz:
example value: 'Lodziarnia Firmowa Gdańsk'


iandees:
I know it's not related to your change, but I'd like to try to resolve the exceptions that popped up here.

matkoniecz:
no problem!

PR #8949, opened 2024-07-26
remove name from hypovereinsbank_de

matkoniecz:
example: HypoVereinsbank Günzburg


matkoniecz:
maybe, maybe it can be `branch` but it would require in-place verification

PR #8951, opened 2024-07-26
coop_food_gb - remove name

matkoniecz:


matkoniecz:

davidhicks:
There is a branch at 189 Monks Rd Lincoln GB which has the following visible external signage[1]:
- coop (as a two-line logo, "co" on the first line, "op" beneath)
- "Lincolnshire coop" ("Lincolnshire" is the only word written above the "coop" two-line logo)
- "Food Store"
- "Open every day, 7am - 10pm"
- "Post Office"

This branch is extracted currently as:
```
addr:city=Lincoln
addr:country=GB
addr:postcode=LN2 5JP
addr:street_address=189 Monks Road
brand=Co-op Food
brand:wikidata=[Q3277439](https://www.wikidata.org/wiki/Q3277439)
location_type=Food
name=Monks Road
nsi_id=[thecooperativegroup-ac8d0d](https://nsi.guide/?id=thecooperativegroup-ac8d0d)
opening_hours=Mo-Su 07:00-22:00
operator=Lincolnshire Co-op
operator:wikidata=Q117202
phone=+44 1522 568608
ref=/store-finder/LN2-5JP/189-monks-road
shop=convenience
website=https://www.coop.co.uk/store-finder/LN2-5JP/189-monks-road
```

The website for this branch location has a title of "Monks Road".

NSI's brand matches Wikidata and Wikipedia and is "Co-op Food".

Is the OSM convention then that `name=coop` (externally visible labeling), `brand=Co-op Food` and `branch=Monks Road` (as designated in brand data/website for that branch location/on a receipt from the branch/etc)?

Given the mostly poor to non-existent externally visible labeling of businesses around the world (perhaps just a graphical logo and nothing else), I assume then that `name` would be rarely used?

[1] https://www.google.com.au/maps/place/189+Monks+Rd,+Lincoln+LN2+5JP,+UK/@53.2308557,-0.5239098,3a,62.2y,206.19h,101.89t/data=!3m7!1e1!3m5!1swbDJH9qmtlA9ZqgoEEv85Q!2e0!6shttps:%2F%2Fstreetviewpixels-pa.googleapis.com%2Fv1%2Fthumbnail%3Fpanoid%3DwbDJH9qmtlA9ZqgoEEv85Q%26cb_client%3Dmaps_sv.share%26w%3D900%26h%3D600%26yaw%3D206.18520606026095%26pitch%3D-11.889468708229273%26thumbfov%3D90!7i16384!8i8192!4m6!3m5!1s0x48785b4701803bb7:0x98ba4466fd30435b!8m2!3d53.2306708!4d-0.5239489!16s%2Fg%2F11c5k5qtx4?coh=205410&entry=ttu

matkoniecz:
> Given the mostly poor to non-existent externally visible labeling of businesses around the world (perhaps just a graphical logo and nothing else), I assume then that name would be rarely used. 

`name` (in OSM at least) is for "commonly used name" which is not always signed, for example many mountain names are not having signs but have names. So if shop has just logo it does not mean it has no name or no `name`

> There is a branch at 189 Monks Rd Lincoln GB which has the following visible external signage[1]:

I would say that it is likely `name=Lincolnshire Coop` or `name=Lincolnshire Co-op` but I am not having local knowledge at all
It is technically possible that its actual used name is radically different from it and it is named Foobar but it is extremely, extremely unlikely
Similarly `name=Monks Road` is also extremely unlikely to be real shop name (fails to identify this specific shop among other on the same street), though it is used as web listing name (as it identifies this shop from other shops run by Coop and gives rough location info, which is used at web listing for specific brand - while giving actual shop name would not be useful as often these are the same across all or nearly all shops for a given brand)

checking OSM, it appears to be not mapped in OSM, see https://www.openstreetmap.org/#map=18/53.23092/-0.52902

> If one were developing a map renderer and deciding how to label businesses, the name key is seemingly far less useful than concatenating the brand and branch keys.

in turn, I am using `brand` for `name` right now and throwing away `name` and `branch` as too many are simply wrong

these PRs are attempt to change this as for some spiders `name` data seems actually real

matkoniecz:
> Given the mostly poor to non-existent externally visible labeling of businesses around the world

This may heavily depends on local culture, but for example in Poland sign with name (or at least shop type where it can be tricky is it name or is shop having no name) is quite common. The same applies across Europe at list, at least parts which I visited.

matkoniecz:
https://github.com/alltheplaces/alltheplaces/blob/master/DATA_FORMAT.md has

> The name of the feature. Ideally the fascia, however this is often a combination of the brand and the branch.

which I interpreted as matching OSM definition, though maybe I missed something here

davidhicks:
> > Given the mostly poor to non-existent externally visible labeling of businesses around the world
> 
> This may heavily depends on local culture, but for example in Poland sign with name (or at least shop type where it can be tricky is it name or is shop having no name) is quite common. The same applies across Europe at list, at least parts which I visited.

I just had a look around on Google Street View in Polish cities and seemingly all the stores/fuel stations/fast food restaurants I looked at had a visible brand name (e.g. BP, KFC, Jysk, RTV Euro AGD, Rossmann, Meble, Netto, Apteka, KIK, Tesco, Kaufland, McDonald's, etc) but no indication of individual branch name. Thus we'd either set `name` and `brand` to be the same value in most cases, or not set `name` at all and just rely on `brand`, as well as `branch` (if we have other information not visible on signs such as title of a store on the brand's website or from their API)?

matkoniecz:
> Thus we'd either set name and brand to be the same value in most cases

I would expect this to hold for Poland, except rare cases chains of brands with many shops have no unique names for specific shops (and even `branch` is rarely being done or applicable)

PR #8976, opened 2024-07-27
b_and_m - skip call centers in UK

matkoniecz:


matkoniecz:

iandees:
Shouldn't the phone numbers be removed, not name?

matkoniecz:
oh dear

PR #8977, opened 2024-07-27
next - skip call centers in UK

matkoniecz:


matkoniecz:

matkoniecz:
no idea why `!= None` is banned, AFAIK it would matter only if some class would implement buggy `__eq__()`

PR #8980, opened 2024-07-27
remove phone instead of name

matkoniecz:


matkoniecz:
and sorry for that bug!

iandees:
No prob!

PR #8987, opened 2024-07-28
leroy_merlin_pl - remove names

matkoniecz:


matkoniecz:
for reason that is not obvious for me this change has failed to fix names

locations/spiders/leroy_merlin_pl.py
```diff
@@ -6,3 +6,7 @@ class LeroyMerlinPLSpider(WoosmapSpider):
     item_attributes = {"brand": "Leroy Merlin", "brand_wikidata": "Q889624"}
     key = "woos-936f7dae-38cd-3785-8a3c-153b43bd33ca"
     origin = "https://www.leroymerlin.pl"
+
+    def post_process_item(self, item, response, ld_data, **kwargs):
```
CloCkWeRX, line 10: Woosmap doesn't have a post_process_item, only parse_item

https://github.com/alltheplaces/alltheplaces/blob/4fbd439f186dd4d00483268124e6367b94826e52/locations/storefinders/woosmap.py#L60

CloCkWeRX, line 10:
```suggestion
    def parse_item(self, item, feature, **kwargs):
```

PR #9048, opened 2024-07-29
Handle 24/7 opening hours for econt_bg.py 

Dimitar5555:


locations/spiders/econt_bg.py
```diff
@@ -33,22 +33,25 @@ def unix_timestamp_to_local_time(timezone, timestamp):
...
-                saturday_start_time = unix_timestamp_to_local_time(timezone, location["halfDayBusinessHoursFrom"])
-                saturday_end_time = unix_timestamp_to_local_time(timezone, location["halfDayBusinessHoursTo"])
-                item["opening_hours"].add_range("Sa", saturday_start_time, saturday_end_time)
-
-            if location["sundayBusinessHoursFrom"] is not None:
-                sunday_start_time = unix_timestamp_to_local_time(timezone, location["sundayBusinessHoursFrom"])
-                sunday_end_time = unix_timestamp_to_local_time(timezone, location["sundayBusinessHoursTo"])
-                item["opening_hours"].add_range("Su", sunday_start_time, sunday_end_time)
+            if "24/7" in item["name"]:
+                item["opening_hours"] = "24/7"
```
davidhicks, line 39: Can you use `OpeningHours.add_days_range(DAYS, "00:00", "23:59")` instead so that the field is an `OpeningHours` object?

locations/spiders/econt_bg.py
```diff
@@ -33,22 +33,25 @@ def unix_timestamp_to_local_time(timezone, timestamp):
...
+            else:
+                item["opening_hours"] = OpeningHours()
+                weekday_start_time = unix_timestamp_to_local_time(timezone, location["normalBusinessHoursFrom"])
+                weekday_end_time = unix_timestamp_to_local_time(timezone, location["normalBusinessHoursTo"])
+                item["opening_hours"].add_days_range(day_range("Mo", "Fr"), weekday_start_time, weekday_end_time)
```
davidhicks, line 44: `day_range("Mo", "Fr")` can be replaced with constant `OpeningHours.DAYS_WEEKDAY`.

davidhicks (changes requested):

locations/spiders/econt_bg.py
Dimitar5555: Thanks, I've updated it.

locations/spiders/econt_bg.py
Dimitar5555: It doesn't convert it to 24/7. Is this the intended behavior?

image

iandees: Yes, back in #6537 we decided we should be more explicit about opening hours and not use `24/7`.

davidhicks (approved):

PR #9062, opened 2024-07-29
fix phone removal condition

matkoniecz:
sorry for inversion


matkoniecz:
Not sure how I missed it.

I was checking output and was happy that phone numbers are gone, without noticing that ones that got kicked out were local ones - not call center ones.

iandees:
Build failure seems to be a problem with AWS, not your code.

PR #9068, opened 2024-07-29
remove generic image

matkoniecz:
https://github.com/alltheplaces/alltheplaces/blob/master/DATA_FORMAT.md

>image
>A URL of an image for the venue. We try to make this specific to the venue and not generic for the brand that is operating the venue.

### DIFF
--- a/locations/spiders/nationwide_gb.py
+++ b/locations/spiders/nationwide_gb.py
@@ -22,7 +22,11 @@ class NationwideGBSpider(CrawlSpider, StructuredDataSpider):
         if "phone" in item and item["phone"] is not None and item["phone"].replace(" ", "").startswith("+443"):
             item.pop("phone", None)
 
-        if "image" in item and item["image"] is not None and "https://www.bhf.org.uk/-/media/images/social/facebook-share-image_v5-1.png" in item["image"]:
+        if (
+            "image" in item
+            and item["image"] is not None
+            and "https://www.bhf.org.uk/-/media/images/social/facebook-share-image_v5-1.png" in item["image"]
+        ):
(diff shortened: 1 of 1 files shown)


locations/spiders/nationwide_gb.py
```diff
@@ -22,4 +22,11 @@ def post_process_item(self, item, response, ld_data, **kwargs):
         if "phone" in item and item["phone"] is not None and item["phone"].replace(" ", "").startswith("+443"):
             item.pop("phone", None)
 
+        if (
+            "image" in item
+            and item["image"] is not None
+            and "https://www.bhf.org.uk/-/media/images/social/facebook-share-image_v5-1.png" in item["image"]
+        ):
```
iandees, line 29: Can we simplify this to

```suggestion
        if item.get("image") == "https://www.bhf.org.uk/-/media/images/social/facebook-share-image_v5-1.png":
```

matkoniecz: no, as it has bunch of FB garbage appended to it that looks unstable (tracking parameter?)

matkoniecz: https://www.bhf.org.uk/-/media/images/social/facebook-share-image_v5-1.png?rev=63f70ab8b3b148edaef074404a4874c1

iandees: How about `"social/facebook-share-image_v5-1.png" in item.get("image")` then?

matkoniecz: should work, unlikely to have higher false positive ratio

PR #9074, opened 2024-07-30
note Mo-Su closed

matkoniecz:
This `opening_hours=Mo-Su closed` can easily confuse people, especially as ATP is not using lifecycle prefixes in such cases (it is still `amenity=fast_food` etc) so warning at least in docs would be nice


matkoniecz:
> POIs that are permanently closed should be omitted by spiders.

Oh, good news, I though that such objects are actually included. Though in such case why we have `end_date=yes` ones?

> `end_date=yes` is applied when given location is closed at unknown date and can be assumed to not operate right now, `end_date` may also have values in year-month-day format, including future dates for planned closures.

Or is it also for objects expected to come back? I understood that it is also for permanently closed ones.

davidhicks:
> > POIs that are permanently closed should be omitted by spiders.
> 
> Oh, good news, I though that such objects are actually included. Though in such case why we have `end_date=yes` ones?
> 
> > `end_date=yes` is applied when given location is closed at unknown date and can be assumed to not operate right now, `end_date` may also have values in year-month-day format, including future dates for planned closures.
> 
> Or is it also for objects expected to come back? I understood that it is also for permanently closed ones.

You're right about `end_date` because #6562 introduces a need (from @Cj-Malone) to keep track of permanently closed locations. This allows some ATP consumers to gradually remove permanently closed locations from other data they hold. Simply deleting/omitting the POI could be interpreted by consumers as ATP failing to complete a full run, as consumers without `end_date` have no idea why the POI used to exist but currently doesn't.

Suggest rewording as:
_POIs that are noted by source data to be permanently closed are returned with the `end_date` field populated. If the POI has been deleted entirely in the source data, ATP will stop returning the former POI._

matkoniecz:
I have done it a bit differently to keep `end_date` docs in `end_date` section. Let me know is it working well or should I use suggested version directly.

davidhicks (approved):

PR #9075, opened 2024-07-30
[Core] Track source urls

Cj-Malone:
This does 2 things, adds the following to counters,

```python
 'atp/item_scraped_host_count/www.trusselltrust.org': 1470,
```

And the following to the POI.

```
@source=https://www.trusselltrust.org/get-help/find-a-foodbank/foodbank-search/?foodbank_s=all&callback=%20
```

`@source` is not guaranteed to be viewable in a browser, it may need a POST request, or special headers. It should just be for us, not anyone else.


Cj-Malone:
I'm now thinking we could rename `source:url` to `@source` to stop someone getting confused.

davidhicks (approved):

locations/middlewares/track_sources.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(self, crawler):
+        self.crawler = crawler
+
+    def _process_item(self, response, item, spider):
+        if isinstance(item, Item):
+            if "@source" not in item["extras"]:
+                item["extras"]["@source"] = response.url
+
+            self.crawler.stats.inc_value(
+                "atp/item_scraped_host_count/{}".format(urlparse(item["extras"]["@source"]).netloc)
```
Cj-Malone, line 25: Technically urlparse isn't safe, I should wrap it.

locations/middlewares/track_sources.py
```diff
@@ -0,0 +1,36 @@
...
+    @classmethod
+    def from_crawler(cls, crawler):
+        return cls(crawler)
+
+    def __init__(self, crawler):
+        self.crawler = crawler
+
+    def _process_item(self, response, item, spider):
+        if isinstance(item, Item):
+            if "@source" not in item["extras"]:
```
Cj-Malone, line 21: Probably not .get

Cj-Malone:
Not currently in ATP, but it could also be an ftp or file uri. However no strong opinion, if you want to change it, feel free.

Cj-Malone:
Thank you

locations/middlewares/track_sources.py
```diff
@@ -0,0 +1,36 @@
...
+from scrapy.item import Item
+
+
+class TrackSourcesMiddleware:
+    """
+    Track the URLs used for each generated item,
+    and overall stats per hostname.
+    """
+
+    def _process_item(self, response, item: Item, spider: Spider):
```
CloCkWeRX, line 13: Do we want to assert this is a Feature, which is our extension to an Item?
Otherwise there's a slight risk Item won't have the key "extras".

Found via https://github.com/alltheplaces/alltheplaces/pull/9253#issuecomment-2267395207

Cj-Malone: Other pipelines require extras, at this stage if a spider isn't outputting it it's an issue with the spider. But this might now be the first an item goes through?

On Item vs Feature, Feature is an Item.  All our pipelines require them even though scrapy can do dicts and other things. Minor preference for Item incase we add other types in future but no strong preference.

PR #9077, opened 2024-07-30
[OpticalCenter] Fix timeout

karen-kerr:


iandees:
Usually when there are timeouts this long its because they are using SSL fingerprinting to block non-browser connections at the CDN. You could try Playwright or add `requires_proxy`

Issue #8519, opened 2024-05-28
Suzuki Marine (shop/boat_parts, shop/boat_repair) (AU) (79 locations)


davidhicks:
### Brand name

Suzuki

### Wikidata ID

Q181642

### Store finder url(s)

https://www.suzukimarine.com.au/find-dealers/dealersByState?state=all

PR #9239, opened 2024-08-03
Suzuki marine au

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/8519

- [x] Opening Hours
- [x] Extra categories?

### DIFF
--- a/locations/categories.py
+++ b/locations/categories.py
@@ -61,6 +61,7 @@ class Categories(Enum):
     SHOP_BICYCLE = {"shop": "bicycle"}
     SHOP_BOOKMAKER = {"shop": "bookmaker"}
     SHOP_BOOKS = {"shop": "books"}
+    SHOP_BOAT_REPAIR = {"shop": "boat_repair"}
     SHOP_BUTCHER = {"shop": "butcher"}
     SHOP_CAMERA = {"shop": "camera"}
     SHOP_CANDLES = {"shop": "candles"}
--- a/locations/categories.py
+++ b/locations/categories.py
@@ -61,7 +61,7 @@ class Categories(Enum):
     SHOP_BICYCLE = {"shop": "bicycle"}
     SHOP_BOOKMAKER = {"shop": "bookmaker"}
     SHOP_BOOKS = {"shop": "books"}
-    SHOP_BOAT_REPAIR = {"shop": "boat_repair"}
+    SHOP_BOAT = {"shop": "boat"}
     SHOP_BUTCHER = {"shop": "butcher"}
     SHOP_CAMERA = {"shop": "camera"}
     SHOP_CANDLES = {"shop": "candles"}
--- a/locations/spiders/suzuki_marine_au.py
+++ b/locations/spiders/suzuki_marine_au.py
@@ -21,17 +21,17 @@ class SuzukiMarineAUSpider(Spider):
 
             if "Services" in feature:
                 if "0" in feature["Services"]:
-                    apply_category(item, Categories.SHOP_BOAT)
+                    apply_category(Categories.SHOP_BOAT, item)
                 if "1" in feature["Services"]:
-                    apply_category(item, {"boat:repair": "yes"})
+                    apply_category({"boat:repair": "yes"}, item)
                 if "2" in feature["Services"]:
-                    apply_category(item, {"boat:parts": "yes"})
+                    apply_category({"boat:parts": "yes"}, item)
 
-            if "ServiceHours" in feature:
+            if "ServiceHours" in feature and feature["ServiceHours"] is not None:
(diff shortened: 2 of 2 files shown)


locations/categories.py
```diff
@@ -61,6 +61,7 @@ class Categories(Enum):
     SHOP_BICYCLE = {"shop": "bicycle"}
     SHOP_BOOKMAKER = {"shop": "bookmaker"}
     SHOP_BOOKS = {"shop": "books"}
+    SHOP_BOAT_REPAIR = {"shop": "boat_repair"}
```
davidhicks, line 64: Perhaps it's better as https://wiki.openstreetmap.org/wiki/Tag:shop%3Dboat and https://wiki.openstreetmap.org/wiki/Key:boat:repair ?

OSM tags for places you can buy boats and boat accessories, and places where you can receive boat repair/service features seem to be fairly vague and overlapping. For example, some of the Suzuki Marine AU locations are nominated as service centres only (`shop=boat_repair`). Other locations appear to sell all sorts of boat accessories, including outboard motors. When buying outboard motors it appears that one turns up with their boat and the shop has a workshop where they'll install the new outboard motor for you. I'm not sure it's designed such that you can rock up with a ute, pick up an outboard motor, and install it yourself/someplace else.

CloCkWeRX, line 64:
```suggestion
    SHOP_BOAT = {"shop": "boat"}
```

locations/spiders/suzuki_marine_au.py
```diff
@@ -0,0 +1,30 @@
...
+from scrapy.http import JsonRequest
+
+from locations.categories import Categories
+from locations.dict_parser import DictParser
+from locations.hours import OpeningHours
+
+
+class SuzukiMarineAUSpider(Spider):
+    name = "suzuki_marine_au"
+    item_attributes = {"brand": "Suzuki", "brand_wikidata": "Q181642", "extras": Categories.SHOP_BOAT_REPAIR.value}
```
CloCkWeRX, line 11:
```suggestion
    item_attributes = {"brand": "Suzuki", "brand_wikidata": "Q181642", "extras": Categories.SHOP_BOAT.value}
```

locations/spiders/suzuki_marine_au.py
```diff
@@ -0,0 +1,30 @@
...
+
+class SuzukiMarineAUSpider(Spider):
+    name = "suzuki_marine_au"
+    item_attributes = {"brand": "Suzuki", "brand_wikidata": "Q181642", "extras": Categories.SHOP_BOAT.value}
+
+    def start_requests(self):
+        yield JsonRequest(url="https://www.suzukimarine.com.au/find-dealers/dealersByState?state=all")
+
+    def parse(self, response, **kwargs):
+        for feature in response.json():
```
davidhicks, line 17: I think we should also probably check `Services` in the source data which can be a value of `0`, `1` or `2` which would determine whether the location is `shop=boat` (you can buy boats or boat parts), `shop=boat_repair` (you can mostly or only service/repair your boat but not buy boats or boat parts) or `shop=boat` + `boat:repair=yes` if you can buy boats and boat parts, and also service/repair boats.

Another approach ATP has used is the `mazda_au` spider for an example of where up to 3 features are extracted per car dealership: a car shop, a car parts shop and a car repair/service workshop. I suppose this same approach could be used for boats. Or spiders like `mazda_au` should have tags changed to be `shop=car` + https://wiki.openstreetmap.org/wiki/Key:service:vehicle:car_repair + https://wiki.openstreetmap.org/wiki/Key:service:vehicle:car_parts so that only one feature is ever extracted for each dealership.

@Cj-Malone @matkoniecz as OSM tagging experts, any preferences for multiple feature extraction or only having a single feature per car/boat dealership that uses a `shop=car` or `shop=boat` primary tag, and then adds part sales and repair/service tags as additional tags as appropriate.

CloCkWeRX: Variations:
"Services":"[\"0\",\"1\",\"2\"]",
"Services":"[\"0\",\"1\"]",
"Services":"[\"0\"]",
"Services":"[\"2\"]",

CloCkWeRX: Well.... it's not immediately obvious which is which.

CloCkWeRX:
Marking as "ready enough", extra need a bit of double checking.

locations/spiders/suzuki_marine_au.py
matkoniecz: I guess it depends. For example if we know that car repair and sale is integrated into one POI then splitting it makes no sense.

But of fuel station and its shop are treated as separate (at least in OSM they are) then splitting makes sense.

I have no on the ground knowledge here.

PR #9240, opened 2024-08-03
Add KFC (ZA)

arrival-spring:
Code largely copied and adapted from KFC AU spider.

### DIFF
--- a/locations/spiders/kfc_za.py
+++ b/locations/spiders/kfc_za.py
@@ -1,0 +1,57 @@
+import scrapy
+from scrapy.http import JsonRequest
+
+from locations.categories import Extras, apply_yes_no
+from locations.dict_parser import DictParser
+from locations.hours import OpeningHours
+from locations.pipelines.address_clean_up import clean_address
+from locations.spiders.kfc_us import KFC_SHARED_ATTRIBUTES
+
+
+class KfcZASpider(scrapy.Spider):
+    name = "kfc_za"
+    item_attributes = KFC_SHARED_ATTRIBUTES
+    start_urls = ["https://orderserv-kfc-eu-olo-api.yum.com/dev/v1/stores/"]
+    tenant_id = "5f89a603781e4df48d31337a6c1252a8"
+    requires_proxy = True  # May require ZA proxy, possibly residential IPs only.
+
+    def start_requests(self):
+        for url in self.start_urls:
+            yield JsonRequest(url=url, headers={"x-tenant-id": self.tenant_id})
+
+    def parse(self, response):
+        for location in response.json():
+            item = DictParser.parse(location)
+            if location["code"] == "1000" or location["code"] == "1001":
+                # Ignore dummy stores used for internal testing/development
+                continue
+            item["ref"] = location["code"]
+            item["street_address"] = clean_address(location["localAddress"][0]["address"]["addressLines"])
+            item["city"] = location["localAddress"][0]["address"]["city"]
+            item["state"] = location["localAddress"][0]["address"]["state"]
+            item["country"] = location["localAddress"][0]["address"]["country"]
+            item["postcode"] = location["localAddress"][0]["address"]["pinCode"]
+            for contact_method in location["contacts"]:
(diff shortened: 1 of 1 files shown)


arrival-spring:
I don't quite understand why this isn't working, could be to do with proxies?

I now realise that the kfc_au spider is not currently working. I should have checked that before copying. The kfc_ca spider also looks like a reasonable target to copy, as it is calling the same API. Any ideas why the AU and CA spiders are so different?

locations/spiders/kfc_za.py
```diff
@@ -0,0 +1,56 @@
...
+            item["country"] = location["localAddress"][0]["address"]["country"]
+            item["postcode"] = location["localAddress"][0]["address"]["pinCode"]
+            for contact_method in location["contacts"]:
+                if contact_method["key"] == "email":
+                    item["email"] = contact_method["value"]
+                elif contact_method["key"] == "phoneNumber":
+                    item["phone"] = contact_method["value"]
+            web_path = item["name"].lower().replace(" ", "-") + "/" + item["postcode"]
+            item["website"] = "https://order.kfc.com.za/restaurants/" + web_path
+            details_url = "https://orderserv-kfc-apac-olo-api.yum.com/dev/v1/stores/details/" + web_path
```
davidhicks, line 40: I'm guessing the domain here should be `orderserv-kfc-eu-olo-api.yum.com`

locations/spiders/kfc_za.py
```diff
@@ -0,0 +1,56 @@
...
+from locations.pipelines.address_clean_up import clean_address
+from locations.spiders.kfc_us import KFC_SHARED_ATTRIBUTES
+
+
+class KfcZASpider(scrapy.Spider):
+    name = "kfc_za"
+    item_attributes = KFC_SHARED_ATTRIBUTES
+    start_urls = ["https://orderserv-kfc-eu-olo-api.yum.com/dev/v1/stores/"]
+    tenant_id = "5f89a603781e4df48d31337a6c1252a8"
+    requires_proxy = True  # May require ZA proxy, possibly residential IPs only.
```
davidhicks, line 15: The error you're getting from CI is due to this `requires_proxy=True` telling the CI process to try and use a Zyte API account that has no balance remaining in it. If you're in ZA and don't need a proxy, can you provide the spider output when you omit `requires_proxy=True`?

davidhicks:
> I don't quite understand why this isn't working, could be to do with proxies?
> 
> I now realise that the kfc_au spider is not currently working. I should have checked that before copying. The kfc_ca spider also looks like a reasonable target to copy, as it is calling the same API. Any ideas why the AU and CA spiders are so different?

`kfc_au` looks better because it extracts more data than `kfc_ca`, such as opening hours.

I suggest that `kfc_au` be used as a base spider with the domain `orderserv-kfc-{some_region_code}-olo-api.yum.com` a parameter that can be modified by subclassed spiders depending on the `{some_region_code}` for the subclassed spider. And also that the `x-tenant-id` can be overridden by subclassed spiders. Then there is not as much code reuse needed between `kfc_au`, `kfc_za`, `kfc_ca` and any other KFC regions that use the Yum API.

davidhicks (changes requested):

locations/spiders/kfc_za.py
arrival-spring: Thank, fixed now

locations/spiders/kfc_za.py
arrival-spring: It works fine from ZA without `requires_proxy=True`

arrival-spring: It works fine from ZA without `requires_proxy=True`

arrival-spring:
It took me a few goes, but I think this is ready for review now.

Are the timeouts something I can do anything about?

iandees:
"timeout" just means that the spider took longer than the 60 seconds we give each spider to do its thing during a PR test. It's not necessarily a bad thing, it just means the spider didn't finish running after ~1 minute.

Issue #9213, opened 2024-08-03
Penningtons (SweetIQSpider)


CloCkWeRX:
### Brand name

Penningtons

Canadian fashion store

### Wikidata ID

Q16956527
https://www.wikidata.org/wiki/Q16956527
https://www.wikidata.org/wiki/Special:EntityData/Q16956527.json

### Store finder url(s)

Official Url(s): http://www.penningtons.com
- [x] I have run the below and if anything in generated, added it.
- [x] I have manually checked for a storefinder
- [x] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q16956527 http://www.penningtons.com`
pipenv run scrapy sf https://locations.penningtons.com/en
```
from locations.storefinders.sweetiq import SweetIQSpider


class PenningtonsCASpider(SweetIQSpider):
    name = "penningtons_ca"
    item_attributes = {
        "brand_wikidata": "Q16956527",
        "brand": "Penningtons",
    }
    allowed_domains = [
        "locations.penningtons.com",
    ]
    start_urls = [
        "https://locations.penningtons.com/en",
    ]
```

PR #9245, opened 2024-08-04
Add Penningtons (CA)

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9213

@davidhicks as a FYI, may want to add the SweetIQ endpoints to the allowed domains in this auto generated spider.


davidhicks:
Thanks you @CloCkWeRX I agree that `api_url_base` needs to be additionally extracted and added to the `allowed_domains` list in the generated spider (ref #6526).

Issue #9241, opened 2024-08-03
maybe throw away empty address fields?


matkoniecz:
small sample of `value.replace("-", "").replace("–", "").strip() == ""` checks:

```
mercedes_benz_group has addr:postcode = -
mercedes_benz_group has addr:state = -
esprit has addr:postcode = --
caltex has addr:street = -
caltex has addr:street = -
caltex has addr:street = -
dr_max has addr:street_address = -
scania has addr:postcode = 
dulux_decorator_centre_gb has addr:city = -
dulux_decorator_centre_gb has addr:city = -
dulux_decorator_centre_gb has addr:city = -
mol has addr:street_address = -
mol has addr:street_address = -
mol has addr:street_address = -
subway_kr has addr:full = -
nespresso has addr:postcode = -
bp has addr:postcode = -
total_energies has addr:postcode = -
```

(very large number of lines, including for spiders mentioned here, were skipped)

davidhicks:
Agreed with resetting to `None` in a pipeline for any of these fields which are blank strings. Although sometimes in ATP, instead of setting fields to `None`, the field gets deleted altogether. I prefer keeping the field though to try and avoid code paths having to check if a field exists before use.

PR #9251, opened 2024-08-04
Address clean up pipeline: Strip small addresses across multiple attributes

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9241 in a very basic way

Later:
* Do we want to assert stripping all \W characters and _ so long as we cooerce the strings into unicode?


matkoniecz:
If for example just `addr:city` or `addr:street_address` has content `-` it will not be dropped.

why not apply `if value.replace("-", "").replace("–", "").strip() == ""` check? It may look oddly specific but actually putting "----" or "-" as null seems quite common across world.

CloCkWeRX:
> If for example just addr:city or addr:street_address has content - it will not be dropped.

Ah, it should be :) min_length < 2

CloCkWeRX:
> why not apply if value.replace("-", "").replace("–", "").strip() == ""

So, this is a really good check for a common human language pattern; and I do agree you have the real world stats to show it.


For targetting specific characters/character sequences as "these 'never' appear in an address", I thought about it and it seemed like targetting via regex all of the non word characters (\W), plus one to two others (like `_` `--` or any repetition of a punctuation character) in a unicode context would be the best approach.

IE "----" might be surprisingly common, but I bet there's a next level down we'll find after we target that exact pattern.

But that's slightly above my regexp golf to go to that level.

matkoniecz:
What about things like `addr:city=Foovar addr:street=Barfoo addr:housenumber=-` ? In such case, despite assembled address being link I would throw away  `addr:housenumber` field.

And you cannot throw away all single-character long housenumber fields :)

iandees:
It doesn't look like this modifies the `addr:housenumber` field, so we don't need to worry there.

Issue #9255, opened 2024-08-04
Waba Grill (Where2GetIt)


CloCkWeRX:
### Brand name

Waba Grill

Fast Food Chain in the United States

### Wikidata ID

Q113163467
https://www.wikidata.org/wiki/Q113163467
https://www.wikidata.org/wiki/Special:EntityData/Q113163467.json

### Store finder url(s)

Official Url(s): https://locations.wabagrill.com/
- [ ] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy links https://locations.wabagrill.com/`
`pipenv run scrapy sitemap --pages https://locations.wabagrill.com/`
`pipenv run scrapy sf --brand-wikidata=Q113163467 https://locations.wabagrill.com/`

```

ubuntu@codespaces-88a392:/workspaces/alltheplaces$ pipenv run scrapy sf https://locations.wabagrill.com
from locations.storefinders.where2getit import Where2GetItSpider


class WaBaGrillUSSpider(Where2GetItSpider):
    name = "waba_grill_us"
    item_attributes = {
        "brand_wikidata": "Q113163467",
        "brand": "WaBa Grill",
    }
    api_endpoint = "https://locations.wabagrill.com/rest/getlist"
    api_key = "7512DE42-E61C-11EE-B773-E61A326043C3"
```
Official Url(s): https://www.wabagrill.com/
- [ ] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy links https://www.wabagrill.com/`
`pipenv run scrapy sitemap --pages https://www.wabagrill.com/`
`pipenv run scrapy sf --brand-wikidata=Q113163467 https://www.wabagrill.com/`
```
```

PR #9309, opened 2024-08-04
Waba Grill

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9255

### DIFF
--- a/locations/spiders/waba_grill_us.py
+++ b/locations/spiders/waba_grill_us.py
@@ -1,7 +1,7 @@
 from locations.storefinders.where2getit import Where2GetItSpider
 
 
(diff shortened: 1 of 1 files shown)


locations/spiders/waba_grill_us.py
```diff
@@ -0,0 +1,11 @@
+from locations.storefinders.where2getit import Where2GetItSpider
+
+
+class WaBaGrillUSSpider(Where2GetItSpider):
```
CloCkWeRX, line 4:
```suggestion
class WabaGrillUSSpider(Where2GetItSpider):
```

PR #9343, opened 2024-08-06
Create 'latest' redirects for spider output

### LINKED ISSUE #3657: Add permalink redirections for latest output of each spider
It's not documented anywhere, but you can fetch the output from a build for a single spider. For example: https://data.alltheplaces.xyz/runs/2022-05-14-13-32-20/output/culvers.geojson

We should also write 302 redirect objects to S3 as part of the weekly build so that "latest/culvers.geojson" will redirect to that URL.

### LINKED ISSUE #4255: "Latest" output endpoint
Currently we have to get the output of a spider on a specific date:

`https://data.alltheplaces.xyz/runs/2022-11-19-13-32-36/output/mcdonalds_fr.geojson`

Could we get something like this to automatically display the latest results?:

`https://data.alltheplaces.xyz/runs/latest/output/mcdonalds_fr.geojson`

Also what are the other available endpoints that could be useful that you use? I remember there was a history.json somewhere. Are they written somewhere? Thanks!

iandees:
Fixes #4255
Fixes #3657 

This adds code to the weekly spider run that writes S3 redirects for each spider's output geojson and the collected zip and parquet outputs.

### DIFF
--- a/ci/run_all_spiders.sh
+++ b/ci/run_all_spiders.sh
@@ -260,6 +260,62 @@ if [ ! $retval -eq 0 ]; then
     exit 1
 fi
 
+# Update the latest/ directory with redirects to the latest run
+touch "${SPIDER_RUN_DIR}/latest_placeholder.txt"
+
+aws s3 cp \
+    --only-show-errors \
+    --website-redirect="${RUN_S3_KEY_PREFIX}/output.zip" \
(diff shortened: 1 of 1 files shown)


Dimitar5555:
Trying to fetch `https://data.alltheplaces.xyz/runs/latest/output/lidl_bg.geojson` results in a 404 error. Is this the correct URL format?

Issue #8050, opened 2024-03-26
Save-on-Foods (JS Blob)


CloCkWeRX:
----
### Brand name

Save-On-Foods

Western Canadian supermarket chain

### Wikidata ID

Q7427974
https://www.wikidata.org/wiki/Q7427974
https://www.wikidata.org/wiki/Special:EntityData/Q7427974.json

### Store finder url(s)

Official Url(s): http://www.saveonfoods.com
- [x] I have run the below and if anything in generated, added it.
- [x] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q7427974 http://www.saveonfoods.com`
```
```

CloCkWeRX:
```
curl 'https://www.saveonfoods.com/sm/pickup/rsid/2210/' \
  -H 'accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7' \
  -H 'accept-language: en-GB,en;q=0.9,en-US;q=0.8,de;q=0.7,es;q=0.6,fr;q=0.5' \
  -H 'cache-control: max-age=0' \
  -H 'cookie: CUSTOMER_SESSION_ID_COOKIE=https://www.saveonfoods.com|7134a445-4606-4c7d-a3f1-54b51609d36f; TIMEOFFSET_COOKIE_NAME=630; __cf_bm=.IYeulR2_68exR57XSuLm_goCFx8_kqrPc3BBuyRblE-1712007790-1.0.1.1-plxzeS_99sAmbpR_gbutzw96vHE0l91H6Br7O_mFzdEA2a_ec2vdMl7Oi6E0RwQpxV9ZOF5Tz8J.lMCPWSaZldFvBnwkQ5um8f2poooike0; MI9_SHOPPING_MODE=pickup; MI9_RSID=2210; cf_clearance=8KZHvZsJaTeTnAYwRgSuQA98JidTu0r0bvZez8uLyEY-1712007794-1.0.1.1-9.tp0P9U_a2eYeW8yIjcF9OzcubUoBPrxjkP_P9K2nrpxB.1mnwzxkc2n_0ikwGPmbSJ2_Ids3sq1vuYpc1mwQ; SHOW_PLACEHOLDER_NOTIFICATIONS=false' \
[...]

PR #8826, opened 2024-07-12
Add save_on_foods_ca spider (188 locations)"

ShinyDog2636:
This is my first ever code commit, please advise if changes are needed

Fixes #8050
Fixes #3096

PR #9355, opened 2024-08-06
Add Save on foods (CA), refactor Cellarbrations (AU), and extra a "Storefrontgateway" spider

CloCkWeRX:
Supercedes #8826
Fixes https://github.com/alltheplaces/alltheplaces/issues/8050
Fixes https://github.com/alltheplaces/alltheplaces/issues/3096


CloCkWeRX:
@davidhicks I didn't research too deeply as to what this API is, potentially one for the autodetection.

CloCkWeRX:
@ShinyDog2636 here's a few adjustments to your original PR. Turns out we had a very similar spider; so I just refactored to use that across both.

Issue #9324, opened 2024-08-05
Tiendas 3B (AgileStoreLocatorSpider)


CloCkWeRX:
----
### Brand name

Tiendas 3B

### Wikidata ID

Q113217378
https://www.wikidata.org/wiki/Q113217378
https://www.wikidata.org/wiki/Special:EntityData/Q113217378.json

### Store finder url(s)

Official Url(s): https://www.tiendas3b.com/
- [x] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q113217378 https://www.tiendas3b.com/`
```
from locations.storefinders.agile_store_locator import AgileStoreLocatorSpider


class Tiendas3BMXSpider(AgileStoreLocatorSpider):
    name = "tiendas_3b_mx"
    item_attributes = {
        "brand_wikidata": "Q113217378",
        "brand": "Tiendas 3B",
    }
    allowed_domains = [
        "tiendas3b.com",
    ]
```

PR #9359, opened 2024-08-06
Add Tiedas 3B

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9324

Names look a bit funky in this one, might need to be checked a bit more


locations/spiders/tiendas_3b_mx.py
```diff
@@ -0,0 +1,12 @@
+from locations.storefinders.agile_store_locator import AgileStoreLocatorSpider
+
+
+class Tiendas3BMXSpider(AgileStoreLocatorSpider):
```
CloCkWeRX, line 4:
```suggestion
class Tiendas3bMXSpider(AgileStoreLocatorSpider):
```

CloCkWeRX:
Input wanted - drop the names? Keep them? They are what is displayed on the original site map.

Addresses - UTF-8 encoding issues with accented characters means there's a lot of `?` replacements. I beleive that's in the source data. Its still mostly useful, just low quality.

iandees:
Drop the names.

Text encoding issues suck, but I don't see a way to get it fixed, so roll with what we get for now.

Issue #7492, opened 2024-02-24
Nasz Sklep (PL) (shop/convenience) (Wordpress)


CloCkWeRX:
### Brand name

Nasz Sklep

Polish convenience store chain

### Wikidata ID

Q62070369
https://www.wikidata.org/wiki/Q62070369
https://www.wikidata.org/wiki/Special:EntityData/Q62070369.json

### Store finder url(s)

Official Url(s): https://www.nasz-sklep.pl/
- [x] I have run the below and if anything in generated, added it.
- [x] I have manually checked for a storefinder
- [x] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q62070369 https://www.nasz-sklep.pl/`
```
from locations.storefinders.wp_store_locator import WPStoreLocatorSpider


class NaszSklepPLSpider(WPStoreLocatorSpider):
        name = "nasz_sklep_pl"
        item_attributes = {
                "brand_wikidata": "Q62070369",
                "brand": "Nasz Sklep",
        }
        allowed_domains = [
                "nasz-sklep.pl",
        ]
```

CloCkWeRX:
https://nasz-sklep.pl/mapa-sklepow/

PR #9369, opened 2024-08-07
Add Nasz Sklep (PL)

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/7492


CloCkWeRX:
Main difference from the parent is... 

`for lat, lon in point_locations(searchable_points_file, "PL"):`

locations/spiders/nasz_sklep_pl.py
```diff
@@ -0,0 +1,33 @@
...
+        "brand": "Nasz Sklep",
+    }
+    allowed_domains = [
+        "nasz-sklep.pl",
+    ]
+    searchable_points_files = [
+        "eu_centroids_20km_radius_country.csv",
+    ]
+    time_format = "%I:%M %p"
+    max_results = 50
```
CloCkWeRX, line 20:
```suggestion
    max_results = 100
```

Cj-Malone:
- name includes html encoding.
- Too many shops. Is this every shop in Poland? https://nasz-sklep.pl/sklepy-wlasne/
- Opening hours are actually in English, but I suspect they are wrong as 100% have `opening_hours=Mo-Fr 09:00-17:00`.

Issue #9176, opened 2024-08-03
Cafe extrablatt (Sitemap, Structured Data)


CloCkWeRX:
### Brand name

Cafe extrablatt

German company

### Wikidata ID

Q1025505
https://www.wikidata.org/wiki/Q1025505
https://www.wikidata.org/wiki/Special:EntityData/Q1025505.json

### Store finder url(s)

Official Url(s): https://cafe-extrablatt.de/
- [x] I have run the below and if anything in generated, added it.
- [x] I have manually checked for a storefinder
- [x] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy links https://cafe-extrablatt.de/`
`pipenv run scrapy sitemap --pages https://cafe-extrablatt.de/`
`pipenv run scrapy sf --brand-wikidata=Q1025505 https://cafe-extrablatt.de/`
pipenv run scrapy sf https://cafe-extrablatt.de/standorte
```
```

CloCkWeRX:
https://cafe-extrablatt.de/standorte/details/cafe-extrablatt-berlin
https://cafe-extrablatt.de/sitemap?sitemap=locations&type=1533906435&cHash=cc74c2023100110ae12fab9dd810aa38

PR #9371, opened 2024-08-07
Add Cafe Extrablatt (DE)

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9176


iandees:
Is this exclusively in Germany? Might be worth renaming to use the DE country code.

PR #9375, opened 2024-08-07
Fix duplicates in WinterthurCHSpider

### LINKED ISSUE #9099: Duplicate id in winterthur_ch
The id `tWidPqOkZrC2-MsLg8PoKAHXqvY=` is assigned twice in spider `winterthur_ch`. I'm not sure if this is the only case of duplicate id but it may be worth checking these in an automated way before publishing the data.

CloCkWeRX:
Fix #9099 ?

### DIFF
--- a/locations/spiders/winterthur_ch.py
+++ b/locations/spiders/winterthur_ch.py
@@ -73,7 +73,7 @@ class WinterthurCHSpider(scrapy.Spider):
         for m in response.json()["FeatureCollection"]["member"]:
             if f := m.get("SitzbankWfs"):
                 ref = f["Banknummer"]
-                if ref in self.ids_seen:
+                if ref in self.seen_ids:
                     raise DropItem()
 
                 coords = f["msGeometry"]["Point"]["pos"].split()
--- a/locations/spiders/winterthur_ch.py
+++ b/locations/spiders/winterthur_ch.py
@@ -2,10 +2,10 @@ import logging
 
 import pyproj
 import scrapy
+from scrapy.exceptions import DropItem
 
 from locations.categories import Categories, apply_category
 from locations.items import Feature
-from scrapy.exceptions import DropItem
 
 
 # Open Data of the City of Winterthur, Switzerland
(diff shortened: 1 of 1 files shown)


locations/spiders/winterthur_ch.py
```diff
@@ -70,6 +72,10 @@ def parse_playgrounds(self, response):
     def parse_benches(self, response):
         for m in response.json()["FeatureCollection"]["member"]:
             if f := m.get("SitzbankWfs"):
+                ref = f["Banknummer"]
+                if ref in self.ids_seen:
```
CloCkWeRX, line 76:
```suggestion
                if ref in self.seen_ids:
```

locations/spiders/winterthur_ch.py
```diff
@@ -88,8 +94,9 @@ def parse_benches(self, response):
                         "material": self.parse_bench_material(btype),
                         "material:wikidata": self.parse_bench_material_wikidata(btype),
                     },
-                    "ref": f["Banknummer"],
+                    "ref": ref,
                 }
+                self.seen_ids.append(ref)
```
CloCkWeRX, line 99:
```suggestion
                self.seen_ids.add(ref)
```

CloCkWeRX:
Maybe not.

vfosnar:
I don't think dropping on id collision is the right thing to do. Implementing id generation based on both ref and position would be a better way to handle this.

CloCkWeRX:
Its the exact same bench twice in the dataset.

iandees:
The dupe detection pipeline step should be doing exactly this. We should figure out why that isn't dropping it as expected.

CloCkWeRX:
It does attempt to run, the problem with this dataset is it's a composite.
The benches have refs.
Others don't.
So, this change targets just the benches component and adds basically the duplication behavior, or is meant to.
Potentially the spider could be split in two.

iandees:
Yea, I would rather see this be two separate spiders than to duplicate the deduplication process.

CloCkWeRX:
Effort kind of high for... one ... doubled bench entry

PR #9379, opened 2024-08-07
Added Spar spider (2223 locations)

franga2000:
Spider for Spar, the European grocery store and supermarket chain.

This spider covers their Austrian, Hungarian, Slovenian and Croatian franchises, which all use the same store finder. I haven't found any others that use the same one, but if any more are found, it's easy to add them to the spider.

**EDIT:** I did some more research and it looks like this site structure is specific to Spar stores operated by [ASPIAG](https://www.aspiag.com/en/countries), which are only on the listed countries and northern Italy, which has a different website. Would it make sense to rename this spider? Something like `spar_aspiag`, maybe with a comment in the file explaining what that is?


iandees:
Yep, I like the idea of renaming the spider as you suggest and adding a comment. Maybe even point back to this PR for a little more context, too.

CloCkWeRX:
Other spar spiders/issues
#346 #448 #567

CloCkWeRX:
Does this close #346 I wonder, or are there other branches still?

iandees:
Yea, thanks for tagging those. It would be great to go through and figure out how this overlaps (if at all) with the other Spar spiders.

Issue #9273, opened 2024-08-04
Telepizza (WPGoMapsSpider, MapsMarkerProSpider?)


CloCkWeRX:
### Brand name

Telepizza

pizza restaurant chain

### Wikidata ID

Q2699863
https://www.wikidata.org/wiki/Q2699863
https://www.wikidata.org/wiki/Special:EntityData/Q2699863.json

### Store finder url(s)

Official Url(s): https://www.telepizza.com.co/
- [ ] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy links https://www.telepizza.com.co/`
`pipenv run scrapy sitemap --pages https://www.telepizza.com.co/`
`pipenv run scrapy sf --brand-wikidata=Q2699863 https://www.telepizza.com.co/`
```
```
Official Url(s): https://www.telepizza.ad/
- [ ] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy links https://www.telepizza.ad/`
`pipenv run scrapy sitemap --pages https://www.telepizza.ad/`
`pipenv run scrapy sf --brand-wikidata=Q2699863 https://www.telepizza.ad/`
```
```
Official Url(s): https://www.telepizza.ec/
- [ ] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy links https://www.telepizza.ec/`
`pipenv run scrapy sitemap --pages https://www.telepizza.ec/`
`pipenv run scrapy sf --brand-wikidata=Q2699863 https://www.telepizza.ec/`
```
```
Official Url(s): http://www.telepizza.com
- [ ] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy links http://www.telepizza.com`
`pipenv run scrapy sitemap --pages http://www.telepizza.com`
`pipenv run scrapy sf --brand-wikidata=Q2699863 http://www.telepizza.com`
```
```
Official Url(s): https://www.telepizza.es/
[...]

PR #9383, opened 2024-08-07
Add Food Delivery Brands

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9273
Or at least partly!


CloCkWeRX:
I might need to work out if these are all telepizza or a mix.

CloCkWeRX:
Coordinates are clearly just the centroid for the country

Issue #9014, opened 2024-07-28
 FABUTAN (Sitemap, Structured Data)


CloCkWeRX:
### Brand name

FABUTAN

Canadian chain of indoor tanning facilities

### Wikidata ID

Q120765494
https://www.wikidata.org/wiki/Q120765494
https://www.wikidata.org/wiki/Special:EntityData/Q120765494.json

### Store finder url(s)
https://fabutan.com/locations/

Official Url(s): https://fabutan.com/
- [x] I have run the below and if anything in generated, added it.
- [x] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q120765494 https://fabutan.com/`
```
```

CloCkWeRX:
https://fabutan.com/locations/manitoba/steinbach/victoria-plaza/
https://validator.schema.org/#url=https%3A%2F%2Ffabutan.com%2Flocations%2Fmanitoba%2Fsteinbach%2Fvictoria-plaza%2F

Structured Data

CloCkWeRX:
https://fabutan.com/sitemap.xml

PR #9402, opened 2024-08-08
Add Fabutan (CA)

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9014


iandees:
We could extract a lat/lon from the static map on each page.

CloCkWeRX:
Coordinates added @iandees

Issue #809, opened 2017-12-23
Airgas


iandees:
https://locations.airgas.com/ (Expand the browse by state thing)

davidhicks:
Seems to be a common self-hosted store finder (new, not sure what product/name) shared with #966, #1005, #1047, `hallmark` and `justice`. Search URL is `https://maps.locations.airgas.com/api/getAsyncLocations?template=search&level=search&search=Kansas%20City,%20KS,%20US&radius=100000&limit=100000`

CloCkWeRX:
Strangely, I get:
```
  File "/home/ubuntu/.pyenv/versions/3.11.8/lib/python3.11/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
               ^^^^^^^^^^^^^^^^^^^^^^
json.decoder.JSONDecodeError: Unterminated string starting at: line 3599 column 31 (char 184653)
2024-03-17 07:55:24 [scrapy.core.engine] INFO: Closing spider (finished)
2024-03-17 07:55:24 [locations.pipelines.duplicates] INFO: Dropped 0 duplicate items
2024-03-17 07:55:24 [scrapy.statscollectors] INFO: Dumping Scrapy stats:
```
But via my browser, I get a valid response.

CloCkWeRX:
```
from locations.categories import Categories, Extras, apply_category, apply_yes_no
from locations.storefinders.rio_seo_spider import RioSeoSpider


class AirgasUSSpider(RioSeoSpider):
    name = "airgas_us"
    item_attributes = {"brand": "Airgas", "brand_wikidata": "Q80635"}
    allowed_domains = ["locations.airgas.com"]
    start_urls = [
        "https://maps.locations.airgas.com/api/getAsyncLocations?template=search&level=search&search=Kansas%20City,%20KS,%20US&radius=100000&limit=100000"
    ]
```

CloCkWeRX:
This one is really frustrating. Started down the path of splitting and parsing each of the valid chunks; so we can drop the last one.
```
        raw_json = Selector(text=map_list).xpath("//div/text()").get()[:-1]
        for raw_json_chunk in raw_json.split("},{"):
            # print(raw_json_chunk)
            location = json.loads("[{{}}]".format(raw_json_chunk))

```

PR #9403, opened 2024-08-08
Add Airgas (US) and make RioSeoSpider more forgiving

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/809

Not ideal and can be dialled in with better radius searches.


locations/storefinders/rio_seo.py
```diff
@@ -38,7 +39,12 @@ def parse_autocomplete(self, response: Response, **kwargs: Any) -> Any:
 
     def parse(self, response, **kwargs):
         map_list = response.json()["maplist"]
-        data = json.loads("[{}]".format(Selector(text=map_list).xpath("//div/text()").get()[:-1]))
+        try:
+            data = json.loads("[{}]".format(Selector(text=map_list).xpath("//div/text()").get()[:-1]))
+        except json.decoder.JSONDecodeError:
```
CloCkWeRX, line 44:
```suggestion
        except json.decoder.JSONDecodeError e:
```

locations/storefinders/rio_seo.py
```diff
@@ -38,7 +39,12 @@ def parse_autocomplete(self, response: Response, **kwargs: Any) -> Any:
...
+            logging.warning(f"Could not parse response - check API output")
```
CloCkWeRX, line 45:
```suggestion
            logging.warning("Could not parse response - check API output: " + e.message)
```

locations/storefinders/rio_seo.py
```diff
@@ -38,7 +39,14 @@ def parse_autocomplete(self, response: Response, **kwargs: Any) -> Any:
 
     def parse(self, response, **kwargs):
         map_list = response.json()["maplist"]
-        data = json.loads("[{}]".format(Selector(text=map_list).xpath("//div/text()").get()[:-1]))
+        try:
+            data = json.loads("[{}]".format(Selector(text=map_list).xpath("//div/text()").get()[:-1]))
+        except json.decoder.JSONDecodeError:
+            logging.warning("Could not parse response - check API output")
```
davidhicks, line 45: Should `self.logger` be used instead of directly using `logging`, per https://docs.scrapy.org/en/latest/topics/logging.html#logging-from-spiders ?

CloCkWeRX, line 45:
```suggestion
            self.logger.warning("Could not parse response - check API output")
```

CloCkWeRX: Done

locations/storefinders/rio_seo.py
```diff
@@ -38,7 +38,14 @@ def parse_autocomplete(self, response: Response, **kwargs: Any) -> Any:
 
     def parse(self, response, **kwargs):
         map_list = response.json()["maplist"]
-        data = json.loads("[{}]".format(Selector(text=map_list).xpath("//div/text()").get()[:-1]))
+        try:
+            data = json.loads("[{}]".format(Selector(text=map_list).xpath("//div/text()").get()[:-1]))
+        except json.decoder.JSONDecodeError:
+            self.logger.warning("Could not parse response - check API output")
```
davidhicks, line 44: This reminded me that the project should document when to use WARNING and ERROR, which at the moment I think is fairly random in ATP.

In this case, I suggest perhaps `self.logger.error` is more appropriate than `self.logger.warning` because the spider will fail to return any results if this branch is encountered.

locations/storefinders/rio_seo.py
```diff
@@ -38,7 +38,14 @@ def parse_autocomplete(self, response: Response, **kwargs: Any) -> Any:
...
+            data = []
+        except TypeError:
```
davidhicks, line 46: Perhaps just catch all exception types (along with `json.decoder.JSONDecodeError`) as they all appear to result in the outcome of "Could not parse response - check API output"

locations/storefinders/rio_seo.py
CloCkWeRX: I chose warning because it will fail.. for that URL

davidhicks: > I chose warning because it will fail.. for that URL

Ah sorry I forgot this storefinder supports more than one start URL :)

PR #9408, opened 2024-08-09
Add greeenway spider (648 locations) 

KonradowyPL:
GreenWay spider for Poland and Slovakia


iandees:
What kind of store is this? If you specify the brand:wikidata, it will try to use Name Suggestion Index to add the correct tags, too.

CloCkWeRX:
charging stations operator in Slovakia and Poland

So in this case operator is probably most correct.

locations/spiders/greenway.py
```diff
@@ -0,0 +1,28 @@
+from scrapy import Spider
+from scrapy.http import Request
+
+from locations.dict_parser import DictParser
+from locations.geo import point_locations
+
+
+class GreenwaySpider(Spider):
+    name = "greenway"
+    item_attributes = {"operator": "GreenWay", "operator_wikidata": "Q116450281"}
```
CloCkWeRX, line 10:
```suggestion
    item_attributes = {"operator": "GreenWay", "operator_wikidata": "Q116450281", extras: {"amenity": "charging_station"}}
```

PR #9411, opened 2024-08-09
speedy_stop_us - remove names (how to remove name in ClosebySpider?)

matkoniecz:


matkoniecz:
wrong fix, names not removed

matkoniecz:
looking at https://github.com/alltheplaces/alltheplaces/blob/86623e37377bd3cade639da84caa0e9a41469714/locations/storefinders/closeby.py#L7 has not helped

iandees:
`ClosebySpider` doesn't call `post_process_item()` so your code here isn't getting called. Also, the `name` is coming from the call to `DictParser.parse()` in there.

matkoniecz:
yeah, guessed that `post_process_item` is just dead code here

not sure what is the proper way to override `parse` from https://github.com/alltheplaces/alltheplaces/blob/86623e37377bd3cade639da84caa0e9a41469714/locations/storefinders/closeby.py#L7

just drop that dependency and depend simply on `Spider` and copy code from `ClosebySpider`? Write own `parse` that would be copy of one from `ClosebySpider` but also would evict `name` and would be subclass of `ClosebySpider`?

CloCkWeRX:
I think it's reasonable to refactor the spider to define an empty post process item method, and in this spider override to as you have done.

CloCkWeRX:
(wider convo, should all storefinder spiders have an interface of pre and post processing so you can consistently override things?)

iandees:
Yes, storefinders should probably follow that pattern.

CloCkWeRX:
After #9419 this should work. Alternatively, cherry-pick just f727b5baec5224c940efc4a1cb876f43ec93a549

CloCkWeRX:
#9436

CloCkWeRX:
@matkoniecz - merge master into this and it'll work now :)

Issue #9348, opened 2024-08-06
Add labels for "value" or "number of locations"?


CloCkWeRX:
Similar to #9347

Spending half a day parsing 7 locations from bespoke HTML that is never mapped is probably more effort than it is worth writing a spider for, vs just mapping them.
7 locations in pristine JSON LD? Worth it.

Spending that same half a day to get a hundred locations in, it's more sensible.

At the moment I'm often putting either an exact number or order of magnitude estimate into an issue title. I think others do as well.

When it's exact, it helps check the PR is correct.
But even if it's just "0-10"; "10-20"; "20-50"; "50-100"; "100+", "300+", that gives a bit of a feel for return on effort

Having GitHub labels for annotating this helps us focus on most impact

CloCkWeRX:
Issue template will help here

Issue #9347, opened 2024-08-06
Add labels for difficulty, quality?


CloCkWeRX:
We have labels for "good first issue", but not as much for "hard work spider" or things in between.

IE is translating a JSON API with DictParser trivial, or will you have to add post processing and clean out closed locations for SEO purposes using a series of obsfucated keys?

Could we look at a broad indicator of "sizing/complexity" (easy, medium, hard, too hard)?

This would allow us to put links in the readme or adjust the sticky post for "how to help" to point new contributors to low hanging fruit.

CloCkWeRX:
@davidhicks I would be interested in your opinion on if this is valuable, as you do a lot of triage and backlog grooming

davidhicks:
The hard part I think is deciding what makes something easy, difficult, etc. I've been using "good first issue" for the simplest ones (e.g. StructuredData, simple store finders, simple HTML parsing of pages from a sitemap) and everything else I'd currently classify as being of a moderate difficulty, except a handful of cases where I'd say it is "hard" because there is ideally a core library of ATP that needs new supporting features, such as handling some more complex types of geographic radius/box searches.

PR #9456, opened 2024-08-11
Ask more optional questions on the new spider form

CloCkWeRX:
Slight variation to #9348 and #9347 

To play with it:
https://github.com/CloCkWeRX/alltheplaces/issues/new?assignees=&labels=new-spider&projects=&template=request.yaml

Sample issue:
https://github.com/CloCkWeRX/alltheplaces/issues/4

Has a 'new-spider' label optionally.

### DIFF
--- a/.github/ISSUE_TEMPLATE/request.yaml
+++ b/.github/ISSUE_TEMPLATE/request.yaml
@@ -1,5 +1,6 @@
 name: New Spider
 description: Request a new spider
+labels: new-spider
 body:
   - type: input
     attributes:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/request.yaml
```diff
@@ -13,10 +14,57 @@ body:
...
+        - Extreme - Obsfucated, brittle behaviour
+  - type: dropdown
+    id: size
+    attributes:
+      label: Number of POI?
+      description: Approximately how many points/locations/branches?
+      options:
+        - 0-10
+        - 10-20
+        - 20-50
```
Cj-Malone, line 52: We really shouldn't encourage spiders this small. We can't maintain the current ones. Big brands, important brands.

CloCkWeRX: Well, depends on what people consume the map data for. If it's about conserving developer resources then I agree.
If it's about significance...

Is there a favourite pub in my city, or coffee chain started my local people that I would personally go out of my way to map on OSM? Absolutely.

Do these people have the same economy of scale to compete with a multinational? Probably no.
Is going to a local business within, say 5km, with a local supply chain better than a hyper efficient org, like say Shien? I would argue yes in a lot of ways (carbon emissions, local jobs, self sustaining communities, etc)

The maintenance we can automate a lot of - do we need a grim reaper bot for proposing killing off busted spiders? Maybe?

The flip side of this is it's just a feature request. If it's 0-10 stores, do any of us go out of our way to build it? Probably no.
500+; different story.

Cj-Malone: You didn't add an option for 500+ though, you added one for 0-10.

We both spend an enormous amount of time mapping in OSM. I love adding independent shops and small local chains. https://www.openstreetmap.org/changeset/155209416 https://www.openstreetmap.org/changeset/86872194 But OSM and ATP are different. We can't scale to a million spiders, but OSM is already past a millions POIs. I want more independent POIs in ATP, but it can't be via a million spiders.

PR #9459, opened 2024-08-11
Add mkdocs, so that it is possible to generate static developer docs

CloCkWeRX:
Example:
image

Annoyingly a few path issues with our pipelines or similar.

To run, pipenv run mkdocs serve


CloCkWeRX:
So, for this one I'm not 100% sure on it. It merges nicely with some of the other markdown docs; but we kind of need to flip it into being hosted on github pages or statically or something; and then review the links we have put in the github markdown files.

I found some of the ways the reference docs presented to be a bit meh.

iandees:
This looks pretty good. I think I would want this to get added to the [alltheplaces/alltheplaces.xyz](https://github.com/alltheplaces/alltheplaces.xyz/) repo in a `docs/` path prefix. Maybe a change to the docs here could run a build step that opens a PR with the output to that repo.

CloCkWeRX:
I will have a go at moving it there, pulling in develop to auto document

PR #9531, opened 2024-08-12
Fix broken Nandos spiders and add BW and MU

arrival-spring:
Several of the Nando's spiders have been broken for a while due to the website format apparently changing.

PK website appears to be currently broken (issue #9533) but other than that this should fix the remainder.

### DIFF
--- a/locations/spiders/nandos_za.py
+++ b/locations/spiders/nandos_za.py
@@ -20,8 +20,12 @@ class NandosZASpider(scrapy.Spider):
         for i in data["locations"]:
             i["name"] = i.pop("locationName")
             i["phone"] = i.pop("primaryPhone")
+            if i["additionalPhone1"]:
+                i["phone"] += ";" + i.pop("additionalPhone1")
             i["province"] = i.pop("administrativeArea")
             i["website"] = self.web_root + i.pop("storeLocatorDetailsShortURL")
+            if i["addressLine2"]:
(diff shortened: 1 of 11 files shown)


locations/spiders/nandos_za.py
```diff
@@ -13,26 +13,33 @@ class NandosZASpider(scrapy.Spider):
...
     ]
+    web_root = "https://store.nandos.co.za/details/"
 
     def parse(self, response):
         data = response.json()
         for i in data["locations"]:
             i["name"] = i.pop("locationName")
             i["phone"] = i.pop("primaryPhone")
+            if i["additionalPhone1"]:
+                i["phone"] += ";" + i.pop("additionalPhone1")
```
arrival-spring, line 24: Actually, is this a good idea?
It gives results like `"phone": "010 216 9143;010 216 9149"`, rather than `"phone": "+27 10 216 9143"`

arrival-spring: Ping @iandees I was just too late commenting it seems

Cj-Malone:
```suggestion
                i["phone"] += "; " + i.pop("additionalPhone1")
```

arrival-spring: Or rather allow it not to have a space, which is common in OSM and seems just as valid. See #9548

PR #9541, opened 2024-08-13
[C&A] improve opening_hours formatting

merylstreet:
This PR merges duplicated opening_hours on multiple days

### DIFF
--- a/locations/spiders/c_and_a.py
+++ b/locations/spiders/c_and_a.py
@@ -5,7 +5,7 @@ import scrapy
 
 from locations.items import Feature
 from locations.pipelines.address_clean_up import clean_address
-
+from locations.hours import DAYS, OpeningHours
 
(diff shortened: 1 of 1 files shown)


locations/spiders/c_and_a.py
```diff
@@ -74,7 +76,7 @@ def parse_city(self, response):
                 "name": store.xpath('./div[@class="addressBox"]/p[@class="store"]/text()').get(),
                 "phone": phone.replace("Tel: ", "") if phone else None,
                 "country": country,
-                "opening_hours": "; ".join(hours),
+                "opening_hours": opening_hours.as_opening_hours(),
```
iandees, line 79: This is fine as is, but just so you know you can use the `OpeningHours` object directly and we'll call `as_opening_hours()` before writing to output.

PR #9548, opened 2024-08-13
Documentation: multiple phone numbers and phone cleaning

arrival-spring:
I spotted in #9531 that multiple phone numbers are being dealt with inconsistently. The split requires `; ` but the merge uses only `;` so I'm making the whitespace optional and updating the documentation for phone numbers to match what is happening.

### DIFF
--- a/DATA_FORMAT.md
+++ b/DATA_FORMAT.md
@@ -41,7 +41,7 @@ Each GeoJSON feature will have a `properties` object with as many of the followi
 | `addr:postcode`       | The postcode part of the address.
 | `addr:country`        | The country part of the address.
 | **Contact**           | _Contact information for the venue_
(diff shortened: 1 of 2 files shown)


arrival-spring:
Ok, maybe I don't know enough about Belgian phone numbers to know if that failing test makes sense to have.

iandees:
It's splitting on both `;` and `/`, so the Beligium test with `02/633` will fail without the whitespace check after the separator. I don't think I want to change that behavior. I don't see a great way to handle the `/` other than to remove support for the `/` as a phone separator, find the spiders that rely on it, and move them to `;`.

We should require lists of strings if we want to support multiple phone numbers, although I think I agree with @Cj-Malone that it might not be worth doing right now.

arrival-spring:
Ok, so I'll no longer change the regex, this is now only about the documentation which still reflects how data is being served.

iandees:
Thanks!

PR #9549, opened 2024-08-13
Remove broken kopps spider: 132 sloc and only 3 locations if it worked!

mjoe999:
I don't think this should be in the code base?


iandees:
It's a hometown store so it's important to me :)

iandees:
But yea, no lat/lon in there anymore.

mjoe999:
Nothing personal :-)

Issue #9373, opened 2024-08-07
Add centroids for all country codes at common breakpoints (500km, 200km, 100km, 50km?)


CloCkWeRX:
See also #6044

- We have eu centroids, which can be filtered by country - but not all spiders support that. Potentially this should be fixed. Couple of PL spiders as recent examples (#9369, #7589)
- ZA would be helpful (#9372)

This is probably to do easy but a bit opaque.

Later on, optimised centroid files described in #6044 would be really neat

CloCkWeRX:
searchable_points/earth_centroids_iseadgg.md apparently are the docs

CloCkWeRX:
@davidhicks I don't... quite understand the pipeline here - looking at say, https://github.com/alltheplaces/alltheplaces/blob/master/locations/searchable_points/de_centroids_iseadgg_50km_radius.meta I can understand there's a shape file used to clip an area, but I'm not really clear where it comes from.

If we were to do this for every country in the world; whats the current process?

https://help.openstreetmap.org/questions/53155/how-to-export-all-countries-boundary-from-osm-to-either-shape-file-or-kml-format ?
https://download.geofabrik.de/ ?

davidhicks:
> @davidhicks I don't... quite understand the pipeline here - looking at say, https://github.com/alltheplaces/alltheplaces/blob/master/locations/searchable_points/de_centroids_iseadgg_50km_radius.meta I can understand there's a shape file used to clip an area, but I'm not really clear where it comes from.

The documentation lacks in this area a bit. It's documented instead in each `.meta` file see for example `us_centroids_iseadgg_458km_radius.meta` which states:

```
# us.shp generation method:
# 1. Download and extract: https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/cultural/ne_10m_admin_0_countries_lakes.zip
# 2. Execute: ogr2ogr -sql "SELECT OGR_GEOMETRY FROM ne_10m_admin_0_countries_lakes WHERE ISO_A2='US'" us.shp ne_10m_admin_0_countries_lakes.shp
```

davidhicks:
I agree with this request. It was on my TODO list to generate these ISEA DGG centroid files for common radii for each country of the world, and commit them all at once. The script to regenerate them would also be provided to allow people to regenerate the centroid files periodically and whenever someone feels like doing it.

davidhicks:
[...]

PR #9378, opened 2024-08-07
Centroid generation scripts; data (20, 40, 120km)

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9373 partially

PR #9568, opened 2024-08-14
Add ISEADGG centroid generator script with 48km radius centroids

davidhicks:
Automates the process of regenerating a common set of ISEADGG centroid CSV files for each country of the world.

Further improvements possible in the future:
* Additional CSV files can be generated for other radiuses.
* Output CSV files can be merged together to generate common groupings of countries/territories such as gg_gb_im_je_centroids_iseadgg_48km_radius.csv
* Worldwide CSV files can be automatically generated for various radiuses.


davidhicks:
Partially fixes #9373 (48km added for now for comment, and will take suggestions for other radiuses to generate CSV files for).

Replaces #9378

davidhicks:
As there are a lot of CSV files added, it's helpful for me to highlight the generation script which is:

https://github.com/alltheplaces/alltheplaces/blob/b4bd86d474e2918fe13f29e15ef83dedc5d730ce/locations/searchable_points/centroids_iseadgg_generator.sh

iandees:
I wonder if we should move these to another repo that runs this script automatically via Github actions 🤔

iandees:
Does the existing code in the repo know how to open multiple CSVs to satisfy requests for multiple country points?

CloCkWeRX:
Yes, a number of spiders iterate over searchable point files as an array/list; though most spiders only target one country.

CloCkWeRX:
> I wonder if we should move these to another repo

Ideally we are going to end up with a fairly static set of files after an initial flurry of activity - how often do country borders change? (Geopolitics, rivers, coastal erosion, continental drift). Are those changes going to be frequent enough or big enough to affect the "search this grid" pattern?
Probably not 

The only other thing that might be more frequently changing would be if we did a lower admin area or targeted population centre version of this

davidhicks:
> Does the existing code in the repo know how to open multiple CSVs to satisfy requests for multiple country points?

As @CloCkWeRX said, yes, sort of. The problem though is if you want to iterate through centroids in a cluster of 6 countries close together, there would be some duplication of requests to servers for the same geographic coordinates that are close to country boundaries. I can add a new function that de-duplicates the centroids from a supplied list of CSV files, avoiding duplicate searches for the same coordinates.

panasonic-mafia:
Do we consider calculating points without storing them in files?

e.g. get country bbox -> calculate grid of points -> cut by country borders -> do requests.

It's quite easy to calculate grid of points (see code below), but I haven't figured a good way of getting bbox coordinates for a country and cutting by country borders (it probably needs cashing wrapper for [Natural Earth datasets](https://www.naturalearthdata.com/downloads/10m-cultural-vectors/) or python wrapper for [location conflation](https://github.com/rapideditor/location-conflation)).

```python
from shapely.geometry import box, Polygon
import numpy as np

class CountryGridGenerator:
    def __init__(self, lat_min, lat_max, lon_min, lon_max, step_lat, step_lon) -> None:
        self.lat_min = lat_min
        self.lat_max = lat_max
        self.lon_min = lon_min
        self.lon_max = lon_max
        self.step_lat = step_lat
        self.step_lon = step_lon
        self.grid = self.generate_grid()

    def generate_grid(self):
        lat_range = np.arange(self.lat_min, self.lat_max, self.step_lat)
        lon_range = np.arange(self.lon_min, self.lon_max, self.step_lon)
        grid = []
        
        for lat in lat_range:
            for lon in lon_range:
                bbox = box(lon, lat, lon + self.step_lon, lat + self.step_lat)
                center_point = bbox.centroid
                grid.append((bbox.bounds, (center_point.x, center_point.y)))
        
        return grid
    
    def print_bboxes(self):
        for i, (bbox, center) in enumerate(self.grid):
            print(f"Box {i}: {bbox}, Center: {center}")

    def iter_centerpoints(self):
        for bbox, center in self.grid:
            yield center

# Bounding box for Thailand
grid = CountryGridGenerator(5.612851, 20.41785, 97.345383, 105.63657, 0.3, 0.3)
for center in grid.iter_centerpoints():
    # do requests
```
image

davidhicks:
> Do we consider calculating points without storing them in files?
> 
> e.g. get country/region bbox -> calculate grid of points -> cut by country/region borders -> do requests.
> 
> It's quite easy to calculate grid of points (see code below), but I haven't figured a good way of getting bbox coordinates for a country and cutting by country borders (it probably needs cache wrapper for [Natural Earth datasets](https://www.naturalearthdata.com/downloads/10m-cultural-vectors/) or python wrapper for [location conflation](https://github.com/rapideditor/location-conflation)).

It's a valid problem for how to generate bounding boxes for the world based on the EPSG:3857 projection as some APIs expect an x1,y1,x2,y2 bounding box to be provided based on the EPSG:3857 projection. This will generally be a rather poor API though because there is a huge variance in bounding boxes between the poles and the equator as Mercator projections (of which EPSG:3857 is one) massively distort distances. For example, in Indonesia, a bounding box of 1 degree by 1 degree at 0N 110E is about 111km x 111km. In Northern Norway, a bounding box of 1 degree by 1 degree at 70N 23E is about 111km x 38km. With such a bounding box approach, one generally wants fewer bounding boxes near the poles (measured in degrees/minutes/seconds) and more bounding boxes near the equator. Some APIs aren't that smart though and have hard limits such as no bounding box can be larger than 1 degree by 1 degree.

ISEA DGG is suited to APIs that expect a radius search within a specified distance of a point. See https://en.wikipedia.org/wiki/Geodesic_grid and https://webpages.sou.edu/~sahrk/dgg/pubs/gdggs03.pdf for how it works. An ISEA DGG approach minimises the number of search coordinates needing to be used to fully search an area, no matter where on Earth this search occurs.

For bounding box searches and radius searches based on ISEA DGG, the following are generally desirable improvements for ATP:
* Don't bother searching oceans or deserts where nothing of interest exists.
* Use a hierarchical geodesic grid to allow densely populated areas to be searched with smaller bounding boxes or smaller radius circles.

davidhicks:
> ```python
>         self.step_lat = step_lat
>         self.step_lon = step_lon
> ```

These step values should probably be calculated with the `vincenty_distance` function of `geo.py` such that the Mercator distortion between poles and the equator for a 1 degree longitude difference are accounted for.

PR #9572, opened 2024-08-14
Lewis stores opening hours and fixes

arrival-spring:
* Ignore closed stores

* Postcodes outside of ZA appeared to be placeholders (9000 and 9999).

* Handle multiple phone numbers.

* Opening hours are a bit messy in the source data, but are now handled.

Follow up to #9566

### DIFF
--- a/locations/spiders/lewis_stores.py
+++ b/locations/spiders/lewis_stores.py
@@ -1,11 +1,12 @@
 import scrapy
 
 from locations.dict_parser import DictParser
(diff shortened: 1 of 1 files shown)


iandees:
This doesn't seem to have any categorization now. Does Name Suggestion Index need to get updated?

arrival-spring:
There was a duplicate Wikidata entry for Lewis. The items have now been merged, but NSI had the merged one. I have updated NSI, so categorisation should come with the next update from there.

PR #9574, opened 2024-08-15
Refactor RioSeoSpider, support more spiders with fewer code paths; Rewrite a bunch of Sitemap+StructuredData spiders to use RioSeoSpider

impiaaa:
Rio SEO websites have a few different access patterns. I tried various combinations of domains, requests, and parameters to find one pattern that worked for all of spiders. I tried to make sure the rewritten spiders have the same or better tagging as the old versions; unfortunately, "clean_eatz_us" lost the email contact for each location. I can revert that spider if requested. The "guess" spider went from 82 results to 390 results because it now includes more brands.


davidhicks (approved):
Thank you for cleaning up this RioSEO storefinder!

locations/storefinders/rio_seo.py
```diff
@@ -17,64 +15,69 @@ class RioSeoSpider(Spider):
...
+
+        if len(data) == self.limit:
+            self.logger.warning(f"Received {len(data)} entries, the limit may need to be raised")
 
         for location in data:
             feature = DictParser.parse(location)
-            feature["name"] = location["location_name"]
-            feature["ref"] = "{}_{}".format(location["fid"], location["lid"])
             feature["street_address"] = merge_address_lines([location["address_1"], location["address_2"]])
+            feature["ref"] = location["fid"]
```
Cj-Malone, line 49: I'm a little uncomfortable changing the ref, I don't know why it was made of both of them, but I assume it was for a reason.

CloCkWeRX:
I wonder if it's worth us running auto spider detection across all existing sites that are structureddata, to detect opportunities for rewrite to this.

Also worth doing as a first pass for #9492 #9490 etc

davidhicks:
> I wonder if it's worth us running auto spider detection across all existing sites that are structureddata, to detect opportunities for rewrite to this.
> 
> Also worth doing as a first pass for #9492 #9490 etc

I think you had already written a script to iterate through Wikidata IDs of brands/operators in NSI data and check any https://www.wikidata.org/wiki/Property:P12454 values for those Wikidata IDs to check that location information URLs are still active and whether ATP is already using any storefinder found to be present at the URL?

And now it'd also be possible to check not just P12454 but also do a check of any https://www.wikidata.org/wiki/Property:P856 values for those Wikidata IDs to then check robots.txt/sitemap.xml/etc to find any structured data in use or find any store locator pages not otherwise specified in https://www.wikidata.org/wiki/Property:P12454

CloCkWeRX:
> I think you had already written a script to iterate through Wikidata IDs of brands/operators in NSI data and check any https://www.wikidata.org/wiki/Property:P12454 values for those Wikidata IDs to check that location information URLs are still active and whether ATP is already using any storefinder found to be present at the URL?

Ah not quite. It pulls main website, but isn't yet aware of P12454. I have been adding them manually to wikidata but its hit and miss.

impiaaa:
> I wonder if it's worth us running auto spider detection across all existing sites that are structureddata, to detect opportunities for rewrite to this.

That is what I did for this PR. There are a few that were autodetected but that I didn't include here because they didn't work. I might come back to those later.

locations/storefinders/rio_seo.py
impiaaa: Yeah, I'd like to know too… oh apparently I wrote that part originally 😅. I don't remember why I chose both. Maybe I was being overly cautious of uniqueness, maybe I couldn't decide which to use, or maybe See's uses that combination on the website. This time I went with only `fid` because it's the one that appears in URLs, which I think makes it a natural `ref`. I verified it is unique within each spider.

PR #9575, opened 2024-08-15
[Our Airports] Fix ele

Cj-Malone:
`ele` is unusually defined as meters only, no option for us adding units like other OSM fields. However we weren't adding units anyway, so we were lying.

### DIFF
--- a/locations/spiders/our_airports.py
+++ b/locations/spiders/our_airports.py
@@ -26,7 +26,7 @@ class OurAirportsSpider(CSVFeedSpider):
         item["name"] = row["name"]
         item["lat"] = row["latitude_deg"]
         item["lon"] = row["longitude_deg"]
-        item["extras"]["ele"] = row["elevation_ft"]
+        item["extras"]["ele:ft"] = row["elevation_ft"]
         item["country"] = row["iso_country"]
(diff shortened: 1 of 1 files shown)


locations/spiders/our_airports.py
```diff
@@ -26,7 +26,7 @@ def parse_row(self, response: Response, row: dict):
         item["name"] = row["name"]
         item["lat"] = row["latitude_deg"]
         item["lon"] = row["longitude_deg"]
-        item["extras"]["ele"] = row["elevation_ft"]
+        item["extras"]["ele:ft"] = row["elevation_ft"]
```
iandees, line 29: Isn't the accepted custom to append "ft" to the value?

```suggestion
        item["extras"]["ele"] = f"{row["elevation_ft"]} ft"
```

Cj-Malone: Usually `'`, https://wiki.openstreetmap.org/wiki/Map_features/Units but ele is explicitly excluded, it has to be meters.  People have tried to remove the exception multiple times https://wiki.openstreetmap.org/wiki/Proposal:Ele_with_units

iandees: Huh, yet another thing that has changed in OSM over the years 😄

PR #9590, opened 2024-08-15
Add IcebolethuZASpider (132 locations)

arrival-spring:
There's nothing I can see to use for `ref`, but I don't seem to be able to leave it out. Using the name seems wrong, so what is a better option?

Website was messier than it first appeared, so some items/properties get dropped.

```
'atp/brand/Icebolethu Funerals': 132,
 'atp/brand_wikidata/Q122594408': 132,
 'atp/category/amenity/mortuary': 29,
 'atp/category/shop/funeral_directors': 103,
 'atp/field/city/missing': 132,
 'atp/field/country/from_spider_name': 132,
 'atp/field/email/missing': 132,
 'atp/field/image/missing': 132,
 'atp/field/lat/missing': 2,
 'atp/field/lon/missing': 2,
 'atp/field/opening_hours/missing': 132,
 'atp/field/operator/missing': 132,
 'atp/field/operator_wikidata/missing': 132,
 'atp/field/phone/invalid': 1,
 'atp/field/phone/missing': 29,
 'atp/field/postcode/missing': 132,
 'atp/field/state/missing': 132,
 'atp/field/street_address/missing': 1,
 'atp/field/twitter/missing': 132,
 'atp/field/website/missing': 132,
 'atp/item_scraped_host_count/www.icebolethugroup.co.za': 133,
```

### DIFF
--- a/locations/categories.py
+++ b/locations/categories.py
@@ -272,6 +272,7 @@ class Categories(Enum):
     VETERINARY = {"amenity": "veterinary"}
     WATER_RESCUE = {"emergency": "water_rescue"}
     ANIMAL_BOARDING = {"amenity": "animal_boarding"}
+    MORTUARY = {"amenity": "mortuary"}
 
     DATA_CENTRE = {"telecom": "data_center"}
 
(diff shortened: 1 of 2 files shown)


CloCkWeRX:
Either... ```no_refs = True``` or I sometimes make a combination of name + address + phone or coordinates to uniquely identify the store within the feed. The latter isn't great, but...

iandees:
If there isn't an identifier from the store, then set `no_refs = True`. I would prefer to not generate one.

arrival-spring:
> `no_refs=true

That's what I needed, but it doesn't seem to be in the docs, and I hadn't come across one yet.

iandees:
Yea, it's a little bit of a hack. If we document it, we probably also need to update DATA_FORMAT.md to mention the case where id might not be unique.

PR #9616, opened 2024-08-17
Refactoring: Add a json blob spider

CloCkWeRX:
Many spiders follow a pattern of having embedded JS or JSON in a storefinder page.

Future:
https://github.com/alltheplaces/alltheplaces/issues?q=is%3Aissue+is%3Aopen+JSON+Blob

Existing:
~27 spiders specifically using chompjs then iterating, probably more.
https://github.com/search?q=repo%3Aalltheplaces%2Falltheplaces%20%22in%20chompjs%22&type=code
https://github.com/search?q=repo%3Aalltheplaces%2Falltheplaces+%22locations+%3D+chompjs%22&type=code

Generally, they consist of a pattern of:

- Extract the JS/JSON from a script tag or similar
- Put it through chompjs or json.loads
- Iterate
- Change the keys into something either item or DictParser understands
- Yield feature

Last week I documented this a little bit at https://github.com/alltheplaces/alltheplaces/blob/master/docs/HARD_WORK_SPIDER.md#site-embeds-json-or-javascript-hash-js-blob ; but to distinguish these from other Hard Work Spiders - the difference between an API call result and ripping out JSON from a script tag is minimal - I feel like having a distinct pattern in place would be useful.

Putting up for discussion with some sample refactorings.
The main benefits are pushing specific methods to have a single responsibility, and it builds in the assumption that you are using DictParser


locations/json_blob_spider.py
```diff
@@ -0,0 +1,42 @@
...
+        for location in locations:
+            self.pre_process_data(location)
+            item = DictParser.parse(location)
+            yield from self.post_process_item(item, response, location) or []
+
+    def pre_process_data(self, location, **kwargs):
+        """Override with any pre-processing on the data, ie normalising key names for DictParser."""
+
+    def post_process_item(self, item, response, location):
+        """Override ith any post process on the item"""
```
CloCkWeRX, line 41:
```suggestion
        """Override with any post process on the item"""
```

CloCkWeRX:
As a second level to this, @davidhicks - thoughts on a general auto detector for "var (stores|branches|locations|markers) =" and just throwing the results through dictparser?
I've had okay success at that

Issue #9635, opened 2024-08-17
SpaceNK (Uberall)


CloCkWeRX:
### Brand name

Space.NK

British-based chain of cosmetics stores

### Wikidata ID

Q7572184
https://www.wikidata.org/wiki/Q7572184
https://www.wikidata.org/wiki/Special:EntityData/Q7572184.json

### Store finder url(s)

Official Url(s): https://www.spacenk.com/uk/en_GB/home-uk
- [x] I have run the below and if anything in generated, added it.
- [x] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q7572184 https://www.spacenk.com/uk/en_GB/home-uk`
pipenv run scrapy sf https://www.spacenk.com/uk/stores
```
```
Official Url(s): http://www.spacenk.co.uk/
- [x] I have run the below and if anything in generated, added it.
- [x] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q7572184 http://www.spacenk.co.uk/`
```
ubuntu@codespaces-88a392:/workspaces/alltheplaces$ pipenv run scrapy sf https://www.spacenk.com/uk/stores
from locations.storefinders.uberall import UberallSpider


class SpaceNKGBSpider(UberallSpider):
    name = "space_nk_gb"
    item_attributes = {
        "brand_wikidata": "Q7572184",
        "brand": "Space NK",
    }
    key = "bdwTQJoL7hB55B0EimfSmXjiMRV8eg"
```

PR #9639, opened 2024-08-17
Add Space NK (GB)

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9635


locations/spiders/space_nk_gb.py
```diff
@@ -0,0 +1,10 @@
+from locations.storefinders.uberall import UberallSpider
+
+
+class SpaceNKGBSpider(UberallSpider):
```
CloCkWeRX, line 4:
```suggestion
class SpaceNkGBSpider(UberallSpider):
```

Issue #9631, opened 2024-08-17
Just Over the Top (Uberall)


CloCkWeRX:
### Brand name

Just Over the Top

### Wikidata ID

Q104890420
https://www.wikidata.org/wiki/Q104890420
https://www.wikidata.org/wiki/Special:EntityData/Q104890420.json

### Store finder url(s)

Official Url(s): https://jott.com
- [x] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q104890420 https://jott.com`
pipenv run scrapy sf https://jott.com/en-eu/pages/storelocator
```
```
Official Url(s): https://justoverthetop.com/
- [x] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q104890420 https://justoverthetop.com/`
pipenv run scrapy sf https://justoverthetop.com//en-eu/pages/storelocator
```
from locations.storefinders.uberall import UberallSpider


class JustOvertheTopSpider(UberallSpider):
    name = "just_over_the_top"
    item_attributes = {
        "brand_wikidata": "Q104890420",
        "brand": "Just Over the Top",
    }
    key = "GebVCV1XBZBklJ6rQ5N2y44PCNLOpB"
```

PR #9640, opened 2024-08-17
Jott

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9631


locations/spiders/just_over_the_top.py
```diff
@@ -0,0 +1,10 @@
+from locations.storefinders.uberall import UberallSpider
+
+
+class JustOvertheTopSpider(UberallSpider):
```
CloCkWeRX, line 4:
```suggestion
class JustOverTheTopSpider(UberallSpider):
```

PR #9646, opened 2024-08-18
Refactor The Good Feet Store, Topaz to a similar pattern to JSON Blob Spider

CloCkWeRX:
These don't quite fit, but this at least labels the "extract json" bit in a consistent fashion

### DIFF
--- a/locations/spiders/the_good_feet_store.py
+++ b/locations/spiders/the_good_feet_store.py
@@ -6,7 +6,7 @@ from locations.hours import OpeningHours
 from locations.items import Feature
 
 
-class TheGoodFeetStoreSpider(Spider):
+class TheGoodFeetStoreSpider(JSONBlobSpider):
     name = "the_good_feet_store"
     item_attributes = {
         "brand": "The Good Feet Store",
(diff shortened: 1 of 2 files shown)


locations/spiders/the_good_feet_store.py
```diff
@@ -1,12 +1,12 @@
...
-from scrapy import Selector, Spider
+from scrapy import Selector
 
 from locations.categories import Categories
 from locations.hours import OpeningHours
 from locations.items import Feature
 
 
-class TheGoodFeetStoreSpider(Spider):
+class TheGoodFeetStoreSpider(JSONBlobSpider):
```
CloCkWeRX, line 9:
```suggestion
class TheGoodFeetStoreSpider(Spider):
```

locations/spiders/the_good_feet_store.py
```diff
@@ -1,12 +1,12 @@
 from chompjs import parse_js_object
-from scrapy import Selector, Spider
+from scrapy import Selector
```
CloCkWeRX, line 2:
```suggestion
from scrapy import Selector, Spider
```

Issue #9654, opened 2024-08-18
Q116232 and mini_be - somehow no matches


matkoniecz:
```
canonical feature failed to detect anything for:
{
    'type': 'Feature',
    'id': 'IH83BFUJTICK-QTrUegshmmcE-g=',
    'properties': {
        'ref': '22540_1',
        '@source_uri': 'https://c2b-services.bmw.com/c2b-localsearch/services/api/v4/clients/BMWSTAGE2_DLO/-/pois?category=MI&maxResults=2000&showAll=true&country=BE',
        '@spider': 'mini_be',
        'addr:street_address': 'Ambachtenlaan, 2a',
        'addr:city': 'Tienen',
        'addr:postcode': '3300',
        'addr:country': 'BE',
        'name': 'A&M Tienen',
        'phone': '+32 16 81 25 39',
        'website': 'http://www.vanosch-tienen.mini.be/nl_BE/home.html',
        'email': 'Tienen.bmw@anmgroup.be',
        'brand': 'Mini',
        'brand:wikidata': 'Q116232'
    },
    'geometry': {'type': 'Point', 'coordinates': [4.972085952758789, 50.80034637451172]}
}
```

mini_be spider has multiple (maybe every single one) entry without main tag

this is confusing as in NSI I see single instance of https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/mini_be.py

why it is not matched? Is failed NSI match resulting in missing main tag logged anywhere?

PR #9655, opened 2024-08-18
Fix Mini BE, SE

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9654


CloCkWeRX:
@matkoniecz - for these; where say "BMW" is a brand of many, many things; just a matter of adding the category

CloCkWeRX:
https://alltheplaces-data.openaddresses.io/runs/2024-08-10-13-32-13/logs/mini_se.txt - has been busted for a bit.

iandees:
Yea, we can probably delete it.

Issue #9556, opened 2024-08-14
some coop_no entries are without main tag


matkoniecz:
example:

```
{
    'type': 'Feature',
    'id': 'xfawI0Jo7tM6g_O3fZ4OMKgfW6E=',
    'properties': {
        'ref': '3476',
        '@source_uri': 'https://www.coop.no/api/content/butikker/elektro/coop-elektro-vikesa-3476',
        '@spider': 'coop_no',
        'addr:street_address': 'Vikesågate 1',
        'addr:city': 'Vikeså',
        'addr:postcode': '4389',
        'addr:country': 'NO',
        'name': 'Coop Elektro Vikeså',
        'phone': '+47 51 45 40 20',
        'website': 'https://www.coop.no/butikker/elektro/coop-elektro-vikesa-3476',
        'opening_hours': 'Mo-Fr 08:00-18:00; Sa 09:00-16:00',
        'brand': 'Coop Elektro',
        'brand:wikidata': 'Q111534601'
    },
    'geometry': {'type': 'Point', 'coordinates': [6.07829677, 58.63194162]}
```

more logs: https://gist.github.com/matkoniecz/4c3897d24a859e40145bb1dfe860e0af

https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/coop_no.py

CloCkWeRX:
For that one specifically, wants SHOP_ELECTRONICS? https://no.wikipedia.org/wiki/Coop_elektro

PR #9657, opened 2024-08-18
Add category to Coop Elecktro

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9556 (Partially? fully? Not sure)


matkoniecz:
> (Partially? fully? Not sure)

does not matter, if fix is partial it will be spotted again in few/several months (depending on volume of other cases)

Issue #9656, opened 2024-08-18
evgo - missing main tags, looks  like code bug


matkoniecz:
```
canonical feature failed to detect anything for:
{
    'type': 'Feature',
    'id': '4TOiHH-odgTD5UYeThAhiB3FIn4=',
    'properties': {
        'ref': '131254',
        'evgo:site_id': 65228,
        '@source_uri': 'https://account.evgo.com/stationFacade/findStationsByIds',
        '@spider': 'evgo',
        'addr:street_address': '2226 S Nellis Blvd',
        'addr:city': 'Las Vegas',
        'addr:state': 'NV',
        'addr:country': 'US',
        'name': 'Winterwood Pavilion',
        'brand': 'EVgo',
        'brand:wikidata': 'Q61803820'
    },
    'geometry': {'type': 'Point', 'coordinates': [-115.06522, 36.14618]}
}
https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/evgo.py
```\


from looking at https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/evgo.py code it looks like all should be marked as charging stations but output file is missing both `amenity` and `man_made`

iandees:
The `evgo_us` spider now applies the `CHARGING_STATION` category (fixed in #16680 and earlier). 🤖

PR #9661, opened 2024-08-18
EVGo - Refactor to DictParser and yield item properly

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9656 ?


iandees:
This seems to be missing a bunch of chargers: https://evgo.com/find-a-charger/

CloCkWeRX:
Screenshot_20240820-064301

CloCkWeRX:
UI says "search by plug share" - https://www.plugshare.com/ - perhaps they are migrating?

iandees:
8 is a bit too far away from the list of chargers they have on their website. I don't see anything about plugshare on the page I linked to. There's a big 'ol JSON blob with what looks like a hard to parse list of all ~200 charging stations on the map.

PR #9667, opened 2024-08-18
avoid conflict with OSM relation type tag

matkoniecz:


matkoniecz:
note: I see no object_type in @scraperbot output. Maybe something went wrong?

CloCkWeRX:
I wonder if going a bit wider than this one, it's worth adding a validation in to indicate that's reserved so we find any similar bugs easily going forward

matkoniecz:
I definitely like such idea.

iandees:
Yea, it might be a good idea to flag in CI...

PR #9668, opened 2024-08-18
African Bank ZA spider (443 locations)

arrival-spring:
```
'atp/brand/African Bank': 443,
 'atp/brand_wikidata/Q4689703': 443,
 'atp/category/amenity/bank': 443,
 'atp/cdn/cloudflare/response_count': 2,
 'atp/cdn/cloudflare/response_status_count/200': 1,
 'atp/cdn/cloudflare/response_status_count/404': 1,
 'atp/field/country/from_spider_name': 2,
 'atp/field/email/invalid': 2,
 'atp/field/email/missing': 443,
 'atp/field/image/missing': 443,
 'atp/field/lat/missing': 2,
 'atp/field/lon/missing': 2,
 'atp/field/operator/missing': 443,
 'atp/field/operator_wikidata/missing': 443,
 'atp/field/phone/invalid': 5,
 'atp/field/phone/missing': 6,
 'atp/field/twitter/missing': 443,
 'atp/item_scraped_host_count/api.locationbank.net': 443,
 'atp/nsi/cc_match': 443, 
```

PR #9684, opened 2024-08-18
Location Bank storefinder and spider

arrival-spring:
Location Bank are used by Nandos in some countries (#9531), as well as African Bank (#9668) so I will plan to refactor those after this has been merged.

Location Bank only have 5 office locations, so I have not added them to NSI.

There are then at least 6 more spiders to be added that use Location Bank that I have found so far.

### DIFF
--- a/locations/spiders/location_bank_company.py
+++ b/locations/spiders/location_bank_company.py
@@ -1,7 +1,12 @@
-from locations.storefinders.location_bank import LocationBankSpider
 from locations.categories import Categories
+from locations.storefinders.location_bank import LocationBankSpider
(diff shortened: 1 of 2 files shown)


CloCkWeRX:
What's the main indicator for these? A call to api.locationbank?
Fyi @davidhicks

arrival-spring:
> What's the main indicator for these? A call to api.locationbank? Fyi @davidhicks

Yes, they call `https://api.locationbank.net/storelocator/StoreLocatorAPI?clientId={self.client_id}` which returns a full store list

PR #9686, opened 2024-08-19
[Fairmont] Add spider

Cj-Malone:


CloCkWeRX:
Is there overlap with the Accor spider now? Not clear, both mention Q1393345

Cj-Malone:
Yeah looks like there is an overlap, but this one has some better metadata.

PR #9688, opened 2024-08-19
[banca_sella_it] new spider (282 locations)

mjoe999:
Saw what @CloCkWeRX had done with the `JSONBlobSpider` and thought we could default `extract_json` to treating the body response as JSON which is a regular pattern and then having an optional single level key to get the location array which is also a common pattern leading into the `DictParser` dance.


CloCkWeRX:
Yeah I was pondering something similar for "look for all of the common `var (name) = ...` to avoid having to do so much xpath. I held off because of the js blob vs js blob[key] pattern, so this is useful

PR #9697, opened 2024-08-19
Add Legalwise ZA spider (37 locations)

arrival-spring:
Needs NSI update for feature tags

On the one hand, the api here returns `"showDetailsAdditionalPhone": false`, despite including data for both primaryPhone and additionalPhone1. However, all primaryPhone are the same number, and only additionalPhone1 is location specific.

So, do we respect what they want displayed and give no phone number, or ignore that, drop the generic number and return location specific phone numbers?


CloCkWeRX:
Location specific phone numbers seem to be the best option if we have them

arrival-spring:
> Location specific phone numbers seem to be the best option if we have them

Even if they are not being publicly displayed?

iandees:
Both is ok.

PR #9709, opened 2024-08-20
[chicken_licken] add 308 primarily southern Africa locations

mjoe999:
`JSONBlobSpider` has cut a few lines of boiler plate here.


CloCkWeRX:
Woo

PR #9715, opened 2024-08-20
[woolworths_za] add spider (466 locations)

mjoe999:


arrival-spring:
Is that what you meant for the website? It's just the store locator page and so I'm not sure it's helpful

mjoe999:
> Is that what you meant for the website? It's just the store locator page and so I'm not sure it's helpful

fair enough ... changed.

PR #9723, opened 2024-08-20
Capitec Bank spider

### LINKED ISSUE #828: Capitec Bank
https://www.capitecbank.co.za/branch-locator

arrival-spring:
Fixes #828 but I'm not sure about this because:
* The robots.txt disallows /api/
* The API only returns a maximum of 100 objects, so we need to do lots of requests to get everything.
* We can't limit to e.g. just branches because there's no way I can see to do separate searches for branches and ATMS, filtering is done client-side after the results are returned

I tried different radii for searching points and got the following:

- 315km radius, 13 requests, 987 objects (199 branches)
- 158km radius, 36 requests, 1975 objects (359 branches)
- 94km radius, 78 requests, 2884 objects (520 branches)
- 79km radius, 98 requests, 3123 objects (567 branches)
- 48 km radius, 262 requests, 3483 objects (635 branches)
- 24km radius, 964 requests, not attempted

Even comparing 48km and 79km radii I can see that there are locations in the 79km search not returned in the 48km search.

The website claims 7000 ATMS, but I don't think we will get that many. I think ATMs at branches are not included in these results and there could be 2 or 3 per branch.

### DIFF
--- a/locations/spiders/capitec_bank_za.py
+++ b/locations/spiders/capitec_bank_za.py
@@ -1,0 +1,57 @@
+from scrapy import Request, Spider
+
+from locations.categories import Categories, Extras, apply_category, apply_yes_no
+from locations.dict_parser import DictParser
+from locations.geo import country_iseadgg_centroids
+from locations.hours import DAYS_WEEKDAY, OpeningHours
+
+DAYS_TERMS = {
+    "OpeningHours": DAYS_WEEKDAY,
+    "SaturdayHours": "Sa",
+    "SundayHours": "Su",
+}
+
+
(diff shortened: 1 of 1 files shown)


CloCkWeRX:
While it seems like a lot of requests, consider that a crawlspider or sitemap spider is 1x request per poi roughly - so getting ~100 hits per api call/5 unique per request is okay-ish.

At some point spiders like this could also be tagged as "run monthly" or similar; so that even if it is doing a lot of work on average it spreads out to be a few per week (assuming you went to 24km radius, ~1k requests/month = 250 a week average).

PR #9724, opened 2024-08-20
Add AlphaPharmZASpider (365 locations)

arrival-spring:


locations/spiders/alpha_pharm_za.py
```diff
@@ -0,0 +1,43 @@
...
+from locations.json_blob_spider import JSONBlobSpider
+
+
+class AlphaPharmZASpider(JSONBlobSpider):
+    name = "alpha_pharm_za"
+    item_attributes = {
+        "brand": "Alpha Pharm",
+        "brand_wikidata": "Q116487265",
+    }
+    allowed_domains = ["alphapharmacies.co.za"]
```
mjoe999, line 14: Minor comment: why bother with `allowed_domains` as this is not a `CrawlSpider` that may spiral out of control.

arrival-spring: Thanks for the tip. I copied from elsewhere and didn't think about it I guess. I've now removed that.

PR #9728, opened 2024-08-21
[usave_za] add more detailed outlet profiling

mjoe999:
@arrival-spring : this represents the last that I had in the cupboard concerning ZA suppliers for now!

### DIFF
--- a/locations/spiders/usave_za.py
+++ b/locations/spiders/usave_za.py
@@ -1,14 +1,53 @@
-from locations.categories import Categories
-from locations.spiders.medirite_za import MediriteZASpider
+import json
+
+import scrapy
+
+from locations.categories import Categories, apply_category
+from locations.dict_parser import DictParser
 
 
-class UsaveZASpider(MediriteZASpider):
+class UsaveZASpider(scrapy.Spider):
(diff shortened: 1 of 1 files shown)


arrival-spring:
Thanks, although this will now overlap with the shoprite and medirite spiders.

I've been working on a spider to replace them, this and also the checkers_za spiders as it turns out we can ping this JSON file on the shoprite holdings website and get results for all the brands in all the countries, with the appropriate parameters.

I'm just stuck on the file encoding being returned for a couple of the countries at the moment, as special characters are not working. Maybe I'll post what I have as a draft later.

mjoe999:
@arrival-spring : sounds like you're on top of the game in South Africa and so are best placed to decide how the various brands in the group should be presented. All good!

arrival-spring:
Well I've just been starting to get into this and adding missing chains that I'm aware of. I've made fairly decent coverage of the main ones in NSI, so that's a good source to start looking for missing brands here in ATP.

arrival-spring:
I have now got it working in #9752, I'm sorry for your effort wasted here, but that gets all brands across many countries in Africa.

PR #9732, opened 2024-08-21
Backport `hours_keys` from autospidergen branch

CloCkWeRX:
https://github.com/alltheplaces/alltheplaces/pull/6526/files#diff-837cba580b823ade53ed7b508177352330903ab135e52a1d1ce4a5622a892064L109

@davidhicks - I wasn't really sure if replacing the 'add range, here's a time format' with add ranges from string was a safe replacement; so omitted it.


davidhicks (approved):
I'll double check why I switched to `OpeningHours.add_ranges_from_string` but I seem to recall it was because I came across a WPStoreLocatorSpider which didn't use a HTML table for opening hours, rather it was a plain text string of opening hours information.

PR #9741, opened 2024-08-21
[absa_za] add spider (551 banks, 2985 ATMs)

mjoe999:
I believe ABSA also has operations in Kenya and Ghana.

Fixes https://github.com/alltheplaces/alltheplaces/issues/827


arrival-spring:
Fixes #827

PR #9745, opened 2024-08-21
willscot: fix and rename to willscot_ca_us

davidhicks:


CloCkWeRX:
I think it's having a senior moment, for some reason!

davidhicks:
pre-commit.ci autofix

iandees:
The name/classname/filename check seems to have added ~40-45 seconds to the pre-commit check for whatever reason. Running it locally is very fast.

PR #9746, opened 2024-08-21
Make JsonBLOBParser dict/array handling more othogonal

mjoe999:
So #9743 improved the world. I was wondering about dict returns which also happen a lot so that's good! We should still allow sub classes to have a go at preprocessing the JSON "feature". Also the key handling is a bit magical in the sense that "id" is picked on yet `DictParser` will have a good go at things. If it failed the spider can always set `ref` from `ATP_FEATURE_KEY`.

@davidhicks and @CloCkWeRX have a little skin in this game and will no doubt comment if need be.

### DIFF
--- a/locations/json_blob_spider.py
+++ b/locations/json_blob_spider.py
@@ -20,6 +20,9 @@ class JSONBlobSpider(Spider):
     # If set then use as a key into the JSON response to return the location data array.
     locations_key: str = None
 
+    # For passing to the subclass when iterating a JSON dict.
+    ATP_FEATURE_KEY = "atp:feature_key"
+
     def extract_json(self, response: Response):
         """
         Override this method to extract the main JSON content from the page. The default
@@ -57,11 +60,9 @@ class JSONBlobSpider(Spider):
             yield from self.post_process_item(item, response, feature) or []
 
     def parse_feature_dict(self, response: Response, feature_dict: dict, **kwargs):
-        for feature_id, feature in feature_dict.items():
-            if not feature.get("id"):
-                feature["id"] = feature_id
-            else:
-                feature["feature_id"] = feature_id
+        for feature_key, feature in feature_dict.items():
+            feature[self.ATP_FEATURE_KEY] = feature_key
+            self.pre_process_data(feature)
             item = DictParser.parse(feature)
             yield from self.post_process_item(item, response, feature) or []
 


davidhicks:
I'm not sure in what circumstance `parse` would ever need overriding? `extract_json` can return an array of dictionaries (each being a feature to extract), or can return a dictionary of dictionaries (where the key is an ID and dictionary within is the rest of the feature to extract). What else could `extract_json` return if it wasn't an array of dictionaries or dictionary of dictionaries?

CloCkWeRX:
I ignored handling dicts in the original change because it was something like a 10:1 ratio of array vs dict when I was doing sample refactorings.
So... I am not fussed, just make sure the docs are up to date and newbie friendly!


I think I still want to focus more on the first step - extract json - and a bit like dictparser just have a top 10 var X = some structure mappings or patterns to try first.

I feel like there are only so many locations|markers|stores|tiendas variable names for unminified JS, followed by either an array or a dict; which has then got a lat/lon pattern in it.

Once we have that, well, then the difference between it doing "extract json" and "autodetect json suitable for extraction" is very low.

And it means that a boat load of spiders no longer have to implement an extract json hook if the default is good enough.

davidhicks:
> I ignored handling dicts in the original change because it was something like a 10:1 ratio of array vs dict when I was doing sample refactorings. So... I am not fussed, just make sure the docs are up to date and newbie friendly!
> 
> I think I still want to focus more on the first step - extract json - and a bit like dictparser just have a top 10 var X = some structure mappings or patterns to try first.
> 
> I feel like there are only so many locations|markers|stores|tiendas variable names for unminified JS, followed by either an array or a dict; which has then got a lat/lon pattern in it.
> 
> Once we have that, well, then the difference between it doing "extract json" and "autodetect json suitable for extraction" is very low.
> 
> And it means that a boat load of spiders no longer have to implement an extract json hook if the default is good enough.

Agreed 100% that `extract_json` should be at least be extended to take a parameter such as `self.js_variable_name` and then automatically find the first JavaScript `<script>` tag in the HTML document (if it is a HTML document rather than a JSON response) that sets this `self.js_variable_name` to either an array or dictionary. That way for most spiders all that needs specifying is a single string such as "tiendas" in `self.js_variable_name`. Bonus points if we can avoid even needing to specify `self.js_variable_name` (just having it as a fallback override) because we try to guess the common variations of `stores`, `locations`, `features`, `tiendas`, etc.

locations/json_blob_spider.py
```diff
@@ -20,6 +20,9 @@ class JSONBlobSpider(Spider):
     # If set then use as a key into the JSON response to return the location data array.
     locations_key: str = None
 
+    # For passing to the subclass when iterating a JSON dict.
+    ATP_FEATURE_KEY = "atp:feature_key"
```
iandees, line 24: It took me a while to figure out what was going on here, so here's an attempt at making the doc clearer.

This variable could also go outside of the class, since it's not an instance variable.

```suggestion
    # When processing a dict of locations in a JSON blob, the dictionary key of the scraped data will get
    # added to the item using this key so that DictParser uses it as a `ref` and the items get deduped
    # correctly.
    ATP_FEATURE_KEY = "atp:feature_key"
```

PR #9777, opened 2024-08-22
Pie City ZA spider

arrival-spring:
This is just the same as tiger_wheel_and_tyre, but I can't find any more of their clients. If/when I do come across more I'll make a dedicated storefinder, rather than cross-inheriting like this.


CloCkWeRX:
Storefinder is "goreview", marketing docs at
https://socialplaces.io/store-locator-local-pages/

PR #9788, opened 2024-08-23
[SallingGroup DK] Add spider

karen-kerr:


CloCkWeRX:
There seems to be one for netto_salling already, is there much overlap? (Q552652)

Issue #9404, opened 2024-08-09
New spider generator from manual command line prompts


CloCkWeRX:
In the ruby world, specifically rails, it's handy to have code generator tools.
Some are more useful than others (ie: rails g migration foo_bar_columns_change)

@davidhicks' autogen branch does a great job for a lot of spiders when it hits an auto detection.
But often I am also just whacking Ctrl+f, searching for (Sitemap, structureddataspider) and copy pasting my way to something half working.

Similar to #7608 it would be a big time saver if I could just

$ scrapy g spider Storemapper Q1234567 foo_bar_us

And have something generate; so I have to only
git checkout -b foo_bar_us
git add locations/spiders/foo_bar_us.py
pipenv run scrapy crawl foo_bar_us

CloCkWeRX:
(while this seems a trivial use case, similar to bash completion it's about lowering the cognitive load of developers; which is the main constraint in my view)

davidhicks:
There's already https://docs.scrapy.org/en/latest/topics/commands.html#std-command-genspider in base Scrapy that kind of does this too, but we'd need a new name for it. I did have a look at use of templates for each generated spider but for a reason that has currently escaped me (probably lack of complex type support in templates?) decided against it and to just print spider output without templates being used.

CloCkWeRX:
Ah nice!

CloCkWeRX:
https://github.com/CloCkWeRX/alltheplaces/tree/new_spider_scaffold first pass of generating the commands.

CloCkWeRX:
TODO: https://github.com/scrapy/scrapy/blob/master/scrapy/commands/genspider.py - our own templates?

iandees:
🤖 `locations/commands/genspider.py` now exists — a spider scaffold/PR generator command named consistently with Scrapy's own `genspider`, added in #9849.

PR #9849, opened 2024-08-25
New spider scaffold/pull request generator

CloCkWeRX:
Very tentative version of https://github.com/alltheplaces/alltheplaces/issues/9404

- Adds a command generator, so that you can generate two steps for "Make branch, spider template, open it in an editor" and "Run, commit, push to your fork"
- Adds templates for JSON Blob, Structured Data, WP Go Maps Spider variants. I chose to start here to avoid overlap with the autospidergen branch; but I can see the two approaches converging.
- Passing more variables to the spider for say, wikidata or other flags

What's missing?
- Chaining together these tools so that *if* you detect, say, structured data; it suggests you generate a spider using the template or autospidergen


templates/spiders/structured_data_spider_sitemap.tmpl
```diff
@@ -0,0 +1,16 @@
+from scrapy.spiders import SitemapSpider
+
+from locations.structured_data_spider import StructuredDataSpider
+
+
+class $classname(SitemapSpider, StructuredDataSpider):
+    name = "$name"
+    #item_attributes = {"brand_wikidata": "Add if known"}
```
CloCkWeRX, line 8:
```suggestion
    #item_attributes = {"brand": $name, "brand_wikidata": "Add if known"}
```

templates/spiders/json_blob.tmpl
```diff
@@ -0,0 +1,33 @@
...
+
+class BurgermeisterSpider(WpGoMapsSpider):
+    name = "burgermeister"
+    item_attributes = {
+        "brand_wikidata": "Q116382535",
+        "brand": "Burgermeister",
+    }
+    allowed_domains = [
+        "burgermeister.com",
+    ]
```
iandees, line 33: Is this supposed to be here?

CloCkWeRX: Fixed

CloCkWeRX:
@davidhicks thoughts on this alongside your branch? I feel like this is good for "I know there is a spider that is common" (jsonblob, structureddata, crawl, Sitemap, basic); though arguably we could go make a bunch of templates for each main Storefinder; for autodetect not working.

davidhicks:
A few initial thoughts:
- Can template names be abbreviated e.g. "sds_crawl" instead of having to type "structured_data_spider_crawl"?
- Templates can/should easily include type hinting on all methods and class attributes?
- Includes "extras" in "item_attributes" with a dummy category that users would have to replace?

panasonic-mafia:
I don't have access to a laptop now to check again, but when i tried to add custom templates last time, scrapy stopped to read standard templates...
https://github.com/scrapy/scrapy/issues/3680

CloCkWeRX:
> Suggest to just transfer the default template file to new TEMPLATES_DIR directory.
Done

CloCkWeRX:
Note to self:
https://github.com/scrapy/scrapy/pull/6470

Cj-Malone:
> * Includes "extras" in "item_attributes" with a dummy category that users would have to replace?

I'm not sure we should encourage this. I know we do it sometimes, but I don't think it's ideal.

davidhicks:
> > * Includes "extras" in "item_attributes" with a dummy category that users would have to replace?
> 
> I'm not sure we should encourage this. I know we do it sometimes, but I don't think it's ideal.

I thought this was encouraged (by mass additions of `{"extras": Categories.WHATEVER.value}` to `item_attributes` in the past approx.  6 months? It allows differentiation between two brands of the same name in the same country that have different primary OSM tags. For example, `Brand123` may be used for both `Categories.FUEL_STATION` and `Categories.SHOP_CONVENIENCE` within the same country. I note this is a somewhat rare situation though!

Cj-Malone:
> > > * Includes "extras" in "item_attributes" with a dummy category that users would have to replace?
> > 
> > 
> > I'm not sure we should encourage this. I know we do it sometimes, but I don't think it's ideal.
> 
> I thought this was encouraged (by mass additions of `{"extras": Categories.WHATEVER.value}` to `item_attributes` in the past approx. 6 months? It allows differentiation between two brands of the same name in the same country that have different primary OSM tags. For example, `Brand123` may be used for both `Categories.FUEL_STATION` and `Categories.SHOP_CONVENIENCE` within the same country. I note this is a somewhat rare situation though.

It works for some cases, because I updated the pipeline to treat `extras` in `item_attributes` in a special way.

Eg

```python
from typing import Any

from scrapy import Spider
from scrapy.http import Response

from locations.categories import Categories, Extras, apply_yes_no


class Brand123Spider(Spider):
    name = "brand_123"
    item_attributes = {"extras": Categories.FUEL_STATION.value}
    start_urls = ["data:,"]

    def parse(self, response: Response, **kwargs: Any) -> Any:
        item = Feature(ref="123")
        apply_yes_no(Extras.SHOWERS, item, True)
        yield item
```
```python
{'extras': {'@source_uri': 'data:,',
            '@spider': 'brand_123',
            'amenity': 'fuel',
            'shower': 'yes'},
 'ref': 'abc'}
```

However in other cases, not the pipeline for example, it doesn't work as expected. eg:

```python
class BrandABCSpider(Spider):
    name = "brand_abc"
    start_urls = ["data:,"]

    def parse(self, response: Response, **kwargs: Any) -> Any:
        item = Feature(ref="abc")
        apply_yes_no(Extras.SHOWERS, item, True)
        item.update(Brand123Spider.item_attributes)
        yield item
```
```python
{'extras': {'@source_uri': 'data:,',
            '@spider': 'brand_abc',
            'amenity': 'fuel'},
 'ref': 'abc'}
```

There is also the possibility that we change how categories work at some point in the future, eg, a flat value instead of dicts (Like was recommended), this would make checking if a poi was a specific category much easier. Using `apply_category` in spider means we could update it without editing individual spiders, just the shared code.

davidhicks:
Any thoughts on inserting an `ATPSpider` subclass that sits between Scrapy's `Spider` and all ATP spider classes that:
- Adds attributes: `brand`, `brand_wikidata`, `operator`, `operator_wikidata` where at least one brand or one operator is required.
- Adds attribute `top_level_tag` that accepts a `Categories.XYZ` dictionary. Required for all spiders.
- Adds attribute `locale` that accepts a `Locale.XYZ` class (will share soon: this is a branch that I've been working on to consolidate all translations for hours, "closed" labels, dictionary keys, etc into language-specific classes). Required for all spiders.

That way we'd largely remove the need for spiders to set `item_attributes` at all. The only time that `apply_category` would generally get used is the spiders which retrieve a variety of features such as convenience stores, fuel stations, etc all jumbled together.

locations/commands/genspider.py
```diff
@@ -0,0 +1,146 @@
...
+            "module": module,
+            "name": name,
+            "url": url,
+            "domain": extract_domain(url),
+            "classname": f"{capitalized_module}Spider",
+        }
+        # Flatten the parameters and merge with the other variables
+        return tvars | self.parameters
+
+    # TODO: If https://github.com/scrapy/scrapy/pull/6470 is merged, much of this can be removed
```
CloCkWeRX, line 120: This did get merged, need to wait for release or point it git version and remove this method

PR #9863, opened 2024-08-25
[BeerStore CA] Flag as needing proxy

CloCkWeRX:


iandees:
The URL used by this spider doesn't seem to exist anymore.

CloCkWeRX:
Delete then?

iandees:
There's an `authorization` header added too and I'm not excited about calculating that.

I think I'll just delete it.

Issue #9809, opened 2024-08-24
Mochachos (WpGoMaps)


CloCkWeRX:
### Brand name

Mochachos

Mexican chicken franchise chain

### Wikidata ID

Q116619117
https://www.wikidata.org/wiki/Q116619117
https://www.wikidata.org/wiki/Special:EntityData/Q116619117.json

### Store finder url(s)

Official Url(s): https://www.mochachos.com/
- [ ] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy links https://www.mochachos.com/`
`pipenv run scrapy sitemap --pages https://www.mochachos.com/`
`pipenv run scrapy sf --brand-wikidata=Q116619117 https://www.mochachos.com/`
```
from locations.storefinders.wp_go_maps import WpGoMapsSpider


class MochachosZASpider(WpGoMapsSpider):
    name = "mochachos_za"
    item_attributes = {
        "brand_wikidata": "Q116619117",
        "brand": "Mochachos",
    }
    allowed_domains = [
        "www.mochachos.com",
    ]

```

PR #9869, opened 2024-08-25
Add mochachos_za

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9809

### DIFF
--- a/locations/spiders/mochachos_za.py
+++ b/locations/spiders/mochachos_za.py
@@ -1,0 +1,12 @@
+from locations.storefinders.wp_go_maps import WpGoMapsSpider
+
+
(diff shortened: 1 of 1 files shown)


arrival-spring:
Great start, but this needs a bit more work.
- There are other brands here, see the locator page for details https://www.mochachos.com/find-a-store/ and use category field to distinguish 
- Some locations are definitely not in ZA but are getting ZA for country
- addr:full needs popping as it is just coordinates
- (Could be done later) use description for phone number(s) and address

CloCkWeRX:
mochachos spider already exists under wider name

PR #9896, opened 2024-08-26
Use proxy for SparZA

arrival-spring:
spar_za has had no results for the past few runs, and was showing 403 error in the logs, however it does run locally.


iandees:
Some of the location data isn't great, but most of it is good enough 👍🏻

PR #9899, opened 2024-08-26
[sos_br] Added Spider (59 items)

ckpiper:
Not quite sure if I should put this as a school or a business. It seems like they went to only teaching computer courses.


locations/spiders/sos_br.py
```diff
@@ -0,0 +1,39 @@
...
+            method="POST",
+            formdata={"isLead": "false", "isMapa": "true", "isTeste": "false"},
+        )
+
+    def parse(self, response: Response, **kwargs: Any) -> Any:
+        for location in response.json()["Unidades"]:
+            item = Feature()
+            item["extras"]["sources"] = [location]
+            item["ref"] = location["ID"]
+            item["name"] = location["NomeFantasia"]
```
CloCkWeRX, line 27: Worth adding anything to dict_parser for non english name variants?

ckpiper: I don't know. I feel like it does not happen a ton and could lead to other stuff getting mapped that should not.

Issue #9829, opened 2024-08-24
Bobablastic


CloCkWeRX:
### Brand name

Bobablastic

Chain of bubble tea food trucks in Oregon

### Wikidata ID

Q108499280
https://www.wikidata.org/wiki/Q108499280
https://www.wikidata.org/wiki/Special:EntityData/Q108499280.json

### Store finder url(s)

Official Url(s): https://bobablastic.com/
- [ ] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy links https://bobablastic.com/`
`pipenv run scrapy sitemap --pages https://bobablastic.com/`
`pipenv run scrapy sf --brand-wikidata=Q108499280 https://bobablastic.com/`
pipenv run scrapy sf https://bobablastic.com/all-locations/
```
from locations.storefinders.wp_store_locator import WPStoreLocatorSpider


class BobablasticSpider(WPStoreLocatorSpider):
    name = "bobablastic"
    item_attributes = {
        "brand_wikidata": "Q108499280",
        "brand": "Bobablastic",
    }
    allowed_domains = [
        "bobablastic.com",
    ]
```

PR #9906, opened 2024-08-27
Add bobablastic

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9829


CloCkWeRX:
This might actually be 50 locations. Storefinder is pretty broken

davidhicks (changes requested):
Suggestions for bringing this spider up to the proposed standard set in #9916:

* Specify `days=DAYS_EN`.
* Remove `time_format`.
* Add `"extras": Categories.FAST_FOOD.value` to `item_attributes`.

Have you tried also adding the following to check if more than 50 locations exist:
```
iseadgg_countries_list = ["US"]
search_radius = 100
max_results = 50
```

If you still see only 50 results with this approach, then seemingly there are indeed 50 stores.

CloCkWeRX:
Oops, lost a store in mexico

locations/spiders/bobablastic_us.py
```diff
@@ -0,0 +1,16 @@
...
+from locations.storefinders.wp_store_locator import WPStoreLocatorSpider
+
+
+class BobablasticUSSpider(WPStoreLocatorSpider):
+    name = "bobablastic_us"
+    item_attributes = {"brand_wikidata": "Q108499280", "brand": "Bobablastic", "extras": Categories.FAST_FOOD.value}
+    allowed_domains = [
+        "bobablastic.com",
+    ]
+    time_format = "%I:%M %p"
```
CloCkWeRX, line 12:
```suggestion
```

davidhicks (approved):

Issue #9222, opened 2024-08-03
Boboli (JSON)


CloCkWeRX:
### Brand name

Boboli

clothing design and manufacturing company

### Wikidata ID

Q39073733
https://www.wikidata.org/wiki/Q39073733
https://www.wikidata.org/wiki/Special:EntityData/Q39073733.json

### Store finder url(s)

Official Url(s): https://www.boboli.es/
- [x] I have run the below and if anything in generated, added it.
- [x] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q39073733 https://www.boboli.es/`
pipenv run scrapy sf https://www.boboli.es/es/content/81-nuestras-tiendas
pipenv run scrapy sf https://www.boboli.es/es/tiendas
```
```

CloCkWeRX:
curl 'https://www.boboli.es/es/tiendas?all=1' \
  -H 'accept: application/json, text/javascript, */*; q=0.01' \
  -H 'accept-language: en-GB,en;q=0.9,en-US;q=0.8,de;q=0.7,es;q=0.6,fr;q=0.5' \
[...]

PR #9921, opened 2024-08-27
Add Boboli 

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/9222 ?

https://www.boboli.es/es/boboli-store claims:
image

So I am not 100% what "store type: offline" means vs that

### DIFF
--- a/locations/spiders/bobali_es_it.py
+++ b/locations/spiders/bobali_es_it.py
@@ -5,7 +5,7 @@ from locations.categories import Categories
 from locations.dict_parser import DictParser
 
 
-class BoboliESIT(Spider):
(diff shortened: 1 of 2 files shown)


locations/spiders/bobali_es_it.py
```diff
@@ -0,0 +1,24 @@
+from scrapy import Spider
+from scrapy.http import JsonRequest
+
+from locations.categories import Categories
+from locations.dict_parser import DictParser
+
+
+class BoboliESIT(Spider):
```
CloCkWeRX, line 8:
```suggestion
class BobaliESITSpider(Spider):
```

PR #9927, opened 2024-08-28
[worldcat] remove spider

panasonic-mafia:
CDN blocks the spider, it should work with proxy.


iandees:
I think I'd rather remove worldcat: it is an aggregator and not a primary source, and the spider has to make a ton of requests. Using proxy means running the spider will be expensive. Have people found this to be useful data?

panasonic-mafia:
> I think I'd rather remove worldcat: it is an aggregator and not a primary source, and the spider has to make a ton of requests. Using proxy means running the spider will be expensive. Have people found this to be useful 

This is an aggregator indeed, agree to remove it.

Issue #923, opened 2018-11-07
Auchan Drive


maxerickson:
https://www.auchandrive.fr/

CloCkWeRX:
Is this the same as https://github.com/alltheplaces/alltheplaces/issues/924 ?

maxerickson:
It's a separate store format. I would guess they changed their website around in the meantime.

CloCkWeRX:
### Brand name

Auchan Drive

Click-and-collect branch of the French retailer Auchan

### Wikidata ID

Q2870659
https://www.wikidata.org/wiki/Q2870659
https://www.wikidata.org/wiki/Special:EntityData/Q2870659.json

### Store finder url(s)

Official Url(s): https://www.auchandrive.fr
- [x] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q2870659 https://www.auchandrive.fr`
```
```

----

iandees:
Closing — already covered. locations/spiders/auchan_fr.py crawls auchan.fr/nos-magasins?types=DRIVE and tags those locations with brand "Auchan Drive" / brand_wikidata Q2870659 specifically, distinct from the main hypermarket/supermarket brand tags.

PR #9936, opened 2024-08-29
[Auchan FR] Add spider

karen-kerr:


CloCkWeRX:
Does this cover... #923 ?

iandees:
I think they're separate things? Can't tell.

CloCkWeRX:
This one does seem to have some sub-branded auchan drive entries when I checked the map.
Whether it warrants a different wikidata ID or not, I'd probably lean towards no

CloCkWeRX:
(https://www.auchandrive.fr redirects to https://www.auchan.fr/)

PR #10021, opened 2024-08-31
Add GoReviewSpider storefinder and HungryLionZASpider and refactor existing spiders

### LINKED ISSUE #9778: New Storefinder: goreview
Storefinder is "goreview", marketing docs at
https://socialplaces.io/store-locator-local-pages/

_Originally posted by @CloCkWeRX in https://github.com/alltheplaces/alltheplaces/issues/9777#issuecomment-2305833368_

arrival-spring:
Such store finder pages have something like `https://hlinfo.goreview.co.za/store-locator/store-information` embedded in a the page and an icon on that page says "Powered by Social Places".

Fixes #9778

### DIFF
--- a/locations/storefinders/go_review.py
+++ b/locations/storefinders/go_review.py
@@ -47,9 +47,7 @@ class GoReviewSpider(CrawlSpider):
             ),
             "phone": response.xpath(
                 '//div[contains(@class, "content-wrapper")]/div[2]/div[3]//a[contains(@href, "tel:")]/@href'
-            )
-            .get()
-            .replace("tel:", ""),
+            ).get(),
             "facebook": response.xpath(
                 '//div[contains(@class, "content-wrapper")]/div[2]/div[5]//a[contains(@href, "facebook.com")]/@href'
             ).get(),
--- a/locations/storefinders/go_review.py
+++ b/locations/storefinders/go_review.py
@@ -3,14 +3,17 @@ import re
 from scrapy.linkextractors import LinkExtractor
 from scrapy.spiders import CrawlSpider, Rule
 
-from locations.hours import OpeningHours
+from locations.hours import DAYS_EN, OpeningHours
 from locations.items import Feature
+from locations.pipelines.address_clean_up import merge_address_lines
 
 
 class GoReviewSpider(CrawlSpider):
     """
     To use this spider, specify one or more start_urls,
-    normally something like "https://hlinfo.goreview.co.za/store-locator/store-information"
+    normally something like "https://hlinfo.goreview.co.za/store-locator"
+
+    Override days if something other than DAYS_EN is required
 
     Also known as socialplaces.io or Social Places
     """
@@ -22,6 +25,7 @@ class GoReviewSpider(CrawlSpider):
             callback="parse",
         )
     ]
+    days = DAYS_EN
 
     def parse(self, response):
         properties = {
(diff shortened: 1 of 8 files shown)


locations/storefinders/go_review.py
```diff
@@ -0,0 +1,69 @@
...
+    ]
+
+    def parse(self, response):
+        properties = {
+            "ref": re.sub(r"\.goreview\.co\.za.*", "", re.sub(r"https:\/\/", "", response.url)),
+            "branch": response.xpath('//div[@class="left-align-header"]/h2/text()')
+            .get()
+            .replace(self.item_attributes["brand"], "")
+            .strip(),
+            "addr_full": re.sub(
```
davidhicks, line 33: This should use `locations.pipelines.address_clean_up.clean_address` instead of reinventing the address segment joining/clean up.

locations/storefinders/go_review.py
```diff
@@ -0,0 +1,69 @@
...
+            "website": response.url,
+        }
+        if maps_links_js := response.xpath('//script[contains(text(), "#apple_maps_directions")]/text()').get():
+            if "&sll=" in maps_links_js:
+                properties["lat"], properties["lon"] = maps_links_js.split("&sll=", 1)[1].split("&", 1)[0].split(",", 2)
+        hours_string = " ".join(
+            response.xpath('//div[contains(@class, "content-wrapper")]/div[2]/div[4]/p//text()').getall()
+        )
+        properties["opening_hours"] = OpeningHours()
+        properties["opening_hours"].add_ranges_from_string(hours_string)
```
davidhicks, line 63: Is it possible for the opening hours to be specified in a language other than English in the future? Should there be a storefinder attribute `days = DAYS_EN` and then use `properties["opening_hours"].add_ranges_from_string(hours_string, days=self.days)` here to allow possible override of opening hours language?

davidhicks (changes requested):

locations/storefinders/go_review.py
arrival-spring: Yep, done

locations/storefinders/go_review.py
arrival-spring: Could be possible, done

davidhicks (approved):

arrival-spring:
There are several other countries that have Hungry Lion,  with exactly the same style website. Is it preferable to add them all into one spider, with a list of URLs, or as separate spiders?

iandees:
> Is it preferable to add them all into one spider, with a list of URLs, or as separate spiders?

If they're exactly the same then you can put them in the same spider. If there's different logic for extracting information from the layout of the page then use separate spiders.

davidhicks:
If they all share the same domain `hlinfo.goreview.co.za` then you can put them in one spider. My preference is to split them though so statistics more readily report on which countries a brand is expanding and contracting within. If the brand uses different domains for each country, create new spiders, that way the spiders can run in parallel.

CloCkWeRX:
TODO: Subdomain search to find any others for *.goreview.co.za. Google yields:

* https://ob.goreview.co.za/
* https://spur41.goreview.co.za/
* https://spur1.goreview.co.za/

CloCkWeRX:
Split off new issue for researching other storefinders/refactoring opportunities.
After looking, I think keeping these all as individual = best; as customers will come and go from this organisation.

arrival-spring:
I've put them as separate ones, but happy to change to a list of start_urls in one spider if that is preferred. Each one has a different prefix, e.g.`https://hlninfo.goreview.co.za/store-locator` vs `https://hlinfo.goreview.co.za/store-locator`

PR #10026, opened 2024-08-31
Add DunnsZASpider

arrival-spring:


CloCkWeRX:
Second one of this store locator recently, should see if we can find out more and add auto detection.

arrival-spring:
Marketing info here https://amai.com/apps/promap-store-locator/

PR #10039, opened 2024-09-01
DoppioZeroSpider (28 locations)

arrival-spring:
Waiting for NSI update for category

I've also included some additions to the email_clean_up to allow for multiple email addresses and to save stripping "mailto:" from extracted links. If that's ok, we could then go through and tidy such things off other spiders.

### DIFF
--- a/locations/pipelines/email_clean_up.py
+++ b/locations/pipelines/email_clean_up.py
@@ -1,19 +1,22 @@
 class EmailCleanUpPipeline:
     def process_item(self, item, spider):
-        email = item.get("email")
+        emails = item.get("email")
 
-        if not email:
+        if not emails:
             return item
 
-        if not isinstance(email, str):
+        if not isinstance(emails, str):
             spider.crawler.stats.inc_value("atp/field/email/wrong_type")
             return item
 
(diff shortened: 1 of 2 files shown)


locations/pipelines/email_clean_up.py
```diff
@@ -1,19 +1,22 @@
...
+        if not emails:
             return item
 
-        if not isinstance(email, str):
+        if not isinstance(emails, str):
             spider.crawler.stats.inc_value("atp/field/email/wrong_type")
             return item
 
-        item["email"] = self.normalize(email, spider)
+        for email in emails.split("; "):
```
CloCkWeRX, line 12: Is it better to prefer emails[0] or first valid email here, on the assumption people would write the "main" email first?
(OSM only really having 1x email attribute, I think)

arrival-spring: I did this because it seemed there was a possibility of multiple emails here, as there were multiple phone numbers. In reality, all of these stores have only one email. 
We allow for multiple phone numbers, in a similar fashion, so I thought that allowing for multiple emails made sense. OSM has a small amount of usage of semicolon separated emails.

PR #10040, opened 2024-09-01
Add OldMutualZASpider

arrival-spring:
Needs NSI update to sort out categories

Changes in locations/hours feel like it would be better done using `time_format` somehow, but this worked for now.


Firefishy:
This probably should be updated to dynamically get the url of the asset file.

It has changed to: https://www.oldmutual.co.za/om-assets/js/page--src--templates--branch-locator-page-vue.75ee934b.js

arrival-spring:
Very good spot, thank you very much

PR #10042, opened 2024-09-01
[DominosPizza GB] Improve regex

TheUKHighStreet:


TheUKHighStreet:
I don't really understand yet how this is creating the stores without a yield. 
However, there is a problem with the original regex as it is missing some stores which have this structure with an extra /:
 <loc>https://www.dominos.co.uk/pizza-near-me/welling/9/9b-upper-wickham-lane</loc>

arrival-spring:
There is no yield because it is using StructuredDataSpider and the default methods for that do not need overriding here, assuming all data is being returned as expected.

TheUKHighStreet:
Thanks - I guessed it was something like that. 
I'm not sure why it is timing out but I assume it doesn't have anything to do with the regex. There seem to be 28 stores which were being missed with the original file.

arrival-spring:
I had the same query about timeout when I started contributing, maybe we should put it in the docs somewhere. Basically though the spiders are only given a minute to run in the PR tests, so timeout is fine for a status here if it runs ok locally.

TheUKHighStreet:
Thanks very much, good to know.

CloCkWeRX:
https://github.com/alltheplaces/alltheplaces/blob/5dfc5bee0ab432e46f9b4a67b86bd95a53904d57/locations/structured_data_spider.py#L301 if you are curious about where it's working from - sitemap > parse_sd > yield from post process item.

TheUKHighStreet:
Thanks very much.

PR #10104, opened 2024-09-04
[TESTS] speed up item_attributes test

panasonic-mafia:
Improved the test so it's not looping over all files for each spider.
Also wrapped iteration of spider classes for all modules into a function.

### DIFF
--- a/tests/test_item_attributes.py
+++ b/tests/test_item_attributes.py
@@ -1,15 +1,10 @@
-from scrapy.utils.misc import walk_modules
-from scrapy.utils.spider import iter_spider_classes
-
-from locations.settings import SPIDER_MODULES
+from locations.exporters.geojson import iter_spider_classes_in_all_modules
 
 
 def test_item_attributes_type():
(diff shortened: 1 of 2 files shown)


tests/test_item_attributes.py
```diff
@@ -1,17 +1,15 @@
...
-    for name in process.spider_loader.list():
-        spider = find_spider_class(name)
-        item_attributes = getattr(spider, "item_attributes", {})
-        assert isinstance(item_attributes, dict)
-
-        if extras := item_attributes.get("extras"):
-            assert isinstance(extras, dict)
+    for mod in SPIDER_MODULES:
+        for module in walk_modules(mod):
+            for spider_class in iter_spider_classes(module):
```
panasonic-mafia, line 10: Wrap in a function?

panasonic-mafia: Wrapped to a separate function. Successfully tested on a small spider - `dataset_attributes` are applied (as expected).

PR #10116, opened 2024-09-04
Add autodetection to Algolia storefinder

impiaaa:


locations/storefinders/algolia.py
```diff
@@ -27,6 +28,18 @@ class AlgoliaSpider(Spider):
     index_name: str = ""
     referer: str | None = None
```
davidhicks, line 29: Does `referer` need to be extracted too via `DetectionRequestRule(headers=...)` or can this be left as None in most cases?

davidhicks (approved):

locations/storefinders/algolia.py
davidhicks: Thanks for confirming! :)

PR #10119, opened 2024-09-05
[Giant Eagle] decompose results into the correct branded outlets

mjoe999:
The situation on the ground is more complex than the original spider assumed. _GetGo_ and _Market District_ are quite significant brands in their own right.

It might be that extra work to pull out Giant Eagle pharmacies at various locations would be useful but not done this time.


locations/spiders/giant_eagle_us.py
```diff
@@ -0,0 +1,113 @@
...
+TYPE_CODE_MAPPING = {
+    "CV": Categories.SHOP_CONVENIENCE,
+    "CW": Categories.CAR_WASH,
+    "FU": Categories.FUEL_STATION,
+    "SM": Categories.SHOP_SUPERMARKET,
+}
+
+
+class GiantEagleUSSpider(Spider):
+    name = "giant_eagle"
```
CloCkWeRX, line 25:
```suggestion
    name = "giant_eagle_us"
```

PR #10124, opened 2024-09-05
Refactor BabyCityZA to AmastyStoreLocator remove DialABedZA

arrival-spring:
As noted in the info for AmastyStoreLocatorSpider the popup_html can be very variable, but it sometimes has phone/email links in, so let's try to get those and distinguish them from a store link.


CloCkWeRX:
Is the existing baby_city.py different?

CloCkWeRX:
Does bash_za significantly overlap dialabed?

arrival-spring:
Not sure how I missed the existing baby city, it is the same and is currently broken due to robots.txt. I'll fix it

As for dial a bed, I'll have a look

arrival-spring:
Dial a bed is covered by bash_za, although data appears to be slightly different. Latest run had 81 DaB in bash, 78 here. Address info is also slightly different, but more detailed in bash.

PR #10132, opened 2024-09-05
Refactor BrowArt23USSpider to JSONBlobSpider and collect opening hours

arrival-spring:
When looking at #10131 I saw this which is basically the same, so have neatened things up a bit.

Not sure if this counts as a storefinder, or what it is?


iandees:
I would say deduplication/DRYing should happen when we find three spiders doing the same thing.

CloCkWeRX:
yith_sl_get_results suggests https://docs.yithemes.com/yith-store-locator-for-wordpress/page-layout/store-locator-page/

PR #10169, opened 2024-09-08
alnatura_de: fix spider

davidhicks:


iandees:
The output doesn't include lat/lon, but there's code in there for it. Does something need to be fixed?

davidhicks:
> The output doesn't include lat/lon, but there's code in there for it. Does something need to be fixed?

Thanks for picking this up. I hadn't realised the coordinates use a comma not a period. Fixed now

PR #10178, opened 2024-09-09
Add Astron Energy spider (338 locations)

arrival-spring:
Also adjust dict_parser to get geometry from geojson files


locations/dict_parser.py
```diff
@@ -246,25 +246,28 @@ def parse(obj) -> Feature:
...
-                "position",
-                "display-coordinate",
-            ],
-        )
-        # If not a good location object then use the parent
-        if not location or not isinstance(location, dict):
-            location = obj
-        item["lat"] = DictParser.get_first_key(location, DictParser.lat_keys)
-        item["lon"] = DictParser.get_first_key(location, DictParser.lon_keys)
+        if obj.get("geometry") and obj["geometry"].get("type") == "Point":
```
iandees, line 249: We support non-point geometries, too.

arrival-spring: Thanks, fixed it now

PR #10186, opened 2024-09-09
Add Toyota SACU (264 items) and NZ spiders (64 items)

arrival-spring:
SACU = South African Customs Union, i.e. ZA, BW, NA, LS, SZ

Also refactor to use shared attributes, except for toyota_us which had been failing, so will be fixed in a separate PR

### DIFF
--- a/locations/spiders/toyota_au.py
+++ b/locations/spiders/toyota_au.py
@@ -3,10 +3,12 @@ from scrapy import Spider
 from locations.categories import Categories, apply_category
 from locations.dict_parser import DictParser
 
+TOYOTA_SHARED_ATTRIBUTES = {"brand": "Toyota", "brand_wikidata": "Q53268"}
+
 
 class ToyotaAUSpider(Spider):
     name = "toyota_au"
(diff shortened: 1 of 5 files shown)


arrival-spring:
I didn't see that toyota_us had been failing before anyway, I'll have a look tomorrow.

arrival-spring:
This would now work, but there must be a better way to scrape the US.

The search page is here: https://www.toyota.com/dealers/
You can search by zip code, name or city/state (not by state alone) and you get the 50 nearest results from their API.

From https://www.toyota.com/dealers/directory/ you can go to a state page, e.g. https://www.toyota.com/dealers/connecticut/, which gives a list of cities for that state. Thing is, there are a total of 29,500 cities on all of those pages and I think only about 1500 dealers in the whole US, so not a good ratio of page requests to items, I think.

Can someone more familiar with the US give some advice? A subset of zipcodes? A minimum population of city and then using the geo library?

arrival-spring:
I've split off the changes to toyota_us, will open a new PR for that

PR #10192, opened 2024-09-10
[Heron Foods GB] Collect opening hours and more attributes

Cj-Malone:
I've also tested all `WordpressHeronFoodsSpider` spiders.

### DIFF
--- a/locations/spiders/heron_foods_gb.py
+++ b/locations/spiders/heron_foods_gb.py
@@ -1,5 +1,12 @@
-from locations.hours import OpeningHours
+from typing import Iterable
+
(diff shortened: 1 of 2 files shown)


TheUKHighStreet:
I don't think this spider is getting all the Heron Foods branches. For example this one is missing: https://heronfoods.com/store/longbenton/ 
I've tried testing by changing the coordinates and the radius but I always get 318 results. 
I think this might be a limit on the number of results returned.
So it might be possible doing 2 calls to the ajax file with different coords and merging the results? - but I'm not sure how to do this.

PR #10216, opened 2024-09-11
[cinnabon_ru] add spider (113 locations)

panasonic-mafia:


panasonic-mafia:
My bad completely overlooked the error, thanks for the fix @iandees.

PR #10217, opened 2024-09-12
Update GBFS: extend and standardize tagging, use info from extra feeds

impiaaa:
* Simplify brand mapping: only include `brand` and `brand_wikidata`, and use a dictionary instead of a list of names
* Use system ID instead of name for brand mapping
* Add more brands by name-matching to NSI (feel free to double-check these)
* Put the system name in `network` instead of `brand`—one brand can have many systems
* Use language suffixes for `network`, `network:short`, `operator`, `name`, and `short_name`, where available
* Get `network`, `opening_hours`, `network:short`, and `operator` from the `system_information` feed where available
* Where available, use the `vehicle_types` feed to determine feature categories
* Set `opening_hours`, `payment:*`, `geometry`, `parking`, `phone`, and `website` for each station
* Use `bicycle_rental=docking_station` for non-virtual stations instead of `physically_present=no` for virtual stations
* Specify capacity per vehicle type where available


iandees:
I agree with pre-commit: this is getting pretty complex. Can you please try splitting this up into smaller functions, reduce the depth of the if statements, etc.

locations/categories.py
```diff
@@ -478,16 +478,14 @@ def apply_category(category: Mapping | Enum, item: Feature | dict) -> None:
...
-                continue
-            existing_values.append(value)
-            existing_values.sort()
-            item["extras"][key] = ";".join(existing_values)
+            existing_values = set(item["extras"][key].split(";"))
+            existing_values.update(values)
+            item["extras"][key] = ";".join(sorted(existing_values))
         else:
-            item["extras"][key] = value
+            item["extras"][key] = values_str
```
Cj-Malone, line 488: This both doesn’t work and isn't good.

Cj-Malone: `        values = set(values_str.split(";"))` is broken.

Cj-Malone: `cuisine` isn't a category, and shouldn't be added using `apply_category`. `amenity=parking` is a "category", `amenity=parking_space` is a "category". `amenity=parking;parking_space` is dog shit.

Cj-Malone: "category"/"feature tags"/"top level tags" aren’t lists. No OSM consumer consumes them as lists, almost no OSM data has them as lists.

Cj-Malone: `cuisine` is a list.

impiaaa: I don't understand why you're commenting here, the behavior didn't change. Maybe when I left this project I should have closed all of my PRs so I don't get emails about how my 2-year-old code is "dog shit."

Cj-Malone: I was excessively annoyed, sorry. FWIW "`amenity=parking;parking_space` is dog shit." wasn't you either.

iandees: Sorry all. My cavalier PR merging the other day broke some stuff for people.

davidhicks: @Cj-Malone how was `amenity=parking;parking_space` appearing for you? I'm guessing it was observed in weekly crawl output? Also guessing the spider might set `Categories.PARKING` and features were getting matched to an NSI entry for `amenity/parking_space` instead, thus when NSI tags are enriched on the item, it ends up with `amenity=parking;parking_space` as a combination of `Categories.PARKING` and NSI matching appending `amenity/parking_space`? If I'm right here, I think this problem could potentially now be fixed with merge of PR #11672 which ensures NSI matching now takes categories into account. No more matching ATP waste baskets to NSI drinking fountains just because they're both operated by the same local government body.

PR #10227, opened 2024-09-12
remove call center phones in card_factory_gb

matkoniecz:


Cj-Malone:
The premise here is false. +443 may or may not be a call centre prefix, but these are unique phone numbers for the POI. It would be like dropping `website` values because it's hosted by AWS.

`$ curl "https://alltheplaces-data.openaddresses.io/runs/2024-09-07-13-32-11/output/card_factory_gb.geojson" | jq '.features[].properties.phone | select(.) | select(. | startswith("+44 3"))' | sort | uniq -c | sort -n`

PR #10236, opened 2024-09-13
[toyota_tw] add 333 locations

mjoe999:
for @arrival-spring

### DIFF
--- a/locations/spiders/toyota_tw.py
+++ b/locations/spiders/toyota_tw.py
@@ -1,0 +1,26 @@
+import scrapy
+
+from locations.categories import Categories, apply_category
+from locations.dict_parser import DictParser
+
+
+class ToyotaTWSpider(scrapy.Spider):
+    name = "toyota_tw"
+    item_attributes = {
+        "brand": "Toyota",
+        "brand_wikidata": "Q53268",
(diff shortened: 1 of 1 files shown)


locations/spiders/toyota_tw.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class ToyotaTWSpider(JSONBlobSpider):
+    name = "toyota_tw"
+    item_attributes = ToyotaUSSpider.item_attributes
+    start_urls = ["https://www.toyota.com.tw/api/location.ashx"]
+    locations_key = "DATA"
+
+    def post_process_item(self, item, response, location: dict):
+        item["website"] = "https://www.toyota.com.tw/"
```
arrival-spring, line 13: I wouldn't include this same website on all locations, if they have no individual website then just leave it out. Instead, make sure it is on Wikidata, with relevant "applies to jurisdiction".

mjoe999: Will do.

mjoe999: ... Toyota wikidata page is semi-protected and I do not yet have enough kudos points :-(

PR #10242, opened 2024-09-13
[starbucks_za] Add spider

arrival-spring:


arrival-spring:
No output because we have to scrape by tight lat, lon pairs and many locations give no results. I didn't run again in full after adding some extra locations, but at least 32 locations.

API appears to give results within approx a 20km radius, up to a maximum of 50 results, so necessarily we miss some unfortunately (same for starbucks_eu).

PR #10259, opened 2024-09-14
[Dreams GB] Fix spider

rjw62:
The dreams_gb spider is currently not returning any results. I don't know if this will fix everything (I'm unable to test code locally), but certainly the LinkedData type that the Dreams website uses has recently changed from LocalBusiness to FurnitureStore.


CloCkWeRX:
If you feel up to it, you could add 1-2 more rules for 
1. Only follow /stores/* urls
2. Extend structuredataspider and specify a wanted_types of FurnitureStore

Cj-Malone:
No coords on the pages

Issue #10241, opened 2024-09-13
Invalid coordinates should make PR checks fail


arrival-spring:
> Added in #9953
> 
> I think this shows an issue with the checking, as the logs on that build show an error but it still says that it was a success.
> 

_Originally posted by @arrival-spring in https://github.com/alltheplaces/alltheplaces/issues/10231#issuecomment-2347979471_

CloCkWeRX:
I feel like the changes to CI surface issues much better now, closable?

iandees:
I was going to add one more check for *invalid* coordinates that trigger an error and a check for the shape of geometry and/or lat/lon before closing this.

iandees:
🤖 This was implemented in #14845 — `ci/run_pr_spiders.sh` now checks the `atp/field/geometry/invalid` stat and fails the PR check when any items have invalid geometry.

PR #10297, opened 2024-09-15
Spider CI improvements

iandees:
Checks the stats JSON for warnings and errors, then includes those in the comment that gets posted.


iandees:
This is designed to help with #10241

PR #10340, opened 2024-09-17
Update omegawatches.py

TheUKHighStreet:
The json was not being parsed correctly.


TheUKHighStreet:
Wanted to check this works. This is all the shops selling Omega watches I think - just needs limiting to the actual Omega shops.

PR #10343, opened 2024-09-17
[afgri_bw_na_za_zw] Add spider

arrival-spring:
Awaiting NSI update for categories


iandees:
Can you link to the NSI PR?

arrival-spring:
Straight commit rather than merge, but it's here: https://github.com/osmlab/name-suggestion-index/commit/4215854f3fbf1cb2347feabb29918673d9b23e32

At that point I hadn't seen that the AU version uses shop=tractor, so I thought agrarian was the best fit for agricultural equipment.

Issue #10032, opened 2024-09-01
Amai Promap store locator


CloCkWeRX:
Marketing info here https://amai.com/apps/promap-store-locator/

_Originally posted by @arrival-spring in https://github.com/alltheplaces/alltheplaces/issues/10026#issuecomment-2323147454_

https://github.com/search?q=repo%3Aalltheplaces%2Falltheplaces%20amaicdn.com&type=code

Indicator:
Request to https://amaicdn.com/storelocator-prod/wtb/*.js

### Others

- [x] https://github.com/alltheplaces/alltheplaces/pull/10913
- [x] https://github.com/alltheplaces/alltheplaces/pull/10912
- [x] https://sl.amaicdn.com/

Via publicwww

### Manufacturers listing resellers

- [x] gestalten.com
- [x] https://beabongiasca.com/pages/boutique ?
- [x] http://discountair1.com/
- [x] http://warmiesshop.com/
- [x] https://ajstuarts.com/
- [x] https://americanmetabolix.com/
- [x] https://biorb.com/
- [x] https://bolfoods.com/
- [x] https://bonanzasatrangi.com/
- [x] https://bootjack.com/
- [x] https://codina.fr/
- [x] https://cult-crew.myshopify.com/
- [x] https://evolutionoutdoor.com/
- [x] https://fila.com.ph/
- [x] https://floafers.com/
- [x] https://granitegold.com/
- [x] https://havit.com.au/
- [x] https://imageskincare.co.uk/
- [x] https://iroirocolors.com/
- [x] https://janeiredale.co.uk/
- [x] https://lxmi.com/
- [x] https://maternatura.it/
- [x] https://mjarsenal.com/
- [x] https://myspringenergy.com/
- [x] https://pinktartan.ca/
- [x] https://proraso-usa.com/
- [x] https://redmond.life/
- [x] https://rieker.ca/
- [x] https://shopbioseaweedgel.com/
- [x] https://shoppeninsulas.com/
- [x] https://sippify.com.au/
- [x] https://smashtess.com/en-gb
- [x] https://solarismexico.com/
- [x] https://sottotono.it/
- [x] https://store.vermontsmokeandcure.com/
- [x] https://suterproducts.com/
- [x] https://thenormalbrand.com/
- [x] https://vegepod.com/
- [x] https://waxrx.com/
- [x] https://www.diffeyewear.com/
- [x] https://www.gigi-fratelli.com/en-gb
[...]

PR #10350, opened 2024-09-18
Add Amai promap auto detection (autospidergen)

CloCkWeRX:
Part of #10032 

- Recently changes from master
- 91685a4b1e8c87986989cb63471fdf858362bfde is the main bit of this PR

@arrival-spring - optionally, we could take out your extract js URL code from the storelocator page for the spiders and just have this set start_urls directly. Any strong opinions?


CloCkWeRX:
@davidhicks

arrival-spring:
Very strong opinions in fact, haha, see #10253 where I had to add the fetching js code. It seems that the js URL changes, maybe when they add/change stores behind the scenes.

CloCkWeRX:
Might be cleaner just to cherry pick https://github.com/alltheplaces/alltheplaces/commit/91685a4b1e8c87986989cb63471fdf858362bfde

PR #10357, opened 2024-09-18
name_suggestion_index: add type hints and simplify normalisation

davidhicks:


locations/name_suggestion_index.py
```diff
@@ -185,93 +190,3 @@ def generate_keys_from_nsi_attributes(nsi_attributes: dict) -> tuple[str, str] |
 
         class_name = f"{class_name}Spider"
         return (key, class_name)
-
-    @staticmethod
-    def normalise(s):
```
Cj-Malone, line 190: This was supposed to match NSI or we will encounter weird bugs.

https://github.com/osmlab/name-suggestion-index/blob/main/lib/simplify.js

davidhicks: Ahh, we weren't doing the same as NSI simplify.js before either.

I've had a quick look and think the key differences we now have are:

1. We don't replace ampersand with "and".
2. We leave underscores in place whereas NSI simplify removes them.
3. npm's `diacritics` is seemingly not as thorough as the python `unidecode` at replacing diacritic characters (we're probably doing a better job and meeting the intent of NSI's simplify.js better).

I can add the first two changes back in and reference NSI's simplify.js in the documentation of this code in ATP.

davidhicks: Should be fixed for the most part in PR #10525 for your review

PR #10363, opened 2024-09-19
[tesco_eu] fix spider

panasonic-mafia:
Tesco revamped their EU sites to use different APIs etc. old API still works, but we lost individual links. I'll comeback to this spider once I figure out auth on the website.


panasonic-mafia:
Ok, it works with proxy.

PR #10366, opened 2024-09-19
[LinkedDataParser] More case insensitive and move linked data opening hours parsing

arrival-spring:
I spotted in #10355 that a lot less data was being parsed than was available. This was because of every item in the linked data starting with a capital letter where we expected lowercase.

To fix this I also had to move the opening hours parsing into linked_data_parser, but that was a TODO in the code anyway.

Therefore I had to make changes in any spider using OpeningHours to parse the linked data hours. A few of those spiders were not working before I touched them, but I felt it was out of scope to investigate and fix them now.

Finally, a bit of tidying on speedy_services_gb and proving that it all now works.


arrival-spring:
Confirm, fine_wine_good_spirits.py was failing before this change anyway, I wasn't planning to investigate that here

PR #10367, opened 2024-09-19
Refactor all Five Guys Spiders to Yext, collect more attributes through YextAnswersSpider

arrival-spring:
Several of the Five Guys spiders have been broken for a while, however they all use Yext, mostly answers and a couple using search.

All now work and there's more data that we can now parse as well.


arrival-spring:
AWS: "refusing to run on more than 15 spiders"

Do you want me to post results of running these locally, or something else?

PR #10379, opened 2024-09-20
Remove dead spider: nandos_pk

### LINKED ISSUE #9533: Nando's Pakistan spider broken
### Spider name

nandos_pk

### Log output

_No response_

### Backtrace (if applicable)

_No response_

No results

_No response_

### Possible causes?

Storefinder page is currently returning an error. They may fix it or change to the same interface as ZA.

https://www.nandos.pk/eat/restaurants

CloCkWeRX:
Fix #9533 temporarily

https://www.nandos.pk/eat/restaurants - broken
image


arrival-spring:
See #9533

I wondered if they would fix it

PR #10396, opened 2024-09-21
Rewrite Sbarro as StructuredData, Crawl

CloCkWeRX:


iandees:
There's lat and lon on the initial page. Can we pull location from there?

CloCkWeRX:
Yeah, let me have a go at it.

CloCkWeRX:
After looking more, because this does the trick to pass the coordinates through the request.meta; there's little saving in refactoring it.

PR #10411, opened 2024-09-21
Refactor lowes_foods to SitemapSpider

CloCkWeRX:


CloCkWeRX:
```
ValueError: time data 'AM' does not match format '%I:%M%p'
```
However, I didn't really change that.

image

PR #10431, opened 2024-09-21
Remove dead spider: Burger King RU

CloCkWeRX:
5 failed runs
https://alltheplaces-data.openaddresses.io/runs/2024-09-14-13-32-11/logs/burger_king_ru.txt

Website looks like it has one location only:
> Moscow, Arbat street, 29


panasonic-mafia:
There are many restaurants on https://burgerkingrus.ru/restaurants, proxy just can not go through their q-rator block. Not sure if im ok with removal of the spider as it might work on some environments.

iandees:
We should add it back if we can find a way to get results.

PR #10478, opened 2024-09-22
brandywine_living finished with a 404 (imperva bot protection) - Trial with proxy

CloCkWeRX:


iandees:
Looks like they might have changed their page structure. It should be starting with https://www.brandycare.com/company/locations.

CloCkWeRX:
Sadly I can't see it to check

PR #10485, opened 2024-09-22
Attempt to fix cava

CloCkWeRX:


CloCkWeRX:
@iandees - keep? Remove? I can't decide

iandees:
The proxy fetch is working, but the spider is broken. I can look at it a bit tomorrow if it's not working for you.

PR #10503, opened 2024-09-22
Fix domain for us_national_guard

CloCkWeRX:
TBA: What's the best category? Old spider did not have it


iandees:
Maybe https://wiki.openstreetmap.org/wiki/Tag:military%3Doffice

CloCkWeRX:
https://nationalguard.com/colorado - displays as both recruitment stations and armories.

iandees:
Yea, that gets tricky. I would argue that `military=office` is close enough for the whole thing and it's probably not worth trying to detect special cases like armories.

PR #10506, opened 2024-09-22
Fix sweetgreen_us domain

CloCkWeRX:
Arguably this could also crawl https://www.sweetgreen.com/locations ; but individual pages don't have coordinates.

The request to hit ratio of this is pretty low; I'm up to 194 pages/0 hits.


CloCkWeRX:
TODO: Check if we are best off just hardcoding:
```
 "variables": {
                                "topLeft": {
                                    "latitude": "45.08900033639594",
                                    "longitude": "-127.4853609375"
                                },
                                "bottomRight": {
                                    "latitude": "28.14945921538994",
                                    "longitude": "-63.94043906250002"
                                },
```

PR #10538, opened 2024-09-23
Update matalan_gb.py

TheUKHighStreet:


TheUKHighStreet:
I still need to add the latitude and longitude.

locations/spiders/matalan_gb.py
```diff
@@ -1,16 +1,33 @@
...
+        ld_item = json.loads(ld)
 
-    def parse_item(self, item, location):
-        if not location.get("c_open_for_shopping"):
-            return
-        if not location["c_open_for_shopping"].get("availability"):
-            return
-        item["website"] = "https://store.matalan.co.uk/" + location["slug"]
-        item.pop("twitter")
+        item = LinkedDataParser.parse_ld(ld_item, time_format="%H:%M:%S")
```
Cj-Malone, line 27: You already have `ld_data`. But you don't need it because the opening hours are parsed anyway.

```suggestion
```

TheUKHighStreet: Thanks - I got in a mess because I didn't realise that it would automatically parse the opening hours from the json so I was trying to do it myself before I realised. And the store data is on the page in two different places in 2 different formats and the latitude is only in the second.

CloCkWeRX:
Ah, is this just an issue with recent Yext regressions, or a long term one?

locations/spiders/matalan_gb.py
```diff
@@ -1,16 +1,27 @@
...
     item_attributes = {"brand": "Matalan", "brand_wikidata": "Q12061509"}
-    api_key = "50f6010c48792a06524b4fe3471d7840"
+    allowed_domains = ["www.matalan.co.uk"]
+    start_urls = ["https://www.matalan.co.uk/stores/uk"]
+    rules = [
+        Rule(LinkExtractor(allow=r"^https://www.matalan.co.uk/stores/uk/[-\w]+$")),
+        Rule(LinkExtractor(allow=r"^https://www.matalan.co.uk/stores/uk/[-\w]+/[-\w]+$")),
+        Rule(LinkExtractor(allow=r"^https://www.matalan.co.uk/store/[-\/\w]+$"), "parse_sd"),
+    ]
+    download_delay = 0.5
```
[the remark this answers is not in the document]
TheUKHighStreet: Sorry - learning by looking at other spiders and not sure yet what is good practice. Have removed.

locations/spiders/matalan_gb.py
TheUKHighStreet: Right - finally think I understand what is going on and how simple this actually is. Have updated. Thanks for the pointers.

locations/spiders/matalan_gb.py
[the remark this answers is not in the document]
iandees: No problem! Thanks for the addition.

iandees (approved):

PR #10546, opened 2024-09-23
[gov_dfe_gias_gb] Try fix

arrival-spring:
Despite working in #10254 it didn't work in the weekly run. I suspect this is because it ran early in the morning and maybe the file wasn't there or wasn't ready. So try getting the previous day's file instead.


arrival-spring:
Websites are user entered, many are without the scheme

Anything outside the UK doesn't get coordinates because it uses the OS national grid, but there are international schools registered through the UK government.

Issue #9735, opened 2024-08-21
Sweaty Betty (EU, US, UK)


CloCkWeRX:
### Brand name

Sweaty Betty

British retailer specialising in women's activewear

### Wikidata ID

Q7654224
https://www.wikidata.org/wiki/Q7654224
https://www.wikidata.org/wiki/Special:EntityData/Q7654224.json

### Store finder url(s)

Official Url(s): https://www.sweatybetty.com
- [ ] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q7654224 https://www.sweatybetty.com`
```
```

davidhicks:
Crawl location pages from https://www.sweatybetty.com/shop-finder

Then on each location page, find the following JavaScript object defined in amongst a lot of other junk:
```
[...]

PR #10547, opened 2024-09-23
[Sweaty betty] add spider

TheUKHighStreet:


TheUKHighStreet:
The opening hours are included in the json as escaped html but I'm not quite sure how to parse them.
I hadn't added this spider because there aren't any coordinates - but I've seen that there are other spiders without coordinates - so is that ok?

CloCkWeRX:
#9735

CloCkWeRX:
Yup, missing coordinates is a bit annoying but still okay

CloCkWeRX:
Also, https://github.com/search?q=repo%3Aalltheplaces%2Falltheplaces%20mobify&type=code - is this similar to the existing/a new Storefinder?

iandees:
I'm going to merge this, but we should definitely look to see if Mobify can be extracted as a store locator.

PR #10549, opened 2024-09-23
[tmobile_cz] Add spider (102 locations)

petrzjunior:
Parsed from [map](https://coveragemap-tmcz.position.cz/#el=TMCZ:Store@pos=49.854414;15.567853@sc=6200000) API, all branches are visible is the screen size is wide enough.
No good "name" field available, I constructed the name from address. Website for each branch not available.
Opening hours parsed correctly.

### DIFF
--- a/locations/spiders/tmobile_cz.py
+++ b/locations/spiders/tmobile_cz.py
@@ -1,0 +1,36 @@
+import json
+import re
+
+from locations.categories import Categories
+from locations.hours import DAYS_CZ, OpeningHours
+from locations.json_blob_spider import JSONBlobSpider
+
+
+class TmobileCZSpider(JSONBlobSpider):
+    name = "tmobile_cz"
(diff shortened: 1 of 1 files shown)


locations/spiders/tmobile_cz.py
```diff
@@ -0,0 +1,36 @@
...
+            yield data
+
+    def pre_process_data(self, feature: dict):
+        feature["street_address"] = feature["address"]
+        del feature["address"]
+        feature["city"] = feature["a_city"]
+        feature["postcode"] = feature["a_psc"]
+
+    def post_process_item(self, item, response, feature: dict):
+        item["name"] = "T-Mobile {}, {}".format(feature["street_address"], feature["city"])
```
iandees, line 31: Generally we don't create our own `name`. It's ok to leave it unset if there isn't a specific branch name in the data.

iandees (changes requested):

iandees (approved):

PR #10492, opened 2024-09-22
Remove duplicate spider: fairway_market 

CloCkWeRX:
fairway_market_us exists; this doesn't run

Fix #2482

PR #10551, opened 2024-09-24
Add back Fairway Market

iandees:
Reverts alltheplaces/alltheplaces#10492


iandees:
The content on this page is quick hard to extract. I'm going to set this down for now. Anyone is welcome to give this a try!

PR #10555, opened 2024-09-24
Create toys_r_us_gb.py

TheUKHighStreet:


TheUKHighStreet:
These are only inside WHSmith stores - so don't think you actually want them.

arrival-spring:
If it still counts as a "place", a notable separate shop in some way, then there is located_in: https://github.com/search?q=repo%3Aalltheplaces%2Falltheplaces%20located_in&type=code and located_in_wikidata for times like this

TheUKHighStreet:
Thanks - I've added those. I haven't been to a WHSmith which has a ToysRUs in it so I'm not really sure how much of a separate shop they are - but the Daily Mail has called them shop-in-shops.

PR #10561, opened 2024-09-24
Use shared attributes for Starbucks spiders

arrival-spring:


arrival-spring:
starbucks_us has been failing with 403s for a while

starbucks_za needs more time to find anything (works in the runs and locally)

PR #10579, opened 2024-09-25
Create beaverbrooks_gb.py

TheUKHighStreet:


iandees:
There seem to be maps on each page. Can you try extracting the location data?

TheUKHighStreet:
Will have a look.

iandees:
Nice, thanks.

PR #10598, opened 2024-09-25
Update naptan_gb.py to trial with proxy

CloCkWeRX:
Link works from browser, so


Cj-Malone:
It'll be either download time or memory issue

Issue #10621, opened 2024-09-27
Sample of cases where image or phone is the same in over 1000 POI, indicating that it is likely not POI-specific


matkoniecz:
* [ ] image = https://dynl.mktgcdn.com/p/SwUIly_6zpFTSdZxbmDDWpdbvAwmIZ-1covWYzFlk4Q/150x150.jpg repeated 1101 times in quiktrip_us
* [ ] phone = +1 844-401-8500 repeated 2130 times in bank_of_america
* [ ] image = https://dynl.mktgcdn.com/p/irgrRvtNhff6VLubiBfb0wXj66qCmjjWiKJVpjROaBo/150x150.jpg repeated 1242 times in dominos_pizza_gb
* [ ] image = https://locations.where2getit.com/burlington/images/new/burlington-details.png repeated 1051 times in burlington_us
* [ ] phone = +27 860 999 119 repeated 1032 times in tymebank_za
* [ ] phone = 20000000 repeated 1019 times in seven_eleven_ph
* [ ] phone = +44 800 440011 repeated 1112 times in coral_gb
* [ ] image = https://locations.baskinrobbins.com/assets/static/capuccino-734598ab.png repeated 2254 times in baskin_robbins_us
* [ ] image = https://dynl.mktgcdn.com/p/vVQvY4pkSyF7r1hz-Md07mKAaUyXFlbiOSBA3l1GXzg/150x150.png repeated 7825 times in taco_bell
* [ ] phone = +49 800 7234870 repeated 2142 times in aldi_nord_de
* [ ] image = https://stores.foodlion.com/permanent-b0b701/assets/images/hero-fallback.3d0b8665.png repeated 1174 times in food_lion_us
* [ ] image = https://assets.rbl.ms/25637742/origin.jpg repeated 2500 times in penske
* [ ] image = https://www.ameriprise.com/binaries/content/gallery/ampcom/ameriprise-social-image.png repeated 1175 times in ameriprise_us
* [ ] phone = +44 800 731 6191 repeated 1442 times in ladbrokes_gb
* [ ] image = https://dynl.mktgcdn.com/p/x09r9louvg69_KeYHPnS2M0C8liZgK2TVvn8GDn7uR8/196x196.jpg repeated 1321 times in big_lots
* [ ] phone = +52 55 4510 9999 repeated 2239 times in tiendas_3b_mx
[...]

PR #10625, opened 2024-09-27
Update seven_eleven_ph.py to remove dud phone

CloCkWeRX:
#10621


matkoniecz:
Maybe only `phone = 20000000` repeated 1019 times should be removed? Not all phone numbers? Maybe now or in future some valid ones will be found?

PR #10631, opened 2024-09-27
Burger King spiders for AL, AR, BO, CO, PR, UY

arrival-spring:
Colombia website works, but no results on their restaurant page. Their facebook account does seem active, so not sure what's going on there


arrival-spring:
pre-commit doesn't like other spiders with `_co` in the name, but that is the country code for Colombia, so I guess they need fixing somehow

PR #10695, opened 2024-09-28
[burger_king_my] Add spider (191 locations) and [burger_king_sg] (68 locations)

arrival-spring:
Also additions to `url_to_coords` because apparently it is valid to put the lat/lon with degrees or to use a space instead of a comma


arrival-spring:
One item has lat/lon the wrong way round in the link, which takes you to the Arctic Ocean in Google Maps

PR #10697, opened 2024-09-28
Create soletrader_gb.py

TheUKHighStreet:


TheUKHighStreet:
This works - but the regex is horrible. I'm still trying to get to grips with regex in Python.

PR #10710, opened 2024-09-28
Add spider for virginia blue ridge distrinct health inspection data

wcedmisten:
Mostly here for discussion since this may be out of scope to include in ATP.

There are 136 districts available on https://inspections.myhealthdepartment.com/ spanning several states, but for now I've hard-coded it to a single district in Virginia.

The main benefit to including this source is having access to recent data "on the ground" from a health inspector. A lot of the restaurants in OSM are quite old, and may or may not still exist.

I've been geocoding this data for the state of Virginia to find missing restaurants in OSM, corroborated with public sources like the restaurant's website.

Although it's not super useful to include by itself because the source is not geocoded, a geocoding step in the pipeline would make this quite valuable for auditing OSM data.

### DIFF
--- a/locations/spiders/virginia_health_inspection_blue_ridge.py
+++ b/locations/spiders/virginia_health_inspection_blue_ridge.py
@@ -35,7 +35,7 @@ class VirginiaHealthInspectionBlueRidgeSpider(Spider):
         for inspection in response.json():
             item = DictParser.parse(inspection)
 
-            item["ref"] = item["ref:myhealthdepartment.com"] = inspection["inspectionID"]
+            item["ref"] = item["ref:myhealthdepartment.com:inspectionID"] = inspection["inspectionID"]
             item["name"] = inspection["establishmentName"]
 
             item["extras"]["checked_date"] = inspection["inspectionDate"]
--- a/locations/spiders/virginia_health_inspection_blue_ridge.py
+++ b/locations/spiders/virginia_health_inspection_blue_ridge.py
@@ -35,11 +35,10 @@ class VirginiaHealthInspectionBlueRidgeSpider(Spider):
         for inspection in response.json():
             item = DictParser.parse(inspection)
 
-            item["ref"] = inspection["inspectionID"]
+            item["ref"] = item["ref:myhealthdepartment.com"] = inspection["inspectionID"]
             item["name"] = inspection["establishmentName"]
 
-            # TODO: should this be check_date or is that controversial to use for a third-party check?
-            item["extras"]["inspectionDate"] = inspection["inspectionDate"]
+            item["extras"]["checked_date"] = inspection["inspectionDate"]
 
             # categorize the inspection to OSM standards
             if inspection["inspectionType"] == "Fast Food":
--- a/locations/spiders/virginia_health_inspection_blue_ridge.py
+++ b/locations/spiders/virginia_health_inspection_blue_ridge.py
@@ -50,6 +50,6 @@ class VirginiaHealthInspectionBlueRidgeSpider(Spider):
             elif inspection["inspectionType"] == "Educational Facility Food Service":
                 item["extras"]["amenity"] = "school"
             else:
-                return
+                continue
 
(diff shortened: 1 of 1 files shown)


locations/spiders/virginia_health_inspection_blue_ridge.py
```diff
@@ -0,0 +1,55 @@
...
+            if inspection["inspectionType"] == "Fast Food":
+                item["extras"]["amenity"] = "fast_food"
+            elif inspection["inspectionType"] == "Full Service Restaurant":
+                item["extras"]["amenity"] = "restaurant"
+            elif inspection["inspectionType"] == "Hospital Food Service":
+                item["extras"]["amenity"] = "hospital"
+            elif inspection["inspectionType"] == "Educational Facility Food Service":
+                item["extras"]["amenity"] = "school"
+            else:
+                return
```
Cj-Malone, line 53: Probably continue, not retuen

locations/spiders/virginia_health_inspection_blue_ridge.py
```diff
@@ -0,0 +1,55 @@
...
+            )
+
+    def parse_location_list(self, response):
+        for inspection in response.json():
+            item = DictParser.parse(inspection)
+
+            item["ref"] = inspection["inspectionID"]
+            item["name"] = inspection["establishmentName"]
+
+            item["extras"] = ({"inspectionDate": inspection["inspectionDate"]},)
```
Cj-Malone, line 41: Not a set

```suggestion
            item["extras"] = {"inspectionDate": inspection["inspectionDate"]}
```

Also would it be worth using a more standard `check_date` or `survey:date`?

locations/spiders/virginia_health_inspection_blue_ridge.py
```diff
@@ -0,0 +1,55 @@
...
+                    },
+                    "task": "searchInspections",
+                },
+            )
+
+    def parse_location_list(self, response):
+        for inspection in response.json():
+            item = DictParser.parse(inspection)
+
+            item["ref"] = inspection["inspectionID"]
```
Cj-Malone, line 38: If there is a OSM ref, you can also set it here. It would make conflation easier.
```suggestion
            item["ref"] = item["extras"]["ref:US:abc"] = inspection["inspectionID"]
```

locations/spiders/virginia_health_inspection_blue_ridge.py
Cj-Malone: Actually don't re init extras
```suggestion
            item["extras"]["inspectionDate"] = inspection["inspectionDate"]
```

wcedmisten: > Also would it be worth using a more standard check_date or survey:date?
I wasn't sure if this would be well received, since it's not actually me checking the location, but a third party. Open to discussion on changing it though!

locations/spiders/virginia_health_inspection_blue_ridge.py
wcedmisten: Not quite sure what this comment means, is this an example or is `ref:US:abc` something specific?

wcedmisten:
Thanks for the quick review!

locations/spiders/virginia_health_inspection_blue_ridge.py
Cj-Malone: That was a place holder. In GB we have similar food hygiene data, we tag OSM with `fhrs:id`.

locations/spiders/virginia_health_inspection_blue_ridge.py
Cj-Malone: I would suggest it's close enough, but no strong preference.

iandees:
Let's add it! Thanks for putting it together.

If you're thinking about adding other districts from this same service, it's a good idea to generalize as much as you can into a "store finder" spider, extending it for each individual district.

PR #10714, opened 2024-09-30
DominosPizzaInternational storefinder and spiders

arrival-spring:
Many countries have a common storefinder for Domino's Pizza. It was already in use for MX, SA and QA so I have refactored those as part of this.

I decided to do each country separately because then I think it is easier to notice when one changes format or stops working than if they were all together. Also as I looked, some of them need extra logic for a few things.

I think the spiders won't run here as there are too many, but they all work locally and here are the baseline stats:

- AE: 67 items
- AW: 4 items
- BH: 16 items
- BR: 285 items
- BS: 11 items
- CH: 22 items
- CO: 147 items
- DO: 44 items
- EC: 27 items
- EG: 37 items
- FH: 2 items
- GT: 169 items
- GU: 7 items
- HN: 8 items
- HR: 5 items
- JM: 18 items
- JO: 11 items
- KE: 9 items
- KH: 10 items
- KN: 1 item
- KY: 4 items
- LB: 11 items
- LC: 4 items
- MU: 7 items
- MX: 937 items (refactor)
- NG: 79 items
- OM: 37 items
- PE: 26 items
- PH: 78 items
- PR: 5 items
- QA: 19 items (split and refactor)
- SA: 312 items (split and refactor)
- SK: 4 items
- SX: 4 items
- TT: 9 items
- UY: 4 items
- VE: 34 items


arrival-spring:
I wonder why we don't generate a list of country codes for the spider naming check, instead of maintaining our own list?

iandees:
> I wonder why we don't generate a list of country codes for the spider naming check, instead of maintaining our own list?

Mostly because I was interested in knowing when new countries show up for the first time.

iandees:
... and also deal with cases where the country codes overlap with business names.

PR #10731, opened 2024-09-30
[Tag Heuer] add spider

TheUKHighStreet:
I think this needs countries adding to get the global stores.


TheUKHighStreet:
I think I need to pass the country into the query as a json param. I'm not sure whether this is possible without changing the AlgoliaSpider.

TheUKHighStreet:
Looking at the map, it doesn't seem to be the country which is limiting the results but the 1000 limit. I'm not sure how to get the next page - looking at the Algolia spider, it looks like it should get the next page automatically.

TheUKHighStreet:
I *think* this is the correct number of stores - but I've had to edit the algolia storefinder to get it to work. Please can someone check this is ok and won't break any other spiders.

Issue #10481, opened 2024-09-22
Spiders that have failed 5 runs and are simply producing no output


CloCkWeRX:
# To be checked

- [ ] Spider bj_restaurant broken silently
- [ ] Spider bunnings broken silently
- [ ] Spider coach broken silently
- [ ] Spider curves broken silently
- [ ] Spider duluth_trading broken silently
- [ ] Spider dusk_au broken silently
- [ ] Spider dusk_nz broken silently
- [ ] Spider elysium_healthcare broken silently
- [ ] Spider euromaster_fr broken silently
- [ ] Spider euromaster_nl broken silently
- [ ] Spider fazolis_us broken silently
- [ ] Spider fedex broken silently
- [ ] Spider five_guys_be broken silently
- [ ] Spider five_guys_kr broken silently
- [ ] Spider frischs broken silently
- [ ] Spider geox broken silently
- [ ] Spider greyhound broken silently
- [ ] Spider h_e_b_us broken silently
- [ ] Spider halfords_nl broken silently
- [ ] Spider hooters broken silently
- [ ] Spider hornbach_de broken silently
- [ ] Spider iceland_foods broken silently
- [ ] Spider intermarche broken silently
- [ ] Spider jaycar_au broken silently
- [ ] Spider kate_spade_ca_us broken silently
- [ ] Spider kbp_foods broken silently
- [ ] Spider larosas broken silently
- [ ] Spider leaders_gb broken silently
- [ ] Spider leeann_chin_us broken silently
- [ ] Spider loews_hotels broken silently
- [ ] Spider marcus_theatres broken silently
- [ ] Spider mavis_us broken silently
- [ ] Spider max broken silently
- [ ] Spider mcalisters_deli broken silently
- [ ] Spider naptan_gb broken silently
- [ ] Spider pizza_ranch_us broken silently
- [ ] Spider planet_fitness broken silently
- [ ] Spider quesada_ca broken silently
- [ ] Spider red_robin broken silently
- [ ] Spider reece broken silently
- [ ] Spider sally_beauty broken silently
- [ ] Spider sendiks broken silently
- [ ] Spider smart broken silently
- [ ] Spider sparkasse_de broken silently
- [ ] Spider staples broken silently
- [ ] Spider sullivans_steakhouse broken silently
- [ ] Spider tanger broken silently
- [ ] Spider tgi_fridays_gr broken silently
- [ ] Spider thrifty_foods_ca broken silently
[...]

PR #10749, opened 2024-10-01
[Refactor] Use deepcopy over copy

Cj-Malone:
`.copy()` doesn't work as expected for nested dicts (`extras`). (cc @davidhicks)

```
>>> a = {"a":"abc"}
>>> b = a.copy()
>>> a["1"] = 123
>>> print(b)
{'a': 'abc'}
>>> print(a)
{'a': 'abc', '1': 123}
```

With shallow dicts/strings, it works as expected, you get a copy.

```
>>> a = {"a": "abc", "extras": {"shop": "yes"}}
>>> b = a.copy()
>>> a["extras"]["shop"] = "no"
>>> print(a)
{'a': 'abc', 'extras': {'shop': 'no'}}
>>> print(b)
{'a': 'abc', 'extras': {'shop': 'no'}}
```

But with dicts you get pointers so changing one, changes both.

### DIFF
--- a/locations/spiders/hyundai_gb.py
+++ b/locations/spiders/hyundai_gb.py
@@ -35,20 +35,20 @@ class HyundaiGBSpider(JSONBlobSpider):
             if service["serviceId"] == "sales" or service["serviceId"] == "predaj":
                 # Features where new (and sometimes used) vehicles can be purchased.
                 # "Predaj" is "Sales" in Slovak.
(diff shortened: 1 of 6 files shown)


CloCkWeRX:
Via https://github.com/alltheplaces/alltheplaces/issues/10481 intermarche probably needs rewrite

PR #10754, opened 2024-10-01
[amtrak_us] Fix and rename

arrival-spring:
Runs in #10732 failed because network is not a main field, so moving it to extras.

This doesn't work locally as I get an empty response on the start_url, but could be a geo issue.

### DIFF
--- a/locations/spiders/amtrak_ca_us.py
+++ b/locations/spiders/amtrak_ca_us.py
@@ -8,8 +8,8 @@ from locations.dict_parser import DictParser
 from locations.pipelines.address_clean_up import merge_address_lines
 
 
-class AmtrakUSSpider(Spider):
-    name = "amtrak_us"
+class AmtrakCAUSSpider(Spider):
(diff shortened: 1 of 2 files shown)


arrival-spring:
I'm not sure then. When I open https://www.amtrak.com/plan-your-trip.html the file https://maps.amtrak.com/services/MapDataService/stations/allStations comes up essentially empty as well.

Any thoughts from anyone?

Cj-Malone:
Works in the weekly, and contains some CA locations.

https://www.alltheplaces.xyz/spiders.html?search=amtrak

Cj-Malone:
Clicking "map" on [here](https://www.amtrak.com/routes/adirondack-train.html) is also broken. I assume they broke there backend, hopefully it will be fixed soon.

arrival-spring:
> Works in the weekly, and contains some CA locations.

Ah yes, I didn't see them as they were only just over the border, I'll rename it again now

PR #10757, opened 2024-10-01
Drop images for burgerville, rewrite to structureddataspider

CloCkWeRX:


locations/spiders/burgerville_us.py
```diff
@@ -14,11 +14,5 @@ class BurgervilleSpider(SitemapSpider):
             "parse_store",
```
CloCkWeRX, line 14:
```suggestion
            "parse_sd",
```

PR #10764, opened 2024-10-01
[burger_king] Add AE, SA, ZA. Fix NZ. Collect more features

arrival-spring:
I think that concludes my Burger King additions for now. There are a few countries that have a very few locations and no storefinder so would require some HTML parsing, but it didn't seem worth it.

Russia eludes me as it works fine in the browser, but even with playwright I can't get anything to work in code.


arrival-spring:
There are quite a lot of empty coordinate hits, which seem to get hit first. It works locally and gets about 9600 locations

PR #10775, opened 2024-10-01
Drop images for rexel_gb, handle exceptions

CloCkWeRX:


locations/storefinders/rexel.py
```diff
@@ -41,9 +41,10 @@ def parse(self, response):
             if not feature["address"]["phone"].replace(" ", "").startswith("+443"):
                 item["phone"] = feature["address"]["phone"]
             # e.g. https://www.denmans.co.uk/den/Bradley-Stoke-Bristol/store/1AR
-            item["website"] = (
-                f'https://{self.base_url}/{feature["address"]["town"].replace(" ", "-")}/store/{feature["ref"]}'
-            )
+            if "address" in feature and "town" in feature["address"]:
```
CloCkWeRX, line 44:
```suggestion
            if "address" in feature and "town" in feature["address"] and feature["address"]["town"] is not None:
```

locations/storefinders/rexel.py
```diff
@@ -41,9 +41,10 @@ def parse(self, response):
             if not feature["address"]["phone"].replace(" ", "").startswith("+443"):
```
CloCkWeRX, line 41:
```suggestion
            if feature["address"]["phone"] is not None and not feature["address"]["phone"].replace(" ", "").startswith("+443"):
```

Issue #10679, opened 2024-09-27
 cases where image is repeated across multiple POIs - more than 20 times, less than 500 times


matkoniecz:
* [x] image = https://magasins.autourdebebe.com/img/PoiDefault.jpg repeated 71 times in autour_de_bebe_fr
* [ ] image = https://s7g10.scene7.com/is/image/KingfisherDigital/66aefe0e1b764a023696e9977d43a2e0978b009d repeated 95 times in castorama_fr
* [ ] image = https://boutiques.carreblanc.com/img/default_store.jpg repeated 66 times in carre_blanc
* [ ] image = https://media-cdn.skoringen.dk/media/hynb31dh/default_store-image.jpg repeated 151 times in skoringen_dk_no
* [ ] image = https://bureaux.kpmg.fr/img/PoiDefault.jpg repeated 27 times in kpmg_fr
* [ ] image = https://dyn.eu.mktgcdn.com/f/0/UvcpynPztrHqOmJPuz3zsfU220k2D7C7DJBPSrFjcmI.png/width=150,height=150 repeated 166 times in allied_irish_banks_ie
* [x] image = https://images.ctfassets.net/d9ybqgejqp0w/7e2fEPiB46UhjDfedBIlUN/f6aa70d4c14510180778c44fac38341e/store-page-generic-new.jpg repeated 120 times in ubreakifix  #10735
* [x] image = https://images.ctfassets.net/16nm6vz43ids/7e2fEPiB46UhjDfedBIlUN/e1c17eb2066a9ed4b196e6f88c48bb21/store-page-generic-new.jpg repeated 171 times in ubreakifix  #10735
* [ ] image = https://dynl.mktgcdn.com/p/iXTl8kSu4fv-uoH6ldspnJ4CFYN_uqvV-DvbqkzjULU/173x173.jpg repeated 31 times in dominos_pizza_gb
* [ ] image = https://gails.com/cdn/shop/files/Bakery_Page_Default_Photo.jpg?v=1710323054&width=330 repeated 40 times in gails_bakery_gb
* [ ] image = https://www.mitra.nl/img/mitra-shop-default.jpg?w=1200&h=900&mode=crop&scale=both repeated 58 times in mitra_nl
* [ ] image = http://www.autonation.com/~/media/47B52E3A3DE44E9CB43B7DA0B0BA7F09 repeated 317 times in auto_nation_us
[...]

PR #10790, opened 2024-10-01
Drop images for keke #10679

CloCkWeRX:
#4757


CloCkWeRX:
https://www.kekes.com/locations/kekes-las-vegas---blue-diamond - no longer structured data

PR #10844, opened 2024-10-01
Ignore duplicates when searching for missing wikidata by country

arrival-spring:
And this brings me down to just 30 missing by wikidata in South Africa. Some of them are defunct though (and correctly marked as such in wikidata), so the next step is to ignore them as well.

### DIFF
--- a/locations/commands/nsi.py
+++ b/locations/commands/nsi.py
@@ -87,10 +87,15 @@ class NameSuggestionIndexCommand(ScrapyCommand):
                         missing.append(item)
         # Assume we are searching by location, either country code or some state geojson string
         else:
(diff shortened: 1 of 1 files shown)


locations/commands/nsi.py
```diff
@@ -87,10 +87,15 @@ def detect_missing(self, args):
                         missing.append(item)
         # Assume we are searching by location, either country code or some state geojson string
         else:
+            seen_wikidata = []  # There can be duplicate wikidata entries across different categories
```
iandees, line 90: It would be more efficient to use a `set()` / `{}` and `.add()` to it below.

Issue #9208, opened 2024-08-03
Booker Group (JSON blob)


CloCkWeRX:
### Brand name

Booker Group

British food wholesale operator
https://www.booker.co.uk/branch-locator/
### Wikidata ID

Q4943180
https://www.wikidata.org/wiki/Q4943180
https://www.wikidata.org/wiki/Special:EntityData/Q4943180.json

### Store finder url(s)

Official Url(s): https://www.booker.co.uk/
- [x] I have run the below and if anything in generated, added it.
- [x] I have manually checked for a storefinder
- [x] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q4943180 https://www.booker.co.uk/`
```
```

CloCkWeRX:
https://www.booker.co.uk/branch-locator/darlington/banks-road

PR #10865, opened 2024-10-02
[Booker GB] add spider

TheUKHighStreet:
Fixes https://github.com/alltheplaces/alltheplaces/issues/9208


TheUKHighStreet:
I think Booker and Makro brands need separating but not sure how to.

PR #10873, opened 2024-10-03
[inxpress] Add spider

arrival-spring:
Waiting on https://github.com/osmlab/name-suggestion-index/commit/392c1229fea72fa98c7628fedabd02b47649db03 for category

```
 'atp/brand/InXpress': 485,
 'atp/brand_wikidata/Q130400315': 485,
 'atp/category/missing': 485,
 'atp/field/city/missing': 485,
 'atp/field/email/invalid': 23,
 'atp/field/email/missing': 6,
 'atp/field/image/missing': 485,
 'atp/field/lat/invalid': 1,
 'atp/field/lat/missing': 10,
 'atp/field/lon/missing': 10,
 'atp/field/name/missing': 485,
 'atp/field/opening_hours/missing': 485,
 'atp/field/operator/missing': 485,
 'atp/field/operator_wikidata/missing': 485,
 'atp/field/phone/invalid': 8,
 'atp/field/phone/missing': 3,
 'atp/field/postcode/missing': 7,
 'atp/field/state/from_reverse_geocoding': 127,
 'atp/field/state/missing': 360,
 'atp/field/street_address/missing': 485,
 'atp/field/twitter/invalid': 1,
 'atp/field/twitter/missing': 480,
 'atp/field/website/invalid': 9,
 'atp/item_scraped_host_count/za.inxpress.com': 485,
 'atp/nsi/brand_missing': 485,
```


arrival-spring:
Remaining 'invalid;' emails are because the item properties checker assumes that we have a single email address, and doesn't accept semi-colon delimited emails.

CloCkWeRX:
Would you consider doing a .split and taking the first as "primary"?

Alternatively if there are only a max of 2... You could kludge one into email and the other into contact:email.

arrival-spring:
There are other places that return multiple emails, so I don't quite see why. I definitely don't like the second suggestion though, I don't think we should return both email and contact:email

arrival-spring:
There are other places that return multiple emails, so I don't quite see why. I definitely don't like the second suggestion though, I don't think we should return both email and contact:email

Issue #10892, opened 2024-10-04
malformed tesla mail


matkoniecz:
`email": {"label": "Store Email", "value": "Beijing_sales@teslamotors.com"}`

in say

```
{"type": "Feature", "id": "Nto7zR72UvuAIPv57Ib5xMkOWLc=", "properties": {"ref": "beijingyizhuang", "shop": "car;car_repair", "@source_uri": "https://www.tesla.com/cua-api/tesla-location?translate=en_US&usetrt=true&id=beijingyizhuang", "@spider": "tesla", "addr:street_address": "\u5317\u4eac\u4ea6\u5e84\u7279\u65af\u62c9\u4e2d\u5fc3, \u5927\u5174\u533a\u4e2d\u548c\u885716\u53f73\u5e621003-1004\u5355\u5143", "addr:state": "\u5317\u4eac", "addr:postcode": "100176", "addr:country": "CN", "name": "\u5317\u4eac\u4ea6\u5e84\u7279\u65af\u62c9\u4e2d\u5fc3", "email": {"label": "Store Email", "value": "Beijing_sales@teslamotors.com"}, "brand": "Tesla", "brand:wikidata": "Q478214"}, "geometry": {"type": "Point", "coordinates": [116.5238, 39.80645]}},
```

PR #10898, opened 2024-10-05
Tesla - Update email if its a dict

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/10892


matkoniecz:
maybe I messed something up, but I keep seeing this bug - see https://github.com/alltheplaces/alltheplaces/issues/11711

Issue #10897, opened 2024-10-04
Allow to specify DAYS in a different language when parsing structured data


arrival-spring:
With a StructuredDataSpider, you can specify time_format, but not any different language for days. It would be good to allow that, to save having to re-parse the hours.

For example, see https://www.virginactive.it/club/milano/milano-vittorio-emanuele

```
"openingHoursSpecification": [
    {
    "@type": "OpeningHoursSpecification",
    "dayOfWeek": [
        "Luned&#236; - Venerd&#236;"
        ],
    "opens": "6.30",
    "closes": "23.00"
    }
,
    {
    "@type": "OpeningHoursSpecification",
    "dayOfWeek": [
        "Sabato - Domenica"
        ],
    "opens": "8.00",
    "closes": "20.00"
    }
```

Cj-Malone:
I don't think that's in spec? Anyway the encoding issue would probably mean we have to do something in spider anyway. I vote to try and clean it in pre_process

davidhicks:
> I don't think that's in spec? Anyway the encoding issue would probably mean we have to do something in spider anyway. I vote to try and clean it in pre_process

Yep schema.org's `OpeningHoursSpecification` has a preset enumeration of English day names only per https://schema.org/OpeningHoursSpecification and https://schema.org/DayOfWeek

For every day the feature is typically open there should be something like:
```
    {
      "@type": "OpeningHoursSpecification",
      "closes": "17:00:00",
      "dayOfWeek": "https://schema.org/Monday",
      "opens": "09:00:00"
    },
```

Ranges and lists of days are not supported. Each day has to be individually listed.

Times also have to be listed only in format `hh:mm:ss[Z|(+|-)hh:mm]` (e.g. seconds included).

Perhaps we could add a helper function that takes a non-conformant `openingHoursSpecification` and creates a string of `{dayOfWeek}: {opens} - {closes}` and passes this to `OpeningHours.add_ranges_from_string` with relevant `DAYS_ZZ` for the language applicable, as `OpeningHours.add_ranges_from_string` is purpose designed to parse random opening hours strings to extract whatever it can.

PR #10899, opened 2024-10-05
Support passing in DAYS_XY when parsing structured data

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/10897

I agree that this not the standard, but from a spider developer point of view there's value in just being able to say _StructuredDataSpider, please parse this using this constant, not your default assumption, because the publisher got it wrong_.

The other alternative is to nag the data producer to conform to the specification... but even the schema.org validator doesn't do that:

https://validator.schema.org/#url=https%3A%2F%2Fwww.virginactive.it%2Fclub%2Fcatania%2Fcatania
image

### DIFF
--- a/locations/linked_data_parser.py
+++ b/locations/linked_data_parser.py
@@ -59,7 +59,7 @@ class LinkedDataParser:
                 return ld_obj
 
     @staticmethod
-    def parse_ld(ld, time_format: str = "%H:%M") -> Feature:  # noqa: C901
+    def parse_ld(ld, time_format: str = "%H:%M", days: {} = DAYS_EN) -> Feature:  # noqa: C901
         item = Feature()
 
         if (
@@ -122,7 +122,7 @@ class LinkedDataParser:
         item["website"] = LinkedDataParser.get_case_insensitive(ld, "url")
 
         try:
-            item["opening_hours"] = LinkedDataParser.parse_opening_hours(ld, time_format=time_format)
+            item["opening_hours"] = LinkedDataParser.parse_opening_hours(ld, time_format=time_format, days=days)
         except ValueError as e:
             # Explicitly handle a ValueError, which is likely time_format related
             logger.warning(f"Unable to parse opening hours - check time_format? Error was: {str(e)}")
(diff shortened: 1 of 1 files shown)


locations/linked_data_parser.py
```diff
@@ -183,22 +183,27 @@ def parse(response, wanted_type, json_parser="json") -> Feature:
...
+        if not isinstance(parsed_days, list):
+            parsed_days = [parsed_days]
 
-        for day in days:
+        for day in parsed_days:
+            # Handle plain text days, or URI references to a DayOfWeek enumeration, ie https://schema.org/Friday
+            parsed_day = day.strip().replace("http://schema.org/", "").replace("https://schema.org/", "")
             oh.add_range(
-                day=day.strip(),
+                day=days[parsed_day],
```
CloCkWeRX, line 206: This is the main bit I am not sure about; as before this just took a string - including "https://schema.org/Friday" - and presumably tolerated that. Now we're validating against DAYS_EN.

locations/linked_data_parser.py
```diff
@@ -183,22 +183,27 @@ def parse(response, wanted_type, json_parser="json") -> Feature:
...
-        if not isinstance(days, list):
-            days = [days]
+        parsed_days = LinkedDataParser.get_case_insensitive(rule, "dayOfWeek")
+        if not isinstance(parsed_days, list):
+            parsed_days = [parsed_days]
 
-        for day in days:
+        for day in parsed_days:
+            # Handle plain text days, or URI references to a DayOfWeek enumeration, ie https://schema.org/Friday
+            parsed_day = day.strip().replace("http://schema.org/", "").replace("https://schema.org/", "")
```
Cj-Malone: Sanitise day should already handle this?

Cj-Malone:
I'm still not fond of this. It's not in spec. I don't think we've encountered a lot of this. We shouldn't be making the core more complex for edge cases, edge cases should be fixed in spider.

iandees:
> edge cases should be fixed in spider

I agree. Would this happen in a pre-process step that renames the day names?

PR #10876, opened 2024-10-03
[nu_za] Add spider

arrival-spring:
Waiting on https://github.com/osmlab/name-suggestion-index/commit/128f57fe54150163727d60b2477dce7281dc2a47 for category

```
{'atp/brand/Nü': 13,
 'atp/brand_wikidata/Q130400175': 13,
 'atp/category/missing': 13,
 'atp/field/city/missing': 13,
 'atp/field/country/from_website_url': 13,
 'atp/field/email/missing': 13,
 'atp/field/image/missing': 13,
 'atp/field/operator/missing': 13,
 'atp/field/operator_wikidata/missing': 13,
 'atp/field/postcode/missing': 13,
 'atp/field/state/missing': 13,
 'atp/field/street_address/missing': 13,
 'atp/field/twitter/missing': 13,
 'atp/item_scraped_host_count/nu10.goreview.co.za': 1,
 'atp/item_scraped_host_count/nu11.goreview.co.za': 1,
 'atp/item_scraped_host_count/nu12.goreview.co.za': 1,
 'atp/item_scraped_host_count/nu13.goreview.co.za': 1,
 'atp/item_scraped_host_count/nu14.goreview.co.za': 1,
 'atp/item_scraped_host_count/nu15.goreview.co.za': 1,
 'atp/item_scraped_host_count/nu16.goreview.co.za': 1,
 'atp/item_scraped_host_count/nu18.goreview.co.za': 1,
 'atp/item_scraped_host_count/nu3.goreview.co.za': 1,
 'atp/item_scraped_host_count/nu5.goreview.co.za': 1,
 'atp/item_scraped_host_count/nu7.goreview.co.za': 1,
 'atp/item_scraped_host_count/nu8.goreview.co.za': 1,
 'atp/item_scraped_host_count/nu9.goreview.co.za': 1,
 'atp/nsi/brand_missing': 13,
```

PR #10910, opened 2024-10-05
GoReviewSpider adjust url regex

arrival-spring:
Not all of the urls have digits at the start, although most do. From a sample, this gets a few extra locations per spider.


arrival-spring:
Nu is waiting on NSI for category, see #10876

PR #10912, opened 2024-10-05
Waterdrop (Multiple countries)

CloCkWeRX:


iandees:
How will this get category info?

Cj-Malone:
Still no category. I hard to believe a chain this big in Europe wouldn't already be in NSI. Are we sure these aren't stockists?

PR #10915, opened 2024-10-05
Rewrite dillards as StructuredDataSpider

CloCkWeRX:


locations/spiders/dillards.py
```diff
@@ -4,46 +4,41 @@
...
-        ldjson = response.xpath('//script[@type="application/ld+json"]/text()[contains(.,"DepartmentStore")]').get()
-        data = json.loads(ldjson)
-
+    def post_process_item(self, item, response, ld_data):
         script = response.xpath('//script/text()[contains(.,"__INITIAL_STATE__")]').get()
         script_data = json.decoder.JSONDecoder().raw_decode(script, script.index("{"))[0]
         lat = script_data["contentData"]["store"]["latitude"]
         lon = script_data["contentData"]["store"]["longitude"]
 
+        # TODO: Why isn't this automatically done by StructuredDataSpider?
```
Cj-Malone, line 27: `dayOfWeek` is a dict, it could probably cleaned in pre process and dropped here entirely.

locations/spiders/dillards.py
```diff
@@ -4,46 +4,41 @@
...
-from locations.items import Feature
+from locations.structured_data_spider import StructuredDataSpider
 
 
-class DillardsSpider(SitemapSpider):
+class DillardsSpider(SitemapSpider, StructuredDataSpider):
     name = "dillards"
     item_attributes = {"brand": "Dillard's", "brand_wikidata": "Q844805"}
     allowed_domains = ["www.dillards.com"]
     sitemap_urls = ("https://www.dillards.com/sitemap/sitemap_storeLocator_1.xml",)
```
Cj-Malone, line 14: While we are here may as well update this `https://www.dillards.com/robots.txt` or `https://www.dillards.com/sitemap/sitemap.xml` in case we get `_2.xml`. `sitemap_filter` can restrict which ones we download, `sitemap_rules` may also want to be updated it seems the urls now include 3 `/`s.

locations/spiders/dillards.py
Cj-Malone: Doesn't work anyway.
`cat dillards.geojson| jq '.features[].properties.opening_hours' | sort | uniq -c | sort -n`
```
      1 "Mo 12:00-19:00"
     27 null
    245 "Mo 11:00-20:00"
```

PR #10922, opened 2024-10-05
Fix halfords_nl

CloCkWeRX:


Cj-Malone (approved):
I would suggest CrawlSpider, but 17 POI aren't worth the time.

PR #10929, opened 2024-10-06
[ribola_hr] add spider (99 locations)

panasonic-mafia:


Cj-Malone:
This isn't getting a name, is this gonna be added to NSI?

panasonic-mafia:
> This isn't getting a name, is this gonna be added to NSI?

Yep

PR #10937, opened 2024-10-07
Skip dissolved brands when checking wikidata for missing

arrival-spring:
That brings me down to just 17 brands missing by wikidata in South Africa.

### DIFF
--- a/locations/name_suggestion_index.py
+++ b/locations/name_suggestion_index.py
@@ -72,14 +72,19 @@ class NSI(metaclass=Singleton):
                     return wikidata_code
         return None
 
-    def lookup_wikidata(self, wikidata_code: str = None) -> dict:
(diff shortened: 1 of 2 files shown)


locations/name_suggestion_index.py
```diff
@@ -72,14 +72,19 @@ def get_wikidata_code_from_url(self, url: str = None) -> str | None:
                     return wikidata_code
         return None
 
-    def lookup_wikidata(self, wikidata_code: str = None) -> dict:
+    def lookup_wikidata(self, wikidata_code: str = None, include_dissolved=False) -> dict | None:
```
iandees, line 75: Include doc for what the new param means.

iandees (changes requested):

locations/name_suggestion_index.py
arrival-spring: Done

iandees (approved):
Thanks

PR #10914, opened 2024-10-05
Add parsing of hasMap to LinkedDataParser

CloCkWeRX:
Fix https://github.com/alltheplaces/alltheplaces/issues/10399

PR #10939, opened 2024-10-07
Update linked_data_parser.py to change order of coordinate detection

CloCkWeRX:


Cj-Malone:
This still doesn't fix stonegate_gb. I would feel better if this was reverted https://github.com/alltheplaces/alltheplaces/pull/10914#issuecomment-2396500141. It's a couple spiders, it doesn't need to be shared code, especially when it has the possibility of breaking things.

CloCkWeRX:
https://validator.schema.org/#url=https%3A%2F%2Fwww.socialpubandkitchen.co.uk%2Fabbey-gloucester

Okay, a map as a type. These examples are pretty easy to fix once demonstrated, it feels like just writing more test coverage would be better to prevent regression than

iandees:
Can you add a test for this change too?

PR #10953, opened 2024-10-07
Trader Joe's: Fix location and add extra tags

impiaaa:
Also move `name` to `branch`, detect warehouses and wine stores, and parse hours differently

### DIFF
--- a/locations/spiders/trader_joes_us.py
+++ b/locations/spiders/trader_joes_us.py
@@ -1,4 +1,5 @@
-from locations.hours import DAYS, OpeningHours
+from locations.categories import Categories, Drink, apply_category, apply_yes_no
+from locations.hours import DAYS, DAYS_FULL, OpeningHours
 from locations.storefinders.where2getit import Where2GetItSpider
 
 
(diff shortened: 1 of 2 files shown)


locations/spiders/trader_joes_us.py
```diff
@@ -7,23 +8,34 @@ class TraderJoesUSSpider(Where2GetItSpider):
...
-        )
-        hours_string = ""
-        for day_abbrev, day in zip(["mon", "tues", "wed", "thurs", "fri", "sat", "sun"], DAYS):
-            if location.get(day_abbrev + "open") and location.get(day_abbrev + "close"):
-                hours_string = (
-                    f"{hours_string} {day}: " + location[day_abbrev + "open"] + "-" + location[day_abbrev + "close"]
-                )
-        item["opening_hours"] = OpeningHours()
-        item["opening_hours"].add_ranges_from_string(hours_string)
+        if location["warehouse"] == "1":
```
arrival-spring, line 17: There is Categories.INDUSTRIAL_WAREHOUSE which you could use here

impiaaa: They're not TJ's branded warehouses

Issue #10439, opened 2024-09-21
mcgrath_au needs rewrite to algolia


CloCkWeRX:
### Spider name

mcgrath_au

### Possible causes?

https://www.mcgrath.com.au/search/office
https://testinga3vxex35ke-dsn.algolia.net/1/indexes/office/query?x-algolia-agent=Algolia%20for%20JavaScript%20(4.22.1)%3B%20Browser%20(lite)&x-algolia-api-key=d1594f8570b9fa5a764d624db7ebd84e&x-algolia-application-id=testingA3VXEX35KE

PR #10960, opened 2024-10-07
McGrath: Update API key

impiaaa:


CloCkWeRX:
Annoying, that was just rewritten two weeks ago: https://github.com/alltheplaces/alltheplaces/issues/10439

May have to look at some kind of auto detection of key if it changes frequently

impiaaa:
To be clear, I was the one who rewrote it (#10571). It wasn't immediately obvious to me where the API key comes from on the webpage, so I'd like to wait and see if it does change frequently before putting more effort into detecting it.

CloCkWeRX:
Yeah, fair enough. You probably already know, but 
https://github.com/alltheplaces/alltheplaces/pull/6526/files#diff-28f4b0075e09fcbd651c6034d03d519fa85fd75f094523bc9103989c7a35954bR29 is the current detection.

From their online samples I'd think they often are in an env variable for a framework; ie https://github.com/search?q=REACT_APP_ALGOLIA_API_KEY&type=code - so we might be able to in a playwright browser evaluate window.app.config.whatever.api_key across many spiders.

PR #10964, opened 2024-10-08
[scouts_gg_gb_im_je] Fix spider

arrival-spring:
They seem to have changed to title case keys, rather than all lower case


Cj-Malone:
Thank you

PR #10965, opened 2024-10-08
[supa_quick] Try longer timeout

arrival-spring:
There have been a lot of timeout errors in the runs of this spider, including leading to some locations never being returned, so trying a longer timeout.


arrival-spring:
I was pausing while I looked at the opening hours, looks fine now though

PR #10974, opened 2024-10-08
remove day_today_gb

matkoniecz:
according to https://github.com/alltheplaces/alltheplaces/discussions/10941#discussioncomment-10883201 research, about 10% entries are bogus


iandees:
10% of entries doesn't seem all that bad, but I'm ok with removing this. I would also be ok with adding it back if we find a way to make it more reliable.

PR #10977, opened 2024-10-08
[ulybka_radugi_ru] add spider (1.6k locations)

panasonic-mafia:


Cj-Malone:
Weirdly when I move to paging this API, instead of `limit=10000` I get duplicates, like they aren't honouring the sorting, or the offset is done pre filtering. Oh well, we can take them all in 1 request.

Cj-Malone:
Thanks

Issue #10694, opened 2024-09-28
Repeated, clearly bogus phone numbers


matkoniecz:
* [ ] phone = + repeated 159 times in hertz
* [ ] phone = (000) 000-00-00 repeated 35 times in opendata_mos_food_ru
* [ ] phone = - repeated 69 times in shop_4f_pl
* [ ] phone = - repeated 499 times in esprit
* [x] #10625
* [ ] phone = 200000000 repeated 24 times in seven_eleven_ph
* [ ] phone = 2000000 repeated 29 times in seven_eleven_ph
* [ ] phone = 0000 repeated 21 times in tim_hortons
* [ ] phone = +7() repeated 25 times in my_documents_ru
* [ ] phone = null repeated 46 times in pret_a_manger
* [ ] phone = (0) 0-0 repeated 27 times in supercuts
* [ ] phone = (000) 000-0000 repeated 24 times in indigo
* [ ] phone = 0 repeated 101 times in shell
* [ ] phone = - repeated 48 times in remax_it
* [ ] phone = - repeated 28 times in top_market_pl
* [ ] phone = 0000000000 repeated 100 times in phillips_66_conoco_76
* [ ] phone = 000-000-0000 repeated 29 times in phillips_66_conoco_76
* [ ] phone = null repeated 177 times in united_colors_of_benetton
* [ ] phone = +43 repeated 31 times in adeg
* [ ] phone = BE repeated 517 times in mol
* [ ] phone = 1111111111 repeated 143 times in bp
* [ ] phone = +48 repeated 23 times in wesola_pani
* [ ] phone = 0000000000 repeated 42 times in travelex
* [ ] phone = 000000000 repeated 30 times in travelex
* [ ] phone = None repeated 144 times in morrisons_gb

matkoniecz:
Note that these can be also thrown out by checking is phone number valid-looking, this would also catch more rare cases.

matkoniecz:
mostly fixed by  #10979

matkoniecz:
(remaining ones remain detectable and will be likely listed together with other highly repeated phone numbers)

PR #10979, opened 2024-10-08
Reject phone numbers with no digits or all zeros

arrival-spring:
Ref #10694

This would get most of them, with the exception of things like "+48" and "2000000"


Cj-Malone:
Thank you

PR #10992, opened 2024-10-09
[TGI Fridays GB] add spider

TheUKHighStreet:


TheUKHighStreet:
There are 51 locations but I can't see a way to get them. The list of all the locations on the locations page which says there are 51 only gives 42.
https://www.tgifridays.co.uk/locations

PR #10996, opened 2024-10-10
brand for real_hu

matkoniecz:
from their website and OSM it appears their brand - at least in Hungary - is rather Reál, not Real


Cj-Malone:
Thanks

PR #10997, opened 2024-10-10
[Core] Make lookup_wikidata safer

Cj-Malone:
`elif "dissolutions" not in wd_json:` was failing due to `wd_json` being `None` when a look up failed.

### DIFF
--- a/locations/name_suggestion_index.py
+++ b/locations/name_suggestion_index.py
@@ -81,10 +81,15 @@ class NSI(metaclass=Singleton):
         """
         self._ensure_loaded()
         wd_json = self.wikidata_json.get(wikidata_code)
+
+        if not wd_json:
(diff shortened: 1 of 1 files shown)


locations/name_suggestion_index.py
```diff
@@ -81,10 +81,15 @@ def lookup_wikidata(self, wikidata_code: str = None, include_dissolved=False) ->
...
         wd_json = self.wikidata_json.get(wikidata_code)
+
+        if not wd_json:
+            return None
+
         if include_dissolved:
             return wd_json
         elif "dissolutions" not in wd_json:
             return wd_json
+
```
iandees, line 92: Thank you for adding whitespace. Makes it easier to read the code!

PR #10999, opened 2024-10-10
[The Fish & Chip Co ZA] Add spider

Cj-Malone:
Rel: https://github.com/osmlab/name-suggestion-index/commit/2b4557713f66e236502cfe7998f9c9fdca452529

### DIFF
--- a/locations/spiders/the_fish_and_chip_co_za.py
+++ b/locations/spiders/the_fish_and_chip_co_za.py
@@ -1,0 +1,33 @@
+from typing import Any, Iterable
+
+from scrapy import Request, Spider
+from scrapy.http import JsonRequest, Response
+
+from locations.dict_parser import DictParser
+
+
+class TheFishAndChipCoZASpider(Spider):
+    name = "the_fish_and_chip_co_za"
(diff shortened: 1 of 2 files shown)


Cj-Malone:
> TheFishAndChipCOZASpider' instead of 'TheFishAndChipCoZASpider

The `CO` isn't a country. I'm OK with faking it though, if we need to please the linter.

arrival-spring:
> > TheFishAndChipCOZASpider' instead of 'TheFishAndChipCoZASpider
> 
> The `CO` isn't a country. I'm OK with faking it though, if we need to please the linter.

When that first came up, other such spiders were renamed to _company

See:
https://github.com/alltheplaces/alltheplaces/pull/10631/commits/e4d3a54f24299c8004823309ccb2529072982256

Cj-Malone:
> When that first came up, other such spiders were renamed to _company

Sounds good to me. Thanks

arrival-spring:
Also, n.b. there is GoReviewApiSpider store finder which this could use. Maybe some more comments need putting there to make it more discoverable?

PR #11060, opened 2024-10-13
fix dsw name

matkoniecz:
disclaimer: `long_name=Designer Shoe Warehouse` is invented by analogy to `short_name` and is not in real use in OSM.


matkoniecz:
> KeyError: 'Feature does not support field: long_name'

oh right

I would be happy to fix it but lets whether someone will veto the entire `long_name` idea

Cj-Malone:
#11208

PR #11062, opened 2024-10-13
fjallraven - set SHOP_OUTDOOR

matkoniecz:
also based on OpenStreetMap tagging

### DIFF
--- a/locations/spiders/fjallraven.py
+++ b/locations/spiders/fjallraven.py
@@ -14,3 +14,5 @@ class FjallravenSpider(Where2GetItSpider):
     api_filter = {
         "brand_store": {"eq": "1"},
     }
+    # Website seems to be generic
+    drop_keys = {"website"}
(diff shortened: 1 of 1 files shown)


matkoniecz:
>     🚨 Website is invalid on 8 items

this seems clearly not related to my change

(I finally figured out I can click on icon, not sure why these are collapsed by default)

locations/spiders/fjallraven.py
```diff
@@ -7,10 +7,12 @@ class FjallravenSpider(Where2GetItSpider):
     item_attributes = {
         "brand": "Fjällräven",
         "brand_wikidata": "Q1422481",
-        "extras": Categories.SHOP_CLOTHES.value,
+        "extras": Categories.SHOP_OUTDOOR.value,
```
Cj-Malone, line 10: This is one of the issues with hardcoding categories, if we had simply no applied one, we would have gotten it from NSI. If NSI changed, we'd have gotten it automatically.
```suggestion
```

PR #11083, opened 2024-10-13
[Shoe Zone GB] Add address and opening hours

TheUKHighStreet:
I'm sure that there must be an easier way to get the structured data spider to get the opening hours from scraping the page.

### DIFF
--- a/locations/spiders/shoe_zone_gb.py
+++ b/locations/spiders/shoe_zone_gb.py
@@ -1,6 +1,7 @@
 import json
 
 from scrapy.spiders import SitemapSpider
+
 from locations.hours import DAYS_3_LETTERS, OpeningHours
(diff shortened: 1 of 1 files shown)


Cj-Malone:
> I'm sure that there must be an easier way to get the structured data spider to get the opening hours from scraping the page.

The structured data is broken, we might be able to fix it in pre process, but it's not worth the effort.

Cj-Malone:
Thanks

TheUKHighStreet:
Thanks for basically rewriting it :) I am gradually getting the hang of it but this is very useful. Thanks. I had tried setting the time format to  "%H.%M" but somewhere had gone wrong and couldn't get it to work.

Cj-Malone:
You're doing great, keep it up.

PR #11090, opened 2024-10-14
[United Surgical Partners International] Fix invalid websites

matkoniecz:


locations/spiders/united_surgical_partners_international.py
```diff
@@ -22,7 +22,7 @@ def parse(self, response):
             if len(item) == 12:
                 website = item[11]
             else:
-                website = "na"
+                website = None
             properties = {
                 "name": item[0],
                 "addr_full": item[3] + " " + item[4],
```
Cj-Malone, line 28: `street_address`

matkoniecz: what `street_address` ?

Cj-Malone: This doesn't look like addr_full. It looks like street address.

PR #11091, opened 2024-10-14
mark that ben_and_jerrys is listing POI aspect, not POIs

matkoniecz:
this source is not listing standalone ice cream POIs from this brand but rather known locations where their ice cream is provided

at least in part

not entirely sure how best handle it

screen

### DIFF
--- a/locations/spiders/ben_and_jerrys.py
+++ b/locations/spiders/ben_and_jerrys.py
@@ -33,4 +33,9 @@ class BenAndJerrysSpider(Where2GetItSpider):
                 item["opening_hours"].set_closed(day)
             else:
                 item["opening_hours"].add_ranges_from_string(day + " " + location.get(day.lower()))
+        # this source is not listing standalone ice cream POIs from this brand
+        # but rather known locations where their ice cream is provided
+        item.pop("amenity", None)
+        item.pop("cuisine", None)
+        apply_yes_no("ice_cream", item, "yes")
         yield item


locations/spiders/ben_and_jerrys.py
```diff
@@ -33,4 +33,9 @@ def parse_item(self, item, location):
                 item["opening_hours"].set_closed(day)
             else:
                 item["opening_hours"].add_ranges_from_string(day + " " + location.get(day.lower()))
+        # this source is not listing standalone ice cream POIs from this brand
+        # but rather known locations where their ice cream is provided
+        item.pop("amenity", None)
```
matkoniecz, line 38: well, this has not worked

matkoniecz: lets leave idea for review before I will start figuring out why code implementing it is broken (likely NSI matching happens after it, so I need to drop these tags in post_process or stop NSI matching)

arrival-spring:
Interesting, I wrote this and used `api_filter = {"icon": {"in": "default,SHOP"}}` which I thought meant that we only got standalone shops (which do exist). Although I remember there were a lot of locations in Portugal, perhaps they do not have the correct type set there, or the filter needs to be changed some other way?

matkoniecz:
Yes, Portugal has unusual density compared to other locations. Fixing so we get actual shops would be obviously better.

locations/spiders/ben_and_jerrys.py
arrival-spring: For reference, see [NSI_MATCHING.md](https://github.com/alltheplaces/alltheplaces/blob/master/docs/NSI_MATCHING.md) for how to disable matching to NSI

arrival-spring:
On the [store locator page](https://www.benjerry.com/ice-cream-near-me/scoop-shops-catering-near-me) this location shows as a "Scoop Shop", just like e.g. "Ben & Jerry's Bondi Beach", which appears to be an actual Ben & Jerry's shop in Sydney.

I can't see an obvious way to distinguish these in the data, except maybe that Bondi Beach has a website and this one doesn't. Maybe all actual shops have websites, and all others don't?

From a glance through some locations, it's a little hard to tell, but skipping items without website I think would definitely skip false locations like this, but may also filter out a few real ones (with only a branch name, it's not easy to tell if it's a real shop).

arrival-spring:
Actually, I think if `jsonshop`/`jsonshopinfo` is null might be an indicator, I'll have a look later

PR #11092, opened 2024-10-14
[kfc_id] Use shared attributes, remove generic website

arrival-spring:


Cj-Malone:
Thanks. Bad coords are in some other projection.

PR #11093, opened 2024-10-14
[kfc_my] Fix parsing of hours

arrival-spring:
This has been finishing with an error in the last few runs, but should now be more robust.

Currently all hours are being parsed


Cj-Malone:
Thanks

PR #11095, opened 2024-10-14
[Godiva JP] Update brand - Needs to be changed in NSI as well.

matkoniecz:
based on provided image, their website and typical markings in OSM: they use shorter name version


Cj-Malone:
Needs to be changed in NSI as well.

matkoniecz:
https://github.com/osmlab/name-suggestion-index/commit/16854c958de4ab3254af47a5c34fbb240ebf838e

PR #11097, opened 2024-10-14
[Landi CH] Change Landi to country store

matkoniecz:
based on feedback from OSM IRC


Cj-Malone:
This needs to be changed in NSI as well.

Cj-Malone:
cc @brawer

brawer (approved):

matkoniecz:
I guess that https://github.com/osmlab/name-suggestion-index/blob/51b949b51712bb4a7f0065d3adb8178c3e822167/data/brands/shop/supermarket.json#L4455 needs to be changed

PR #11134, opened 2024-10-15
Add Papa Johns in various countries

arrival-spring:
Refactor papa_johns_middle_east into a storefinder and then implement it for AE, AZ, JO, KE, KG, KZ, QA, SA and UZ

There may be a better name for it, especially since having checked the websites at https://www.papajohns.com/international/ there are clearly one or two other standardised websites going on, which could have spiders written for them.


arrival-spring:
pytest failing is unrelated to this, but is because of changes in Wikidata (https://www.wikidata.org/wiki/Q38076)

Likewise exception in papa_johns is unrelated, I simply changed to shared attributes

PR #11138, opened 2024-10-16
[better_gb] drop generic social media links

RedAuburn:
but keep unique ones, there are 5 unique facebook links, and all current twitter links are generic.

image


RedAuburn:
it's correctly returning 5 with Facebook links. 
non rate-limited run with this PR:
[better_gb.zip](https://github.com/user-attachments/files/17388200/better_gb.zip)

PR #11145, opened 2024-10-16
Add taco_palenque spider

elevenchars:
Adds a spider for [Taco Palenque](https://tacopalenque.com/) a fast food chain in the US and Mexico.

The addresses from the store locator are very messy, so I had to try to extract the state from the string to help the geocoder disambiguate restaurants that are very close to the US-Mexico border. This could be avoided by only matching within the country boundaries, but I'm not sure if this is possible.

Additionally, I have manually applied the fast food category, since NSI has not released since I added it (https://github.com/osmlab/name-suggestion-index/pull/10041)

This is my first spider, so please let me know if I am reinventing the wheel with the state matching!

### DIFF
--- a/locations/spiders/taco_palenque.py
+++ b/locations/spiders/taco_palenque.py
@@ -1,0 +1,43 @@
+import re
+from html import unescape
+
+from geonamescache import GeonamesCache
+from scrapy import Selector
+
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Thanks

PR #11151, opened 2024-10-16
Fix McDonald's lookup test for NSI

iandees:
The NSI test was failing because the Wikidata record for McDonald's got other social media identities than the one we were testing for. Website seems to come back pretty reliably, so switching to that.


iandees:
This test is really just checking that NSI lookup worked, not that NSI data is correct. Maybe it should just check that the result of the lookup is non-null?

PR #11158, opened 2024-10-16
[TGI Fridays GB] Add website

TheUKHighStreet:


TheUKHighStreet:
I'm wondering whether it is worth doing a bit more to get the opening hours from the store pages.

TheUKHighStreet:
I assume that this timeout is just because it is now crawling all the individual pages rather than a problem.

Cj-Malone:
Thanks.

PR #11167, opened 2024-10-17
[rituals] Add spider (1165 stores)

hfs:
Add [Rituals](https://www.rituals.com/) cosmetics stores worldwide.

### DIFF
--- a/locations/spiders/rituals.py
+++ b/locations/spiders/rituals.py
@@ -1,0 +1,15 @@
+from scrapy.spiders import SitemapSpider
+
+from locations.structured_data_spider import StructuredDataSpider
+
+
+class RitualsSpider(SitemapSpider, StructuredDataSpider):
+    name = "rituals"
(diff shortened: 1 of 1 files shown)


hfs:
The CI pipeline is run with `CLOSESPIDER_TIMEOUT=60`, but that’s not enough for this spider. The spider makes on the order of 5000 requests. The sitemap lists also non-Rituals shops which sell their products, which we are not interested in. Only the pages for their Rituals branded shops have the Structured Data, so the StructuredDataSpider is extracting exactly the right subset of shops.

Maybe you can run the spider locally to review it?

iandees:
Can you tighten up the sitemap filter so that it only makes requests for pages that we want?

hfs:
Unfortunately the wanted shops cannot be detected from the URL. I *can* exclude the most common non-wanted shops based on their brand name, though. Would this be an acceptable solution?

iandees:
Sure, that sounds good.

PR #11169, opened 2024-10-18
Fix Netto Salling

jhutton1:
Sitemap for DE and DK from netto_salling  changed and were no longer providing data.  Only locations for PL were coming from spider.  
I created new spiders for DE and DK, and renamed original netto_salling spider to represent PL only.


panasonic-mafia:
@jhutton1 May i kindly ask to avoid code duplication and re-use DE class for DK spider, please?

PR #11174, opened 2024-10-19
[wuerth_de] Add spider (598 stores)

hfs:
Add [Würth](https://www.wuerth.de/), a B2B hardware store chain in DE.


Cj-Malone:
There is no requirement to use DictParser. In times like this it may have been quicker and easier to just populate a `Feature` manually.

PR #11176, opened 2024-10-20
Update philz_coffee_us.py

AbhiKamp:
country info added as "US".


impiaaa:
Why is this needed? CountryCodeCleanUpPipeline gets the country from the spider name for both of these.

PR #11178, opened 2024-10-20
[ernstings_family] Add spider (1937 shops)

hfs:
[Ernsting's family](https://www.ernstings-family.de/) is a clothes store chain in DE and AT.


iandees:
This is likely failing because you're asking for too much data at a time and their query is timing out. Can you build a couple smaller queries instead?

Cj-Malone:
Thank you

PR #11197, opened 2024-10-21
[sushi_daily] Add spider (1584 locations)

hfs:
Add [Sushi Daily](https://sushidaily.com/), a sushi store-in-store chain in GB, FR, IT, NL, ES, DE, PT, BE, MX, SE, FI, AE.


Cj-Malone:
If this type of data is wanted in OSM, it should also make it into NSI.

PR #11210, opened 2024-10-22
[Smashburger] Remove duplicates

Cj-Malone:
This wasn't actually broken, just getting 403s so this probably wont fix the spider. It runs fine locally from GB, the issue was duplicates, the page contains Microdata for both `Place` and `Restaurant` both were producing POIs. Josh was right `Place` to too generic to be included in the default wanted types.

```python
{'atp/brand/Smashburger': 204,
 'atp/brand_wikidata/Q17061332': 204,
 'atp/category/amenity/fast_food': 204,
 'atp/cdn/cloudflare/response_count': 261,
 'atp/cdn/cloudflare/response_status_count/200': 259,
 'atp/cdn/cloudflare/response_status_count/403': 2,
 'atp/field/branch/missing': 204,
 'atp/field/email/missing': 204,
 'atp/field/opening_hours/missing': 6,
 'atp/field/operator/missing': 204,
 'atp/field/operator_wikidata/missing': 204,
 'atp/field/twitter/missing': 204,
 'atp/item_scraped_host_count/smashburger.com': 204,
 'atp/nsi/perfect_match': 204,
 'downloader/request_bytes': 166538,
 'downloader/request_count': 261,
 'downloader/request_method_count/GET': 261,
 'downloader/response_bytes': 2600751,
 'downloader/response_count': 261,
 'downloader/response_status_count/200': 259,
 'downloader/response_status_count/403': 2,
 'dupefilter/filtered': 133,
 'elapsed_time_seconds': 2.812895,
 'feedexport/success_count/FileFeedStorage': 1,
 'finish_reason': 'finished',
 'finish_time': datetime.datetime(2024, 10, 22, 13, 47, 45, 808990, tzinfo=datetime.timezone.utc),
 'httpcache/hit': 261,
 'httpcompression/response_bytes': 14526788,
 'httpcompression/response_count': 261,
 'item_scraped_count': 204,
 'log_count/INFO': 10,
 'log_count/WARNING': 1,
 'memusage/max': 247898112,
 'memusage/startup': 247898112,
 'request_depth_max': 3,
 'response_received_count': 261,
 'robotstxt/request_count': 1,
 'robotstxt/response_count': 1,
 'robotstxt/response_status_count/200': 1,
 'scheduler/dequeued': 260,
 'scheduler/dequeued/memory': 260,
 'scheduler/enqueued': 260,
 'scheduler/enqueued/memory': 260,
[...]

PR #11213, opened 2024-10-22
[Core] Drop Place from default wanted_types

Cj-Malone:
Rel: https://github.com/alltheplaces/alltheplaces/pull/11210
> the page contains Microdata for both Place and Restaurant both were producing POIs. Josh was right Place to too generic to be included in the default wanted types.

I would also like to drop some of the rare types from the default list as well, but I don't actually have a valid reason right now other than personal preference.

### DIFF
--- a/locations/spiders/hsbc_tw.py
+++ b/locations/spiders/hsbc_tw.py
@@ -1,14 +1,15 @@
 import scrapy
 from requests_cache import Response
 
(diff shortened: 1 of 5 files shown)


panasonic-mafia (approved):

iandees:
For what it's worth, I would prefer that spiders specify the types and that there not be a default list.

PR #11233, opened 2024-10-25
[Casey's General Store] Fix spider

Cj-Malone:
I've not ran the spider though, it's very heavy on the geoblocking, but looking at the last run, the pages were fetched, just not parsed correctly.


iandees:
Can you add a pre_process to check for the special hours case showing up in the log?

PR #11250, opened 2024-10-28
[fio_banka_atm] Add spider (242 locations)

petrzjunior:
Follow-up to #11249, this one contains ATMs.

Peculiarities:
- No ref-like field, using goo.gl identifier where available, street as a fallback
- City is parsed from header by removing notes in parantheses
- Some branches contain Google Maps links, coordinates parsing as a fallback
- Most branches are non-stop
- There is currently no way to specify holiday (PH) opening hours

### DIFF
--- a/locations/spiders/fio_banka_atm.py
+++ b/locations/spiders/fio_banka_atm.py
@@ -1,0 +1,97 @@
+import re
+import time
+
+from scrapy.linkextractors import LinkExtractor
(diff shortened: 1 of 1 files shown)


iandees:
Can you remove building a `ref` and set `no_refs = True` instead?

petrzjunior:
@iandees Fixed

Issue #11286, opened 2024-10-30
Nike missing stores


TheUKHighStreet:
### Spider name

nike

### Example bad data or results

The spider is missing stores - example the Bournemouth store. It is in the store finder.
https://storeviews-cdn.risedomain-prod.nikecloud.com/store-locations-static.json
efa568a3-e2aa-4977-bffb-1718b406990b

I can't see why it is missing at the moment.

TheUKHighStreet:
The list of stores in the storefinder has 3628 stores in the storesGeo list. There are 1812 in the spider.

TheUKHighStreet:
I've narrowed it down to the opening hours.

TheUKHighStreet:
This is a problem when there is a single digit minute
       closing_m = closing[len(closing) - 3 :].replace("M", "") if "M" in closing else "0"

TheUKHighStreet:
I can fix this by removing the "H".

PR #11287, opened 2024-10-30
[Nike] Fix missing stores

TheUKHighStreet:


TheUKHighStreet:
This is to fix this issue https://github.com/alltheplaces/alltheplaces/issues/11286

PR #11288, opened 2024-10-30
[Moss GB] Add spider

TheUKHighStreet:


TheUKHighStreet:
The store that doesn't have coordinates seems to be on a cruise ship. https://www.moss.co.uk/store/510-harding-brothers
Is it possible to remove a store in the post_process_item if there are no coordinates?
Or once an item exists it can it not be deleted?

PR #11289, opened 2024-10-30
[Montblanc] Fix Spider

mohit-sharma2194:
`Fixes : opening_hours has been removed to fix spider`

{'atp/brand/Montblanc': 444,
 'atp/brand_wikidata/Q142691': 444,
 'atp/category/shop/fashion_accessories': 444,
 'atp/field/branch/missing': 444,
 'atp/field/city/missing': 444,
 'atp/field/country/from_reverse_geocoding': 26,
 'atp/field/country/invalid': 1,
 'atp/field/email/missing': 444,
 'atp/field/image/missing': 444,
 'atp/field/lat/invalid': 1,
 'atp/field/lat/missing': 6,
 'atp/field/lon/missing': 6,
 'atp/field/opening_hours/missing': 444,
 'atp/field/operator/missing': 444,
 'atp/field/operator_wikidata/missing': 444,
 'atp/field/phone/missing': 444,
 'atp/field/postcode/missing': 444,
 'atp/field/state/from_reverse_geocoding': 33,
 'atp/field/state/missing': 386,
 'atp/field/street_address/missing': 444,
 'atp/field/twitter/missing': 444,
 'atp/item_scraped_host_count/stores.montblanc.com': 444,
 'atp/nsi/perfect_match': 444,
 'downloader/request_bytes': 1592,
 'downloader/request_count': 2,
 'downloader/request_method_count/GET': 1,
 'downloader/request_method_count/POST': 1,
 'downloader/response_bytes': 92173,
 'downloader/response_count': 2,
 'downloader/response_status_count/200': 1,
 'downloader/response_status_count/404': 1,
 'elapsed_time_seconds': 4.902041,
 'finish_reason': 'finished',
 'finish_time': datetime.datetime(2024, 10, 30, 15, 56, 11, 272388, tzinfo=datetime.timezone.utc),
 'httpcompression/response_bytes': 573460,
 'httpcompression/response_count': 2,
 'item_scraped_count': 444,
 'log_count/DEBUG': 863,
 'log_count/INFO': 9,
 'response_received_count': 2,
 'robotstxt/request_count': 1,
 'robotstxt/response_count': 1,
 'robotstxt/response_status_count/404': 1,
 'scheduler/dequeued': 1,
 'scheduler/dequeued/memory': 1,
 'scheduler/enqueued': 1,
 'scheduler/enqueued/memory': 1,
[...]

PR #11293, opened 2024-10-31
[MontBlanc] Remove opening times

karen-kerr:
Spider was failing due to error in parsing now non existent opening times. Code removed.


impiaaa:
This looks like a duplicate of https://github.com/alltheplaces/alltheplaces/pull/11289

PR #11295, opened 2024-11-01
[whataburger] move @itemprop="streetAddress" from addr_full to street_address

matkoniecz:


matkoniecz:
Not looked at details, but 'fastned', 'ljsilvers', 'wells_fargo', 'coffee_like' were also flagged as suspect in a similar way.

PR #11333, opened 2024-11-10
[Robert Dyas GB] Add spider

TheUKHighStreet:


TheUKHighStreet:
Is it ok to add the opening hours like this? Since they already existed in the correct format - they just aren't grouped.

PR #11352, opened 2024-11-11
[Lyle and Scott] Add spider

TheUKHighStreet:


locations/spiders/lyle_and_scott.py
```diff
@@ -0,0 +1,42 @@
+# this is very dependent ono the structure of the page not changing
```
Cj-Malone, line 1: Indeed, looks fragile. And for just 14 locations... I'd recommend just updating OSM.

PR #11373, opened 2024-11-13
auto_nation_us: requires US proxy

davidhicks:


davidhicks:
PR#9993 already tried this. Zyte has [blocked](https://docs.zyte.com/zyte-api/usage/errors.html#invalid-requests) autonation.com from being scraped and we don't have any other proxy services pre-configured to be used with ATP.

PR #11375, opened 2024-11-13
[Best & Less AU] Fix opening hours parsing

karen-kerr:


Cj-Malone:
Thanks. I've moved the try statement into the main loop, this way we don't have to pass the url or other ref·

PR #11399, opened 2024-11-15
[Kroger US] Fix phone number error

karen-kerr:


Cj-Malone:
Thanks, FYI this could have been done in 1 line `(location.get("phoneNumber") or {}).get("raw")`

PR #11406, opened 2024-11-15
[Change Group] Add spider

TheUKHighStreet:
The opening hours are a mess.


TheUKHighStreet:
Some of the branches are missing locations in the json file.

PR #11426, opened 2024-11-18
[klier_hair_group] Add spiders (728 locations)

hfs:
Klier Hair Group is a hairdresser chain in DE with several brands:

- Cosmo
- Essanelle
- Hair Express
- Klier
- Supercut

They share the same website structure, so they can be scraped using a common storefinder class.

### DIFF
--- a/locations/spiders/cosmo_de.py
+++ b/locations/spiders/cosmo_de.py
@@ -1,0 +1,7 @@
+from locations.storefinders.klier_hair_group import KlierHairGroupSpider
+
+
+class CosmoDESpider(KlierHairGroupSpider):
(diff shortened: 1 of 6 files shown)


hfs:
Let me know if it’s fine to add them in one Pull Request, or if I should split them.

hfs:
Some locations have invalid coordinates 28.6478898,180.0. Please advise if I should

* leave them invalid as they are,
* remove these locations altogether, or
* keep the locations but remove the invalid coordinates

iandees:
I believe the lat/lon filtering code will remove the bad lat/lons.

PR #11439, opened 2024-11-20
Fix broken postbank_bg spider

Dimitar5555:


Dimitar5555:
It works on my machine. I'm not sure why it's giving 403 error here.

iandees:
Likely getting blocked because of the IP range we request from.

iandees:
I think the proxy we normally use also doesn't know what to do with this domain. We'll leave it for now and hopefully they resolve it.

Dimitar5555:
> Likely getting blocked because of the IP range we request from.

Their site might be restricted for users outside of Bulgaria... It's not the first local bank that does it.

PR #11479, opened 2024-11-24
[new_england_dot_us] add 353 cameras in the US north east

mjoe999:
The start of a more generic mechanism for calling back in the pipeline code to a spider provided hook? There may be better ways but nothing apparent in the framework.

### DIFF
--- a/locations/pipelines/state_clean_up.py
+++ b/locations/pipelines/state_clean_up.py
@@ -1,5 +1,6 @@
 import reverse_geocoder
 from geonamescache import GeonamesCache
+from scrapy.exceptions import DropItem
 
 from locations.items import get_lat_lon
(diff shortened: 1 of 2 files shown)


iandees:
Why not just set `state` in `item_attributes`?

mjoe999:
I can only  get the state from the lat/lon as the "New England" crew do not supply anything else. To get the state I would have to call/replicate the pipeline code. Now I could do this but I kind of thought we have a lot of heavy lifting pipeline code in the project and it's more than likely people want to get callbacks at various points. See the TODO comments in #11483 for another example, granted in the same "domain".

davidhicks:
Replaced this proposed spider with #11543 which has more information available in the alternate API call made, including the state the camera is located within.

PR #11482, opened 2024-11-24
Add Wisconsin DOT cameras

iandees:


mjoe999:
Don't get addicted to this :-) This spider should simplify if you inherit from NevadaDotUSSpider ... see #11480

PR #11490, opened 2024-11-25
[pennsylvania_dot_us] add 1321 cameras in Pennsylvania

mjoe999:
@iandees here's another one and I'd like your view on us extending the pipeline code to handle a naming convention of `_state_country.py` i.e. here I would suggest `dot_pa_us.py`. The pipeline in the rare case that we did not want it to stamp out the state / country if not already set could be disabled i.e. `skip_auto_cc_spider_name`. The US is where this works well and it would enable things like:

``
$ ls locations/spiders/*_al_us.py
``

.. the more we organise the spiders by convention the better? Let me know what you think and I am happy to work something up either in this PR or outside it.

### DIFF
--- a/locations/spiders/pennsylvania_dot_us.py
+++ b/locations/spiders/pennsylvania_dot_us.py
@@ -1,0 +1,19 @@
+from typing import Iterable
+
+from scrapy.http import Response
+
+from locations.items import Feature
+from locations.spiders.nevada_dot_us import NevadaDotUSSpider
+
+
+class PennsylvaniaDotUSSpider(NevadaDotUSSpider):
+    name = "pennsylvania_dot_us"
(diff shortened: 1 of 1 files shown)


iandees:
I'm still pondering these.

The `super()` call from the post process callback here feels a bit weird and non-obvious to me. I don't like inferring data from the filename for the spider[^1], so that doesn't feel like a good solution.

For now, I think I would prefer that we just not set state if we can't easily get it from the data source.

[^1]: Back when I was adding the code that checks for spider name, class name, and filename, I kind of wanted to remove the convention altogether.

davidhicks:
Replaced by proposed TravelIQSpider in #11538

JoeDixon-TomTom:
All fine and in this case the more obvious thing was to set state in `item_attributes` ..I'll try harder :-)

PR #11491, opened 2024-11-25
[new_york_dot_us] add 2028 camera in New York state, and some outside

mjoe999:
Good to merge I think but once again would benefit from better "state" handling.


JoeDixon-TomTom:
I've edited the pipeline code ot expose the reverse state geocode method. Not the end of this story but still useful I think.

davidhicks:
Thanks! I've replaced this proposed spider with a new TravelIQ spider introduced in #11538 (TravelIQ is the common software used by many North American transport agencies)

PR #11538, opened 2024-12-01
Add TravelIQ and TravelIQWebCameras storefinders

davidhicks:
Add a new TravelIQ storefinder for a common traffic information system/software used extensively throughout Northern America.

Add a number of new spiders using this TravelIQ system.

Convert some existing 511 spiders to this new TravelIQ store finder.


mjoe999:
Ah, I'm catching up now. Looks fine. My only comment would be we are now requiring API keys which is always unfortunate but this does not make the spiders unique in that respect.

iandees:
Yea, I'm a little weary of the API key too, but at least some of them appear to be fairly static. @davidhicks did you really go get all these DOT's apps and extract API keys or are you finding them on their website?

davidhicks:
I've only found this API and the API keys being used by the Android applications (which generally Travel-IQ create for the state transportation department).

The web application (for desktop browsers) doesn't use this API, instead it uses an alternative 100-at-a-time paged API that I had to write a different spider for at https://github.com/alltheplaces/alltheplaces/pull/11543

I preferred the mobile API because it is less work for the server--only 3 requests needed (robots.txt, developer documentation page to find list of features provided, and API endpoint for listing traffic cameras). Add one request for each other type of feature we could extract in the future such as bridges or variable message signs. If the web application API at https://github.com/alltheplaces/alltheplaces/pull/11543 were used we'd be looking at 30-50 requests for the larger states to get all traffic camera information.

PR #11542, opened 2024-12-01
virginia_dot_us: change to JSONBlobSpider and extract more webcam feeds

davidhicks:


iandees:
I don't know that I like the `;`-concatenated list of URLs for the webcam. I'd rather just pick one and use separate tags for the others if they're interesting enough.

PR #11545, opened 2024-12-01
Add Illinois DOT (US) (1544 traffic cameras)

davidhicks:


davidhicks:
Combines cameras in Illinois with a few other cameras in states of Indiana, Wisconsin, Kentucky, etc. Most of these other states already have a different non-shared traffic camera map/list which is authoritative. This Illinois spider therefore needs to find a smart way to drop/ignore cameras from other states.

davidhicks:
Replaced by #12390 12390

PR #11560, opened 2024-12-02
[Carrefour BR] Add requires proxy

karen-kerr:


karen-kerr:
Closing as now running ok.

PR #11575, opened 2024-12-04
Add Alabama DOT (US) (588 traffic cameras)

davidhicks:


locations/spiders/alabama_department_of_transportation_us.py
```diff
@@ -0,0 +1,29 @@
...
+    allowed_domains = ["api.algotraffic.com"]
+    start_urls = ["https://api.algotraffic.com/v3.0/Cameras"]
+
+    def post_process_item(self, item: Feature, response: Response, feature: dict) -> Iterable[Feature]:
+        item["name"] = " ".join(
+            [feature["location"]["displayRouteDesignator"], feature["location"]["displayCrossStreet"]]
+        )
+        item["lat"] = feature["location"]["latitude"]
+        item["lon"] = feature["location"]["longitude"]
+        item["extras"]["contact:webcam"] = ";".join(filter(None, [feature["hlsUrl"], feature["imageUrl"]]))
```
iandees, line 27: We can do this later, but I'd really like for the value of `contact:webcam` to be a single URL, not `;`-concatenated URLs. We should come up with another key if the URLs lead to different images.

davidhicks: Should we use https://wiki.openstreetmap.org/wiki/Multiple_values#Numbered_suffix_in_key or https://wiki.openstreetmap.org/wiki/Lanes#Two_driving_directions (two semicolons in the tag name) as an alternative to https://wiki.openstreetmap.org/wiki/Multiple_values#Semi-colon_value_separator ?

By comparison, `contact:phone` seems to use semicolons the most, almost never uses https://wiki.openstreetmap.org/wiki/Multiple_values#Numbered_suffix_in_key and occasionally uses https://wiki.openstreetmap.org/wiki/Lanes#Two_driving_directions for `contact:phone:US` vs. `contact:phone:GB` etc.

PR #11580, opened 2024-12-04
Add Castle Rock OneWeb storefinder and convert/add known ATIS spiders

davidhicks:


davidhicks:
Tennessee also uses this ATIS and has a site at https://smartway.tn.gov/traffic however this is currently not loading for me so I cannot add it to this pull request for now.

iandees:
The TN site has a notice across the top that reads:

> Notice: Smartway cameras will be undergoing maintenance this Wednesday, 12/4 from 9 am until 2 pm CT. Users may experience disruptions at this time. Thank you for your patience.

PR #11588, opened 2024-12-04
McGrath: Rewrite to use new API 

impiaaa:


locations/react_server_components.py
```diff
@@ -0,0 +1,39 @@
+import array
+import itertools
+import json
+from typing import Any, Iterable, Iterator
+
+ARRAY_TYPES = {"O": "b", "o": "B", "S": "h", "s": "H", "L": "l", "l": "L", "G": "f", "g": "d", "M": "q", "m": "Q"}
+
+
+def parse_rsc(data_raw: Iterable[int]) -> Iterator[tuple[int, Any]]:
```
davidhicks, line 9: Thanks for writing this deserialization function! I've tried to track down the authoritative reference and seemingly it is the parseModelString() function at React's [/packages/react-server/src/ReactFlightReplyServer.js](https://github.com/facebook/react/blob/main/packages/react-server/src/ReactFlightReplyServer.js)

If this is correct, could a reference be provided to the React source code identified above, as it seems to document the meaning of all these abbreviations such as "G", "f", "g", "d", etc. I haven't found any authoritative React documentation in human-readable format that describes this serialization format.

davidhicks (changes requested):

locations/react_server_components.py
davidhicks: Thank you, the documentation added is great! I'm not a React person but it's astonishing to me how poorly documented that serialization format appears to be.

davidhicks (approved):

iandees:
Nice, thanks for adding the parser!

PR #11603, opened 2024-12-06
[DOCS] Add apply_category info

Cj-Malone:
`item_attributes = {"extras": Categories.XYZ.value}` should be discouraged.

1. In cases where we do `item.update(SpecificSpider.item_attributes)` it will nuke any already given `extras`.
2. And for future proofing `apply_category` should be centralised so we can change it once for everything not have to edit 4k spiders. Otherwise we'll be heading back to when some spiders we're assigning things like `amenity:atm=True`.

### DIFF
--- a/templates/spiders/crawl.tmpl
+++ b/templates/spiders/crawl.tmpl
@@ -22,6 +22,6 @@ class $classname(CrawlSpider):
 
     def parse_item(self, response):
         item = Feature()
-        # item["name"] = response.xpath('//div[@id="name"]').get()
+        # item["ref"] = response.xpath('//div[@id="store_id"]').get()
         # apply_category(Categories.SHOP_XYZ, item)
-        return item
+        yield item
--- a/templates/spiders/crawl.tmpl
+++ b/templates/spiders/crawl.tmpl
@@ -3,6 +3,7 @@ from scrapy.linkextractors import LinkExtractor
 from scrapy.spiders import CrawlSpider, Rule
 
 from locations.categories import Categories, apply_category
+from locations.items import Feature
 
 
 class $classname(CrawlSpider):
@@ -20,7 +21,7 @@ class $classname(CrawlSpider):
     rules = (Rule(LinkExtractor(allow=r"Items/"), callback="parse_item", follow=True),)
 
     def parse_item(self, response):
-        item = {}
+        item = Feature()
         # item["domain_id"] = response.xpath('//input[@id="sid"]/@value').get()
         # item["name"] = response.xpath('//div[@id="name"]').get()
         # item["description"] = response.xpath('//div[@id="description"]').get()
--- a/templates/spiders/crawl.tmpl
+++ b/templates/spiders/crawl.tmpl
@@ -12,8 +12,7 @@ class $classname(CrawlSpider):
         # "brand": "$brand",
         # "brand_wikidata": "$brand_wikidata",
         # "operator": "$operator",
-        # "operator_wikidata": "$operator_wikidata",
-        # "extras": Categories.SHOP_XYZ.value
+        # "operator_wikidata": "$operator_wikidata"
     }
 
     rules = (Rule(LinkExtractor(allow=r"Items/"), callback="parse_item", follow=True),)
--- a/templates/spiders/crawl.tmpl
+++ b/templates/spiders/crawl.tmpl
@@ -22,8 +22,6 @@ class $classname(CrawlSpider):
 
     def parse_item(self, response):
         item = Feature()
-        # item["domain_id"] = response.xpath('//input[@id="sid"]/@value').get()
         # item["name"] = response.xpath('//div[@id="name"]').get()
-        # item["description"] = response.xpath('//div[@id="description"]').get()
         # apply_category(Categories.SHOP_XYZ, item)
         return item
--- a/templates/spiders/crawl.tmpl
+++ b/templates/spiders/crawl.tmpl
@@ -2,6 +2,8 @@ import scrapy
 from scrapy.linkextractors import LinkExtractor
 from scrapy.spiders import CrawlSpider, Rule
 
+from locations.categories import Categories, apply_category
+
 
 class $classname(CrawlSpider):
     name = "$name"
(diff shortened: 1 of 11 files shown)


davidhicks:
I have a branch almost ready to propose which revamps `locations/pipelines/apply_nsi_categories.py` to improve NSI matching. The most significant change is to require a top level category (new criteria) in addition the existing requirement to supply a brand/operator Wikidata item, and checking that ATP items have a country and state corresponding to NSIs ISO 3166-1 alpha-2/ISO 3166-2 codes. What I have found is that NSI may know only of a government department operating parks/playgrounds, but the same department also operates public sports fields. When an ATP spider is proposed for extracting public sports fields operated by that government department, our current matching mechanism will apply parks/playgrounds tags because NSI is not aware that the same Wikidata item operates sports fields.

Thus the minimum matching criteria would be a tuple `(operator_wikidata:str or brand_wikidata:str, top_level_categories:list[dict]], country:str or state:str)` where `state` is the first level subdivision and ATP's `state` field would be converted to an ISO 3166-2 code for matching against NSIs `locationSet`.

I was wondering whether we could promote a top level category of the feature to be an attribute of `Feature`, and eventually move top level attribute information out of `Extras`?

davidhicks:
My comment above only impacts `CATEGORIES.md` documentation and `templates/spiders/basic.tmpl` also needing an `add_category` call as the other templates have.

templates/spiders/crawl.tmpl
```diff
@@ -13,14 +16,12 @@ class $classname(CrawlSpider):
...
 
     rules = (Rule(LinkExtractor(allow=r"Items/"), callback="parse_item", follow=True),)
 
     def parse_item(self, response):
-        item = {}
-        #item["domain_id"] = response.xpath('//input[@id="sid"]/@value').get()
-        #item["name"] = response.xpath('//div[@id="name"]').get()
-        #item["description"] = response.xpath('//div[@id="description"]').get()
+        item = Feature()
+        # item["name"] = response.xpath('//div[@id="name"]').get()
```
davidhicks, line 25: Suggest changing this default example to `ref` rather than `name` because `ref` is more-or-less mandatory, but `name` isn't.

templates/spiders/csvfeed.tmpl
```diff
@@ -14,16 +17,15 @@ class $classname(CSVFeedSpider):
...
     #    return response
 
     def parse_row(self, response, row):
-        i = {}
-        #i["url"] = row["url"]
-        #i["name"] = row["name"]
-        #i["description"] = row["description"]
+        i = Feature()
+        # i["website"] = row["url"]
+        # i["name"] = row["name"]
```
davidhicks, line 29: Suggest changing this default example to `ref` rather than `name` because `ref` is more-or-less mandatory, but `name` isn't.

templates/spiders/json_blob.tmpl
```diff
@@ -11,7 +12,6 @@ class $classname(JSONBlobSpider):
         # "brand_wikidata": "$brand_wikidata",
         # "operator": "$operator",
         # "operator_wikidata": "$operator_wikidata",
-        # "extras": Categories.SHOP_XYZ.value
     }
     start_urls = ["$url"]
     # no_refs = True
```
davidhicks, line 17: I don't mind `no_refs` being listed but it is a very rare occasion we'd be using it, so may not be necessary to include in templates?

templates/spiders/xmlfeed.tmpl
```diff
@@ -13,12 +16,11 @@ class $classname(XMLFeedSpider):
...
-        # "extras": Categories.SHOP_XYZ.value
     }
 
     def parse_node(self, response, selector):
-        item = {}
-        #item["url"] = selector.select("url").get()
-        #item["name"] = selector.select("name").get()
-        #item["description"] = selector.select("description").get()
+        item = Feature()
+        # item["name"] = selector.select("name").get()
```
davidhicks, line 23: Suggest changing this default example to `ref` rather than `name` because `ref` is more-or-less mandatory, but `name` isn't.

davidhicks (changes requested):
Minor recommendations added for templates for consideration.

Cj-Malone:
I've added more refs and removed `no_refs` from the templates, makes sense to me thanks. I've no issue some templates not using `apply_category` after all, most spider don't.

Cj-Malone:
As for the NSI pipeline. I've considered it before, but we'd have to update ~3k spiders to make category a requirement. + We lose the benefit of automatically changing category when NSI moves something, the maintenance to keep `brand` in sync is already not being done.

We absolutely should not be matching to the wrong category (park in your example) when we output something else (sports fields). I think this only happens when NSI has 1 entry, but we have a different category.

I don't want to move in the direction of adding `amenity`, `shop`, `tourism` and all the rest to `Feature`. It's just Schrödinger's categories. I actually want to move towards something like `item["category"] = "amenity/parking"`. This is how NSI does it. It's how we were warned to do it. The output would still be `"amenity": "parking"`. But before we get anywhere, we need to be using the shared `apply_category`.

davidhicks:
+1 to `item["category"]` or similar for consideration another day. I like it over `apply_category` because:

1. We have a lot of spiders (mostly for storefinders) that do not currently require implementation of a `post_process_item` function. If they do have to have such a function just to call `apply_category`, all of these spiders become longer and a little bit more effort to write. It would be easier to write `item_attributes = {"operator": "XYZ", "operator_wikidata": "Q123", "category": "amenity/parking"}` so that our NSI matching has everything it needs to fill in any additional OSM tags from the correctly matching NSI item.
2. The enumerations under `Categories` (categories.py) are an indirection and complexity which I'm not convinced adds much value over just specifying `amenity/parking` as a text string or `("amenity", "parking)` as a tuple.

My only concern remaining with this branch was the concept of an item not having a category specifically designated, yet many of NSIs operators needing a category to disambiguate an operator of multiple types of things--public rubbish bins, public restrooms, playgrounds, etc. I can see how this doesn't impact NSI brands because it would be unusual for the same brand Wikidata item to open two different types of shop under the same brand name in the same region.

I'll send through my other proposed branch for improving NSI matching and we can further consider the issue there.

PR #11606, opened 2024-12-07
[Topps Tiles GB] Test spider

TheUKHighStreet:


TheUKHighStreet:
I removed  requires_proxy = True to see what it was getting stuck on and it seems to work.

iandees:
Nice.

PR #11608, opened 2024-12-07
feat: add Coop Centro Italia/Superconti (IT, 76)

### LINKED ISSUE #10453: superconti_it needs rewrite
### Spider name

superconti_it

### Log output

_No response_

### Backtrace (if applicable)

_No response_

_No response_

_No response_

### Possible causes?

Now using:

curl 'https://storelocator.stagingfattoria.it/api/stores' \
  -H 'accept: */*' \
  -H 'accept-language: en-GB,en;q=0.9,en-US;q=0.8,de;q=0.7,es;q=0.6,fr;q=0.5' \
  -H 'content-type: application/x-www-form-urlencoded; charset=UTF-8' \
  -H 'dnt: 1' \
  -H 'origin: https://www.superconti.eu' \
  -H 'priority: u=1, i' \
  -H 'referer: https://www.superconti.eu/' \
  -H 'sec-ch-ua: "Chromium";v="128", "Not;A=Brand";v="24", "Google Chrome";v="128"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "Windows"' \
  -H 'sec-fetch-dest: empty' \
  -H 'sec-fetch-mode: cors' \
  -H 'sec-fetch-site: cross-site' \
  -H 'sec-gpc: 1' \
  -H 'user-agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/128.0.0.0 Safari/537.36' \
  --data-raw 'access_token=d116ed29bd172fc7e82b43589237a5c1b10861b6&insegna=superconti&store_id='

McCio:
Fixes #10453


matkoniecz:
Does it make sense to list departments as ;-delimited shop value?

For example `shop=newsagent;supermarket` is still simply `shop=supermarket`

The same for say `shop = books;electronics;supermarket`

Or maybe those are fully separate shops at that location? (no idea) Then it would be better to emit multiple points

McCio:
@matkoniecz 
Yeah that was a strange result I saw, in OSM I'd prefer the [sells:*=yes proposal](https://wiki.openstreetmap.org/wiki/Proposal:Sells:) - I have a plan of completing the Coop Italia scrapers, if it makes sense I'll make the change at that point
The reason for specifying those tags is that (in Italy at least) they are uncommon supermarket items (to the point that the chain itself specifies which stores are selling them)

matkoniecz:
`sells:` seems more reasonable to me, even for items unusual to be found in supermarkets in general

;-delimited makes not really possible to distinguish shops with multiple roles from multiple different shops emitted as one entry, and supermarket really can be though as selling wide range of stuff (shop selling specifically flowers and honey and nothing else would be far tricker than supermarket selling flowers and honey among its products)

PR #11629, opened 2024-12-11
[opentransportdata.swiss] Change dataset attribution to optional

brawer:
The upstream data source has waived the attribution requirement for All The Places (”database”) and its data consumers (“licence holders of the database”).

* https://opentransportdata.swiss/en/authorised-databases/
* https://opentransportdata.swiss/en/terms-of-use/#511_Exemption_for_databases


brawer:
The spider crashes because the feed format has slightly changed last Sunday. I'll look into this, but the change in attribution (this change here) is independent of this.

Cj-Malone:
Thank you and look forward to the fix.

PR #11644, opened 2024-12-13
Add espressolab spider (~250 stores)

styrowolf:
Espressolab is a coffee shop that has around 250 stores in Turkey, Iraq, Egypt, Morocco, etc.

Wikidata: https://www.wikidata.org/wiki/Q97599059


locations/spiders/espressolab.py
```diff
@@ -0,0 +1,58 @@
...
+from locations.categories import Categories, apply_category
+from locations.pipelines.address_clean_up import clean_address
+
+STORES_PAGE_URL = "https://espressolab.com/subeler/"
+WEBSITE_ROOT = "https://espressolab.com"
+STORE_LINKS_XPATH = "/html/body/form/div[4]/div[4]/div[1]/div/div/div/div[1]/div/div/div[2]/ul/li/a/@href"
+STORE_NAME_XPATH = "//h1/text()"
+STORE_ADDRESS_XPATH = "//div[contains(@class, 'address')]"
+MAP_SCRIPT_XPATH = "/html/head/script[7]/text()"
+MAP_SCRIPT_REGEX = re.compile(r"google\.maps\.LatLng\([ ]*([-0-9.]*)[ ]*,[ ]*([-0-9.]*)[ ]*\);")
```
iandees, line 17: No need to pull these all out into constants. It makes it harder to read the code below.

iandees (changes requested):

locations/spiders/espressolab.py
styrowolf: I've put these into the code below in b073d8fb73606c1e29d75b3ab85a239158859e36.

styrowolf:
I don't get a 403 on the first web page on my computer. This website uses Cloudflare so that might be the reason for blocking. What should I try to get it working on the scraperbot?

Cj-Malone (commented):
We should be outputting a Feature, not a dict. Proxy may be needed for the 403. We probably don't need lxml here.

styrowolf:
> We should be outputting a Feature, not a dict. Proxy may be needed for the 403. We probably don't need lxml here.

I use lxml to extract the text since the text is arbitrarily nested in different elements depending on the store page. What should I use instead?

impiaaa:
Here's how I'd do it: https://github.com/alltheplaces/alltheplaces/commit/e551df8b6b21bcc8c1ce5c8c7e310a2493dac520 If you want I can put up a separate PR.

iandees:
Sure, if you want to make another PR that'd be helpful.

PR #11646, opened 2024-12-13
[Selfast ZA] Fix non running spider

karen-kerr:


iandees:
Are the other spiders that use `WPStoreLocatorSpider` also broken? If so, we should fix those too.

PR #11651, opened 2024-12-13
[ALDI FR] Restore name and brand

Cj-Malone:
This is the quick fix. Long term we should support "FX" from NSI.


iandees:
What is "FX"?

Cj-Malone:
A fake country code for just France, not it's colonies.

https://location-conflation.com/?locationSet=%7B%20include%3A%20%5B%27fr%27%5D%20%7D%0A

PR #11672, opened 2024-12-18
Rewrite and improve NSI category matching for improved accuracy

davidhicks:
* Handle NSI entities that are only located in first level subdivisions (such as "US-TX" being a valid location, but "US" is not because the brand/operator doesn't operate across the entire country).
* Ensure that NSI operator entities (such as local government bodies) are only matched if the ATP item declares a category which NSI knows the operator is associated with. Without this more pedantic matching, it is very common for current NSI matching to think an ATP item without a category is a sports pitch, simply because NSI doesn't yet know the same operator also operates playgrounds.
* Significantly increase test coverage for NSI category matching of ATP items.
* Improve country code extraction from the spider name to correctly stop matching at the first word which is NOT an ISO 3166-1 alpha-2 code. Some spiders may have suffixes such as "_ok_fr_gb" where only "FR" and "GB" are valid codes. Country code extraction is now smart enough to extract "FR" and "GB" but ignore "OK".


davidhicks:
Note to self: Need to fix `country_code_from_spider_name()` to handle case insensitivity and special manual replacements currently provided in UNHANDLED_COUNTRY_MAPPINGS.

davidhicks:
Ready for review/merge.

Note that I didn't need to ensure `country_code_from_spider_name()` recognises `UNHANDLED_COUNTRY_MAPPINGS` because this would have only applied to `UK` -> `GB` and we don't have any spiders remaining with a suffix of `_uk` (they've all been fixed to use `_gb` instead to match ISO 3166-1 alpha-2 codes).

PR #11688, opened 2024-12-23
[Invitro RU] Try Fixing  Spider by flagging as requires proxy

ehsan-firdaus:
[Website](https://www.invitro.ru/) itself is not accessible. Perhaps only available for certain regions, might require proxy.

image


panasonic-mafia (approved):
Yea, proxy should help, it works from RU ip.

PR #11703, opened 2024-12-25
put street address into street address

matkoniecz:


matkoniecz:
>     ⚠️ Latitude or Longitude is missing on 2 items

seems unlikely to be caused by this PR

PR #11704, opened 2024-12-25
retail_apparel_group - this looks like addr_street, not addr:full

matkoniecz:


matkoniecz:
```
  File "/codebuild/output/src1840015659/src/github.com/alltheplaces/alltheplaces/locations/spiders/retail_apparel_group.py", line 131, in parse
    item["addr_street"] = location["address"]["addr_full"]
    ~~~~^^^^^^^^^^^^^^^
  File "/home/ubuntu/.pyenv/versions/3.11.11/lib/python3.11/site-packages/scrapy/item.py", line 98, in __setitem__
    raise KeyError(f"{self.__class__.__name__} does not support field: {key}")
KeyError: 'Feature does not support field: addr_street'
```

matkoniecz:
many include also shop location within mall but ATP has no clear schema for this from what I see

PR #11706, opened 2024-12-26
[billa] tag values should be strings, also ones containing numbers

matkoniecz:


matkoniecz:
>     🚨 Category is not set on 271 items

Likely not my fault

matkoniecz:
maybe catchall pipeline would be nice rather doing it in each affected spider separately

iandees:
We don't make a promise that property values are always strings. I can see why it's useful when the destination is OSM tags, but I can also see it being useful to keep integers around for analysis. 🤔 

I suppose it's easy to convert from string back to integer in the analysis case.

matkoniecz:
interesting, https://github.com/alltheplaces/alltheplaces/issues/4878#issuecomment-1749686729 seems to suggest to me that plan is/used to be a different?

having consistency in either direction would be nice, or at least having it documented that numeric values may appear both as stringified number or as numbers in https://github.com/alltheplaces/alltheplaces/blob/master/DATA_FORMAT.md

(on my side I have already https://codeberg.org/matkoniecz/list_how_openstreetmap_can_be_improved_with_alltheplaces_data/src/branch/master/qa.py#L20 - it was detected by my ATP data quality checker as potential simple PR intended to be helpful)

matkoniecz:
@iandees should I make PR to add info to [DATA_FORMAT.md](https://github.com/alltheplaces/alltheplaces/blob/master/DATA_FORMAT.md) that numeric fields can be both integer or string?

PR #11707, opened 2024-12-26
[nravizza_by] drop website=null

matkoniecz:
Maybe it can be generally dropped somewhere

```
{
    '@source_uri': 'https://nravizza.by/page9609244.html',
    '@spider': 'nravizza_by',
    'amenity': 'fast_food',
    'cuisine': 'pizza',
    'takeaway': 'yes',
    'addr:street_address': 'Лида, ул. Рыбиновского, 78',
    'addr:country': 'BY',
    'name': 'Nravizza Pizza',
    'website': 'null',
    'brand': 'Nravizza',
    'brand:wikidata': 'Q125747982',
    'nsi_id': 'nravizzapizza-381458'
}
```

### DIFF
--- a/locations/spiders/nravizza_by.py
+++ b/locations/spiders/nravizza_by.py
@@ -19,6 +19,6 @@ class NravizzaBYSpider(Spider):
         ):
             item = DictParser.parse(location)
             item["street_address"] = item.pop("name")
-            if item.get("website") == 'null':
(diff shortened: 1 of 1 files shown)


iandees:
In this case `'null'` seems to be specific to the spider, so I'll merge this.

A field is not included in the output if we set it to `None` (the Python `null`) in the spider.

PR #11733, opened 2025-01-01
[otr_au] it is street_address

matkoniecz:


matkoniecz:
it also somehow includes brand name in address field which is a separate issue not tried to be fixed here

PR #11737, opened 2025-01-01
[vallarta_us] - this is addr:street rather than addr:full

matkoniecz:
for example, this value: 13715 Vanowen St


matkoniecz:
I am confused why output map is empty

iandees:
> I am confused why output map is empty

It's empty because all 54 items are missing geometry data.

matkoniecz:
`addr:street_address` fiddling looks reasonable also across this sample

matkoniecz:
and I probably have not broken geometry extraction

PR #11738, opened 2025-01-01
[brico_ok_it] - this is addr:street rather than addr:full

matkoniecz:
for example, this value: Corso Traiano 4


matkoniecz:
Looking at it more: Polish ones are in form like `OPOLSKA 12, 48-231 Lubrza` which should rather get Poland as part of address and form proper `addr:full`

BTW, why spider has `it` suffix?

Italian were actually proper `addr:full`

I have not looked into detail at Russian hieroglyphics.

PR #11739, opened 2025-01-01
[barrhead_travel_gb] - this is addr:street rather than addr:full

matkoniecz:
for example, this value: 305A Provan Walk


matkoniecz:
on review it was an isolated problem, I should check how prevalent problem is before opening PR

PR #11750, opened 2025-01-02
[wells_fargo] - this is addr:street_address rather than addr:full

matkoniecz:
for example, this values: 1000 LAKES DR
RTE 4 & RTE 41
HWY 73 MAIN ST
HWY 70 W
HWY 158 & 801
HIGHWAY 191
HIGHWAY 160
HIGHWAY 150
9 N MAIN ST
99 HARVEY RD


matkoniecz:
If that would be useful/needed for this PR I can present some deeper analysis of this (currently) broken address field.

iandees:
Not needed, thank you. I was just AFK.

PR #11762, opened 2025-01-03
[English Heritage GB] Add spider

TheUKHighStreet:


TheUKHighStreet:
Thanks for sorting that out @Cj-Malone Should English Heritage be tagged as tourism:attraction in NSI? It seems a bit strange that it defaults to car park?

Cj-Malone:
Defaulting to parking is because of [this entry](https://nsi.guide/index.html?t=operators&k=amenity&v=parking&tt=englis#englishheritage-c49703) as an operator.

I also considered flipping the spider to operator instead of brand, but I don't think either are perfect so I left it as brand.

I don't think of `tourism=attraction` as a real category any more, it's more of an attribute. I don't imagine NSI would add a `tourism/attraction` tree for this reason. I left it in the spider, because it's the least bad option for ATP right now. But I don't think 100% of these should end up in OSM with `tourism=attraction`, and I don't think any of them should exclusively have it as a category.

+ NSI is great for normal brands, where each brand is only in one or 2 categories, but where each POI has a different category (like here or US health care brands), NSI isn't great.

PR #11787, opened 2025-01-05
Add Hobby Lobby spider (1040 locations)

AndrewBelt:
This has a bug that I can't figure out. 
The `ld_data` has
```py
'openingHoursSpecification': [..., {'dayOfWeek': 'Monday', 'opens': '09:00 AM', 'closes': '08:00 PM', '@type': 'OpeningHoursSpecification'}, ...]
```
but if I set
```py
time_format = "%H:%M %p"
```
the closing time is parsed as `08:00`. The "PM" is ignored. Is this a bug with `StructuredDataSpider` or `OpeningHours`?

Also, this is my second spider I've written. Could I request general feedback?

### DIFF
--- a/locations/spiders/hobby_lobby.py
+++ b/locations/spiders/hobby_lobby.py
@@ -1,0 +1,44 @@
+from urllib.parse import urlparse
+
+from scrapy.spiders import SitemapSpider
+
+from locations.categories import Categories, apply_category
+from locations.structured_data_spider import StructuredDataSpider
+
+
+class HobbyLobbySpider(SitemapSpider, StructuredDataSpider):
+    name = "hobby_lobby"
+    allowed_domains = ["hobbylobby.com"]
+    item_attributes = {
+        "brand": "Hobby Lobby",
+        "brand_wikidata": "Q5874938",
+    }
+    sitemap_urls = ["https://www.hobbylobby.com/sitemap-Store.xml"]
+    sitemap_rules = [
+        (r"/stores/search/(\d+)$", "parse_sd"),
+    ]
+    time_format = "%H:%M %p"
+    custom_settings = {
+        # Site has no rate limiting
+        "DOWNLOAD_DELAY": 0.1,
+    }
+    # There are only social links for branch, not location
+    search_for_twitter = False
+    search_for_facebook = False
+
+    def pre_process_data(self, ld_data, **kwargs):
+        ld_data.pop("name", None)
+        # Broken image link
+        ld_data.pop("image", None)
+
+        # ID/ref should be store number, not main website URL
+        url = urlparse(ld_data["url"])
+        ref = url.path.split("/")[-1]
+        ld_data["@id"] = ref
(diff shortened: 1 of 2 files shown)


locations/spiders/hobby_lobby.py
```diff
@@ -0,0 +1,44 @@
...
+    allowed_domains = ["hobbylobby.com"]
+    item_attributes = {
+        "brand": "Hobby Lobby",
+        "brand_wikidata": "Q5874938",
+    }
+    sitemap_urls = ["https://www.hobbylobby.com/sitemap-Store.xml"]
+    sitemap_rules = [
+        (r"/stores/search/(\d+)$", "parse_sd"),
+    ]
+    time_format = "%H:%M %p"
```
davidhicks, line 20: `%H` is for 24 hour time, `%I` is for 12 hour time.

You want this line to be:

`time_format = "%I:%M %p"`

locations/spiders/hobby_lobby.py
```diff
@@ -0,0 +1,44 @@
...
+    custom_settings = {
+        # Site has no rate limiting
+        "DOWNLOAD_DELAY": 0.1,
```
davidhicks, line 23: Generally use of `DOWNLOAD_DELAY` within spiders is reserved for increasing the download delay from a default value to avoid triggering a rate limit. If someone running the spider (or many spiders) wants to speed things up, they'd set the default `DOWNLOAD_DELAY` via settings.py to be a smaller value than the current default of `1` (1 second).

This custom setting should thus be removed from the spider.

locations/spiders/hobby_lobby.py
```diff
@@ -0,0 +1,44 @@
+from urllib.parse import urlparse
+
+from scrapy.spiders import SitemapSpider
+
+from locations.categories import Categories, apply_category
+from locations.structured_data_spider import StructuredDataSpider
+
+
+class HobbyLobbySpider(SitemapSpider, StructuredDataSpider):
+    name = "hobby_lobby"
```
davidhicks, line 10: This brand only seems to exist in the United States so the spider `name` should be `hobby_lobby_us` and the class name should be `HobbyLobbyUSSpider`. The file should be renamed to match the spider `name`.

locations/spiders/hobby_lobby.py
```diff
@@ -0,0 +1,44 @@
...
+    allowed_domains = ["hobbylobby.com"]
```
davidhicks, line 11: Ideally I think this should be:

`allowed_domains = ["www.hobbylobby.com"]`

(e.g. restrict to a specific subdomain `www` that is being used)

davidhicks (changes requested):
A very good second spider. Some review comments added :)

AndrewBelt:
Thanks @davidhicks, everything makes sense.

davidhicks (approved):
Thank you looks good!

PR #11821, opened 2025-01-06
pipelines/check_item_properties: improve country code matching, stats

davidhicks:
* Ensure country codes for items are valid (assigned) ISO 3166-1 alpha-2 codes and not unassigned codes such as "ZZ".

* Collect statistics on the number of items extracted per country code. This is helpful to debug multinational spiders to determine if features of all expected countries are being extracted. It is also helpful to alert to spiders named e.g. "_gb" for the UK, but also including features in territories such as "IM" (Isle of Man), and thus the spider should be renamed.

### DIFF
--- a/locations/pipelines/check_item_properties.py
+++ b/locations/pipelines/check_item_properties.py
@@ -1,7 +1,7 @@
 import math
-import pycountry
 import re
 
+import pycountry
 from scrapy import Spider
 
 from locations.hours import OpeningHours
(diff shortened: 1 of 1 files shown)


Cj-Malone (commented):
At this stage, should it even be regex?

I agree in principle with this change, thanks.

iandees:
Agree that it'd be nice to find a way to do this without a regex.

PR #11857, opened 2025-01-06
[fastned] - this is addr:street_address rather than addr:full

matkoniecz:
for example, this values: Bospoort A4
't Haasje A2
De Horn A7
Kruisoord A7
Rijksweg 471
Ruwiel A2
Akermaat A9
Haarrijn A2
1 Tolvaddon Rd
Mandelan A32


matkoniecz:
looking more, I see some with settlement name - but in general without it, and I see none with say country info

PR #11858, opened 2025-01-06
[snap_fitness] - this is addr:street_address rather than addr:full

matkoniecz:
for example, this values: 92 Main Street
66A Church Street
1/130 High Street
32 Alice Street
1/53 Indwarra Avenue
250 Olsen Ave
10 McCormacks Road
14 Mount Street
22 Queen Street
20 Station Street


matkoniecz:
format varies a bit between areas :(

but still, this seems clearly better

PR #11882, opened 2025-01-07
Fix sitemap and HTML scraper from updated website

AndrewBelt:
Website and sitemap were updated within the last 3 years. Wrote HTML scraper of new location page.

No lat/lon that I can find unfortunately

### DIFF
--- a/locations/spiders/texas_roadhouse.py
+++ b/locations/spiders/texas_roadhouse.py
@@ -1,9 +1,11 @@
-import json
+import re
+from urllib.parse import urljoin, urlparse
 
 import scrapy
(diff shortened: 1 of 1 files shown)


iandees:
The lat/lon are available in the API here (for example): https://www.texasroadhouse.com/restaurants/near?lat=44.0&long=-93.0&radius=1000&limit=20&nomnom=calendars&nomnom_calendars_from=20250109&nomnom_calendars_to=20250120&nomnom_exclude_extref=

Check other spiders that make request using `nomnom` parameters – there might be an example to follow.

PR #11883, opened 2025-01-07
[systeme_u] - this is addr:street_address rather than addr:full

matkoniecz:
for example, this values: 46 Avenue Du Midi
Avenue Des 4 Saisons
50 RUE ORFILA
1 Rue D Assas
2 Avenue De Sabres
11 Avenue Du Mans
Rd 225
1 Avenue Mas Viel
Avenue Du 8 Mai 1945
6 Rue Du Stade


matkoniecz:
not sure why it is getting timeout and is it my fault

iandees:
This is an old PR, but in case we haven't talked about it yet the "timeout" status in the CI happens when the spider takes longer than the 60 seconds we give it during CI. It's not a failure because the spider just wants to run longer. We have a time limit so that we can get feedback to PR authors faster.

PR #11891, opened 2025-01-07
[Prix Quality IT] Fix json parsing in non running spider

karen-kerr:


iandees:
Can you fix the merge conflict here please? I'm not sure which change makes the most sense to include.

PR #11901, opened 2025-01-08
[unicredit_bulbank_bg] Use local brand for local bank

matkoniecz:
@Dimitar5555 - review welcomed!


Dimitar5555 (changes requested):
The Latin name is correct, I will fix it in NSI later today. 

Their offices, website and documents are branded in Latin,

matkoniecz:
hmm, so OSM is also wrong? From what I was seeing non-latin form appears often there.

Maybe too many people followed NSI instead of reporting problem? Or they rebranded?

Dimitar5555:
It's my fault. I've added quite a few Bulgarian brands to NSI without properly checking if they are written in Latin or in Cyrillic. See for example my last ~10 commits there (https://github.com/osmlab/name-suggestion-index/commits/main/?author=Dimitar5555).

PR #11912, opened 2025-01-09
Add Rosetta APR "storefinder" for Australian electricity features

davidhicks:
Rosetta APR is used by most Australian energy grid operators for displaying maps of electricity grid assets such as substations and lines. Rosetta APR and this type of data is generally made publicly available.

Depending on the energy grid operator, different types of files may be served by the Rosetta APR platform such as geojson files (for features with coordinates) and csv or zipped csv spreadsheet files for detailed data on features such as capacities of substations or transformers. This storefinder provides a means to make second and subsequent requests for additional data files to allow extracted features to be supplemented with more detailed data such as capacity ratings of transformers.

### DIFF
--- a/locations/categories.py
+++ b/locations/categories.py
@@ -317,8 +317,15 @@ class Categories(Enum):
 
     ANTENNA = {"man_made": "antenna"}
     MONITORING_STATION = {"man_made": "monitoring_station"}
+    SUBSTATION = {"power": "substation"}
+    SUBSTATION_GENERATION = {"power": "substation", "substation": "generation"}
+    SUBSTATION_MINOR_DISTRIBUTION = {"power": "substation", "substation": "minor_distribution"}
+    SUBSTATION_INDUSTRIAL = {"power": "substation", "substation": "industrial"}
+    SUBSTATION_TRACTION = {"power": "substation", "substation": "traction"}
+    SUBSTATION_TRANSMISSION = {"power": "substation", "substation": "transmission"}
+    SUBSTATION_ZONE = {"power": "substation", "substation": "distribution"}
     SURVEILLANCE_CAMERA = {"man_made": "surveillance", "surveillance:type": "camera"}
-
+    TRANSFORMER = {"power": "transformer"}
 
 def apply_category(category, item: Feature):
     """
@@ -357,6 +364,7 @@ def apply_category(category, item: Feature):
 
 
 top_level_tags = [
+    "aeroway",
     "amenity",
     "club",
     "craft",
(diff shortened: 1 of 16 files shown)


locations/categories.py
```diff
@@ -368,13 +377,13 @@ def apply_category(category, item: Feature):
     "leisure",
     "man_made",
     "office",
+    "power",
     "public_transport",
     "shop",
+    "telecom",
     "tourism",
-    "aeroway",
```
Cj-Malone, line 374: why have you removed this?

davidhicks: Just reordered alphabetically :) It's still in the list of top level tags.

locations/settings.py
```diff
@@ -115,6 +115,7 @@
     "locations.pipelines.closed.ClosePipeline": 650,
     "locations.pipelines.apply_nsi_categories.ApplyNSICategoriesPipeline": 700,
     "locations.pipelines.check_item_properties.CheckItemPropertiesPipeline": 750,
+    "locations.pipelines.geojson_multipoint_simplification.GeoJSONMultiPointSimplificationPipeline": 760,
```
Cj-Malone, line 118: How frequently will this be used? Should it be a spider thing?

davidhicks: For the last weekly run, the only spider with geometry types other than `Point` is `bern_ch` which uses `LineString` and `MultiLineString`. So currently this new pipeline would only be applicable to Rosetta APR spiders some of which return `MultiPoint` (but is really just `Point` as there never appears to be more than one point returned).

In saying that, I think we could possibly do with a few other GeoJSON type conversion pipelines to support future growth of ATP into features other than points.
* A similar pipeline for converting a `MultiLineString` to a `LineString` if there is only one line present in the geometry.
* A similar pipeline for converting a `MultiPolygon` to a `Polygon` if there is only one polygon present in the geometry.

If ATP ventures in features that aren't just points--examples being roads, administrative borders, etc, these pipelines could be more useful. Potentially other geometry simplifications could be investigated too.

But I can also remove this pipeline and turn it into a shared simplification function in perhaps `geo.py` so that the simplification occurs not in a pipeline but rather has to be dealt with explicitly by each spider.

iandees:
This is fine, but a couple smaller/organizational things should follow:

- The `*_utils.py` files should get moved into a new `locations/utils/` package instead of sitting in the root `locations/` directory.
- The single MultiPoint geometry to single Point geometry pipeline step should be made more generic to handle other single multi-geometries.

locations/settings.py
Cj-Malone: It's also possible to only enable pipelines for a specific spider via `custom_settings`.

PR #11915, opened 2025-01-09
Add Powercor (power/(substation|transformer)) (AU) (93779 features)

davidhicks:
Powercor and Citipower are separate electricity grid operators but have common ownership and common ICT systems. As a result, this spider combines assets such as substations and transformers under the Powercor brand only, as there is no method in the source data for determining whether an asset is a Powercor asset or Citipower asset.

Feature types extracted:
 'atp/category/power/substation': 97,
 'atp/category/power/transformer': 93682,

### DIFF
--- a/locations/spiders/powercor_au.py
+++ b/locations/spiders/powercor_au.py
@@ -10,8 +10,18 @@ class PowercorAUSpider(RosettaAPRSpider):
     item_attributes = {"operator": "Powercor", "operator_wikidata": "Q7236677"}
     start_urls = ["https://dapr.powercor.com.au/"]
     data_files = [
-        RosettaAPRDataFile(url="https://content.rosettaanalytics.com.au/citipower_powercor_layers_serve/CitiPower_Powercor_Zone_Substations.geojson", file_type="geojson", encrypted=True, callback_function_name="parse_zone_substations"),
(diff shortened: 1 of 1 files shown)


davidhicks:
The failure appears to be `content.rosettaanalytics.com.au` seemingly returning an error message or captcha challenge or similar when it detects an origin IP address is a cloud/data centre IP range or VPN service or similar. Perhaps there is a generic firewall in place for the entire `rosettaanalytics.com.au` that broadly attempts to block "threats" or whatever the firewall vendor has decided to add to a filter set they provide to their customers.

This spider is working fine from residential IPs within Australia and the URLs are also accessible through the Internet Archive Wayback machine.

davidhicks:
I suspect there is a limit with Zyte not being able to return 15MB responses, or timing out quickly, etc, so the file is being truncated by the time it arrives at Scrapy, and this truncated file cannot be decrypted as there are blocks missing.

davidhicks:
https://docs.zyte.com/zyte-api/faq.html#is-there-a-response-size-limit states the response body returned by Zyte is truncated at 10MB. The file this spider is unable to download and parse is 17.5MB

If an Australian IP address is used other than via Zyte, or the Internet Archive Wayback Machine is used (maybe they have Australian IPs?), the spider is working fine.

PR #11919, opened 2025-01-09
Add Ausnet (power/transformer) (AU) (64222 transformers)

davidhicks:


iandees:
It's likely the data center IPs.

iandees:
If we know that Zyte won't work we should remove the requires_proxy for now.

davidhicks:
Thanks, I missed this feedback last week and have just addressed it by removing `requires_proxy` and replacing it with documentation instead.

PR #11921, opened 2025-01-09
Add Ausgrid (power/substation) (AU) (342 substations)

davidhicks:


davidhicks:
Looks like we need a pipeline to check for `MultiPoint` geometry and if there is only one point, convert it from a `MultiPoint` to just a single `Point`.

PR #11933, opened 2025-01-10
[Naf Naf Grill US] Add spider

Cj-Malone:
No coords :(

### DIFF
--- a/locations/spiders/naf_naf_grill_us.py
+++ b/locations/spiders/naf_naf_grill_us.py
@@ -1,0 +1,23 @@
+from typing import Any
+
+from scrapy import Spider
+from scrapy.http import Response
+
+from locations.items import Feature
+
+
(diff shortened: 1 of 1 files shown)


davidhicks:
Coordinates could be obtained by following the Google URL shortener URLs and observing the redirected page URL that contains coordinates?

```
...
        yield Request(url=short_google_maps_url, meta={"item": item}, callback=self.add_coordinates)

    # Follow shortened URL to find full Google Maps URL containing coordinates
    def add_coordinates(self, response: Response) -> Iterable[Feature]:
        item = response.meta["item"]
        item["lat"], item["lon"] = url_to_coords(response.url)
        yield item
```

`bauhaus_ch` and `tgi_fridays_au` are some examples of this approach.

Cj-Malone:
I don't scrapy Google.

PR #11940, opened 2025-01-11
[pandora] emit only Pandora branded as being Pandora branded

matkoniecz:
it is half clarification request: what it even means? Does it carry even a potentially useful info?


hfs:
> what it even means?

`secondary=yes` means that these are not Pandora stores where it says “Pandora” on the store front sign, but rather other retailers such as jewelers or department stores selling Pandora products.

matkoniecz:
in such case they should not get `{"brand": "Pandora", "brand_wikidata": "Q2241604"}` as they are not Pandora-branded.

PR #11958, opened 2025-01-13
Raley's, Food City AZ: Make an additional request to get session cookies

impiaaa:


locations/spiders/raleys_us.py
```diff
@@ -1,40 +1,47 @@
...
 from locations.json_blob_spider import JSONBlobSpider
 
-BRAND_WIKIDATA = {
-    "Bel Air": "Q112922067",
-    "Nob Hill Foods": "Q121816894",
-    "Raley's": "Q7286970",
-    "Raley's ONE Market": "Q7286970",
+BRAND_MAP = {
+    "Bel Air": {"brand": "Bel Air", "brand_wikidata": "Q112922067"},
+    "Nob Hill Foods": {"name": "Nob Hill Foods", "brand": "Nob Hill Foods", "brand_wikidata": "Q121816894"},
```
Cj-Malone, line 9: Did you create [Q121816894](https://www.wikidata.org/wiki/Q121816894)? It has been deleted

locations/spiders/raleys_us.py
```diff
@@ -1,40 +1,47 @@
...
+        if brand := BRAND_MAP.get(location["brand"]["name"]):
+            item["name"] = None
+            item.update(brand)
+        else:
+            self.logger.error("Unexpected brand: {}".format(location["brand"]["name"]))
         item["street_address"] = item.pop("street")
 
         oh = OpeningHours()
         # TODO: Is it safe to assume that all stores are open 7 days?
         oh.add_ranges_from_string(f"Mo-Su {location['storeHours'].removeprefix('Between ')}")
```
Cj-Malone, line 42: I would have assumed this meant today, but I'm not American so I've no idea what to assume.

locations/spiders/raleys_us.py
impiaaa: There used to be a Wikipedia article for Nob Hill Foods, and that was the QID, but the article got deleted for being not notable.

locations/spiders/raleys_us.py
```diff
@@ -1,40 +1,42 @@
...
         item["ref"] = location["number"]
         item["website"] = response.urljoin(f"/store/{location['number']}")
-        item["brand"] = item["name"] = location["brand"]["name"]
-        item["brand_wikidata"] = BRAND_WIKIDATA.get(location["brand"]["name"])
+        if brand := BRAND_MAP.get(location["brand"]["name"]):
+            del item["name"]
+            item.update(brand)
+        elif self.name == "raleys_us":
+            # Only log an error for this spider. FoodCityArizonaUSSpider is a subclass that doesn't use the brand mapping.
+            self.logger.error("Unexpected brand: {}".format(location["brand"]["name"]))
```
iandees, line 37: This is pretty smelly. If we're starting to perform checks like this we probably don't want the subclass relationship.

Cj-Malone: I didn't know it was using this class when I added it, and I agree it's not a good idea. Whenever we are doing anything specific in here, it shouldn't be used by multiple spiders, however since we are already applying a category, I've embraced it and added the Food City branding here instead of the other spider.

Cj-Malone (approved):

PR #11964, opened 2025-01-14
[Vitalia] Fix non running spider

karen-kerr:


iandees:
Is it expected that there are no coordinates now?

PR #11989, opened 2025-01-16
follow OpenStreetMap opening_hours over midnight

matkoniecz:
see https://github.com/alltheplaces/alltheplaces/discussions/4959 this way opening hours now are not having a change from OSM syntax previously ATP was using ; in meaning of ,
this mattered only for opening hours extending past midnight and only there ATP was changed to use ,

far more complicated processing can be used to avoid using , in rare cases of over midnight opening hours where ; can be used anyway - but instead much simpler one was used

Disclaimer: that falls under work that was funded by EU via NGI 0 via nl.net grant (I can also process it on my side, but others would not benefit from it - so I hope that it is fine to send it here)

see also https://openingh.ypid.de/evaluation_tool/?EXP=Mo-Sa%2010%3A00-02%3A00%2C%20Su%2009%3A00-02%3A00&lat=48.7769&lon=9.1844&mode=0 and https://openingh.ypid.de/evaluation_tool/?EXP=Mo-Sa%2010%3A00-02%3A00%3B%20Su%2009%3A00-02%3A00&lat=48.7769&lon=9.1844&mode=0 for why ; vs , may matter

### DIFF
--- a/locations/hours.py
+++ b/locations/hours.py
@@ -951,10 +951,39 @@ class OpeningHours:
         self.days_closed.discard(day)
         self.day_hours[day].add((open_time, close_time))
 
+    @staticmethod
+    def time_struct_instance(hour, minutes):
+        return time.struct_time((1900, 1, 1, hour, minutes, 0, 0, 1, -1))
+
     def as_opening_hours(self) -> str:
         day_groups = []
         this_day_group = None
 
+        # usually ; works fine as a separator between different days
+        # but it has an annoying quirk, in OpenStreetMap opening_hours syntax
+        # it overrides previous definitions
+        # so for example
+        # Mo-Sa 10:00-02:00; Su 09:00-02:00
+        # means that object is closed on midnight between Saturday and Sunday
+        # Mo-Sa 10:00-02:00; Su 00:00-02:00,09:00-02:00
+        # in turn means that object is open also during early Sunday hours
+        #
+        # All the Places is using a small section of entire opening_hous syntax
+        # so we need only check whether time goes over midnight and split it
+        # in two regular ranges
+        day_hours_midnight_split = defaultdict(set)
+        for index, day in enumerate(DAYS):
+            for h in self.day_hours[day]:
+                if h[0].tm_hour * 60 + h[0].tm_min > h[1].tm_hour * 60 + h[1].tm_min:
+                    # start hour is greater than end hour, indicating that it is
+                    # an over-midnight range
+                    day_hours_midnight_split[day].add((h[0], OpeningHours.time_struct_instance(23, 59)))
+                    next_day = DAYS[(index + 1) % len(DAYS)]
+                    day_hours_midnight_split[next_day].add((OpeningHours.time_struct_instance(0, 0), h[1]))
+                    if next_day in self.days_closed:
+                        del self.days_closed[next_day]
+                else:
+                    day_hours_midnight_split[day].add(h)
         for day in DAYS:
             if day in self.days_closed:
                 hours = "closed"
@@ -965,7 +994,7 @@ class OpeningHours:
                         time.strftime("%H:%M", h[0]),
                         time.strftime("%H:%M", h[1]).replace("23:59", "24:00"),
                     )
-                    for h in sorted(self.day_hours[day])
+                    for h in sorted(day_hours_midnight_split[day])
                 )
 
             if not this_day_group:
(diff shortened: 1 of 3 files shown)


locations/hours.py
```diff
@@ -980,15 +980,33 @@ def as_opening_hours(self) -> str:
...
+        # usually ; works fine as a separator between different days
+        # but it has an annoying quirk, in OpenStreetMap opening_hours syntax
+        # it overrides previous definitions
+        # so for example
+        # Mo-Sa 10:00-02:00; Su 09:00-02:00
+        # means that object is closed on midnight between Saturday and Sunday
+        # Mo-Sa 10:00-02:00, Su 09:00-02:00
+        # in turn means that object is open also during early Sunday hours
+        #
+        # All the Places is using a small section of entire opening_hous syntax
```
impiaaa, line 992: typo "opening_hous"/"opening_hours"

matkoniecz, line 992:
```suggestion
        # All the Places is using a small section of entire opening_hours syntax
```

matkoniecz: thanks, fixed

locations/hours.py
```diff
@@ -980,15 +980,33 @@ def as_opening_hours(self) -> str:
...
+        # so for example
+        # Mo-Sa 10:00-02:00; Su 09:00-02:00
+        # means that object is closed on midnight between Saturday and Sunday
+        # Mo-Sa 10:00-02:00, Su 09:00-02:00
+        # in turn means that object is open also during early Sunday hours
+        #
+        # All the Places is using a small section of entire opening_hours syntax
+        # so we need only check whether any time goes over midnight,
+        # in all other cases ; can be used safely
+        separator = ";"
```
davidhicks, line 995: Is there an advantage of ever using `;` instead of always just using `,`? Having both separators in play makes it slightly harder for ATP data consumers than just a single separator `,`.

matkoniecz: There are few things here:

- for start `,` is already in use also for separating different time ranges on the same day (`Mo-Fr 10:00-20:00,21:00-22:00, Sa 10:00-20:00` - I can give code generating it in ATP)
- some simple parser can handle `Mo-Fr 10:00-20:00,21:00-22:00; Sa 10:00-20:00` but will fail on `Mo-Fr 10:00-20:00,21:00-22:00, Sa 10:00-20:00`  
- in OSM at least use of `;` is preferred if possible and far more common
- midnight-crossing is likely requiring special support anyway and only it would be affected by getting `,` separator
- changing all `;` to `,` will break existing code relying on `;` in ATP output (and it could not been using proper OH parser as then it would be broken for over-midnight)

Though in general either is fine, actually first version of my code just had `,` everywhere - I added that `;` keeping later.
I am also fine with `,` everywhere, let me know should I change it (note to self: change also examples of OH values, if any, and nearly all OH tests)

davidhicks: If I recall all the previous arguments at ATP about `,` vs `;` in relation to hours spanning past midnight, the sole use of `,` was one argument put forward (along with not splitting hours across midnight into hours for day 1 and hours for day 2). I think the arguments were of the nature of trying to maintain the format the source data provided the hours in rather than convert it to one which is easiest and least ambiguous for ATP consumers to work with.

Ideally I prefer splitting hours at midnight so source data of `Mo 20:00-02:00; Tu 08:00-16:00` becomes `Mo 20:00-24:00; Tu 00:00-02:00, 08:00-16:00`. This way, a parser can guarantee that `;` always acts as a separator between days of the week and `,` always acts as a separator between multiple opening times on the same day of the week.

This proposal appears to generate an output of `Mo 20:00-02:00. Tu 08:00-16:00` which means a parser encountering a `;` knows this is always a day separator (great!), and if it encounters a `,`, it may be a day separator and/or second opening time for the same day. The parsing logic for an ATP consumer is therefore more complex due to the handling of `,` separators.

matkoniecz: > Ideally I prefer splitting hours at midnight so source data of Mo 20:00-02:00; Tu 08:00-16:00 becomes Mo 20:00-24:00; Tu 00:00-02:00, 08:00-16:00

this should be fairly simple, I can try implementing this

matkoniecz: hmmm

how current `Mo-Fr 10:00-02:00; 09:00-02:00` should be represented?

`Mo-Fr 10:00-02:00; Su 00:00-02:00,09:00-02:00` - still has override, just that it is repaired on spot ([triggers warning in validator](https://openingh.ypid.de/evaluation_tool/?EXP=Mo-Fr%2010%3A00-02%3A00%3B%20Su%2000%3A00-02%3A00%2C09%3A00-02%3A00&lat=48.7769&lon=9.1844&mode=0&DATE=1737197100000))

would be `Mo-Fr 00:00-02:00,10:00-02:00; Su 00:00-02:00,09:00-02:00` be also OK? Or should last day get special rule that it is not split?

What about `Mo 23:00-22:00` - should it be converted to `Mo 23:00-24:00, Tu 00:00-22:00` ? 

`Mo-Fr 00:00-02:00,10:00-24:00; Su 00:00-02:00,09:00-24:00` - never goes over midnight, more verbose

For that matter, can/should `Mo-Fr 10:00-23:00; Su 10:00-03:00` be converted to `Mo-Fr 00:00-03:00,10:00-23:00; Su 10:00-24:00` ?

davidhicks: If the source has `Mo-Fr 10:00-02:00; Sa-Su 09:00-02:00` then I'd expect OSM opening hours to be `Mo-Fr 00:00-02:00, 10:00-24:00; Sa 09:00-24:00; Su 00:00-02:00, 09:00-24:00`. This is due to the source using `;` as a separator so that `00:00-02:00` on Saturday morning is ignored and replaced with `09:00-24:00` instead.

davidhicks: > What about Mo 23:00-22:00 - can/should it be converted to Mo 23:00-24:00; Tu 00:00-22:00

That is my proposal, yes.

davidhicks: > For that matter, can/should Mo-Fr 10:00-23:00; Su 10:00-03:00 be converted to Mo-Fr 00:00-03:00,10:00-23:00; Su 10:00-24:00

This should be interpreted instead as `Mo 00:00-03:00, 10:00-23:00; Tu-Fr: 10:00-23:00; Sa: closed; Su: 10:00-24:00`?

locations/hours.py
```diff
@@ -951,10 +951,39 @@ def add_range(self, day, open_time, close_time, time_format="%H:%M", closed=CLOS
         self.days_closed.discard(day)
         self.day_hours[day].add((open_time, close_time))
 
+    @staticmethod
+    def time_struct_instance(hour, minutes):
+        return time.struct_time((1900, 1, 1, hour, minutes, 0, 0, 1, -1))
```
matkoniecz, line 956: I cannot believe there is no nicer way to do this :(

I guess I can go through string?

matkoniecz: Code directly above actually uses strptime so maybe it should be treated as preferred method?

matkoniecz:
Is there anything that I can do here to progress/unlock it?

(I have idea how to make workaround on my side, and it would not take long time to do this - though I would be happier to fix it in general for everyone, using already written code)

davidhicks (approved):
Reviewed test cases and looks good to me!

iandees:
Sorry I was slow on this.

Originally I preferred to represent the hours in the format that the business listing showed, even if they didn't match OSM. This would make it easier to confirm correctness. But I think I've talked myself into modifying the source data so that it matches OSM more closely.

iandees:
Thank you for the good test coverage, too.

matkoniecz:
> Sorry I was slow on this.

no problem, I was not (yet) even blocked by it

(and even if I would be blocked by it, that does not mean that anyone else is obligated to act on it)

PR #12012, opened 2025-01-18
DATA_FORMAT.md: document handling of translated field values

davidhicks:
Add documentation to clarify how translations of field values are handled. Typically translatable field values (name, branch, addr:*) are supplied in the primary local language of the feature. Available translations are then supplied as extras fields.

Example:
`country`=`FR`
`name`=`Emplacement aléatoire` (for ISO 639-1 alpha-2 code of "FR") `extras`:{`name:en`: `Random location`}


iandees:
The extras fields should be merged in with the other properties, not within an `extras` field.

`extras` is a side-effect of Scrapy objects requiring hard-coded field names. The output pipelines will take the dictionary in the `extras` field and merge it into the output properties for the feature.

davidhicks:
> The extras fields should be merged in with the other properties, not within an `extras` field.
> 
> `extras` is a side-effect of Scrapy objects requiring hard-coded field names. The output pipelines will take the dictionary in the `extras` field and merge it into the output properties for the feature.

Thanks, I've fixed this error and also significantly expanded/improved the DATA_FORMAT documentation at the same time, hopefully making it easier for consumers of ATP data to understand how ATP and OSM tagging differs and what nuances to be aware of when interpreting properties extracted by ATP.

PR #12053, opened 2025-01-23
[Vinnies AU] Remove brand

matkoniecz:
https://www.wikidata.org/wiki/Q7592120

https://en.wikipedia.org/w/index.php?title=St._Vincent_de_Paul_Thrift_Store&oldid=647143127 before it was merged

### DIFF
--- a/locations/spiders/vinnies_au.py
+++ b/locations/spiders/vinnies_au.py
@@ -1,14 +1,15 @@
 from scrapy import Spider
 from scrapy.http import JsonRequest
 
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Needs an NSI update.

matkoniecz:
https://github.com/osmlab/name-suggestion-index/pull/10389#issuecomment-2612051564 seems to indicate that this NSI entry has deeper problem with branding, not checked yet is spider giving nonsense values

PR #12067, opened 2025-01-24
Update wikidata id in migros_tr

matkoniecz:
I admit that I am less sure here and it is based mostly on https://www.openstreetmap.org/way/762328412/history

>  Fixed supermarkets using wrong brand:wikidata tag (Q680727->Q1754510). Discussed in the Turkey OSM community before the change. The incorrect tagging of these supermarkets was due to a bug in the iD editor, which has also been fixed.

### DIFF
--- a/locations/spiders/migros_tr.py
+++ b/locations/spiders/migros_tr.py
@@ -10,20 +10,20 @@ from locations.items import Feature
 
 CATEGORY_MAPPING = {
     1: {"brand": "Macrocenter", "brand_wikidata": "Q123194881", "extras": Categories.SHOP_SUPERMARKET.value},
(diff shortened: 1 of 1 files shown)


panasonic-mafia:
@matkoniecz the is spider is already in line with [Migros NSI entries for TR](https://nsi.guide/index.html?t=brands&tt=q1754510) (Q1754510). Let's close the PR if you don't have objections?

Cj-Malone:
Your quote says Q680727->Q1754510 but for commit does the opposite.

matkoniecz:
that makes no sense

I will investigate what is going on and reopen if it makes sense

PR #12075, opened 2025-01-27
[Lili GR] Update brand

matkoniecz:
see https://lilidrogerie.gr/

see https://www.openstreetmap.org/node/8623418446/history and other places in OSM if mapped

https://www.openstreetmap.org/search?query=lilly%252C%20greece finds single one mapped this way


Cj-Malone (changes requested):
Needs to be in sync with NSI

matkoniecz:
https://github.com/osmlab/name-suggestion-index/pull/10404#event-16083015482 got merged

Cj-Malone (approved):

PR #12084, opened 2025-01-27
Sync brands with NSI

Cj-Malone:
I've occasionally done some tests on the output, while in general I think it's the better place, we have final categories, it's not just `item_attributes` etc. As @panasonic-mafia [mentioned](https://github.com/alltheplaces/alltheplaces/pull/12071#issue-2811015095) a test I gave it a go. It can can be helpful, and massively more routine if included in CI, and includes none functioning spiders.

As is, I've reviewed A->~C~ ~P~ Z, it's mind-numbingly boring and usually just needs to be updated here. I have a few edits to NSI as well though. I've not included the test yet as it is obviously currently failing.

I'm happy to work on this branch until I've done them all, or do it in several PRs. If it's not merged by the time I next do some, I'll continue on this branch.


iandees:
I think chunks like this is ok.

There's probably a way to automate this, maybe with a pre-commit check? Or we could do something in the output of the CI check that puts the comments on PRs here?

Cj-Malone:
I'm talking myself out of pytest because we don't want to have failing tests for a week if we are waiting on an NSI release. But yeah I think getting the feedback in the PR is vital, we ca workout the details in that PR.

Cj-Malone:
While working on this, I can't help thinking about the enum of brands @mjoe999 wanted. Maybe it would have been unmaintainable like I thought, but this isn't ideal either. Iterating though all brands, not just `item_attributes` would make this more complete.

It may have also made it easier to have multilingual brands. Right now we define 1 brand string in the spider and hope NSI sorts us out, in spiders that cover different languages, we define 0 and hope NSI sorts us out. Having `brand:en` and `brand:fr` may have been helpful.

matkoniecz:
> I'm talking myself out of pytest because we don't want to have failing tests for a week if we are waiting on an NSI release. But yeah I think getting the feedback in the PR is vital, we ca workout the details in that PR.

Maybe it can use not-yet-released NSI for tests? And/or check only spider modified by a given PR?

Or if test is failing then check is it fixed by using latest, not yet released NSI?

locations/spiders/orlen.py
```diff
@@ -13,7 +13,7 @@ class OrlenSpider(scrapy.Spider):
     name = "orlen"
     custom_settings = {"ROBOTSTXT_OBEY": False}
     brand_mappings = {
-        "ORLEN": {"brand": "Orlen", "brand_wikidata": "Q971649"},
+        "ORLEN": {"brand": "Orlen Paczka", "brand_wikidata": "Q110457879"},  # ?
```
matkoniecz: Orlen Paczka is for parcel lockers (though they also should rather use `brand:wikidata` pointing to Orlen brand)

Orlen is for fuel stations.

Cj-Malone:
I think this is good to go for now. A few left but more complex and can be done in separate PRs, I'll add the test separately as well.

PR #12091, opened 2025-01-28
volg_ch: fix spider

davidhicks:


davidhicks:
Note: dupefilter will go into overdrive on this one because a geographic search is being done with a 24km radius where the nearest X locations are returned (even if they're much further away). Hence lots of locations get found multiple times in different radius searches and will get ignored by the dupefilter.

PR #12129, opened 2025-02-01
Add Melbourne City Council parking spaces (3307 parking spaces)

davidhicks:
* Only includes parking spaces with occupancy sensors

* Source data of parking restrictions for each parking space is not the best quality data (even though official). Sometimes provides contradictory restrictions (e.g. overlapping restrictions for the same day where one says it is free parking, the other says it is paid parking). This spider ensures both overlapping/contradictory restrictions are extracted so that users can decide how to handle the ambiguity.


davidhicks:
Note: parking restrictions are quite complex to parse and if more spiders need to worry about complex parking restrictions, we can move much of this MCC-specific code into it's own shared API that multiple spiders can use.

PR #12131, opened 2025-02-02
[fasol_ru] add spider (1293 locations)

panasonic-mafia:


panasonic-mafia:
Rel: https://github.com/osmlab/name-suggestion-index/pull/10444

PR #12134, opened 2025-02-03
Add Energex power/transformer (AU) (52074 transformers)

davidhicks:


davidhicks:
> 💭 We should probably make a general ArcGIS REST spider. I've been thinking about moving OpenAddresses' sources into something powered by All the Places...

I've been thinking this too and some of these bigger spiders are now leading me further in this direction (in preference to just JSONBlobSpider) because of the need to handle pagination of results. Without pagination, most of the benefit appears to be in query building to avoid a hard-to-read 300 character long URL with all the query arguments included. I'll look into it. It's going to be easy to convert existing JSONBlobSpider to ArcGISFeatureServerSpider (or whatever name).

PR #12153, opened 2025-02-04
[Radley GB] Add Spider

TheUKHighStreet:


TheUKHighStreet:
Is is ok without the addresses?

iandees:
> Is is ok without the addresses?

Yup, it's ok to add a spider without addresses. It's best to have as much information as you can get, but sometimes it's not worth the effort or the data just doesn't exist.

PR #12167, opened 2025-02-05
[sim_br] Add petrol stations in Brazil (362 items)

ckpiper:


locations/spiders/sim_br.py
```diff
@@ -0,0 +1,40 @@
...
+}
+
+
+class SimBRSpider(Spider):
+    name = "sim_br"
+
+    start_urls = ["https://www.simrede.com.br/unidades"]
+
+    def parse(self, response):
+        stations = json.loads(response.text)
```
Cj-Malone, line 22: Presumably can be `stations = response.json()`

PR #12176, opened 2025-02-06
Split aggregator spiders

Cj-Malone:
Rel: https://github.com/alltheplaces/alltheplaces/pull/12099

TODO:

- [x] Assign values in `dataset_attributes` automatically
- [x] Docs
- [x] Move more spiders
- [ ] Shared class with licenses, enum? builder? (Different PR?)


locations/settings.py
```diff
@@ -15,7 +15,11 @@
 
 BOT_NAME = "locations"
 
-SPIDER_MODULES = ["locations.spiders"]
+SPIDER_MODULES = [
```
davidhicks, line 18: There are some challenges of classification to address in splitting the `spiders` namespace, for example:

1. An `aggregation spider` (such as a government "food safety" database requiring restaurants to register) could include small businesses (single restaurants not part of a chain) as well as chains of restaurants that otherwise may have their own `brand spider`. How should this situation be handled?
  1a. Prefer the primary source (a `brand spider` crawling the main website of a brand) if it provides better or similar accuracy, currency and completeness of data.
  1b. Allow both the `aggregation spider` and `brand spider` to capture the same (or potentially different) data on the same location. Potentially ATP after crawl execution can perform a de-duplication/merge process allowing ATP data consumers to get a single feature that has a source field nominating more than one source of data.
  1c. Prefer an `aggregation spider` if it provides better or similar accuracy, currency and completeness of data.

2. Could ATP handle features that NSI doesn't care about, or may not ever care about? For example, bays, mountain peaks and addresses (as suggested by @iandees at https://github.com/alltheplaces/alltheplaces/pull/12134#issuecomment-2631539615)? Does the categorisation scheme allow such future expansion?

3. Can ATP handle features that ATP doesn't currently care about but NSI does, mostly transit/*? I was planning to introduce a GTFS spider (General Transit Feed Specification) allowing ATP to capture train stations, bus stops, etc so this PR comes at a good time allowing this possible addition to be thought about.

Cj-Malone: 1. That would be government spider, not an aggregator. I have no issue with 1 POI being a "duplicate" of another spider, this already happens. As you point out, each have their own attributes which are useful. I do have beef with spiders that contain duplicates, but that also already exists.
2. As is, we (wrongly) already output POIs with no category. They should be categorised as a good category or generic POI. Then things legitimately like addresses could flow freely. Other categories are already supported, as there is really no difference between a shop and a mountain peek.
3. I've written a GTFS spider before, the category conversion is a pain. But also, in terms of OSM, there are more specific tools available to handle more of the edge cases. In terms of ATP, I wouldn't object, but also I wouldn't find it interesting.

davidhicks:
I don't understand the intention(s) for splitting out ATP spiders by brands/*, operators/*, transit/*, aggregators/* or whatever is decided. Is it just to reduce the number of files in the `spiders` directory and use file system paths to try and make it easier to navigate/search for/filter on specific spiders?

Cj-Malone:
The goal isn't to shrink that directory, although it is too big. The goal is to help document the origin (which I haven't done yet, but will be part of this PR).

davidhicks:
Is the intent to classify spiders by the source of data for:
1. Quality purposes (e.g. primary vs. secondary vs. tertiary sources)
2. Copyright purposes
3. Import compatibility into OSM reasons (a combination of (1), (2), maybe other OSM reasons)
4. ??

For definitions of `aggregator` and `government`:

As an example, is an `aggregator` something like a hotel price comparison website or perhaps even a government website which is a directory of local businesses, as the source data in these examples is not authoritative? And `government` is something like an official business register that is authoritative because laws require the register to exist and be used as an authoritative source of information?

A more complex example (first one I thought of and could find in a search) could be the Fair Trade Federation (FTF) which has a member locator widget at https://www.fairtradefederation.org/browse-ftf-members/stores-cafes/ and where FTF is the authority on who are members of FTF. Is a FTF spider an `aggregator` or `government` spider? Seemingly it's `aggregator` in the level of real world accuracy/currency/completeness for where member stores/cafes are located, but it's also the authoritative source for who is an FTF member. Maybe then we should be addressing quality of data on a field-by-field basis rather than categorisation of entire spiders... or implement both options?

locations/settings.py
impiaaa: > I was planning to introduce a GTFS spider (General Transit Feed Specification) allowing ATP to capture train stations, bus stops, etc

[I made one](https://github.com/impiaaa/alltheplaces/blob/gtfs/locations/spiders/gtfs.py) but never put up a PR because it produced a *lot* of data points (est. 3 million), many more than any other current spider.

Cj-Malone:
I'm not playing the definition game, we already have edge cases, we always will. That's OK. I don't want us to start scraping any random aggregators, I'm thinking [SOCi/Brandify](https://wiki.openstreetmap.org/wiki/Brandify_Imports) or [Uberall/Navads](https://wiki.openstreetmap.org/wiki/Navads_Imports). For gov data, see spiders prefixed with `gov_` I'll move them in this PR, I'm not thinking about a spider that would scrape government offices or speed cameras, they are normal "brand" spiders.

Not literally quality, brand spiders are also of questionable quality. And different spiders have different strengths, opening hours from `sainsburys` should be undisputed, but `fhrs:id` and other tags (`fhrs:rating`) should be better coming directly from [gov data](https://ratings.food.gov.uk/business/518687/sainsbury).

We already document copyright in dataset_attributes, this will not replace that.

I've spent way too much of my life worrying about OSM, they have the data and loads more, it they use it is up to them. But I'm not wasting any more time on discussing imports.

davidhicks:
Thanks! The `government` spiders make sense to me with those examples highlighted. Hence a `government` spider is for situations where a government has compiled a list of locations with data that businesses have self-supplied, generally for purposes such as licensing.

For the `aggregator` spider--are there any current examples? The examples provided above seem to be dead / I couldn't find those brands publishing aggregated location data online. Is `our_airports` an example?

Cj-Malone:
Yeah, I'll move our_airports in this PR. I've also written/half written a spider for OpenChargeMap. I'll do that in a follow up PR.

Cj-Malone:
> I'm thinking [SOCi/Brandify](https://wiki.openstreetmap.org/wiki/Brandify_Imports) or [Uberall/Navads](https://wiki.openstreetmap.org/wiki/Navads_Imports).

Assuming they give us permission/wrote the spider themselves. I am not and will not endorse scraping aggregators that don't want to be scraped. If they want to sell data, we won't steal it.

This is somewhat of a flip for "brand" spiders, Walmarts terms of service may say we aren't allowed to scrape the website, but they actually mean "don't offer price comparison sites", they do want us to advertise there locations. They literally pay Brandify to do it.

davidhicks:
> > I'm thinking [SOCi/Brandify](https://wiki.openstreetmap.org/wiki/Brandify_Imports) or [Uberall/Navads](https://wiki.openstreetmap.org/wiki/Navads_Imports).
> 
> Assuming they give us permission/wrote the spider themselves. I am not and will not endorse scraping aggregators that don't want to be scraped. If they want to sell data, we won't steal it.
> 
> This is somewhat of a flip for "brand" spiders, Walmarts terms of service may say we aren't allowed to scrape the website, but they actually mean "don't offer price comparison sites", they do want us to advertise there locations. They literally pay Brandify to do it.

Are there any brand agencies publishing open data on behalf of their clients? https://wiki.openstreetmap.org/wiki/Brandify_Imports and https://wiki.openstreetmap.org/wiki/Navads_Imports both look to be years out of date and neither current company appears to still publish open data on behalf of their clients? There are many paid data services seemingly duplicating (or potentially using ATP data) listed at https://datarade.ai/search/products?keywords=Business+Location+Data and I agree these aren't scrapable by ATP even if this was desired (not because of copyright but because doing so requires subscriptions and a contract/EULA on how such data can be used).

Cj-Malone:
> Are there any brand agencies publishing open data on behalf of their clients? https://wiki.openstreetmap.org/wiki/Brandify_Imports and https://wiki.openstreetmap.org/wiki/Navads_Imports both look to be years out of date and neither current company appears to still publish open data on behalf of their clients?

The 2 examples tried to be involved in OSM, but were chased off, and no longer exist due to mergers and acquisitions.

(This comment is not about any specific companies)

"open data" is a foreign phrase to a lot of this these companies, but they are often "open access, if you know who to ask" and provided basically as "no rights reserved, perpetual, none exclusive". CC0/PD in our world, some might have a minor attribution requirement. I work with data from several of these companies and really really want them to write there own spiders. Optimistically this is set up for them.

Cj-Malone:
Alternatively this could become a place Foursquare OS Places or Overture Places spiders. I've written both, but that would be more of an infrastructure question than a "which directory" question.

locations/settings.py
matkoniecz: > That would be government spider, not an aggregator. I have no issue with 1 POI being a "duplicate" of another spider, this already happens.

Should I then close #12190 and #12186 as "it may happen, data consumers should deal with it somehow"
And maybe document it somewhere (I was surprised by it)

hfs:
> Are there any brand agencies publishing open data on behalf of their clients?

https://rocketdata.io/ imports locations into OSM. See https://wiki.openstreetmap.org/wiki/Rocketdata_import and e.g. https://community.openstreetmap.org/t/importing-data-from-company-owners-using-the-rocketdata-io-service/91712/30 (they have one forum thread per country).

panasonic-mafia (approved):

docs/SPIDER_TYPE.md
```diff
@@ -0,0 +1,23 @@
+## Types of spiders
+
+The vast majority of spiders are collecting first party data about a specific brand/chain.
+We sometimes call these "brand spiders", these should be well categorised and branded by us, and have original data from
```
davidhicks, line 4: Can examples be provided, as well as what should be excluded? Obviously this includes retail chains and restaurant chains. But what about government-operated assets that are POIs people are usually interested in such as public playgrounds, public toilets, etc? And what about branded bike rental locations, car share locations and electric car charging stations (which may be colocated with another feature such as a hotel or shopping centre)? At what point does something stop being a POI and becomes an `infrastructure_spider` -- park benches, rubbish bins, drinking fountains, etc?

docs/SPIDER_TYPE.md
```diff
@@ -0,0 +1,23 @@
...
+the brand.
+We store them in [locations/spiders](../locations/spiders). and  they contain `spider:lineage=S_ATP_BRANDS` in the outputted
+`dataset_attributes`.
+
+We also collect government Open Data.
```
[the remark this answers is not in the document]
davidhicks: If a government were to publish a list of named mountain peaks, would this fit in `government_spiders` or would a separate new category for natural features be better to include?

docs/SPIDER_TYPE.md
```diff
@@ -0,0 +1,23 @@
...
+These should also contain license details in `dataset_attributes`.
+These spiders are typically bigger than brand spiders, often collecting all of a specific category for a country,
+the age of the dataset, or records in the dataset, may be uncomfortably old.
+We store them in [locations/government_spiders](../locations/government_spiders), there lineage is `S_ATP_GOVERNMENTS`.
+
+We also collect infrastructure data, these wouldn't typically be called Points of Interest.
+These may be anything from trees to electrical sub stations.
```
davidhicks, line 16: Some ideas for other examples to include: road and public transport features, council assets such as street trees, lamp posts, rubbish bins and park benches, utility assets such as manholes, electrical substations, pumping stations, etc.

Depending on where more useful POIs (such as playgrounds, public toilets, public showers near a beach, etc) end up, it might be worth also noting what should not be included under `infrastructure`.

docs/SPIDER_TYPE.md
```diff
@@ -0,0 +1,23 @@
+## Types of spiders
+
+The vast majority of spiders are collecting first party data about a specific brand/chain.
+We sometimes call these "brand spiders", these should be well categorised and branded by us, and have original data from
+the brand.
+We store them in [locations/spiders](../locations/spiders). and  they contain `spider:lineage=S_ATP_BRANDS` in the outputted
+`dataset_attributes`.
+
+We also collect government Open Data.
+These should also contain license details in `dataset_attributes`.
```
davidhicks, line 10: alltheplaces.xyz states the downloadable weekly run result is CC-0 so setting a license in `dataset_attributes` sends mixed messages. In jurisdictions that ATP users seem to be within, a database of facts is not copyrightable. But if there are some creative fields in source data such as `company motto`, then perhaps this shouldn't be extracted at all so as to not pollute the weekly download on alltheplaces.xyz?

docs/SPIDER_TYPE.md
```diff
@@ -0,0 +1,23 @@
+## Types of spiders
```
davidhicks, line 1: Could we add a general comment that it is OK for multiple spiders of different types to extract duplicate features, regardless of whether different fields of data are available. For example, a brand spider may extract a feature from a hotel chain website, and the same hotel feature may be additionally extracted from a government spider for registration of hotels, and additionally extracted from a hotel aggregator service spider.

davidhicks (changes requested):
Added some suggestions for documentation improvement to help clarify the intended purpose of each type of spider.

docs/SPIDER_TYPE.md
Cj-Malone: Maybe, but it's not new to this PR and I don't see it flowing in this file.

docs/SPIDER_TYPE.md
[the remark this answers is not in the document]
davidhicks: As another thought--there are non-profit organisations and industry associations which may maintain lists of their member's offices or POIs. For example, a database of vineyards from a wine growers association. Could these be included under `government_spiders` too?

iandees (commented):
Apologies for taking so long to give feedback on this.

We have a lot of spiders and we're starting to bring in non-POI data (see https://github.com/alltheplaces/alltheplaces/discussions/12498) where our consumers do not expect it. At a high level, moving spiders into separate directories and using "build tagging" of some kind is a good way to solve these problems, so thanks for putting this together.

A couple requests:
- Instead of `locations/spiders/` and `locations/government_spiders/`, etc. Can the directory structure be changed to `locations/spiders/` and `locations/spiders/government/`? (With the corresponding change to the settings)
- Unless there's a particular reason for the existing naming scheme, can we use something more Pythonic for the `Lineage` enum? Something like `LINEAGE_AGGREGATOR = "atp.aggregators"` and `LINEAGE_UNKNOWN = "unknown"`?
- In addition to applying the lineage by file path, could we also do it via class field? This would allow us to be more flexible with directory structure in the future.
- I don't know if "lineage" is the best term for this. I can't think of a better term and want to keep this moving. If we're using "lineage", we should also refer to it that way in the documentation.

Cj-Malone:
> Apologies for taking so long to give feedback on this.

No worries, I took this out of draft today.

> Instead of locations/spiders/ and locations/government_spiders/, etc. Can the directory structure be changed to locations/spiders/ and locations/spiders/government/? (With the corresponding change to the settings)

Can do, this actually stays in the same "module" so the settings change wont be needed.

> Unless there's a particular reason for the existing naming scheme, can we use something more Pythonic for the Lineage enum? Something like LINEAGE_AGGREGATOR = "atp.aggregators" and LINEAGE_UNKNOWN = "unknown"?

Changing the string `S_ATP_BRANDS` would break compatibility for us, I can do it if it's a hard requirement. We have several others with `S_TT_*`, that aren't relevant here. Happy to change the variable names.

> In addition to applying the lineage by file path, could we also do it via class field? This would allow us to be more flexible with directory structure in the future.

I can do that.

> I don't know if "lineage" is the best term for this. I can't think of a better term and want to keep this moving. If we're using "lineage", we should also refer to it that way in the documentation.

100% docs should line up, I'll rename that file.

panasonic-mafia:
I agree with the whole approach here. Code also looks good to me. I think documentation related questions probably can be addressed in another PRs, but IMO is already sufficient enough.

davidhicks:
With #12726 and #12727 merged into this branch there would only be a small number of "infrastructure", "government" and "aggregator" spiders remaining in the parent directory for sorting through.

iandees:
I'm OK to merge this. My hold up was the enum values showing up in our output data. I'm ok with merging this, but reserve the right to change my mind. I'd be curious how much work patching over any breakage changing that format would cause, @Cj-Malone.

Cj-Malone:
As far as I can tell, there shouldn't be any crashes. It's a bunch of little changes on our end, links/url formatters will break. After 5 runs and a few code changes we'd be out the other end.

PR #12178, opened 2025-02-06
[Kia US] Add category

karen-kerr:


karen-kerr:
^^ I'm not sure why category is not being set here.
I've removed setting a service centre as this is not possible to determine from the data.

PR #12256, opened 2025-02-15
Add Ergon Energy power poles (AU) (992k poles)

davidhicks:


davidhicks:
Why are there so many power poles? Queensland if it were a country would have the 20th largest land area amongst all countries.

PR #12280, opened 2025-02-17
Add Greater Geelong City Council crossings (AU) (172 crossings)

davidhicks:
Formal children/school crossings of roads.


iandees:
🤔 We're really stretching the concept of All the Places here ...

I don't particularly have a problem with including all this data, but I worry that consumers of the data might not be expecting to see this kind of stuff in here.

PR #12295, opened 2025-02-20
Espressolab: Try proxy

impiaaa:


iandees:
You could try setting a user agent instead?

impiaaa:
It works fine for me locally without a user-agent set. I thought that meant that the UA is fine, it just needs a proxy to run on the server.

PR #12338, opened 2025-02-24
Black Bear Diner: Fix spider by making regex more tolerant

impiaaa:
also use named matches which I should have been doing already


impiaaa:
see also  #12345

PR #12345, opened 2025-02-25
black_bear_diner_us: fix spider

davidhicks:


impiaaa:
see also #12338

davidhicks:
Sorry I missed #12338 and hadn't seen this spider was already fixed.

I think we need to merge both together as there are some things each of us missed. For example, you extracted phone numbers in a much better way (which I ignored improving). And also allowed addresses to have multiple lines, something I also didn't improve. And then in my merge the branch name is more correctly extracted by removing a prefix, not a suffix (most locations appear to be "Black Bear Diner" with no branch name or "Black Bear Diner - Branch Name").

impiaaa:
I assumed that the suffix in the caption was the branch name, but it seems to just be the city name. I've updated my PR to use the branch name from the "title" key like you did.

davidhicks:
Thanks I'll close this duplicate PR.

PR #12365, opened 2025-02-27
[CleanStringsPipeline] Add pipeline to unescape strings

Cj-Malone:
I've thought about this before, and was a little nervous that `html.unescape` could be destructive to none escaped strings. I don't think this is the case.

Tests still need to be written.

### DIFF
--- a/locations/pipelines/clean_strings.py
+++ b/locations/pipelines/clean_strings.py
@@ -10,9 +10,12 @@ def clean_string(val: str) -> str:
 
 
 class CleanStringsPipeline:
+    skipped_keys = {"ref", "website"}
 
     def process_item(self, item: Feature, spider: Spider):
         for key, value in item.items():
+            if key in self.skipped_keys:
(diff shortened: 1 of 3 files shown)


locations/pipelines/clean_strings.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def clean_string(val: str) -> str:
+    return html.unescape(val).strip()
+
+
+class CleanStringsPipeline:
+
+    def process_item(self, item: Feature, spider: Spider):
+        for key, value in item.items():
```
panasonic-mafia, line 15: Maybe some fields can be excluded, e.g. ones which might contain URL and `ref`s.

Cj-Malone: Seems reasonable, on it.

Cj-Malone: Done

PR #12370, opened 2025-02-27
[CleanStringsPipeline] exclude fields that can have URLs from clean

panasonic-mafia:
Rel: #12365


iandees:
Does the nsi_id have a URL in it? I thought it was a simple string.

Cj-Malone (approved):
It wont have a url, but it's also a little pointless to check.

davidhicks:
I thought social media fields weren't meant to be URLs but are instead meant to be the account name/handle?

PR #12374, opened 2025-02-27
[insights command] drop a few long in the tooth options

mjoe999:
Given the march of time and in particular #12365 it is a good moment to trim some of the less relevant options to the `insights` command where the function / data is pretty much in the stats JSON output. Probably a little bit more to come in this area.


Cj-Malone (approved):
Thank you, both of these are now in pipelines.

PR #12389, opened 2025-03-01
settings: refactor and remove circular dependency with geojson exporter

davidhicks:
Most of the Scrapy settings which accept class references as strings also accept the class type itself. The difference is lazy loading of extensions, pipelines and exporters (and similar) with class references as strings, versus eager loading of these classes when Scrapy is started. Almost everything of substance that settings.py was going to lazy load is eventually loaded in every ATP crawl anyway, so lazy loading provides little to no benefit. _IF_ there was an expensive pipeline in the future which took considerable time to load then it would make better sense to selectively load it using a spider's overridden SETTINGS attribute, rather than apply that pipeline to every single item from every spider.

For the few classes settings.py called out to (such as exporters) which would typically not always be loaded, these are extremely simple in nature and the performance benefit of lazy loading these simple classes will be negligable.

The benefit of eager loading from settings.py is the ability to more readily detect syntax errors in extensions, pipelines and exporters (and similar) without having to wait for a complete crawl to complete, and without having to attempt a crawl of each combination of extensions, pipelines and exporters (and similar) to ensure all are free of syntax errors.

To enact this change, extensions, pipelines and exporters (and similar) can no longer import from settings.py as this would create a circular redundancy. The GeoJSON exporter was unnecessarily importing from settings.py and this has had to be removed in favour of an alternative method that doesn't create a circular dependency.

### DIFF
--- a/locations/exporters/geojson.py
+++ b/locations/exporters/geojson.py
@@ -1,11 +1,11 @@
-import base64
-import datetime
-import hashlib
-import io
-import json
-import logging
-import uuid
-from typing import Any, Generator, Type
+from base64 import urlsafe_b64encode
+from datetime import datetime
+from hashlib import sha1
+from io import StringIO
+from json import dump
+from logging import warning
+from uuid import uuid1
+from typing import Any
 
 from scrapy import Item, Spider
 from scrapy.exporters import JsonItemExporter
@@ -13,7 +13,7 @@ from scrapy.utils.misc import walk_modules
 from scrapy.utils.python import to_bytes
 from scrapy.utils.spider import iter_spider_classes
 
-from locations.settings import SPIDER_MODULES
+from locations.items import Feature
 
 mapping = (
     ("addr_full", "addr:full"),
@@ -53,8 +53,10 @@ def item_to_properties(item: Item) -> dict[str, Any]:
     # Add in the extra bits
     if extras := item.get("extras"):
         for key, value in extras.items():
-            if value is not None and value != "":
-                # Only export populated values
+            if value is not None and value != "" and key != "@spider" and key != "@spider_classes":
+                # Only export populated values and ignore certain attributes
+                # which are output in the GeoJSON header as
+                # non-feature-specific attributes.
                 props[key] = value
 
     # Bring in the optional stuff
@@ -86,11 +88,11 @@ def item_to_geometry(item: Item) -> dict:
                 "coordinates": [float(item["lon"]), float(item["lat"])],
             }
         except ValueError:
-            logging.warning("Couldn't convert lat (%s) and lon (%s) to float", lat, lon)
+            warning("Couldn't convert lat (%s) and lon (%s) to float", lat, lon)
     return geometry
 
 
-def item_to_geojson_feature(item: Item) -> dict:
+def item_to_geojson_feature(item: Feature) -> dict:
     feature = {
         "id": compute_hash(item),
         "type": "Feature",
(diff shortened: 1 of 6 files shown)


davidhicks:
Inspired by https://github.com/alltheplaces/alltheplaces/commit/10e5bcd1429a5a0aa363e7dcef90d2c72df517b6#diff-f62e80b00a49c4c9ab1806ef0d0972551c39f52e52aee12e9dc6cea72c51111eR108

locations/exporters/geojson.py
```diff
@@ -101,72 +101,93 @@ def item_to_geojson_feature(item: Item) -> dict:
...
+    if item_ref and item_spider_name:
+        sha1_hash = sha1(item_ref.encode("utf8"))
+        sha1_hash.update(item_spider_name.encode("utf8"))
+        return urlsafe_b64encode(sha1_hash.digest()).decode("utf8")
 
-    return base64.urlsafe_b64encode(sha1.digest()).decode("utf8")
+    # If a spider has no_refs = True, generate a GeoJSON feature identifier
+    # as a random UUID that will change each time a crawl and export is
+    # completed.
+    return str(uuid1()).encode("utf8")
```
Cj-Malone, line 119: Please lets not break ids. This is such a headache for me.

locations/exporters/geojson.py
```diff
@@ -53,8 +51,10 @@ def item_to_properties(item: Item) -> dict[str, Any]:
     # Add in the extra bits
     if extras := item.get("extras"):
         for key, value in extras.items():
-            if value is not None and value != "":
-                # Only export populated values
+            if value is not None and value != "" and key != "@spider" and key != "@spider_classes":
+                # Only export populated values and ignore certain attributes
+                # which are output in the GeoJSON header as
+                # non-feature-specific attributes.
```
Cj-Malone, line 57: This is a major change.

locations/pipelines/apply_source_spider_attributes.py
```diff
@@ -0,0 +1,14 @@
+from inspect import getmro
+
+from scrapy.utils.trackref import object_ref
+
+
+class ApplySourceSpiderAttributesPipeline:
+    def process_item(self, item, spider):
```
Cj-Malone, line 7: Is Python introspection good enough to do this per POI?

locations/exporters/geojson.py
davidhicks: Logic was meant to be unchanged here. Both old and new code calculate the GeoJSON ID as `sha1(item_ref + spider_name)` or if `no_refs=True` (item has no reference ID) then it'll be a random UUID instead. I've just spotted an error though--if `no_refs=True` the old code would emit `sha1(uuid1())` and not emit `uuid1()` directly. I'll get that fixed.

locations/exporters/geojson.py
davidhicks: The difference here is in GeoJSON output to not emit `properties["extras"]["@spider"]` for every single POI but rather to only emit `@spider` for an entire GeoJSON FeatureSet. We guarantee elsewhere in the exporter that a single GeoJSON output is not the product of two or more spiders.

Is there anyone using `properties["extras"]["@spider"]` at the moment in GeoJSON output? I can add this back in if it's potentially needed for backwards compatibility.

locations/pipelines/apply_source_spider_attributes.py
davidhicks: The problem is Scrapy's `BaseItemExporter` doesn't receive an instance of the spider per https://docs.scrapy.org/en/latest/_modules/scrapy/exporters.html#BaseItemExporter -- it only receives the scraped `Feature`.

Previously the GeoJSON exporter would search `locations/spiders/*.py` for all classes, and test each class name found to see if it matches `Feature.extras["@spider"]`. This new approach should now avoid the need to scan 3500+ files within `locations/spiders/*.py` once per every spider executed to find whichever file/spider class is applicable.

There's an example at https://docs.scrapy.org/en/latest/topics/exporters.html#using-item-exporters where a Scrapy Item Pipeline is used instead to conduct an export, but this looks less than ideal compared to adding the spider class types (multiple inheritance could apply) into `Feature` so the exporter can then access these class attributes.

Cj-Malone:
+1 on big diff. I get confused easily, and I find this hard to review.

locations/exporters/geojson.py
Cj-Malone: Yes we do. Also everyone who uses https://www.alltheplaces.xyz/map does. Why do you need backwards compatibility when the "forward" change isn't needed nor requested?

locations/exporters/geojson.py
panasonic-mafia: IDs are pretty important. I'm suggesting adding a unit test for this. 
Also, could this refactoring be done in a separate PR please?

davidhicks:
I can split this change into two parts, the second dependent on the first:
1. Remove the circular redundancy with the GeoJSON exporter currently obtaining from `settings.py` the path to scan (`locations/spiders/*.py`) to find the spider class that matches the `item["extras"]["@spider"]` value seen by the GeoJSON exporter. This may improve performance by not requiring 3500+ Python files to be read from disk every crawl.
2. (being debated) Whether `settings.py` should use lazy loading (old status quo) or eager loading.

Can I get a straw poll on whether the two parts identified above should still go ahead, or only one, or none of them?

If any of the above are decided, I can also change:
1. Keep emitting `properties["@spider"]` for each feature in GeoJSON output as ATP consumers rely on this per-feature duplicated attribute currently, and are not taking the `@spider` attribute from the GeoJSON `FeatureSet` instead, avoiding duplication in GeoJSON output and larger GeoJSON files.
2. Add a unit test for the `compute_hash` function for an ATP `Feature` which generates an ATP-specific identifier for each `Feature`. I would then propose that this `compute_hash` function become a method of the `Feature` class, avoiding the need to import a static function from a utils module. Longer term (and for discussion later and elsewhere) I think we should migrate to a different more stable identifier which ignores the ATP spider name and instead relies upon the Wikidata identifier of the brand or operator of a feature, the OSM category of an feature and the identifier of the feature from source data.

PR #12485, opened 2025-03-02
[Homebase GB] Fix Spider

TheUKHighStreet:


Cj-Malone:
New count seems accurate https://en.wikipedia.org/wiki/Homebase

PR #12540, opened 2025-03-06
Add Greater Metropolitan Cemeteries Trust graves (AU) (551k graves)

davidhicks:


Cj-Malone (changes requested):
How are you intending to use this data?

davidhicks:
> How are you intending to use this data?

I don't use any ATP data currently other than admiring the ever growing map at alltheplaces.xyz

Government open data is released at least in Australia often without anyone requesting it or wanting to use it. Then at some GovHack style event a few years later, someone will find a creative way to use the data. The immediate benefit of this particular spider (grave sites in cemeteries) is to capture historical data that could be more useful to future users of the data. For example, some researcher comes along and wants to understand how burial practices have changed in the past X years. Are more or less people choosing to be buried on top of each other as opposed to X years ago. And Wikidata and Wikipedia articles on cemeteries would now have more accurate data, consistently tagged across cemeteries, on how many graves exist, how many people are buried in those graves, be able to display a chart of graves/persons buried over time, etc. Perhaps someone if interested could create an opencemeteriesmap similar to https://opentrees.org/ for council managed trees.

PR #12541, opened 2025-03-06
Revert "Bump pygeohash from 1.2.0 to 2.1.0"

Cj-Malone:
Reverts alltheplaces/alltheplaces#12495

I've also seen some weirdness with this.


Cj-Malone:
"Weirdness" being log level not being respected.

iandees:
Interesting that this happens. I don't see anything in the pygeohash diff that would cause logging behavior changes.

PR #12562, opened 2025-03-10
Add IGA (AU) (1257 supermarkets)

davidhicks:


locations/spiders/iga_au.py
```diff
@@ -0,0 +1,98 @@
...
+                url=google_maps_url_cid,
+                callback=self.parse_coordinates_cid,
+                meta={"properties": properties, "handle_httpstatus_list": [429]},
+            )
+        elif self.follow_google_maps_urls and google_maps_url_short:
+            yield Request(
+                url=google_maps_url_short,
+                callback=self.parse_coordinates_short,
+                meta={"properties": properties, "handle_httpstatus_list": [429]},
+            )
```
iandees, line 71: Not comfortable taking Google data here.

iandees (changes requested):

locations/spiders/iga_au.py
davidhicks: Updated to remove Google Maps CID link following. The goo.gl redirector service will be shut down soon so hopefully it forces the brand to fix their store finder page to include coordinates on the page (even if just a Google Maps URL that includes coordinates in the URL).

If anyone wants to follow Google Maps CID links, they could write a Scrapy middleware to do so.

PR #12568, opened 2025-03-10
Add Dave's Hot Chicken (US) (amenity/fast_food) (256 restaurants)

davidhicks:


locations/spiders/daves_hot_chicken_us.py
```diff
@@ -0,0 +1,35 @@
...
+    start_urls = ["https://daveshotchicken.com/api/v2/locations"]
+    locations_key = "locations"
+
+    def post_process_item(self, item: Feature, response: Response, feature: dict) -> Iterable[Feature]:
+        if not feature["available"]:
+            return
+        item["street_address"] = item.pop("addr_full", None)
+        item["opening_hours"] = OpeningHours()
+        for day_hours in feature["hours"]:
+            item["opening_hours"].add_range(day_hours["day"].title(), day_hours["open"], day_hours["close"])
```
Cj-Malone, line 24: output looks a bit broken: `Mo-Th 11:00-23:00; Fr 11:00-24:00; Sa 00:00-00:00,11:00-24:00; Su 00:00-00:00,11:00-23:00`

locations/spiders/daves_hot_chicken_us.py
```diff
@@ -0,0 +1,35 @@
...
+        apply_category(Categories.FAST_FOOD, item)
+        apply_yes_no(Extras.DRIVE_THROUGH, item, feature["services"]["drivethru"], False)
+        apply_yes_no(Extras.DELIVERY, item, feature["services"]["delivery"], False)
+        apply_yes_no(Extras.TAKEAWAY, item, feature["services"]["pickup"], False)
+        apply_yes_no(Extras.INDOOR_SEATING, item, feature["services"]["dinein"], False)
+        card_types = feature["payments"]["cardtypes"]
+        apply_yes_no(PaymentMethods.AMERICAN_EXPRESS, item, "American Express" in card_types, False)
+        apply_yes_no(PaymentMethods.DISCOVER_CARD, item, "Discover" in card_types, False)
+        apply_yes_no(PaymentMethods.MASTER_CARD, item, "MasterCard" in card_types, False)
```
Cj-Malone, line 33: False is awful brave here, what happens if they have `Mastercard` instead?

Cj-Malone (changes requested):

locations/spiders/daves_hot_chicken_us.py
davidhicks: I can add a warning if a different card type is discovered. But I also wonder if we should also keep track of stats for payment methods and then at the end of a spider's crawl, check for unlikely events such as a brand with a common point of sale system accepting MasterCard at 100% of stores but only accepting Visa at 80%.

locations/spiders/daves_hot_chicken_us.py
matkoniecz: ops

I will add one more test and fix this

locations/spiders/daves_hot_chicken_us.py
Cj-Malone: Or just remove `, False` so we don't lie.

Cj-Malone: > opening hours

We don't set `x Closed` on the absence of day x...

> For this spider though, I've just checked whether any of the 256 features reject any of the 4 payment methods and none do.

The argument isn't for now, it's for next week when nobody is looking at the spider. Or are you gonna check it every week?

This isn't just me wanting to be safer, I've already had to do PRs for exactly this.

> `source:xx`

Why are you overcomplicating this? Why would you want to be liable for saying you can't use MasterCard at a location when you aren't 100% sure.

locations/spiders/daves_hot_chicken_us.py
matkoniecz: trued to reproduce it, failed so far. For example I based one of test cases on output here

```
o = OpeningHours()
o.add_range("Mo", "11:00", "23:00")
o.add_range("Tu", "11:00", "23:00")
o.add_range("We", "11:00", "23:00")
o.add_range("Th", "11:00", "23:00")
o.add_range("Fr", "11:00", "24:00")
o.add_range("Sa", "11:00", "24:00")
o.add_range("Su", "11:00", "23:00")
o.as_opening_hours()
```

it gives

`Mo-Th 11:00-23:00; Fr-Sa 11:00-24:00; Su 11:00-23:00`

I will try to dig deeper into this.

matkoniecz: I opened #12611 - but it merely ads test mentioned above as I have not managed to trigger the bug

I now tried to run code from this PR and it actually creates this `00:00-00:00` so at least it is not looking like a bizarre bug

-------------------------------

I added log in this PR:

```
        for day_hours in feature["hours"]:
            print("add_range", day_hours["day"].title(), day_hours["open"], day_hours["close"])
            item["opening_hours"].add_range(day_hours["day"].title(), day_hours["open"], day_hours["close"])
        print(item["opening_hours"].as_opening_hours())
        print()
        print()
        print()
```

it reveals that 

```
add_range Thu 10:30 0:00
add_range Fri 10:30 1:00
add_range Sat 10:30 1:00
add_range Sun 10:30 0:00
add_range Mon 10:30 0:00
add_range Tue 10:30 0:00
add_range Wed 10:30 0:00
Mo-Fr 00:00-00:00,10:30-24:00; Sa-Su 00:00-01:00,10:30-24:00
```

results in this silliness

I guess that `0:00` instead of `24:00` confuses code, I will look into what is going on here

matkoniecz: I think I found root cause.

`if close_time == "24:00" or close_time == "00:00":` is not accounting for possible different formatting of close_time

https://github.com/alltheplaces/alltheplaces/blob/ecc5657b288de77d988b93ae24e70b231a66ec72/locations/hours.py#L945

this was introducing some weirdness before (`o.add_range("Mo", "11:00", "0:00")` being interpreted differently than `o.add_range("Mo", "11:00", "00:00")`) but my PR made it worse

if someone would actually use `time_format` it would be even worse

matkoniecz: and PR that actually fixes root issue (but NOT its sibling, not sure yet how to do so) is at https://github.com/alltheplaces/alltheplaces/pull/12613

locations/spiders/daves_hot_chicken_us.py
impiaaa: To me it seems like a break from ATP's goals. At least, I've tried to write my spiders to only include information that the website states explicitly. If the website doesn't _say_ they accept Mastercard, that doesn't mean that the establishment does _not_ accept Mastercard, so the spider shouldn't say one way or the other. Someone on the ground would have to go there to verify. This is also the case in OSM: if a store has a sign out front that says they accept AmEx, that doesn't mean you should tag it with `payment:discover_card=no`, you'd have to go inside and try. And I don't think comparison with other locations is enough to imply that either. Just like one locations' Discover sign might have fallen off their window, the person who set up the website might have forgotten to add Mastercard to the payment list on one location.

Cj-Malone (approved):

PR #12613, opened 2025-03-13
harden close_time parsing

matkoniecz:
adds a regression test and fixes problem reported in https://github.com/alltheplaces/alltheplaces/pull/12568#discussion_r1987046526

this problem existed in past (`10:00-0:00` and `10:00-00:00` was resulting in a different output) but become more noticeable with #11989

I am quite unhappy about this fix as it leaves close_time ending at 24:00 (specified in unusual formats) with the same bug

but problem here is that using `time.strptime` on `24:00` makes it explode (at least with `"%H:%M"`) so you cannot parse and then check what is going on

but I have no better ideas right now

### DIFF
--- a/locations/hours.py
+++ b/locations/hours.py
@@ -950,6 +950,11 @@ class OpeningHours:
             open_time = time.strptime(open_time, time_format)
         if not isinstance(close_time, time.struct_time):
             close_time = time.strptime(close_time, time_format)
+            if close_time.tm_hour == 0 and close_time.tm_min == 0:
+                # weird format not caught by checks above
+                # may be 0:00 or even more divergent if time_format
+                # parameter was used with some exotic value
+                close_time = time.strptime("23:59", "%H:%M")
 
         self.days_closed.discard(day)
         self.day_hours[day].add((open_time, close_time))
--- a/locations/hours.py
+++ b/locations/hours.py
@@ -942,9 +942,9 @@ class OpeningHours:
         if isinstance(close_time, str):
             if close_time.lower() in closed:
                 return
-            if close_time == "24:00" or close_time == "00:00":
+            if close_time in ("24:00", "00:00", "0:00"):
                 close_time = "23:59"
-            if close_time == "24:00:00" or close_time == "00:00:00":
+            if close_time in ("24:00:00", "00:00:00"):
                 close_time = "23:59:00"
         if not isinstance(open_time, time.struct_time):
             open_time = time.strptime(open_time, time_format)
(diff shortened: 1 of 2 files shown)


matkoniecz:
oh, and in general I suspect that using something designed to hold hours may be better than a bit of hack of using datetime structure here

but I am not signing up for this refactoring for now

davidhicks (approved):
Thanks!

locations/hours.py
```diff
@@ -942,14 +942,19 @@ def add_range(self, day, open_time, close_time, time_format="%H:%M", closed=CLOS
         if isinstance(close_time, str):
             if close_time.lower() in closed:
                 return
-            if close_time == "24:00" or close_time == "00:00":
+            if close_time in ("24:00", "00:00", "0:00"):
                 close_time = "23:59"
-            if close_time == "24:00:00" or close_time == "00:00:00":
+            if close_time in ("24:00:00", "00:00:00"):
                 close_time = "23:59:00"
```
davidhicks, line 948: On second check, `OpeningHours.as_opening_hours()` only does a midnight check on `23:59` not `23:59:00`. Perhaps this line should be `close_time = "23:59:59"` and this added to the midnight check in  `OpeningHours.as_opening_hours()`? That way the code is accommodating both precisions of time--to the minute and to the second.

davidhicks (changes requested):

locations/hours.py
matkoniecz: I will start from adding this case to tests.

matkoniecz: Note that 

```
    o = OpeningHours()
    o.add_range("Mo", "11:00:00", "24:00:00", time_format="%H:%M:%S")

    assert o.as_opening_hours() == "Mo 11:00-24:00"
```

and

```
    o = OpeningHours()
    o.add_range("Mo", "11:00:00", "00:00:00", time_format="%H:%M:%S")

    assert o.as_opening_hours() == "Mo 11:00-24:00"
```

are passing. Not relooked into code yet.

matkoniecz: oh, I see - that is because second part of the hack does `time.strftime("%H:%M", h[1]).replace("23:59", "24:00")`


which BTW seems to changing also say `23:59:30` into `24:00`

matkoniecz: see

https://github.com/alltheplaces/alltheplaces/blob/4fb79648ae39fc26261e7778fa4e6bfdbf64f0df/locations/hours.py#L997

in general the entire `struct_time` looks poorly fitting here as it cannot hold `24:00` while having say year field

davidhicks (approved):
Thanks for the extra test cases :)

PR #12615, opened 2025-03-13
[Pizza Hut TR][KFC TR] Remove out of business spiders

Cj-Malone:
It sounds like these are no more,

https://munsifdaily.com/kfc-turkey-bankruptcy-gaza-boycott/


panasonic-mafia (approved):
LGTM, thanks. NSI will probably need to reflect it, I'll do the update.

PR #12631, opened 2025-03-14
[MBPLC GB] Add brands

TheUKHighStreet:
Added some brands, added some wikidata tags, changed some categories.


TheUKHighStreet:
Note that this store finder hasn't been updated since 17-06-2024 so some of these pubs are now closed. I'm not sure where this should be noted.

PR #12632, opened 2025-03-16
Fixed yettel_bg spider

Dimitar5555:


locations/spiders/yettel_bg.py
```diff
@@ -1,8 +1,7 @@
-import io
-
-from openpyxl import load_workbook
```
iandees, line 3: Can you check to see if this library is used elsewhere?

Dimitar5555: It's currently being used in `dbe_gov_schools_za.py` and `dbe_gov_ecd_za.py`.

PR #12680, opened 2025-03-21
Generate Wikidata links for all Wikidata tags in preview

brawer:


iandees:
What uses this preview HTML?

brawer:
Oh, this isn't the page that displays the preview for a spider run, eg. when someone sends a pull request?

I noticed that in those pull request previews, there are nicely formatted links for values of). `brand:wikidata`. But for `operator:wikidata`, `network:wikidata` and similar tags, no such links are generated. So I searched for JavaScript code in the ATP repo that looked like it generates the preview. I didn't actually test it (because I wouldn't know how).

iandees:
You are correct to think that it's the map that gets used in PR comments. I forgot that I renamed it when I uploaded it to S3 (preview.html -> map.html).

brawer:
So this file should be renamed to map.html? I'll be glad to do that, but perhaps in a separate PR?

iandees:
Nope as it stands right now I'll have to manually upload this and remember to change the file name.

In the future we could move this to https://github.com/alltheplaces/alltheplaces.xyz/ and change the link that is generated for pull requests to point to alltheplaces.xyz/preview.html

iandees:
This is deployed. Thanks for the PR.

PR #12691, opened 2025-03-23
[Salvation Army GB] Fix spider

TheUKHighStreet:


TheUKHighStreet:
The robots.txt file contains broken links. The charity shops only seem to be on page 3 of the site map so I've replaced the robots.txt file with page 3 of the site map.

I'm not sure what the failing text means.

PR #12692, opened 2025-03-23
fibank_bg use Latin brand

Dimitar5555:
Ref: https://github.com/osmlab/name-suggestion-index/commit/8fdb52092e3fa6c2ee27b3075a0ba20ddce86385


Cj-Malone:
Oh, pytest is failing not because this brand isn't matching the current NSI release but because the other isn't matching. And #12693 is failing because of this brand change.

PR #12724, opened 2025-03-26
Switch ref:google tag to ref:google:place_id

davidhicks:
There are other types of Google IDs such as Google Maps CID which differ from Google Maps Place IDs. Split the ref:google tag into more specific tags to allow for both CIDs and Place IDs.

Note that from
https://taginfo.openstreetmap.org/keys/ref%3Agoogle#values it appears Google Maps redirect URLs are used for ref:google. These appear to be deprecated, as will the goo.gl redirector some time later in 2025. There appears to be no other existing ref tags used for extracing Google Maps CIDs or Place IDs so we're free to make something up.


Cj-Malone:
Please noone add these to OSM

PR #12743, opened 2025-03-28
Add new storefinder, NomNom, and move Blaze Pizza, Panda Express, and Wingstop to it

impiaaa:


locations/storefinders/nomnom.py
```diff
@@ -0,0 +1,92 @@
...
+    def parse_locations(self, response: TextResponse) -> Iterable[Feature]:
+        for location in response.json()["restaurants"]:
+            item = DictParser.parse(location)
+            item["ref"] = location["extref"]
+            item["branch"] = location["name"].replace(location["storename"], "").strip()
+            item["name"] = location["storename"]
+
+            apply_yes_no(Extras.DELIVERY, item, location["candeliver"] or location["supportsdispatch"])
+            apply_yes_no(Extras.TAKEAWAY, item, location["canpickup"] or location["supportscurbside"])
+            apply_yes_no(Extras.DRIVE_THROUGH, item, location["supportsdrivethru"])
```
davidhicks, line 72: All these brands appear to block out-of-country access so I can't check myself--should these have `, False` added to the end of the parameters list if the source data is always `True` or `False`? The code you have hear indicates it's unknown whether a location *does not* have a drive through--we'd only ever know if it *does* have a drive through.

locations/storefinders/nomnom.py
```diff
@@ -0,0 +1,92 @@
...
+                calendars = []
+            for calendar in calendars:
+                if calendar["type"] == "business":
+                    item["opening_hours"] = self.parse_opening_hours(calendar)
+                elif key := self.CALENDAR_KEYS.get(calendar["type"]):
+                    item["extras"][key] = self.parse_opening_hours(calendar).as_opening_hours()
+
+            yield from self.post_process_item(item, response, location)
+
+    def parse(self, response: TextResponse) -> Iterable[Feature]:
```
davidhicks, line 87: Despite the name `TextResponse`, [documentation](https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.TextResponse) indicates this class is only used for binary data. Should it be `Response` instead?

Also combine `parse_locations` into `parse`, unless there is some future need to have another function?

impiaaa: > Despite the name `TextResponse`, [documentation](https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.TextResponse) indicates this class is only used for binary data. Should it be `Response` instead?

The documentation says (emphasis mine):

> [TextResponse](https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.TextResponse) objects *adds* encoding capabilities to the base [Response](https://docs.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Response) class, *which* is meant to be used only for binary data

It's worded confusingly but to me it does say that `Response` is for binary data and `TextResponse` is for text data. I think the reason I used `TextResponse` because I used a type checker to check my work and it pointed out that `.json` is only available on the latter.

> Also combine `parse_locations` into `parse`, unless there is some future need to have another function?

It's used by the Panda Express spider which (maybe?) needs to make a different request first to get a cookie, then call the main `parse_locations` with a callback later. Though I could make the initial request use a different callback instead. Also I'm not sure it really needs the cookie request since it worked once without it and seemed to fail in the same way with it.

locations/storefinders/nomnom.py
impiaaa: It seems to me like this software solution was built primarily as an online-ordering platform, with the store finder only built because you need to be able to select a store to order from. So I think these attributes are used to tell whether that mode is available for online ordering. So, if `supportsdrivethru` is `True`, for example, that means that you can order online to pick up at the drive-through. They might still have a drive through that doesn't allow online ordering.

locations/storefinders/nomnom.py
davidhicks: Ah gotcha, I guess Scrapy chooses the class based on the `Content-Type` of the response returned by the server.

If you were to use `JsonRequest` instead of `Request` within `start_requests`, do you still get a response and if yes, is it still `TextResponse` instead of just `Response`?

(note this comment is not a review comment--I'm just curious how common `TextResponse` is)

davidhicks: I can't run this spider myself (geoblocking) so am unable to validate my uncertainty of the `TextResponse` type hint.

I think the safest option generally is to pick `Response` base class across most Scrapy spiders because Scrapy dynamically picks the more specific response class (`TextResponse`, `JsonResponse`, etc) based on the response received from the server. See https://github.com/scrapy/scrapy/blob/9f99da8f865efff11f6c8736567b8fbd4413091c/scrapy/responsetypes.py#L22   Perhaps a server may use the wrong MIME type (such as `text/html` instead of `application/json`) or MIME type checking may confuse `application/javascript` and `application/json`.

Otherwise because we're working with a more fixed storefinder where MIME detection is probably unlikely to randomly change or be confused, seemingly `TextResponse` should instead be `JsonResponse` as the type hint?

locations/storefinders/nomnom.py
```diff
@@ -0,0 +1,89 @@
...
+        for row in calendar["ranges"]:
+            if row["weekday"] not in DAYS_3_LETTERS:
+                continue
+            day = DAYS[DAYS_3_LETTERS.index(row["weekday"])]
+            oh.add_range(day, row["start"], row["end"], time_format="%Y%m%d %H:%M")
+        return oh
+
+    CALENDAR_KEYS = {
+        "drivethru": "opening_hours:drive_through",
+        "dispatch": "delivery",
```
Cj-Malone, line 60: I'm not fond of this, [`delivery`](https://taginfo.openstreetmap.org/keys/delivery#values) is 99% classic OSM boolean. There is actually a conditional with more usage than traditional OSM opening hours.

[`opening_hours:delivery`](https://taginfo.openstreetmap.org/keys/opening_hours%3Adelivery#overview) has basically no usage, but I'd prefer it.

impiaaa: Ok, I didn't look at actual usage, just documentation: https://wiki.openstreetmap.org/wiki/Key:delivery

Cj-Malone: I'm really surprised by that.. idk

PR #12769, opened 2025-04-01
[Chore] Use shared apply_category

Cj-Malone:
This is far from done, but an improvement. This gives us better consistency over things like [amenity=ice_cream](https://taginfo.openstreetmap.org/tags/amenity=ice_cream) vs [shop=ice_cream](https://taginfo.openstreetmap.org/tags/shop=ice_cream).


davidhicks:
As an aside, another chore on my TODO list has been implementation of a `Sports` `Enum`, now needed for disambiguating two types of sports pitches both operated by the same council/organisation. I'm happy if you get to it before I do, or I'll use this change to remind myself to implement it soon when I have time.

locations/categories.py
```diff
@@ -796,6 +803,62 @@ def apply_clothes(clothes: [Clothes], item: Feature):
...
+    if item["extras"].get("vending"):
+        current = item["extras"]["vending"].split(";")
+    else:
+        current = []
+
+    for v in vending if isinstance(vending, list) else [vending]:
+        if v.value not in current:
+            current.append(v.value)
+
+    item["extras"]["vending"] = ";".join(current)
```
iandees, line 831: It would be nice to factor out the "check extras and apply enum" code so that it can be used for other types of categories as well.

PR #12783, opened 2025-04-03
[starbucks_th] fix by enabling proxy

panasonic-mafia:


panasonic-mafia:
Works on my locals with proxy, so it should be fine.

PR #12800, opened 2025-04-04
[ecobulpack_bg] Add spider

Dimitar5555:
Based on https://gist.github.com/jleedev/7fbeb9f497c7a0c04424b8c3cf48e8f9

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -2,4 +2,4 @@
 line_length = 120
 multi_line_output = 3
 include_trailing_comma = True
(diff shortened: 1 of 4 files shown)


Dimitar5555:
> [locations/spiders/the_works.py](https://github.com/alltheplaces/alltheplaces/blob//locations/spiders/the_works.py)

This is not my spider. I haven't even touched this file...

iandees:
> > [locations/spiders/the_works.py](https://github.com/alltheplaces/alltheplaces/blob//locations/spiders/the_works.py)
> 
> This is not my spider. I haven't even touched this file...

There is code that runs these spiders when you touch the `pyproject.yaml` or `uv.lock` files. It looks like I need to adapt it to handle the case where a spider is also touched.

PR #12842, opened 2025-04-10
uv lockfile: add missing pdfplumber dependency

davidhicks:
e5b8ed8a69b732d13a147138ecc36df7dc485e36 added pdfplumber as a dependency to ATP. The uv lockfile was forgotten to be updated.


panasonic-mafia (approved):
Thanks

PR #12847, opened 2025-04-10
Explicitly open file with utf-8 in check_spider_naming_consistency

panasonic-mafia:


iandees:
Was this causing a problem somewhere?

panasonic-mafia:
Windows complains when using pre-commit hooks as it tries to open files with default 1251 encoding.

PR #12858, opened 2025-04-13
Increase max-zoom during run_all_spiders again, again

iandees:
Continuation from #12833

The gravestones are causing individual tiles to get very busy and fail the build.


davidhicks:
It is possible to quickly test by generating tiles only for one of `greater_metropolitan_cemeteries_trust_graves_au`, `niagara_falls_city_council_graves_ca` and `army_cemeteries_explorer_us`? All will be roughly equally dense, and the densest features ATP currently has to extract.

iandees:
Yea, I should have adjusted the build to ignore these failures and proceed without generating the tiles, but I already started another pass.

Unfortunately each of these runs costs ~$25-30 in Zyte calls, so we'll probably miss out on the ones that require proxy.

PR #12920, opened 2025-04-21
Batteries Plus Bulbs: Update API key

impiaaa:


davidhicks:
Replaced with #12952 by accident/duplication of efforts. #12952 does away with `api_endpoint` which is currently geoblocked and not necessary because it seemingly just proxies requests to the hosted Where2GetIt SaaS API endpoint.

This older PR also tries to turn the `bho` field from a string to JSON format, but #12952 (maybe due to Where2GetIt hosted SaaS API) finds that the `bho` field is already in JSON format and shouldn't be converted.

Happy to reopen if there are other changes I missed that should be in #12952, or added in another commit after changes proposed by #12952.

PR #12921, opened 2025-04-21
[Robinsons Brewery GB] add spider

TheUKHighStreet:


TheUKHighStreet:
250 seems like a very round number - I'm not sure whether this is really the full list or not.

TheUKHighStreet:
It does state on their website that there are 250 pubs and hotels.

PR #12934, opened 2025-04-22
Frankston City Council (AU) spiders: change from Opendatasoft to Pozi

davidhicks:
It appears Opendatasoft has been replaced with Pozi (https://pozi.com/showcase/). The old Opendatasoft website is now password gated. Most of the same information is available through the Pozi embedded maps on the council website, with the main map URL being https://frankston.pozi.com/

### DIFF
--- a/locations/spiders/infrastructure/frankston_city_council_disabled_parking_spaces_au.py
+++ b/locations/spiders/infrastructure/frankston_city_council_disabled_parking_spaces_au.py
@@ -20,5 +20,5 @@ class FrankstonCityCouncilDisabledParkingSpacesAUSpider(FlatGeobufSpider):
 
     def post_process_item(self, item: Feature, response: Response, feature: dict) -> Iterable[Feature]:
         apply_category(Categories.PARKING_SPACE, item)
-        apply_category({"parking_space": "disabled"}, item)
+        item["extras"]["parking_space"] = "disabled"
         yield item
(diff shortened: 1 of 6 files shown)


locations/spiders/infrastructure/frankston_city_council_disabled_parking_spaces_au.py
```diff
@@ -4,18 +4,16 @@
...
+    item_attributes = {"operator": "Frankston City Council", "operator_wikidata": "Q132472668", "state": "VIC", "nsi_id": "N/A"}
+    allowed_domains = ["connect.pozi.com"]
+    start_urls = ["https://connect.pozi.com/userdata/frankston-publisher/Community/Accessible_Carpark.fgb"]
 
     def post_process_item(self, item: Feature, response: Response, feature: dict) -> Iterable[Feature]:
-        item["ref"] = str(feature["asset_id"])
-        item["state"] = "VIC"
         apply_category(Categories.PARKING_SPACE, item)
-        item["extras"]["parking_space"] = "disabled"
+        apply_category({"parking_space": "disabled"}, item)
```
Cj-Malone, line 23: These aren't categories.

davidhicks: > These aren't categories.

I've switched back to `item["extras"]["..."] = ""` syntax instead to avoid overloading use of `apply_category`. There will be a fair few other spiders already overloading use of the `apply_category` function.

Perhaps in the future we should implement an `append_tag` function that `apply_category` uses internally, but can also be used for more obscure tags that don't have widespread use yet. Then the `apply_category` function can have type hints changed to only accept `Categories` Enum members. The new `append_tag` function would handle appending multiple values together into the same tag (semicolon delimited) and deduplicating values, as `apply_category` does today.

PR #12958, opened 2025-04-23
[Telkom ZA] Fix category

karen-kerr:


iandees:
Are these categories not correct because the Name Suggestion Index is not correct?

PR #12970, opened 2025-04-24
[Coventry Building Society GB] Add spider

TheUKHighStreet:


TheUKHighStreet:
I've added an NSI push for Coventry Building Society so the category should be fixed.

Cj-Malone:
NSI will add it to bank and atm, we will need to define a category in spider.

TheUKHighStreet:
> NSI will add it to bank and atm, we will need to define a category in spider.

Thanks - added.

PR #12984, opened 2025-04-27
[Roman Originals GB] Add spider

TheUKHighStreet:


TheUKHighStreet:
The Hornsea store is missing a minus sign on the longitude - I have sent them an email to let them know.

PR #13047, opened 2025-05-02
move libraries out of infrastructure and weightbridges into infrastructure

matkoniecz:


matkoniecz:
It is hard for me to give generally applicable situation but I would consider rubbish bins, manholes, street lamps, monitoring stations, etc as infrastructure.

No idea about trees.

PR #13051, opened 2025-05-03
[sopharmacy_bg] Try to bypass Cloudflare

Dimitar5555:


locations/spiders/sopharmacy_bg.py
```diff
@@ -3,6 +3,8 @@
 from locations.hours import DAYS_BG, OpeningHours
 from locations.items import Feature
 
+requires_proxy = "US"  # Cloudflare bot protection used
```
Cj-Malone, line 6: None of these were used because `requires_proxy` is needed to be a property on the spider. (Move below line 13)

Dimitar5555: :man_facepalming: Thanks! I've completely overlooked that part.

Dimitar5555:
Should be ready for merging, the data looks good.

PR #13075, opened 2025-05-07
Add spiders for Associated Supermarket Group brands

riordan:
Introduce spiders for Associated Supermarket, Compare Foods, Met Foodmarket, and Pioneer Supermarket, along with a base class to streamline the extraction of store locations from their websites. Code formatting has also been improved.


Cj-Malone:
These brands aren't in NSI, should they be?

Cj-Malone:
> These brands aren't in NSI, should they be?

Probably not, they are really low counts.

PR #13108, opened 2025-05-12
[Nandos GB IE] Removed closed restaurants

TheUKHighStreet:


TheUKHighStreet:
I don't understand why it is parsing "https://www.nandos.co.uk/restaurants"
This seems to be why there is one store without coordinates.

PR #13130, opened 2025-05-14
add Rural King store

riordan:


riordan:
hmm how might I get around this timeout? Worked on my earlier version with hacks on state capitals but that's not really representative

locations/spiders/rural_king_us.py
```diff
@@ -0,0 +1,316 @@
...
+        # Check if a specific URL was passed for testing
+        if hasattr(self, "url") and self.url:
+            yield scrapy.Request(self.url, callback=self.parse_api)
+        else:
+            # Use searchable points to query for stores
+            # This approach allows for complete coverage across the US
+            # and accommodates future expansion to new regions
+
+            with open("./locations/searchable_points/us_centroids_50mile_radius.csv") as points_file:
+                next(points_file)  # Skip the header
```
iandees, line 166: Instead of iterating over the CSV yourself, you can use [`point_locations`](https://github.com/alltheplaces/alltheplaces/blob/master/locations/geo.py#L41C5-L41C30) or [`country_seadgg_centroids`](https://github.com/alltheplaces/alltheplaces/blob/master/locations/geo.py#L41C5-L41C30)

locations/spiders/rural_king_us.py
```diff
@@ -0,0 +1,316 @@
...
+
+            # Store the raw services list for reference
+            extras["store_services"] = services
+
+            item["extras"] = extras
+
+            return item
+        except Exception as e:
+            self.logger.error(f"Error creating store item: {str(e)}")
+            return None
```
iandees, line 296: Generally I prefer to not wrap parsing functions in try/catch and let any errors bubble up to cause the spider to fail.

Cj-Malone:
Assuming 143 is the right number: https://www.ruralking.com/wcs/resources/store/10151/storelocator/latitude/0/longitude/0?maxItems=1000&radius=2500&siteLevelStoreSearch=false

PR #13173, opened 2025-05-19
add tom wahls

riordan:
more rochester chainlets


Cj-Malone (approved):
This is a lot of code, for so few POIs. Personally, I'd just map them in OSM.

PR #13195, opened 2025-05-21
[Avis US] Fix category

Cj-Malone:


Cj-Malone:
Is "dupe filter" meant to be duplicated requests? That's kinda how CrawlSpider works, scrapy skips them. Or is it meant to be duplicated POIs? The pipeline that checks for matching refs?

Cj-Malone:
Because every [page](https://www.avis.com/en/locations/us/tx/abilene/abi) links to the [city page](https://www.avis.com/en/locations/us/tx/abilene), [state page](https://www.avis.com/en/locations/us/tx) and [country page](https://www.avis.com/en/locations/us) the CrawlSpider LinkExtractors add requests to the queue, scrapy drops these requests. This is a framework thing, not a ATP thing.

This is not doing anything with the refs, I think it was meant to, but the wrong stat was used.

iandees:
Yep, looks like I used the wrong stat: https://github.com/scrapy/scrapy/blob/master/scrapy/dupefilters.py#L155

panasonic-mafia (approved):

PR #13196, opened 2025-05-21
[Core] Add stat for count of duplicates

Cj-Malone:
Rel: https://github.com/alltheplaces/alltheplaces/pull/13195#issuecomment-2897793979

### DIFF
--- a/locations/pipelines/duplicates.py
+++ b/locations/pipelines/duplicates.py
@@ -1,6 +1,9 @@
 import logging
 
+from scrapy import Spider
 from scrapy.exceptions import DropItem
+
+from locations.items import Feature
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


davidhicks:
I agree that it should be counting duplicate features extracted and not duplicate requests. Duplicate requests are in the "who cares" category for me--they never get executed/sent and therefore don't impact the performance of an ATP crawl.

davidhicks:
It's worth noting that for numerous storefinders and spiders using a radius search method, there is often a lot of overlap and duplicate features dropped. Especially for APIs that return X nearest features to a pin, and we need to search for a small radius to get all densely packed locations in cities. In rural areas we'll get close to 100% duplicate/drop rate.

There is an alternative stats scheme used for such spiders:
* `atp/geo_search/hits`
* `atp/geo_search/misses`
* `atp/geo_search/max_features_returned`

It should be (and is currently) an error if `atp/geo_search/max_features_returned` is the same as the API limit for X nearest features returned.

Then perhaps if `atp/geo_search/hits` and `atp/geo_search/misses` are present, an alternative warning should be created if the hits to misses ratio is unusually low? Then the usual duplicate check warning can be suppressed for such spiders?

I don't think this should hold up merging of this change. It's more a future improvement to CI warnings to prevent them going off (and needing to be manually checked which is more time consuming) for geographic radius search spiders.

iandees:
Yep, that check was meant to check duplicates filtered out by the `ref` check. Thanks for writing this.

iandees:
@davidhicks I agree that it would be helpful to have more information about geo/gridded searches and how much overlap is happening as a result. That's one of the primary reasons for this check.

PR #13271, opened 2025-05-28
Add LCBO (shop/alcohol) (CA) (688 stores)

### LINKED ISSUE #7005: LCBO (CA) (shop/alcohol)
### Brand name

LCBO

### Wikidata ID

Q845263

### Store finder url(s)

https://www.lcbo.com/en/stores/

davidhicks:
Fixes #7005

### DIFF
--- a/locations/storefinders/amasty_store_locator.py
+++ b/locations/storefinders/amasty_store_locator.py
@@ -60,7 +60,6 @@ class AmastyStoreLocatorSpider(Spider):
 
         if self.pagination_mode and not self._crawl_completed:
             # Continue the crawl by requesting the next page of features.
-            print(response.url)
             if m := re.search(r"\?p=(\d+)$", response.url):
                 next_page_number = int(m.group(1)) + 1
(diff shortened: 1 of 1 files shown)


davidhicks:
I did go straight for `AmastyStoreLocator` as the first attempt but this storefinder is not currently setup to handle pagination of results. I recall seeing pagination with `AmastyStoreLocator` when first playing with it and thought I recalled  observing random page numbers such as 2384678 and 128371 would return different random results. I've just tried again and it seems the logic, at least for the LCBO Amasty Store Finder implementation is:

1. Parse and remember results from https://www.lcbo.com/en/amlocator/index/ajax/?p=1
2. Request the next page by incrementing the page number by 1
3. Parse results for the requested page
4. Compare new results against those from step (1). If they match, stop crawling as the limit has been reached.
5. Repeat steps (2)-(4) until (4) triggers a stop to the crawl.

We'd need to check this logic matches other `AmastyStoreLocator` spiders that have a maximum number of results returned per page. If it works out everywhere, `AmastyStoreLocator` could be fairly easily updated with a new attribute that toggles pagination mode for websites that need it, such as LCBO.

PR #13389, opened 2025-06-10
[David Jones AU NZ] Add spider

davidhicks:


davidhicks:
Thanks @Cj-Malone for the sitemap improvement. I missed this David Jones sitemap existing as I'd just finished writing myer_au which doesn't have a sitemap available.

PR #13414, opened 2025-06-13
[Texaco GB] Add spider

TheUKHighStreet:
Not sure whether this should be called valero_gb or texaco_gb


TheUKHighStreet:
Do I need to edit something in the Valero spider to be able to do this?

PR #13415, opened 2025-06-14
Update Tesla to parse JSON more reliably

iandees:


Cj-Malone:
The site uses http2, which in theory a proxy should be using (But I'm not sure it actually is), but even with http2 there is some bot thing happening somewhere. It's not javascript, but without the cookie, they are hiding. IDK why Tesla wants to hide.

Issue #13488, opened 2025-06-21
Points are clustered even at the highest zoom level


vfosnar:
Hi, I'm encountering an issue where points, maybe duplicates, maybe just nearby pois, are clustered even when zoomed all in. This makes it that I can't inspect neither of the entries displayed on the map. Example: https://alltheplaces.xyz/map.html#16.67/50.055241/15.762519

iandees:
Looks like the pmtiles output from tippecanoe is clustering too much. I adjusted the tippecanoe command so that it doesn't do that. We'll see what it looks like at the next run.

iandees:
Ok, this should be resolved now. You should be able to see all points at the highest vector tile zoom level (15) and beyond.

PR #13489, opened 2025-06-21
Set cluster-maxzoom in tippecanoe

iandees:
This leaves all data at maxzoom so that we can view clusters, even when points are exactly overlapping and would otherwise get clustered together.

Part of #13488


iandees:
This will probably cause the build to fail because tippecanoe can't fit everything in a zoom 14 tile, but it's a start.

PR #13536, opened 2025-06-24
[english_heritage_gb][national_trust_gb] switch brand to operator

panasonic-mafia:
It feels not right to have brands on tourist attractions.


davidhicks:
Should it be `operator` instead of `brand`, or does National Trust not operate the features, rather, just supports their operation in some way (financial support or whatever)?

PR #13571, opened 2025-06-25
[The Whisky Shop GB] Add Spider

TheUKHighStreet:


locations/spiders/the_whisky_shop_gb.py
```diff
@@ -0,0 +1,30 @@
...
+import re
+
+from locations.categories import Categories, apply_category
+from locations.dict_parser import DictParser
+from locations.hours import DAYS_FULL, OpeningHours
+from locations.json_blob_spider import JSONBlobSpider
+from locations.pipelines.address_clean_up import merge_address_lines
+
+
+class TheWhiskyShopGBSpider(JSONBlobSpider):
```
panasonic-mafia, line 11: Syntax of JSONBlobSpider is not utilised, so it's probably easier to just use Spider:

```suggestion
import json
import re

from locations.categories import Categories, apply_category
from locations.dict_parser import DictParser
from locations.hours import DAYS_FULL, OpeningHours
from scrapy import Spider
from locations.pipelines.address_clean_up import merge_address_lines


class TheWhiskyShopGBSpider(Spider):
```

PR #13577, opened 2025-06-26
[Euro ARGO ERIC] Add spider for ocean monitoring floats (~4k floats)

davidhicks:


iandees:
The Soul Buoy that originally marked Null Island (mentioned in the [Null Island Wikipedia article](https://en.wikipedia.org/wiki/Null_Island#Soul_buoy)) is no longer there, so there shouldn't be any buoys at 0,0 anymore.

iandees:
Oh, the change you made also allows 0 lat _or_ lon. Got it.

PR #13592, opened 2025-06-27
[OceanOPS Global Drifter Program] Add spider for ocean drifters

davidhicks:


iandees:
Are these drifting in the sense that they are designed to constantly move? If so, I don't think that's useful for All the Places.

davidhicks:
> Are these drifting in the sense that they are designed to constantly move? If so, I don't think that's useful for All the Places.

Yes, same as #13577 (floaters) they move around the ocean but have quite a few years of lifespan. Drifters will report their position more regularly as they're on the surface, whereas floaters sink and rise again a week or however long later to make a report, then repeat their cycle.

My thinking with these moving buoys/floaters is the value to ATP consumers comes from:
- Easily allow drawing of a map of all floater/drifter locations on a single day (for inclusion in e.g. Wikimedia Commons/Wikipedias)
- Using historical ATP run data to create an approximate time lapse image/video of floater/drifter locations (for inclusion in e.g. Wikimedia Commons/Wikipedias)
- An event occurs (e.g. underwater volcano) and someone immediately wants to see what sensors were nearby in the ocean that may have detected the phenomena. They could do this today by visiting a dashboard such as ocean-ops.org, but would have essentially have to revert to writing a spider such  as these ATP spiders to extract positions into QGIS or some other tool for other analysis.

Note #13573 is for moored buoys that are anchored and don't move around the ocean.

iandees:
Closing this — moving buoys don't fit well as location data for AllThePlaces since they're not fixed locations.

PR #13593, opened 2025-06-27
[Schuh GB] Add Spider

TheUKHighStreet:
Not sure whether this is possible - it is blocking with 403 errors.


TheUKHighStreet:
The latitude and longitude don't seem to be available.

PR #13601, opened 2025-06-29
[BHF GB] Add missing stores

TheUKHighStreet:


TheUKHighStreet:
This is missing a lot of shops that don't have structured data on the store page.
Example: view-source:https://www.bhf.org.uk/what-we-do/find-bhf-near-you/abergavenny-bhf-shop

TheUKHighStreet:
The dupe filter is removing 100% dupes because the store pages which do have structured data have it twice.

PR #13607, opened 2025-06-30
[OMV] Removed opening hours parsing causing errors bad data

karen-kerr:
In data, each week had 8 days.

### DIFF
--- a/locations/spiders/omv.py
+++ b/locations/spiders/omv.py
@@ -5,6 +5,7 @@ from scrapy.http import Response
 
 from locations.categories import Categories, Extras, Fuel, FuelCards, PaymentMethods, apply_category, apply_yes_no
 from locations.dict_parser import DictParser
+from locations.hours import DAYS, OpeningHours
 
 # List of brands and countries where they operate:
 # https://www.omv.com/en/customers/services/filling-stations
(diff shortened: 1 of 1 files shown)


Cj-Malone (approved):
I restored the function, no valid justification for it. Just feel optimistic that it could be re-enabled at some point in the future.

Cj-Malone:
There is about 35 with a value other than, "8/TRUE/00:00/00:00" eg `ref=AT.1661.8`. In this case, it's the same as Su. Public Holidays sounds good, but I'm not sure. I'll leave it open for a while if you wanna take a look David, but otherwise I'll merge it and it can be revisited in the future.

davidhicks:
I read the client side JavaScript code that handles the opening hours string and found the "day 8" entry is never used.

From https://app.wigeogis.com/kunden/omv/webcomponent/js/app.js :

```
        var index = currentDayOfWeek - 1;
        if (currentDayOfWeek === 0) {
            index = 6;
        }

       var todaysOpening = realDays[index];
```

(where "realDays" is setup to be 8 elements long, but is never read beyond the 7th element)

I think this spider could also just ignore the 8th item in the list with:
```
for rule_str in opening_hours.split("#")[0:6]:
```

I'll send a PR for this branch

karen-kerr:
Thanks both. I'll close this PR.

davidhicks:
Good catch on the range missing Sunday @Cj-Malone . Thanks

PR #13631, opened 2025-07-03
[Pizzeria 105 PL] Ref from url regex group

starsep:
Note it also changes `url` to `response.url` in `StructuredDataSpider`.
For pizzeria_105_pl `url` is https://105.pl for every place, `response.url` changes e. g. https://105.pl/pizzeria-warszawa-wlochy/

I haven't checked impact on other spiders.
Perhaps configuration only for this spider should be modified.

`'extras': {'@source_uri': 'https://105.pl/pizzeria-warszawa-stegny/'},` contains full url so `ref` could be calculated from it.
I am not sure whether the same `url` is a bug or a feature :smile:


Cj-Malone:
Yeah, I'm not comfortable changing this. It could blow up other spiders. This issue is Pizzeria 105 having wrong canonical urls, we'll fix it in the spider.

PR #13635, opened 2025-07-04
[Super Star Car Wash US] Add spider

Cj-Malone:
Rel: https://github.com/osmlab/name-suggestion-index/pull/11119


panasonic-mafia (approved):
Thank you

PR #13674, opened 2025-07-09
[Brico] Add spider

karen-kerr:


locations/spiders/brico.py
```diff
@@ -0,0 +1,26 @@
+import scrapy
+
+from locations.dict_parser import DictParser
+
+
+class BricoSpider(scrapy.Spider):
+    name = "brico"
```
davidhicks, line 7: Could this be `brico_be` to differentiate from existing spiders `brico_io_it` and `brico_ok_it`?

davidhicks (changes requested):

locations/spiders/brico.py
karen-kerr: Thanks, 
 'atp/country/BE': 146,
 'atp/country/LU': 2,

davidhicks (approved):
thanks!

PR #13694, opened 2025-07-11
[Localcoin] Add spider for cryptocurrency ATMs in CA, AU, NZ, HK

davidhicks:


locations/pipelines/clean_strings.py
```diff
@@ -6,7 +6,7 @@
 
 
 def clean_string(val: str) -> str:
-    return html.unescape(val).strip()
+    return html.unescape(val).strip().replace("\u200b", "")
```
davidhicks, line 9: Just for reference, this is https://en.wikipedia.org/wiki/Zero-width_space which is a non-printing character used to indicate a word boundary. It's meant to be used in text as a hint for where a word can split across lines.

PR #13703, opened 2025-07-12
[Circle K Ireland] Fix spider

TheUKHighStreet:


Cj-Malone:
This spider was just fixed so I don't think this is necessary

PR #13727, opened 2025-07-15
[Albert Heijn] Split spider

Cj-Malone:
The spider is [wobbly](https://alltheplaces.xyz/spiders.html?search=albert_heijn), recently it's not fetched both countries, the time it "worked" it only got 1 country. Albert Heijn is keeping the connection open and the spider is never timing out, this is bad enough with 1 spider, it takes up a slot in the weekly for the full 8 hours, with 2 spiders the potential is even worse, 2x8 hour blocks wasted.

However with the 2 spiders, I've not hit the issue locally. Hopefully this works for us.

### DIFF
--- a/locations/spiders/albert_heijn_be.py
+++ b/locations/spiders/albert_heijn_be.py
@@ -1,0 +1,7 @@
+from locations.spiders.albert_heijn_nl import AlbertHeijnNLSpider
+
+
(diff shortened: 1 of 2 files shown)


panasonic-mafia (approved):
LGTM, thanks

PR #13754, opened 2025-07-19
Using removeprefix() instead of lstrip(). 

arfatkh:
Changed `lstrip()` to `removeprefix()`.

`lstrip()`'s argument is a string specifying the set of characters to be removed **not a prefix.**
`removeprefix()`is a method that will remove a single prefix string rather than all of a set of characters. 

For example in our case:

```
site = "www.wikipedia.com"

site.lstrip("www.wikipedia.com")
# Outputs  >>  ikipedia.com

site.removeprefix("www.wikipedia.com")
# Outputs  >>  wikipedia.com

```

For Reference:  https://docs.python.org/3/library/stdtypes.html#str.lstrip

### DIFF
--- a/locations/name_suggestion_index.py
+++ b/locations/name_suggestion_index.py
@@ -66,7 +66,7 @@ class NSI(metaclass=Singleton):
         for wikidata_code, org_parameters in self.wikidata_json.items():
             for official_website in org_parameters.get("officialWebsites", []):
                 official_website_domain = urlparse(official_website).netloc
-                if official_website_domain.lstrip("www.") == supplied_url_domain.lstrip("www."):
+                if official_website_domain.removeprefix("www.") == supplied_url_domain.removeprefix("www."):
                     return wikidata_code
         # Last attempt to find a fuzzy match for registered domain (excluding subdomains)
         for wikidata_code, org_parameters in self.wikidata_json.items():


Cj-Malone (approved):
Thank you, I'm not sure who uses that code. I think David added it a while ago

davidhicks:
> Thank you, I'm not sure who uses that code. I think David added it a while ago

It was added in https://github.com/alltheplaces/alltheplaces/commit/1be8ef7d6b and seems to only be used in `commands/genspider.py` which is intended to make writing new ATP spiders quicker by using templates to draft pre-filled spiders.

It was seemingly a bit temporary pending merge of https://github.com/alltheplaces/alltheplaces/pull/6526 (still open but not updated in a while) and a few other changes (since closed by the author) which attempt to automate detection/drafting of common spiders. For example, give a storefinder/location map URL and the Wikidata/NSI information would be detected and pre-filled, and a draft spider returned to the user to customise and submit.

Whilst all this work hasn't really been touched in the better part of 2 years, it wouldn't be an enormous amount of work to get it merged. My view though is perhaps it wouldn't get much use if merged because:

1. ATP already implements many/most spiders that would be detected by this code.
2. It's not that hard to manually detect and draft a spider for a common pattern/storefinder so the amount of effort saved is not particularly high.

PR #13763, opened 2025-07-22
[Avia FR] Fix non running spider

karen-kerr:


panasonic-mafia:
Seems file on https://www.avia-france.fr/wp-admin/admin-ajax.php?action=get_avia_csv got new columns, may I kindly ask to re-map fuels when you have time please?

PR #13808, opened 2025-07-28
[YextAnswers] Fix opening_hours parsing

impiaaa:
Commit aed3b661c3f97000133f35e9f921c51690e32235 caused `TypeError: Object of type OpeningHours is not JSON serializable` errors.

Should be ok to call `parse_opening_hours().as_opening_hours()` in these cases because `parse_opening_hours` only returns `None` if the input is `None`, and that's already checked for.


davidhicks:
Thanks, my bad for not checking `item["extras"]` keys are type checked / converted before conversion into the output format as applies to fields such as `item["opening_hours"]`.

PR #13821, opened 2025-07-29
[Costco AU] Add spider

davidhicks:


davidhicks:
Oops, had my spiders sorted the wrong way and I missed `costco_global` that already includes AU locations.

PR #13918, opened 2025-08-05
[YHA AU GB] Fix brand names and Wikidata IDs

davidhicks:
NSI/Wikidata both split "YHA" into country-specific brands/items. Use the correct names and Wikidata IDs per NSI.


davidhicks:
`yha_gb` is blocked by Cloudflare so can't fix it at the moment

PR #13961, opened 2025-08-07
[BCI AQ CL] Add spider

karen-kerr:


iandees:
Yea, AQ is the country code for Antarctica. We could add it to the list of allowed codes, but this is a fine solution too.

PR #13974, opened 2025-08-09
Matsukiyo spider draft

j9d3it:
I'm making a new spider, at least I got the location info working. Now just to get brand Wikidata info, opening hours, payment methods, etc., and the correct store URL for each store. I mostly want to use this draft to see how it looks on a map and the tags for each store.

### DIFF
--- a/locations/spiders/matsukiyo.py
+++ b/locations/spiders/matsukiyo.py
@@ -1,0 +1,18 @@
+from datetime import datetime
+from typing import Any
+
+from scrapy import Spider
+from scrapy.http import Response
+
+from locations.dict_parser import DictParser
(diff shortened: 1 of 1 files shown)


j9d3it:
Why is it timing out? Works fine on my computer...

Cj-Malone:
What is this? Matsukiyo owns these brands? They should probably be in NSI.

j9d3it:
I will make a pull request for NSI, yes they own all these brands. Some of them don't have Wikidata IDs too so they'll need to be added.

j9d3it:
@Cj-Malone I figured out how to add a new branch

PR #13976, opened 2025-08-10
[QUICPay JP] add spider

j9d3it:
lots of categories that have multiple shop types. first, testing with only a couple pages to see if it works.


Cj-Malone:
What is the license for this data? I'm not comfortable scraping random aggregators

PR #14034, opened 2025-08-17
[Allergan Beauty clinics JP] add spider

j9d3it:


locations/spiders/allerganbeauty_jp.py
```diff
@@ -0,0 +1,75 @@
...
+    def parse(self, response: Response, **kwargs: Any) -> Any:
+        for store in response.json()["shop_list"]:
+
+            item = DictParser.parse(store)
+            item["ref"] = store["key"]
+            try:
+                item["website"] = store["施設URL"]
+            except:
+                pass
+            try:
```
iandees, line 48: Instead of using a `try/except` to catch the `KeyError`, use the `get()` function:


```suggestion
            item["website"] = store.get("施設URL")
```

iandees (changes requested):

locations/spiders/allerganbeauty_jp.py
j9d3it: > The try/except are still in there.

The store.get function didn't work for some of the stores, since it would fail if they didn't have a website.

j9d3it: > The try/except are still in there.

I have updated it, the phone number works fine but I still have to use the try except with the specialty function.

PR #14075, opened 2025-08-20
[Skylark restaurants JP] add spider

j9d3it:


locations/spiders/skylark_jp.py
```diff
@@ -0,0 +1,157 @@
...
+                case "夢庵":
+                    item.update({"brand_wikidata": "Q11253593"})
+                case "ジョナサン":
+                    item.update({"brand_wikidata": "Q11310628"})
+                case "ステーキガスト":
+                    item.update({"brand_wikidata": "Q92599119"})
+                case "むさしの森珈琲":
+                    item.update({"brand_wikidata": "Q116758676"})
+                case "から好し（単独店）":
+                    item.update({"brand_wikidata": "Q115008407"})
```
panasonic-mafia, line 39: I suggest to apply_category for these cases as well. If someone adds a toilet/camera (etc.) operators to NSI for a brand - our NSI matching will fail.

locations/spiders/skylark_jp.py
```diff
@@ -0,0 +1,157 @@
...
+                    apply_category(Categories.RESTAURANT, item)
+                    item["extras"]["cuisine"] = "yakiniku"
+                case "資さんうどん":
+                    item["brand"] = store["marker"]["ja"]["name"]
+                    item.update({"brand_wikidata": "Q11634979"})
+                    apply_category(Categories.RESTAURANT, item)
+                    item["extras"]["cuisine"] = "udon"
+                case _:
+                    item = {}
+                    yield item
```
panasonic-mafia, line 127: Could you please explain what you're trying to do here? After this `yield` code will proceed anyway and apply some fields to `{}` instead of `Feature` after `DictParser`.

j9d3it: > Could you please explain what you're trying to do here, so I could help you?
> 
> After this `yield` code will proceed anyway and apply some fields to `{}` instead of `Feature` after `DictParser`.
> 
> But at the end those records will end up as errors in the log:
> 
> ```
> Traceback (most recent call last):
>   File "/codebuild/output/src707041446/src/github.com/alltheplaces/alltheplaces/.venv/lib/python3.11/site-packages/twisted/internet/defer.py", line 1088, in _runCallbacks
>     current.result = callback(  # type: ignore[misc]
>                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/codebuild/output/src707041446/src/github.com/alltheplaces/alltheplaces/.venv/lib/python3.11/site-packages/scrapy/utils/defer.py", line 390, in f
>     return deferred_from_coro(coro_f(*coro_args, **coro_kwargs))
>                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/codebuild/output/src707041446/src/github.com/alltheplaces/alltheplaces/locations/pipelines/duplicates.py", line 19, in process_item
>     ref = (spider.name, item["ref"])
>                         ~~~~^^^^^^^
> KeyError: 'ref'
> 2025-08-20 20:10:48 [scrapy.core.scraper] ERROR: Error processing {}
> ```

There are some of the から良し restaurants that are inside Gusto so they would show up as duplicates of Gusto on the map. So any that didn't match the fields I had I wanted to delete. I'll think about a better way to do this.

panasonic-mafia:
Thank you

Issue #12690, opened 2025-03-22
lululemon Spider doesn't include European stores (and probably others)


peternewman:
### Spider name

lululemon

### Example bad data or results

Missing e.g. London, Amsterdam etc

### Expected behaviour or results

Includes all stores worldwide ideally

### Possible causes?

I think there is different data available on different websites unfortunately.

E.g. this URL gets stores within a 500 radius (it started at 15 in their default):
https://www.eu.lululemon.com/on/demandware.store/Sites-EU-Site/en_LU/Stores-FindStores?showMap=true&radius=5000&lat=50&long=0

I don't know if that URL includes the Middle East stuff here:
https://www.lululemon.me/en-ae

peternewman:
See also #670 and #1910 for the original creation/issues

iandees:
The `lululemon_ca_us` spider only crawls `shop.lululemon.com` which has CA/US sitemaps. European stores appear to be on a different domain (lululemon.com/en-gb/ etc.) which times out from here. The spider needs to be extended to cover those sitemaps, or a separate spider added. The main lululemon.com site is not accessible without a proxy. 🤖

PR #14080, opened 2025-08-20
Rename lululemon spider to lululemon_ca_us

impiaaa:
because the spider only covers those countries, and to better communicate the spider's coverage. Related to, but does not fix, #12690. Seemingly the websites for every country other than CA and US each have a Demandware storefinder that has global coverage, but is protected by Cloudflare.


panasonic-mafia:
Thanks, makes sense.

PR #14118, opened 2025-08-27
Implement Camoufox spiders with Cloudflare CAPTCHA bypass

davidhicks:
Implement a CamoufoxSpider which can be inherited by a spider to cause all requests to be made using Camoufox, a variant of Firefox designed specifically for web scraping whilst avoiding bot blocking. Also implement a Cloudflare Turnstile CAPTCHA bypass mechanism, one of the most common mechanisms in use.

Also implement a PlaywrightSpider class to generally improve the ease of developing Playwright-using spider. The legacy method of using spider attribute "is_playwright_spider" is kept in place for now, pending conversion of spiders with this attribute to inherit from PlaywrightSpider or CamouSpider.


iandees:
We do some extra prep work in CI if the `is_playwright_spider` field is detected in any of the spiders that need to run. Is there another way we can detect this situation using this setup?

davidhicks:
> We do some extra prep work in CI if the `is_playwright_spider` field is detected in any of the spiders that need to run. Is there another way we can detect this situation using this setup?

Should be fixed now. Thanks for the heads up!

The only thing I'm unsure of with in hindsight with Camoufox is the lack of development community as the developer put up a notice in March explaining an illness that has kept the developer off the project and then providing an estimate of late 2025 to get things back up and running again. Patchright (for WebKit browsers only) appears to be the next favourite and supported option at the moment, and patchright-python is a drop-in replacement for playwright-python which makes things easy. There's a scrapy-patchright fork or two of scrapy-playwright that allows patchright to be used, but the forks don't appear to be maintained in any serious way.

Even so--Camoufox in this branch currently bypasses at least Cloudflare Turnstile which is very commonly used, and also bypasses some less sophisticated anti-bot mechanisms of older anti-bot technologies. More sophisticated measures such as Kasada (very common for hotel, airline, etc websites) no longer work with Camoufox, possibly because an outdated Camoufox (not ESR edition either) is quite the giveaway.

locations/spiders/costco_global.py
```diff
@@ -7,11 +7,12 @@
 from locations.categories import Categories, Extras, apply_category, apply_yes_no
 from locations.dict_parser import DictParser
 from locations.pipelines.address_clean_up import merge_address_lines
+from locations.spiders.costco_ca_gb_us import COSTCO_SHARED_ATTRIBUTES
 
 
 class CostcoGlobalSpider(Spider):
     name = "costco_global"
-    item_attributes = {"brand_wikidata": "Q715583"}
+    item_attributes = COSTCO_SHARED_ATTRIBUTES
```
Cj-Malone, line 15: This breaks brand in places like JP.

davidhicks: Hmmm, `costco_global` will need some changes to handle native brand/name fields in JP, ZH and TW. I can submit a PR

iandees:
Yea that's totally fair. In my mind we started to step over that line when we sometimes stopped honoring robots.txt and mimicked user agents, but crossed it when we started to use the proxy. If we want to try harder to get "good bot" status we should disable proxy usage, go back to using our own user agent, and only ignore robots.txt when its broken on the site's side.

PR #14148, opened 2025-08-31
[great_wall_motor] Fix and rename spider

arrival-spring:
This now covers more countries than the previous spider, much of southern Africa, despite being through the South Africa website. However, most other countries seem to only have one dealership


davidhicks:
Maybe name it `great_wall_motor_africa` to differentiate better from other regions/countries this brand operates in?

davidhicks (approved):

PR #14162, opened 2025-09-02
[SOpharmacy BG] Fix spider (first Cloudflare Interstitial spider)

davidhicks:


iandees:
Closing — the entire sopharmacy.bg domain (including robots.txt) is now behind a Cloudflare non-interactive JS challenge (`cType: 'non-interactive'`), which Camoufox can no longer bypass. The proxy-based spider on master is also returning 0 results for the same reason. The site is currently unscrapable by automated means. Will reopen if the situation changes. 🤖

PR #14187, opened 2025-09-04
[Core] Add data bbox stats

Cj-Malone:
Possibly related to https://github.com/alltheplaces/alltheplaces.xyz/issues/67, possibly pointless. This adds the following stats which then can be used to build a bbox to show what area the data covers.

```python
{'atp/bbox/lat_max': 55.92434295,
 'atp/bbox/lat_min': 50.704424,
 'atp/bbox/lon_max': 1.383928383,
 'atp/bbox/lon_min': -4.142793656,
```

```python
    def close_spider(self, spider: Spider):
        print(
            "http://bboxfinder.com/#{},{},{},{}".format(
                spider.crawler.stats.get_value("atp/bbox/lat_min"),
                spider.crawler.stats.get_value("atp/bbox/lon_min"),
                spider.crawler.stats.get_value("atp/bbox/lat_max"),
                spider.crawler.stats.get_value("atp/bbox/lon_max"),
            )
        )
```

`http://bboxfinder.com/#50.704424,-4.142793656,55.92434295,1.383928383`

### DIFF
--- a/locations/spiders/benchmarx_gb.py
+++ b/locations/spiders/benchmarx_gb.py
@@ -8,7 +8,6 @@ from locations.hours import OpeningHours
 from locations.pipelines.address_clean_up import merge_address_lines
 
 
-
 class BenchmarxGBSpider(Spider):
     name = "benchmarx_gb"
     item_attributes = {"brand": "Benchmarx", "brand_wikidata": "Q102181127"}


davidhicks:
I like the idea, but think it'd more useful if the longitude calculations factored in https://en.wikipedia.org/wiki/180th_meridian#Software_representation_problems and returned the _smallest_ bounding box encompassing all points.

For a spider that captures locations in Fiji (which is split by the 180th meridian), I think it'd be more useful for lon_min to be (approximately) 178 for the Easternmost point of Fiji (https://www.wikidata.org/wiki/Q2268568) and lon_max to be (approximately) -176 for the Westernmost point of Fiji (https://www.wikidata.org/wiki/Q13155855).

Cj-Malone:
I have renamed them, please check I renamed them right, I got very confused while doing it. Also you're welcome to add the 180 thing, I think I understand the concept but it doesn't work on bboxfinder or in JOSM so IDK

davidhicks:
I haven't tried this, but this is hopefully the solution I was suggesting:

```
def close_spider(self, spider: Spider):
  lon_min = spider.crawler.stats.get_value("atp/bbox/lon_min")
  lon_max = spider.crawler.stats.get_value("atp/bbox/lon_max")
  if (180 + lon_min) > (180 - lon_max):
    # Example of Fiji is lon_min=-176 (near easternmost point of Fiji) and lon_max=178 (near westernmost point of Fiji)
    # Bounding box should use lon_nw=178 and lon_se=-176
    spider.crawler.stats.set_value("atp/bbox/lon_nw", lon_max)
    spider.crawler.stats.set_value("atp/bbox/lon_se", lon_min)
    spider.crawler.stats._stats.pop("atp/bbox/lon_max", None)
    spider.crawler.stats._stats.pop("atp/bbox/lon_min", None)
  else:
    # Example of somewhere in the USA of lon_min=-92 (more west) and lon_max=-90 (more east)
    # Bounding box should use lon_nw=-92 and lon_se=-90
    spider.crawler.stats.set_value("atp/bbox/lon_nw", lon_min)
    spider.crawler.stats.set_value("atp/bbox/lon_se", lon_max)
    spider.crawler.stats._stats.pop("atp/bbox/lon_max", None)
    spider.crawler.stats._stats.pop("atp/bbox/lon_min", None)
```

Then in addition, it's just renaming lat_min to "lat_se" and lat_max to "lat_nw".

Unfortunately there seems to be no nice way with Scrapy's stats API to delete a stat without getting into the private dictionary behind the exposed API interface.

davidhicks:
Actually I didn't think it through well enough.

It's not possible with just incrementing lat/lon counters to distinguish between a spider that has collected only points in Fiji (-176 < lon < 178) and a spider that has collected points everywhere else in the world except Fiji (also -176 < lon < 178). It's also necessary to look into the list of points extracted to see whether Fiji is included or excluded in the list of points.

Cj-Malone:
My idea was to have the reality of the data, not the ideals of the spider. That does mean [some](https://www.greateranglia.co.uk/travel-information/station-information/buu) will have bad areas, but anything defined in the spider will be out of date straight away.

image

iandees:
I agree with what the data should represent (the actual bbox of the captured data), I just disagree with using the stats framework to store the numbers. Instead of storing them as stats, we could add a `bbox` property to a pipeline and then add the bbox to the output data at the spider level in the same way we write out the spider name or the lineage information.

Issue #13598, opened 2025-06-28
Lego


TheUKHighStreet:
### Spider name

lego

### Example bad data or results

{"type": "Feature", "id": "0hGPGxNEUnzZ-25xdNvR4SIuzoI=", "properties": {"ref": "https://www.lego.com/stores/stores/gb/brighton", "@source_uri": "https://www.lego.com/api/graphql/StoreSearch", "shop": "toys", "@spider": "lego", "addr:street_address": "West Quay Shopping Centre, 8-10 Portland Terrace", "addr:city": "Southhampton", "addr:state": "England", "addr:postcode": "SO15 1QF", "addr:country": "GB", "branch": "Southampton", "phone": "+44 23 8268 6906", "website": "https://www.lego.com/stores/stores/gb/brighton", "opening_hours": "Mo-Fr 09:00-20:00; Sa 09:00-19:00; Su 11:00-17:00", "brand": "Lego", "brand:wikidata": "Q1063455"}, "geometry": {"type": "Point", "coordinates": [-1.4051426, 50.9041446]}},

### Expected behaviour or results

Somehow the Brighton and Southampton stores have been merged together.

TheUKHighStreet:
Bit more investigation. These are the results from the website for the two stores:

[...]

PR #14282, opened 2025-09-21
[Lego] Fix spider

TheUKHighStreet:
Try a fix for https://github.com/alltheplaces/alltheplaces/issues/13598


TheUKHighStreet:
This has added Brighton and Southampton as two separate stores.

PR #14303, opened 2025-09-26
[Bambu] Fix Spider

mohit-sharma2194:
**_Flag bambu as require proxy to fix spider_**


impiaaa:
It passed CI in #14262, I don't think this is necessary

PR #14437, opened 2025-10-21
Implement ty type checking for ATP core

davidhicks:
* Use the ty type checker for Python to automatically check types on changes to core libraries/files within ATP. ty is written in Rust and is very fast (almost instant for checking ATP's core).
* Fix type hinting for ATP core so that ty passes successfully.
* Change deprecated non-async start_requests to async start function per latest changes in Scrapy upstream for ATP storefinders.


locations/categories.py
```diff
@@ -825,21 +825,32 @@ class Clothes(Enum):
...
+        current = item["extras"]["clothes"].split(";")
+    else:
+        current = []
+
+    for v in clothes if isinstance(clothes, list) else [clothes]:
+        if v.value not in current:
+            current.append(v.value)
+            apply_yes_no(f"clothes:{v.value}", item, True)
+
+    current.sort()
```
Cj-Malone: Thanks for removing apply_category.

I'm not sure we should be doing sort here (and in similar functions) though. In OSM these should be ordered by importance to the POI. Nobody cares about `clothes`, so I'm not gonna cry, but this wouldn't fly for `cuisine`.

PR #14441, opened 2025-10-22
Replace deprecated start_requests function in 'a' spiders

davidhicks:
For all spiders with a name commencing 'a' that use the now-deprecated start_requests(...) function, replace it with Scrapy's async start function.

### DIFF
--- a/locations/spiders/adeg.py
+++ b/locations/spiders/adeg.py
@@ -1,6 +1,6 @@
 from typing import AsyncIterator
 
-from scrapy import Spider
+from scrapy import Selector, Spider
 from scrapy.http import Request
 
 from locations.hours import OpeningHours
@@ -25,7 +25,7 @@ class AdegSpider(Spider):
     async def start(self) -> AsyncIterator[Request]:
         yield self.get_page(1)
 
-    def get_page(self, n):
+    def get_page(self, n: int) -> Request:
         return Request(
             f"https://www.adeg.at/services/maerkte-oeffnungszeiten?tx_solr[page]={n}&type=7382&distance=1000",
             meta={"page": n},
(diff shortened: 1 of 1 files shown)


Cj-Malone (approved):
Not specific to this PR, (or `start` as it also applies to `start_requests` in new scrapy), but all the requests are generated before network activity. This is fine most the time but may cause issues if there are a LOT of requests (memory), or the requests use some kind of time auth (it may have expired by the time the request is actually sent).

Cj-Malone:
Do you want this merging, or are you going to add `b`s to this PR?

davidhicks:
> Not specific to this PR, (or `start` as it also applies to `start_requests` in new scrapy), but all the requests are generated before network activity. This is fine most the time but may cause issues if there are a LOT of requests (memory), or the requests use some kind of time auth (it may have expired by the time the request is actually sent).

`starbucks_eu` queues 18-19k requests in the `start` (`start_requests`) method and overall has a maximum memory consumption of ~500MB (not too large in the grand scheme of some spiders, some which consume 1-2GB maximum memory usage as they approach 1M POIs extracted).

`amazon_locker` queues ~23k requests in the `start` method and overall has a maximum memory consumption of ~600MB. This uses a postcode search method, hence the relatively high number of requests made.

The most number of requests I could find for ISEADGG geographical search spiders was <10k so they're all much smaller than `starbucks_eu` and `amazon_locker`.

Based on this research, I think we're OK for now with spiders and the number of requests they make upfront. Seemingly no more than 25k requests upfront, and those spiders seem to still have manageable maximum memory usage. The number of POIs extracted by a spider seems to be much more important than number of requests initially queued when it translates to maximum memory usage.

PR #14466, opened 2025-10-26
[run all spiders][run pr spiders] give scrapy more time for graceful shutdown

panasonic-mafia:
Run all spiders sends close signal after 8h then forcefully shutdown after 15s if spider is still open, leaving file malformed (not closed properly at the end of the file). There are ~20 spiders like that, e.g. [choice_hotels](https://alltheplaces-data.openaddresses.io/runs/2025-10-18-13-32-41/output/choice_hotels.geojson), or [edward_jones](https://alltheplaces-data.openaddresses.io/runs/2025-10-18-13-32-41/output/edward_jones.geojson).

### DIFF
--- a/ci/run_all_spiders.sh
+++ b/ci/run_all_spiders.sh
@@ -34,7 +34,7 @@ mkdir -p "${SPIDER_RUN_DIR}"
 (>&2 echo "Write out a file with scrapy commands to parallelize")
 for spider in $(uv run scrapy list -s REQUESTS_CACHE_ENABLED=False)
 do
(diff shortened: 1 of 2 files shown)


iandees:
For context: I added `timeout` because there were some spiders that would just hang and never exit. The `CLOSESPIDER_TIMEOUT` scrapy setting we use _should_ shut down the spider after some amount of time, but that didn't work. I added a timeout to send the SIGINT, but they wouldn't ever gracefully exit, so I added the `-k` option to send SIGKILL and forcefully exit a hung spider. In theory, the SIGKILL is the fallback after a couple attempts at graceful exit.

I'm happy to merge this, but we should keep an eye out for spiders that do this and find a way to fix them.

PR #14493, opened 2025-10-30
[NSI] Pin NSI v6

Cj-Malone:
NSI is in the process of migrating to v7, we can't (and shouldn’t) be reading HEAD. Once v7 is out, we should migrate to it.


panasonic-mafia:
Thank you

PR #14500, opened 2025-11-01
Add missing websites to bhf_gb.py spider output

rjw62:
Manually parsed store pages are missing the "website" property. But since each entry comes from an individual web page, we know what the website should be.


TheUKHighStreet:
Can you add the correct link for the Portsmouth store and I'll delete my pull request?

PR #14587, opened 2025-11-18
[Belgium Addresses] Add readme

Cj-Malone:
I've flagged all as CC4 (even though Brussels could be CC0) because we are downloading from BeSt, not individual datasets.


Cj-Malone:
I really want to put my GB readme here, but it may be confusing without any spiders :/

PR #14619, opened 2025-11-24
[DOCS] Add family tree

Cj-Malone:
image

### DIFF
--- a/DATA_FORMAT.md
+++ b/DATA_FORMAT.md
@@ -1,4 +1,4 @@
-# All The Places Data Format
+# All the Places Data Format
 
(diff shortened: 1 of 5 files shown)


iandees:
Makes sense to me. We should write some extra context to help frame that.

davidhicks:
In addition to the datasets ingested into ATP, perhaps Wikidata and NSI could be included to show that third party data can be enriched through the thousands of properties present in Wikidata for each brand. E.g. want to add more tags to ATP output data to include business registration numbers, official brand websites, etc--easy to do because ATP links features to Wikidata items of the brands/operators.

PR #11758, opened 2025-01-02
[kay_jewelers] - this is addr:street_address rather than addr:full

matkoniecz:
for example, this values: 3300 Broadway
1344 Lum Road
2529 Main St.
3401 Dale Rd.
400 Plaza Dr.
20 City Drive
2424 U.S. 6
415 Adams Dr.
1901 Nw Expy.
6100 O Street

PR #14642, opened 2025-11-28
[hotter_gb] - this is addr:street_address rather than addr:full

matkoniecz:
for example, this values: 465 Lord Street
71 Mostyn Street
3 Feasegate
105 High Street
11 Mill Lane
30 Bridge Street
26 Cricklade Street
113 Princes Street
16 East Street
32 High Street


matkoniecz:
btw, #11758 is quite similar

PR #14774, opened 2025-12-17
Replace deprecated start_requests function in 'o'->'r' spiders

davidhicks:
For all spiders with a name commencing '0' through 'r' that use the now-deprecated start_requests(...) function, replace it with Scrapy's async start function.


davidhicks:
> So the ty workflow runs because categories.py was edited, but I thought it was passing now?

ty was updated to a new version and it now has smarter type checking. Will need to fix up more type hints in the core of ATP to quiet the latest version of ty.

PR #14812, opened 2025-12-24
[Maes Mobility BE NL] add Maes operated fuel stations with different brands in Belgium and the Netherlands (1738 locations)

JanCallewaert-TomTom:


JanCallewaert-TomTom:
> It looks like the country code is getting set incorrectly in some cases, which causes the phone number normalizer to fail. Can you double-check that?

It seems that the country isn't always correctly returned by the website. I tried to rely on reverse geocoding, but fuel stations at the country border are not always correctly identified.

For example, the following should be the Netherlands and not Belgium:

    {
        "id": 21923,
        "title": "ESSO Dirksland",
        "lat": "51.7497451",
        "long": "4.1211094",
        "slug": "esso-dirksland",
        "street": "KORTEWEEEGJE",
        "nr": "75",
        "postalcode": "3247 BH",
        "city": "DIRKSLAND",
        "country": "BELGI\u00cb",
        "carwash": 0,
        "shop": 0,
        "phone": "3177-3999200",
        "fuels": [
            {
                "product_id": null,
                "price": null
            },
            {
                "product_id": null,
                "price": null
            },
            {
                "product_id": null,
                "price": null
            },
            {
                "product_id": null,
                "price": null
            },
            {
                "product_id": null,
                "price": null
            },
            {
                "product_id": null,
                "price": null
            }
        ],
        "payment_methods": [
            "MAES EUROPE CARD",
            "MAES PREPAID",
            "TANX",
            "MAES HYBRID CARD"
        ]
    },

What's the preferred way to handle this?

PR #14843, opened 2026-01-02
[discover.swiss] Fix spider

brawer:
The Swiss OpenStreetMap organization has an agreement with Hotellerie
Suisse (who owns and runs the discover.swiss platform) that we’re
allowed to read this feed for keeping OSM up to date.
    
The site uses URLs with continuation tokens that are longer than
Scrapy’s default limit of 2083 characters.


brawer:
OK to simply ignore the “website is invalid” error, or should I run my own validation regexp on the websites (and email addresses) taken from the upstream feed?

PR #14844, opened 2026-01-02
[OpenTransportData.swiss] Fix spider

brawer:


brawer:
Hm, is it intentional that Continuous Integration fails upon spider timeout? On my laptop, this spider completes in 143 seconds, producing a  205 MB GeoJSON file containing 198074 features. I think it’s perfectly reasonable to terminate spiders in CI after 60s, just wondering if CI should really fail in this case.

iandees:
This is an edge case I haven't been able to pin down yet. I think it happens when the spider times out and produces no output after getting killed by the timeout. I think the CI code that looks for the number of results fails to find anything and this is what we get.

PR #14845, opened 2026-01-03
Type hint fixups particularly for storefinders

davidhicks:
As ty has now moved into beta versioning and continues to be improved
rapidly, it can now identify additional type hint problems with ATP's
core and storefinder code.

This changeset allows the latest version of ty to validly verify ATP's
core and storefinder code.

Significant code style improvements and documentation improvements have
also been conducted for all storefinders, and storefinders are now more
robust with providing useful error messages if a spider fails to
correctly specify required storefinder attributes. There is a small
chance that approx. < 10 spiders will be broken by this change as they
may have incorrectly specified multiple API keys, start_urls, etc for
use with a storefinder. These spiders, if they exist, will need to be
split into separate spiders (easy task to do, generally one spider can
inherit from the other and the only difference will be an API key or
start_url or similar).

### DIFF
--- a/ci/run_pr_spiders.sh
+++ b/ci/run_pr_spiders.sh
@@ -238,7 +238,7 @@ do
             # Warn if items are missing geometry
             missing_geometry=$(jq '."atp/field/geometry/missing" // 0' "${STATSFILE}")
             if [ $missing_geometry -gt 0 ]; then
-                STATS_WARNINGS="${STATS_WARNINGS}<li>⚠️  Geometry is missing on ${missing_geometry} items</li>"
+                STATS_WARNINGS="${STATS_WARNINGS}<li>⚠️ Geometry is missing on ${missing_geometry} items</li>"
             fi
 
             # Error if items have invalid geometry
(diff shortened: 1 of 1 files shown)


Cj-Malone:
:(

davidhicks:
I think it's a reference to #14837?

If there's any problems with changes I'm happy to fix. I've checked exporters and a sample spider for numerous major storefinders and can't see any faults introduced on my end.

The changes here do tweak some stats around lat/lon fields vs. geometry field so maybe @Cj-Malone and team are impacted by changes to these stats for invalid geometry? Geometry handling and checking is still pretty poor overall so I'd suggest refactoring in this space, particularly if it can allow proper ATP support for extracting more than just Point geometry. I think a discussion about _when_ to validate extracted information is first needed though as ATP currently will extract invalid geometry into lat/lon fields before validation and transformation into GeoJSON compatible geometry within a pipeline, and I wonder whether each spider should instead handle invalid geometry earlier through the set_lat_lon function and raising warnings within the spider, rather than rely on pipelines.

PR #14847, opened 2026-01-04
Alter natwest_gb spider to Return Banking Hub and Mobile branches as well

rjw62:
The current NatWest spider does not return Mobile branches, or the 'branches' that are in Banking Hubs where a NatWest community banker is present (generally one day a week). It would be useful to have these included in the ATP output. This change should achieve this.

The natwest_gb.py script is changed to include the other branch types in the query, and then there are a few other changes later on to make sure the correct properties are applied to each type, and any missing info is handled gracefully.

### DIFF
--- a/locations/spiders/natwest_gb.py
+++ b/locations/spiders/natwest_gb.py
@@ -60,8 +60,16 @@ class NatwestGBSpider(Spider):
                 else None
             )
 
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Is there any specific tagging for them?

PR #14932, opened 2026-01-14
Add located_in mapping for ATM spiders

JanCallewaert-TomTom:
Sorry for the big PR, bit it seemed easier to do it like this than create all those separate small PRs. Let me know if you want me to split this up.

### DIFF
--- a/locations/extensions/filter_stats.py
+++ b/locations/extensions/filter_stats.py
@@ -1,0 +1,21 @@
+from scrapy import signals
+from scrapy.crawler import Crawler
+
+
+class FilterStatsExtension:
+    crawler: Crawler
+
+    def __init__(self, crawler: Crawler):
+        self.crawler = crawler
+        crawler.signals.connect(self.stats_spider_closing, signal=signals.spider_closed)
+
+    @classmethod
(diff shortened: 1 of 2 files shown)


JanCallewaert-TomTom:
> Looks good to me, although I did not test every single spider which you added the new code in. I like the examples in `brand_utils.py`, but as there is a `test_country_utils.py`, we should probably also have a `test_brand_utils.py`.

I have added a test.

locations/brand_utils.py
```diff
@@ -0,0 +1,92 @@
...
+        >>> # Mix modes per mapping
+        >>> mappings = [
+        ...     (["PREMIER"], {"brand": "Premier", "brand_wikidata": "Q7240340"}, "equals"),
+        ...     (["WALGREENS"], {"brand": "Walgreens", "brand_wikidata": "Q1591889"}, "contains"),
+        ... ]
+    """
+    name_upper = name.upper().strip()
+
+    for mapping in mappings:
+        # Unpack tuple - support both 2 and 3 element tuples for backwards compatibility
```
Cj-Malone: backwards compatible with what?

PR #14953, opened 2026-01-16
[Rawson ZA] Fix addr_full error to get more POIs

karen-kerr:


matkoniecz:
it seems to not work well? is `addr:full=407 Inanda Road` intentional?

https://alltheplaces.xyz/preview.html?show=https://alltheplaces-data.openaddresses.io/ci/alltheplaces-fetch:70f18a50-6b36-4552-9a4a-e00bd757850e/rawson_za/output.geojson

PR #14960, opened 2026-01-17
[Target US] Update stale API key

jhanna99:
The API key added in #14859 appears to be returning empty responses, resulting in only 12 stores being scraped instead of ~1,980.

Updated to a working API key extracted from Target's store pages.

Tested: API now returns full store data for valid store IDs.

### DIFF
--- a/locations/spiders/target_us.py
+++ b/locations/spiders/target_us.py
@@ -14,7 +14,7 @@ class TargetUSSpider(SitemapSpider):
     item_attributes = {"brand": "Target", "brand_wikidata": "Q1046951"}
     allowed_domains = ["target.com"]
     sitemap_urls = ["https://www.target.com/sitemap_stores-index.xml.gz"]
-    custom_settings = {"ROBOTSTXT_OBEY": False, "CONCURRENT_REQUESTS": 1}
+    custom_settings = {"ROBOTSTXT_OBEY": False, "CONCURRENT_REQUESTS": 1, "DOWNLOAD_DELAY": 0.5}
     api_key = "c661e81f01c77ef088bfea0cea9242ad30f02858"
 
     def sitemap_filter(self, entries: Iterable[dict[str, Any]]) -> Iterable[dict[str, Any]]:


jhanna99:
The timeout is expected - the spider successfully scraped 96 stores before hitting the CI time limit, confirming the API key fix works. The full ~1,980 stores will be captured during scheduled runs.

locations/spiders/target_us.py
```diff
@@ -14,8 +14,8 @@ class TargetUSSpider(SitemapSpider):
     item_attributes = {"brand": "Target", "brand_wikidata": "Q1046951"}
     allowed_domains = ["target.com"]
     sitemap_urls = ["https://www.target.com/sitemap_stores-index.xml.gz"]
-    custom_settings = {"ROBOTSTXT_OBEY": False, "CONCURRENT_REQUESTS": 1}
-    api_key = "8df66ea1e1fc070a6ea99e942431c9cd67a80f02"
+    custom_settings = {"ROBOTSTXT_OBEY": False, "CONCURRENT_REQUESTS": 1, "DOWNLOAD_DELAY": 0.5}
```
davidhicks: Download delay should always be unspecified and take on the default (1s), or specified to be >1s unless justified as needing to be faster. See #14967

locations/spiders/target_us.py
```diff
@@ -14,8 +14,8 @@ class TargetUSSpider(SitemapSpider):
...
+    api_key = "c661e81f01c77ef088bfea0cea9242ad30f02858"
```
davidhicks: https://www.target.com/store-locator/find-stores still uses API key of 8df66ea1e1fc070a6ea99e942431c9cd67a80f02 in my observation right now. Which page did the API key of c661e81f01c77ef088bfea0cea9242ad30f02858 come from?

davidhicks (changes requested):

jhanna99:
I'm fine with closing this as long as we're now getting a reasonable number of Target stores from the ATP spider.

PR #14963, opened 2026-01-18
[Walmart US] Switch to sitemap-based approach for complete coverage

jhanna99:
The previous GraphQL API approach used an `accessTypes` filter that excluded stores without pickup services, resulting in only ~4,434 stores instead of the full ~4,618.

Changes:
- Use sitemap at sitemap_store_main.xml which indexes all store types (supercenter, neighborhood market, discount, other)
- Extract store data from __NEXT_DATA__ JSON embedded in store pages
- Reduced DOWNLOAD_DELAY from 10s to 5s (sitemap approach is gentler)

The sitemap is the authoritative source for all Walmart US store URLs and should provide more reliable, complete coverage.

Expected improvement: ~4,434 → ~4,618 stores (+4%)

### DIFF
--- a/locations/spiders/walmart_us.py
+++ b/locations/spiders/walmart_us.py
@@ -1,90 +1,91 @@
 import json
-from typing import Any, AsyncIterator
-from urllib.parse import urlencode
+from typing import Any
 
-from scrapy import Spider
-from scrapy.http import JsonRequest, Response
+from scrapy.http import Response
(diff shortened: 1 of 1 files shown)


jhanna99:
I'm currently in Kissimmee Florida and there are 7 Walmarts in the town.  The old approach found only 1 Walmart.  The new approach finds all 7 which also matches OSM for Kissimmee Walmarts.

PR #14977, opened 2026-01-19
[national_bank_of_greece_gr] fix address classification

matkoniecz:


matkoniecz:
maybe there should be `addr:something` for such data not clearly going into any specific well defined address type field?

iandees:
I think "street address" is a good spot for that sort of thing.

PR #14996, opened 2026-01-20
TreePlotter spiders: fix spiders

davidhicks:
APIs have changed slightly.

Also remove Brockport Village Board spider as this TreePlotter account is disabled/no longer available. Maybe they've moved to an alternative system or no longer track tree assets.


iandees:
Closing — could you split this into smaller, more focused PRs? That would make it easier to review and merge incrementally.

PR #15022, opened 2026-01-20
[green_motion] - this is addr:street_address rather than addr:full

matkoniecz:
for example, this values: Fuglavik 43
Via Europa 90
4111 NW 27th Street
Skogarhlio 16
Ratakatu 5
ZAVNOBiH-a 13
Levintie 259
Liepojos pl 1
Zvejų g. 24
1025 W Laurel St


matkoniecz:
output and change looks fine to me on additional check

though note that `add:city` is part of some - say `ParkDIA, 25200 E 68th Ave, 80019, Aurora`

PR #15023, opened 2026-01-20
[mcdonalds_latin_america] - this is addr:street_address rather than addr:full

matkoniecz:
for example, this values: Chaco752
Alberdi N° 25
Alberdi N° 25
Alberdi N° 25
Córdoba 1643
Paraná 3745
Paraná 3745
Paraná 3745
Paraná 3745
Barrio 31


Cj-Malone:
Rel: https://github.com/alltheplaces/alltheplaces/pull/11863

PR #15047, opened 2026-01-22
Check for NaN in coordinate.

ckpiper:
I found that `tuev_nor_de` spider was giving a NaN in the coordinates. You can check the [output](https://alltheplaces-data.openaddresses.io/runs/2026-01-17-13-32-44/output/tuev_nord_de.geojson) and search for NaN. Now I thought this would have been nulled out and put as an invalid geometry. But I believe it would pass the `check_geom` as the `lat/lon_untyped` would be "nan". That means `float(nan)` does not fail, out of range lat/lon check does not fail (as `nan > 90` is False and `nan <= 90` is False), and finally `math.fabs(nan) < 3` is also False. 

Unfortunately, when I run the spider there is no issue like that. But I did add a unit test.

### DIFF
--- a/tests/test_pipeline_check_item_properties.py
+++ b/tests/test_pipeline_check_item_properties.py
@@ -34,3 +34,8 @@ def test_country_field_check():
     item = Feature(country="00")
     pipeline.process_item(item)
     assert spider.crawler.stats.get_value("atp/field/country/invalid")
+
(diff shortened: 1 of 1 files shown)


Cj-Malone:
A: Was this already fixed by something? I can't reproduce on master. #14845?
B: WTF is happening in `check_geom`? Surely `shapely` can do this for us.

PR #15066, opened 2026-01-23
[MediaMarkt] create spider for MediaMarkt and Saturn

JanCallewaert-TomTom:
The spiders for Germany and Belgium were failing. They do follow the same pattern, just like the websites for AT, CH, ES, HU, IT, NL. Saturn is part of the same group and follows the same pattern. We can therefore remove the saturn_de spider.

### DIFF
--- a/locations/spiders/mediamarkt.py
+++ b/locations/spiders/mediamarkt.py
@@ -12,6 +12,7 @@ SATURN = {"brand": "Saturn", "brand_wikidata": "Q2543504"}
 
 class MediamarktSpider(StructuredDataSpider):
     name = "mediamarkt"
+    requires_proxy = True
     allowed_domains = [
         "www.mediamarkt.at",
         "www.mediamarkt.be",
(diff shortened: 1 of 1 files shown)


juice928:
👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **1 point that might be worth attention** (could be false positives, please use your judgment):

1. <strong>Consider using a more robust JSON parsing method instead of regex for store discovery</strong>
   - [locations/spiders/mediamarkt.py:L47](https://github.com/alltheplaces/alltheplaces/blob/780f7ed53d0b83bab38189497bcb76ed77387cf3/locations/spiders/mediamarkt.py#L47): `locations/spiders/mediamarkt.py:L47`
   - Impact: Regex parsing is sensitive to changes in JSON key order or whitespace, which might lead to data collection issues if the source formatting shifts.
   - Suggestion: Consider parsing the script content with `json.loads()` or utilizing the existing GraphQL/Sitemap interfaces for a more stable data flow.

If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

PR #15096, opened 2026-01-27
[The Interiors Group] Move to Stockist spider and rename from carpet_court_nz

applecuckoo:
note: the NZ Carpet Court does not have anything to do with the Australian Carpet Court

the phone number uniqueness warning is fine since most Carpet Courts/Curtain Studios share the same premises


applecuckoo:
this PR is ready to merge

PR #15140, opened 2026-01-31
[Wolseley GB] Fix spider

TheUKHighStreet:


TheUKHighStreet:
Works with the proxy but is using a lot of requests.

TheUKHighStreet:
I think the data is on the branches page - which would just need one proxy call. I'll have a look tomorrow.

PR #15158, opened 2026-02-02
[Bed Bath N' Table] fix spider (188 locations)

JanCallewaert-TomTom:


JanCallewaert-TomTom:
https://github.com/alltheplaces/alltheplaces/pull/15106 seems like a better fix for this.

PR #15159, opened 2026-02-02
Fix LocallySpider implementations with multiple URLs

JanCallewaert-TomTom:
[Columbia US] 163 locations
[Crocs EU] 269 locations


davidhicks:
I propose replacing this PR with #15165 and #15166

Location counts are different as the new PRs apply filtering of Columbia, Crocs and ASICS stores to only returned branded stores and not stockists.

PR #15186, opened 2026-02-05
[AED Locations NZ] add spider

applecuckoo:
## Summary by CodeRabbit

* **New Features**
  * Added support for collecting AED location data across New Zealand.
  * Captures facility names, addresses, coordinates, availability details, and defibrillator categorization.
  * Includes locations available 24/7 when that information is provided.

### DIFF
--- a/locations/spiders/aed_locations_nz.py
+++ b/locations/spiders/aed_locations_nz.py
@@ -1,0 +1,37 @@
+from typing import AsyncIterator
+
+from scrapy import Spider
+from scrapy.http import JsonRequest
+
+from locations.categories import Categories, apply_category
+from locations.items import Feature
+
+
+class AedLocationsNZSpider(Spider):
+    name = "aed_locations_nz"
+    item_attributes = {"country": "NZ"}
+    custom_settings = {"ROBOTSTXT_OBEY": False}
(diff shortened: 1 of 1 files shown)


applecuckoo:
hmm, this spider is running a bit slow probably because of the request for each AED

iandees:
🤖 The CodeBuild run was failing because:

1. **`mapmysites.com`'s `robots.txt` returns HTTP 503**, which blocked the spider from starting. Added `ROBOTSTXT_OBEY: False` to the custom settings.

2. **17,791 individual API requests** was causing a CodeBuild timeout - the spider was fetching a separate `.geojson` endpoint for each AED location. The viewport endpoint already returns all the data we need (name, address, coordinates, 24/7 flag), so I simplified to a single request.

3. **DictParser** was not correctly parsing the GeoJSON `properties` dict - name, address, and phone weren't being extracted. The new code parses the `properties` dict directly.

Other changes:
- Added `emergency=defibrillator` category
- Set `opening_hours=24/7` for locations flagged as `available_24_7`
- Set country=NZ in item_attributes

The rewritten spider yields ~17,800 items from a single HTTP request. CI should now pass.

PR #15196, opened 2026-02-06
[Bunnings] fix spider

applecuckoo:


davidhicks:
It looks like they're now using Imperva Incapsula anti-bot protection for the API, but not for their stores sitemap and individual store pages.

The easiest fix from what I can see is to change this spider to a SitemapSpider/JSONBlobSpider parsing https://www.bunnings.com.au/stores.xml (AU) and https://www.bunnings.co.nz/stores.xml (NZ) for store URLs and then looking inside for `response.xpath('//script[@id="__NEXT_DATA__"]')` which appears to contain the same information the API was returning.

I suggest splitting to two spiders, one AU and one NZ, with NZ inheriting from AU and just changing the sitemap URL.

PR #15291, opened 2026-02-16
[Beauty Success] Improve address tagging

matkoniecz:
for example, this values: 331 avenue du Prado
3 GRANDE RUE
55 D820
25 avenue d'Alsace
2 Rue Guépin
120 grand rue
41 avenue du Chater
10 grande rue
Shop A24
15 avenue du cep

note to self:

look into /media/mateusz/OSM_cache/ATP_matcher_cache/entire_atp/2026-02-14-13-32-33/output/

## Summary by CodeRabbit

* **Data Improvements**
  * Removed the optional full-address field from processed location records for cleaner output.


Cj-Malone (changes requested):
`street_address` is assigned twice, which is right?

iandees:
🤖 Looking at the diff here: this PR sets `item["street_address"]` twice, and the second assignment (`item.pop("addr_full", None)`) overwrites the first. I tested `DictParser.parse()` against a sample feature and confirmed it auto-populates `addr_full` from only the `address" field — it doesn't know about `address2`. So the second line actually drops `address2` from the output, which is a regression from the original (correct) line that concatenates `address` + `address2`.

If the goal is just to remove the redundant `addr_full" key from the output without touching `street_address`, it should be `item.pop("addr_full", None)` on its own line, discarding the value rather than reassigning it.

Closing as stale — the fix as written would drop data. Please rebase and reopen if you want to apply the corrected version.

Cj-Malone (approved):

PR #15293, opened 2026-02-16
[Betfred GB] Add spider

ntzm:
Adds all ~1233 Betfred locations. Takes about an hours on my machine.

I've compared the output to the premises-license-register, there are some postcode mismatches but always close by, and some shops that have been closed since, even searching for the exact postcode doesn't come up with the branch.

### DIFF
--- a/locations/spiders/betfred_gb.py
+++ b/locations/spiders/betfred_gb.py
@@ -16,7 +16,6 @@ class BetfredGBSpider(JSONBlobSpider):
     allowed_domains = ["www.betfredgroup.com"]
     item_attributes = {"brand": "Betfred", "brand_wikidata": "Q4897425"}
     requires_proxy = True
-    seen = set()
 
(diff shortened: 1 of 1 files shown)


locations/spiders/betfred_gb.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def extract_json(self, response: TextResponse) -> list:
+        if script := response.xpath('//script[contains(text(), "const json")]/text()').get():
+            return chompjs.parse_js_object(script[script.index("const json") :])
+        return []
+
+    def post_process_item(self, item: Feature, response: TextResponse, feature: dict) -> Iterable[Feature]:
+        if item["ref"] in self.seen:
+            return
+        self.seen.add(item["ref"])
```
Cj-Malone: We have a pipeline that does this, so we don't need to do it here

PR #15296, opened 2026-02-17
ty 0.0.17: lightly refactor ATP core to avoid newly detected type errors

davidhicks:
Fix type hints in ATP core to support improved type hint checking introduced from ty 0.0.15 to ty 0.0.17


locations/middlewares/track_sources.py
```diff
@@ -35,10 +35,8 @@ def _process_item(self, response: Response, item: Item) -> None:
                 "atp/item_scraped_host_count/{}".format(urlparse(item["extras"]["@source_uri"]).netloc)
             )
         except ValueError:
-            self.crawler.spider.logger.error(  # ty: ignore [possibly-missing-attribute]
-                "Failed to parse @source_uri: {}".format(item["extras"]["@source_uri"])
-            )
             self.crawler.stats.inc_value("atp/parse_error/@source_uri")
+            raise RuntimeError("Failed to parse @source_uri: {}".format(item["extras"]["@source_uri"]))
```
Cj-Malone: This isn't a type hint change

Cj-Malone: I don't see this breaking anything, I have only set `@source_uri` inside spiders to known good urns or URLs. But I think you understand my cautious around core changes inside lint commits by now.

(The comment was vague, and its real intent was to let Ian know and had sean this and will review it when I can concentrate. Sorry, I'll be explicit next time)

davidhicks: I've updated the name of this PR to include "refactor" which is the more accurate description of what these ty-induced PRs are trying to achieve.

Sometimes ty improvements might just need a type hint modified in ATP core (should now be 100% completed and unlikely to appear again in the future) and sometimes the best way to fix the type error is to refactor a function to change parameters, or something else similarly invasive.

I'd like to say that these ty-induced PRs should be smaller and smaller as ty continues to improve in what it can detect. And I had observed that was the case until #15287 which introduced ty changes that are a little larger than other recent ty-induced PRs.

locations/pipelines/apply_spider_name.py
```diff
@@ -14,8 +14,12 @@ def from_crawler(cls, crawler: Crawler):
         return cls(crawler)
 
     def process_item(self, item: Feature):
+        spider_name = getattr(self.crawler.spider, "name", False)
+        if not isinstance(spider_name, str):
+            raise RuntimeError("Spider is missing a 'name' attribute. ApplySpiderName pipeline cannot operate.")
```
Cj-Malone: scrapy is removing the name requirement, I imagine we'll keep the requirement, at least for now.

locations/storefinders/nomnom.py
```diff
@@ -45,7 +45,7 @@ def _append_calendar_param(self, url: str) -> str:
             parsed = urlparse.urlparse(url)
             params = dict(urlparse.parse_qs(parsed.query))
             today = date.today()
-            params["nomnom"] = "calendars"
+            params["nomnom"] = ["calendars"]
```
Cj-Malone: Is this intentional?

Cj-Malone (approved):
After the above comments, I'm OK with this.

locations/pipelines/apply_spider_level_attributes.py
```diff
@@ -14,12 +14,13 @@ def from_crawler(cls, crawler: Crawler):
...
             return item
-
         item_attributes = self.crawler.spider.item_attributes
+        if not isinstance(item_attributes, dict):
+            return item
 
-        for key, value in item_attributes.items():  # ty: ignore [unresolved-attribute]
-            if key == "extras":
-                extras = item.get("extras", {})
+        for key, value in item_attributes.items():
```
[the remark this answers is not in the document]
davidhicks: I think we'd need to move these checks to `open_spider` for the pipeline so we get a single error and the spider is terminated before any items are scraped. Within `process_item` the only acceptable exception is `DropItem`, so any warning or error we add here will be generated once per feature scraped, which doesn't make sense as the error is not feature-specific.

davidhicks:
Changing to draft until upstream (or us) have fixed recent versions of ty causing resource exhaustion.

PR #15388, opened 2026-02-25
[The Entertainer GB] Add spider

TheUKHighStreet:


Cj-Malone:
I looked at this data (years ago) and thought it was too bad to be useful. Have you looked at the data?

PR #15432, opened 2026-03-03
Fix lilly_rs spider (241 locations)

Dimitar5555:


Dimitar5555:
> Account Suspended","status":403,"detail":"Your account has been suspended. Check your Zyte API subscription details at https://app.zyte.com/o/settings/billing-subscriptions for more information. If you have reached your spending limit, increasing your spending limit will immediately lift your account suspension 

It works locally but I'm in Bulgaria so it works without proxy too.

PR #15497, opened 2026-03-09
[Benchmarx GB] Deleted benchmarx_gb as brand covered in travis_perkins_gb

mohit-sharma2194:
_**Deleted benchmarx_gb as brand covered in travis_perkins_gb**_


rjw62:
Could we instead consider reinstating this spider and perhaps removing the results from Travis Perkins output? 

The results for Benchmarx in the Travis Perkins results contain the URL and contact details (website, email, telephone) for the parent Travis Perkins store, rather than the specific details for the Benchmarx concession. As you can see from https://www.travisperkins.co.uk/branch-locator/0136 and https://www.benchmarxkitchens.co.uk/branches/leatherhead these values are different. If would be better for ATP to return the specific Benchmarx data for those stores.

PR #15648, opened 2026-03-24
[Molton Brown] Add spider

TheUKHighStreet:
This still needs pages adding.


TheUKHighStreet:
This would be so much easier if I could limit the results to just Molton Brown stores to begin with rather than having to download all the stockists - but I can't see a way to do that. 6883 results but just 27 actual shops if this is correct.

TheUKHighStreet:
Using the pages is missing a lot of stores. It is possible to get nearby stores using geometry:
https://api.moltonbrown.com/kaowebservices/v2/moltonbrown-gb/stores/?latitude=50.821&longitude=-0.139&distance=1000

But the bounding box is quite small. Not sure how this can be used worldwide.

iandees:
🤖 A couple of answers to the questions in this thread:

1. **Filtering to just Molton Brown stores**: the `kaowebservices/v2/moltonbrown-gb/stores` path is already brand-scoped (it's Kao's multi-brand platform, but the `moltonbrown-gb` segment restricts to this brand) — the large stockist count you saw was likely from paginating a different/broader endpoint rather than a missing brand filter.
2. **Worldwide radius search**: this API only supports lat/lon-radius queries, not a full listing, which is a common pattern in this repo. Rather than a single bounding box, spiders in this situation use `city_locations()` from `locations/geo.py` to generate a systematic grid of query start points (city centroids) and dedupe results as they come in — see any spider that imports `city_locations` for the pattern (e.g. `locations/spiders/barclays_gb.py`).
3. **Skipping bad pages (400 errors)**: rather than omitting a page number by hand, add an `errback` to those requests that logs and moves on, so a future 400 on a different page doesn't silently break the crawl.

Closing as stale for now. Please rebase and reopen if you want to pick this back up with that approach.

PR #15682, opened 2026-03-27
Add OpenAddresses spider

Cj-Malone:
Proof of concept for IF OpenAddresses gets folded into ATP somehow. I would expect it to be done in few stages:

1. We read OA output (eg this PR)
2. We read OA json and process source data
3. We remove OA json and have "native" ATP spiders eg https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/addresses/us/mn/us_mn_hennepin_addresses_us.py

I'm not expecting this to be merged, just food for thought.

This is not currently functional because there is no AWS key to read OA.


iandees:
🤖 Closing as stale — per the discussion in this thread, there's already agreement this experimental exporter approach isn't intended to be merged (spider sources should move directly into ATP spiders instead). Please rebase and reopen if you want to revisit this.

PR #15685, opened 2026-03-28
[New Look GB] Test spider

TheUKHighStreet:
I think there should be far more than 10 stores.


TheUKHighStreet:
I think it either needs to use a proxy or be deleted?

iandees:
🤖 Closing as stale. The `new_look_gb` spider currently has no open dead-spider issue and was last fixed 2024-11-22, so master appears to be working fine without this change. Please rebase and reopen if you find it's actually still broken.

PR #15821, opened 2026-04-06
[Central England Co-operative] Add name

TheUKHighStreet:
The name is missing for the convenience stores - I'm still not sure what the correct way to add it is.


TheUKHighStreet:
Some of them seem to have names:
https://stores.centralengland.coop/essex/blackmore/1-fingrith-hall-lane
This is The Co-operative Blackmore so I assume that is branded as The Co-operative Food.
But it has been added to the spider as a funeral directors - so it's all a bit of a mess. I'll have to have another look tomorrow.

PR #15830, opened 2026-04-07
[Posten] fix categories for post_i_butikk and pakkeutlevering

VincentPenalba-TomTom:
Changed Categories.GENERIC_POI to Categories.POST_OFFICE for:
- enhetstype 4 (Post i butikk) - these are post partners operating inside shops
- enhetstype 19 (Posten pakkeutlevering) - these are post partners with parcel pickup

Both are post office services per https://www.posten.no/kart and should not be tagged as generic POI.

All other category mappings unchanged.


VincentPenalba-TomTom:
Closing as out of scope for ATP

PR #15858, opened 2026-04-08
[Core] Add option to skip address spiders

Cj-Malone:
Again, just food for thought. This could be used in weekly to skip address spiders if they want to be ran at a different cadence, or only manually, while staying in this repo.

### DIFF
--- a/locations/commands/list.py
+++ b/locations/commands/list.py
@@ -1,0 +1,30 @@
+import argparse
+
+from scrapy.commands.list import Command
+from scrapy.spiderloader import get_spider_loader
+
+from locations.exporters.geojson import find_spider_class
+
+
+class ListCommand(Command):
+
+    def run(self, args: list[str], opts: argparse.Namespace) -> None:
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Yeah, I could see trees and stuff wanting to be ran less regularly. This PR isn't really meant to be merged, it's just here to show that some spiders can be ignored easily, with minimal changes to the current infra.

Another (possibly better) way would be to have 2 settings.py, one for the normal weekly, one for the addresses. But that would need a real separate module eg `locations.address_spiders` not part of `locations.spiders`, so `SPIDER_MODULES` can be different. I think this would be a "more scrapy" way. But again, this is food for thought, no action needed right away.

iandees:
I am thinking about tackling this as part of a bigger plan to split the output from the various categories of spiders and remove the "all together" output. This is a fairly large change so I'm still working through it, but that's where I'm heading.

Cj-Malone:
#15996

PR #14405, opened 2025-10-15
[Costco] Deleted broken costco Spiders and created costco worldwide spider

mohit-sharma2194:
**_Fixes : deleted broken costco spiders of AU,ES,FR,IS,JP,KR,MX,NZ,SE,TW and merge into costco spider_**

```python
{'atp/brand/Costco': 132,
 'atp/brand/コストコ': 37,
 'atp/brand/コストコ\u3000ガスステーション': 27,
 'atp/brand/好市多': 14,
 'atp/brand/好市多加油站': 3,
 'atp/brand_wikidata/Q715583': 213,
 'atp/category/amenity/fuel': 72,
 'atp/category/shop/wholesale': 141,
 'atp/clean_strings/branch': 1,
 'atp/clean_strings/city': 4,
 'atp/clean_strings/email': 2,
 'atp/clean_strings/postcode': 2,
 'atp/country/AU': 31,
 'atp/country/ES': 10,
 'atp/country/FR': 3,
 'atp/country/IS': 2,
 'atp/country/JP': 64,
 'atp/country/KR': 20,
 'atp/country/MX': 61,
 'atp/country/NZ': 2,
 'atp/country/SE': 3,
 'atp/country/TW': 17,
 'atp/field/city/missing': 108,
 'atp/field/country/from_website_url': 2,
 'atp/field/email/missing': 145,
 'atp/field/image/missing': 213,
 'atp/field/opening_hours/missing': 182,
 'atp/field/operator/missing': 213,
 'atp/field/operator_wikidata/missing': 213,
 'atp/field/phone/invalid': 2,
 'atp/field/phone/missing': 66,
 'atp/field/postcode/missing': 73,
 'atp/field/state/missing': 152,
 'atp/field/state/wrong_type': 61,
 'atp/field/twitter/missing': 213,
 'atp/item_scraped_host_count/www.costco.com.au': 213,
 'atp/lineage': 'S_?',
 'atp/nsi/cc_match': 213,
 'downloader/request_bytes': 4560,
 'downloader/request_count': 11,
 'downloader/request_method_count/GET': 11,
 'downloader/response_bytes': 86211,
 'downloader/response_count': 11,
 'downloader/response_status_count/200': 11,
 'elapsed_time_seconds': 29.971592,
 'finish_reason': 'finished',
 'finish_time': datetime.datetime(2025, 10, 15, 6, 39, 49, 515090, tzinfo=datetime.timezone.utc),
 'httpcompression/response_bytes': 1574249,
[...]

PR #15862, opened 2026-04-09
Consolidate costco country spiders into single worldwide spider

iandees:
## Summary
- Consolidates 10 separate Costco country spiders (AU, ES, FR, IS, JP, KR, MX, NZ, SE, TW) into a single `costco` spider
- Uses a unified JSON API approach for all countries
- Handles country-specific quirks (state as dict, fuel types in multiple languages)

Based on #14405 by @mohit-sharma2194, rebased onto current master to resolve merge conflicts.

Closes #14405


davidhicks:
I'm not too keen on merging these because if Costco leave a country, or a country-specific website fails for another reason, the entire spider for the world also fails to complete the crawl. The previous approach of splitting/inheriting country-specific spiders from a common storefinder avoids this possibility so each country spider crawls independently.

Is there something I'm missing, such as all country-specific domains pointing to a single backend that has a single global rate limit applied?

iandees:
This PR was just resolving merge conflicts in #14405.

It seems that lots of the individual country spiders were broken and Mohit was trying to fix those. If we can solve them without deleting I'd be ok with that.

davidhicks:
Proposed replacement PR @ #15937

PR #15883, opened 2026-04-10
[easybox] Restore parcel locker images

Dimitar5555:
All photo URLs are unique, so there shouldn't be any duplicates.

I also updated the code so each national version uses the correct domain.

### DIFF
--- a/locations/spiders/easybox_bg.py
+++ b/locations/spiders/easybox_bg.py
@@ -13,7 +13,7 @@ class EasyboxBGSpider(scrapy.Spider):
     custom_settings = {"ROBOTSTXT_OBEY": False}
     requires_proxy = "BG"
     no_refs = True
-    
+
     def get_tld(self):
         return "bg"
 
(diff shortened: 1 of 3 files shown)


locations/spiders/easybox_bg.py
```diff
@@ -13,13 +13,16 @@ class EasyboxBGSpider(scrapy.Spider):
...
+    def get_tld(self):
+        return "bg"
 
     def parse(self, response, **kwargs):
         for location in response.json()["data"]:
             item = DictParser.parse(location)
             apply_category(Categories.PARCEL_LOCKER, item)
-            item["image"] = "https://sameday.bg" + location["photo"]
+            if location["photo"] is not None:
+                item["image"] = f"https://sameday.{self.get_tld()}{location['photo']}"
```
Cj-Malone: The following might work, if so, get_tld isn't needed.
```suggestion
            if location["photo"] is not None:
                item["image"] = response.urljoin(location['photo'])
```

Dimitar5555: The photo parameter is being ignored

` 'image': 'https://sameday.bg/wp/wp-admin/admin-ajax.php?action=get_ooh_lockers_request&country=Bulgaria',`

Cj-Malone: I think that'd only happen if photo was null, which you already have a check for. I'll take a look tomorrow

Cj-Malone: photo was `""` not `None` so we were always doing urljoin.

PR #15937, opened 2026-04-15
Costco spiders: various fixes for opening hours, field extraction

davidhicks:
* Opening hours extraction is now localised according to the language of the Costco country website.
* Additional address fields are extracted where the raw data populates the fields.
* Minor bug fixes to avoid errors in some spiders resulting for data format differences/changes. Each country website has some quirks about how they have implemented what otherwise looks like a common data model.

### DIFF
--- a/locations/hours.py
+++ b/locations/hours.py
@@ -773,6 +773,23 @@ DAYS_JP = {
     "日": "Su",
 }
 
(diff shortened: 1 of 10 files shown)


davidhicks:
Replaces PR #15862

PR #15973, opened 2026-04-16
feat: add and extend SylinderSpiders

Johannes-Andersen:
Add a few more store chains that Sylinder APIs offer, expand Sylinder data with shopServices, and cleanup names by moving much of it over to branch info


Cj-Malone:
Mix is currently in OSM as shop=convenience, so I've change this to that. If you feel strongly we can change back. Also ideally it'd be in NSI, it's big enough.

PR #15984, opened 2026-04-17
[Footasylum GB] Add name

TheUKHighStreet:


TheUKHighStreet:
The opening hours are all wrong.

Cj-Malone:
I think the lack of name was caused by it not being in NSI, NSI released the other day so it's sorted now.

PR #16048, opened 2026-04-23
Add OpenAddresses compatible exporter

Cj-Malone:
Just for funsies for now, possibly more important in the future, but I have some open questions before then.

### DIFF
--- a/locations/exporters/openaddresses.py
+++ b/locations/exporters/openaddresses.py
@@ -1,0 +1,53 @@
+import hashlib
+import json
+import uuid
+from typing import Any
+
+from scrapy.exporters import JsonLinesItemExporter
+from scrapy.item import Item
+
+from locations.exporters.geojson import item_to_geometry
+
+mapping = (
+    ("housenumber", "number"),
+    ("street", "street"),
+    ("addr:unit", "unit"),
+    ("city", "city"),
+    (None, "district"),
+    ("state", "region"),
+    ("postcode", "postcode"),
+    ("ref", "id"),
+    (None, "accuracy"),
+)
+
+
+def item_to_properties(item: Item) -> dict[str, Any]:
+    props = {}
+    for map_from, map_to in mapping:
+        if map_from is None:
+            props[map_to] = ""
+        elif map_to == "id":
+            props[map_to] = str(item.get(map_from) or "")
+        elif map_from in item.fields:
+            props[map_to] = item.get(map_from) or ""
+        else:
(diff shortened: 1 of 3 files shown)


locations/exporters/openaddresses.py
```diff
@@ -0,0 +1,53 @@
...
+mapping = (
+    ("housenumber", "number"),
+    ("street", "street"),
+    ("addr:unit", "unit"),
+    ("city", "city"),
+    (None, "district"),
+    ("state", "region"),
+    ("postcode", "postcode"),
+    ("ref", "id"),
+    (None, "accuracy"),
```
Cj-Malone, line 20: TODO: mapping

locations/exporters/openaddresses.py
```diff
@@ -0,0 +1,53 @@
...
+        else:
+            props[map_to] = item["extras"].get(map_from) or ""
+
+    return props
+
+
+def compute_hash(props: dict, fingerprint: str) -> str:
+    h = hashlib.sha1(fingerprint.encode("utf8"))
+    h.update(json.dumps(sorted(props.items()), separators=(",", ":")).encode("utf8"))
+    return h.hexdigest()[:16]
```
Cj-Malone, line 42: We don't actually need to be as autistic about breaking hashes because OA regenerates the hash on every run, unless a fingerprint is supplied, but it looks like none do. We could potentially use the normal ATP hashes, but then we'd sometimes have stable ids.

tests/test_exporter_openaddresses.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def test_hash():
+    item = Feature()
+    item["street"] = "Forelands Field Road"
+    item["housenumber"] = "32"
+    item["postcode"] = "PO35 5TR"
+    item["city"] = "Bembridge"
+
+    assert compute_hash(item_to_properties(item), "457f15bf-1843-4054-b737-8d2719c641db") == "4ab9fc09b8c9a3c9"
```
Cj-Malone, line 32: I didn't actually get as far as testing this is a match to OA, but see above, I don't think we need to.

tests/test_exporter_openaddresses.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def test_unexpected_fields_are_dropped():
```
Cj-Malone, line 35: Maybe not ideal in the long run, eg GB addresses have 2 refs often that downstream consumers would want.

PR #16058, opened 2026-04-24
[ChargePoint] Set Chrome User-Agent to bypass UA blacklist

iandees:
- ChargePoint's map-prod API is fronted by a Cloudflare UA-substring blacklist that 403s some automated UAs while letting real browsers through. Setting `USER_AGENT = CHROME_LATEST` via `custom_settings` gets clean 200s without needing proxy.
- Verified locally: 178 items scraped in 25s with all 200 responses on the map API; no proxy required.

seen in #16020


Cj-Malone:
This runs from Azure with the default useragent :(

image

PR #16088, opened 2026-04-28
[Hall & Woodhouse GB] Try JSON

TheUKHighStreet:
The current spider is using SitemapSpider and several pubs are missing both coordinates and postcode. This gets all the data in one page and has coords.


TheUKHighStreet:
Don't know how to get this to work so I'll delete.

PR #16122, opened 2026-04-30
[Yours Clothing GB] Fix Spider

mohit-sharma2194:
_**Added proxy to fix spider**_


TheUKHighStreet:
Looks like they've removed the structured data from their store pages.

TheUKHighStreet:
I think the store data is actually available in an array on the store finder page.

PR #16128, opened 2026-04-30
[Stadt Zürich Trees CH] Add spider

brawer:


brawer:
On my local machine, the spider completes in 109 seconds, yielding 80952 trees.

PR #16135, opened 2026-05-01
[Costcutter GB] Fix website

TheUKHighStreet:


TheUKHighStreet:
I really don't understand this spider. Do you think it is possible that Costcutter have just changed their site layout completely since it last updated? The store URLs do not match the rule in the spider, there is no ConvenienceStore on the pages and I've only added one line which I don't think should have completely broken the spider?

Cj-Malone:
It does look like the data is different to last Sunday, weird timing.

Cj-Malone:
I vote to wait a few days to see if the site is now stable, before merging this

TheUKHighStreet:
Thanks for rewriting it :)

PR #16178, opened 2026-05-05
Fix Bargain booze Spider

rjw62:
Bargain booze has changed the URL format of its store pages, so the current spider will fail on its next run. This PR fixes the regex to find the new pages. We also add defaults to the item.pop calls to avoid errors if the entries don't exist. (twitter doesn't at the moment.)


Cj-Malone:
Same as Costcutter

PR #16189, opened 2026-05-06
Fix deprecation warning for scrapy.utils.misc.walk_modules

brawer:
Fixes the following warning:

```
alltheplaces/locations/exporters/geojson.py:150: ScrapyDeprecationWarning: The scrapy.utils.misc.walk_modules function is deprecated and will be removed in a future version of Scrapy. Use scrapy.utils.misc.walk_modules_iter instead.
```


Cj-Malone:
Thank you, I kept meaning to do it

PR #16230, opened 2026-05-12
[Best-one GB] Fix spider

karen-kerr:


Cj-Malone:
Looks like the site is the same as Costcutter+Bargain Booze, the coord extraction in them might be simpler.

PR #16258, opened 2026-05-14
[Core] Promote unit to a direct field in Feature

Cj-Malone:
This is kindof prep for getting closer to OpenAddresses, kindof just house cleaning. In regards to OpenAddresses, the other big one is "district", but I kicked the bucket on it because I don't really know what it is, it's different per source, and so needs a bit more thought.


davidhicks:
Just noticed that `CheckItemPropertiesPipeline` is perhaps missing a check for `unit` to be a string, and not some other Python type such as an integer. This pipeline is also missing `housenumber` so it might be an opportunity to double check that the type checks being performed by this pipeline apply to all hardcoded attributes of `Item`.

davidhicks:
I'm guessing updating `DictParser` to automatically extract `unit` and related-named fields from dictionaries is ruled out? `unit` is such a generic variable name that it's possible it's often used not in the context of addresses. Perhaps such variable name is also used for the unit of currency of a shop or ATM, for example. It might only be safe to extract if a parent dictionary is called `address` or similar?

Cj-Malone:
I don't think it's worth the effort, DictParser is already commonly wrong with street address vs full address. Anything with unit will also be a PINA, also it's pretty infrequent, hence we only have a few uses to update after years and years.

PR #16280, opened 2026-05-16
Added Spider: Handlar'n, Tempo and Matöppet (323 new entries)

matthiasfeist:
I added all spiders in the same PR as both supermarket chains use the same CMS and therefore the spiders are essentially identical

### DIFF
--- a/locations/spiders/handlarn_se.py
+++ b/locations/spiders/handlarn_se.py
@@ -5,7 +5,7 @@ from locations.categories import Categories, apply_category
 from locations.json_blob_spider import JSONBlobSpider
 
 
-class HandlarnSeSpider(JSONBlobSpider):
+class HandlarnSESpider(JSONBlobSpider):
     name = "handlarn_se"
     item_attributes = {
         "brand": "Handlar'n",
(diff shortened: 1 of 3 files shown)


matthiasfeist:
Please note that I also here based my PR on the wrong branch initially. it was fixed (hence the force-push). But the comment by "scraper-bot" above is outdated

locations/spiders/handlarn_se.py
```diff
@@ -0,0 +1,31 @@
...
+    }
+    start_urls = ["https://www.handlarn.se/butiker-och-oppettider"]
+
+    def extract_json(self, response):
+        return json.loads(response.css('script#locations[type="application/json"]::text').get())
+
+    def pre_process_data(self, feature):
+        feature["street_address"] = capwords(feature.pop("Street", "") or "")
+        if city := feature.get("City"):
+            feature["City"] = capwords(city)
```
Cj-Malone, line 22: I don't usually title case the raw data, because we'll never be perfect. Different languages are weird, different accents are weird, different punctuation is weird. We'll always be wrong somewhere. I'd rather leave it for a human somewhere.

But in the case, it's probably a very small number of cities, and I assume you checked them. So 👍

PR #16345, opened 2026-05-20
[Postbank BG] Fix broken async start function

davidhicks:
Incorrect return type for async start function fixed.

HTTP 403 experienced outside of country so the spider still may fail. Proxy was previously flagged as being required.


davidhicks:
Zyte proxies also appear to be blocked, hence reason for new failure of this spider.

PR #16376, opened 2026-05-24
[Burson Auto Parts AU] Try Camoufox

davidhicks:


davidhicks:
Works locally in Australia from a residential IP, and without a stealth browser. I'd guess a residential AU IP may be needed

PR #16432, opened 2026-05-26
[Hilti US] Fix spider

karen-kerr:
Embedded Apollo state changed shape: store entries moved from direct values of `apollo.state` to `apollo.state.ROOT_QUERY.hiltiCenters(...).results[i].hiltiCenter`, so the `__typename == 'HiltiCenter'` filter matched nothing.


Cj-Malone:
#16473

PR #16485, opened 2026-05-28
[CoinFlip] Fix spider

karen-kerr:
The ld+json `dayOfWeek` field changed from URL arrays (`["https://schema.org/Monday", ...]`) to plain strings (`"Monday"`). The custom `int(day_time["dayOfWeek"][0])` parser raised ValueError on every store, killing per-store parse.

Dropped the custom `openingHoursSpecification` loop and the manual `hasMap` regex — `LinkedDataParser` now populates `geo`-based lat/lon and parses the new dayOfWeek strings natively. Also dropped the redundant `wanted_types = ["AutomatedTeller"]` and `USER_AGENT = BROWSER_DEFAULT` overrides (defaults match the site).

Added `apply_category(Categories.ATM)` + `apply_yes_no("currency:XBT", item, True)` to match the sibling `coin_flip_us.py`. Picked up per-store `branch` by stripping the brand prefix from the ld+json name, and cleared the brand-level phone/email/twitter/facebook per the unique-contact-fields rule.

Local crawl: 3 items (Cloudflare rate-limited from local IP); the structural fix means the production fetches that already succeed will now yield items instead of ValueError-ing.

### DIFF
--- a/locations/spiders/coinflip.py
+++ b/locations/spiders/coinflip.py
@@ -1,13 +1,12 @@
-import re
+from typing import Any, Iterable
 
 from scrapy.http import Response
 from scrapy.linkextractors import LinkExtractor
 from scrapy.spiders import CrawlSpider, Rule
 
-from locations.hours import DAYS_FULL, OpeningHours
+from locations.categories import Categories, apply_category, apply_yes_no
 from locations.items import Feature
 from locations.structured_data_spider import StructuredDataSpider
-from locations.user_agents import BROWSER_DEFAULT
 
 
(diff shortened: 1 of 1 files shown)


karen-kerr:
Closing pending deeper investigation. While the dayOfWeek parsing fix is structurally correct, the spider has several issues that need a focused look:

- Heavy Cloudflare rate-limiting (88 of 98 fetches 429 in CI)
- 200% item dupe rate within the 6 items that reached the pipeline
- Site now 301-redirects most requests to `otc.coinflip.tech/unsupported`
- Overlaps with `coin_flip_us` (same `brand_wikidata`, different source)

PR #16517, opened 2026-06-01
[Kate Spade] Fix spider

karen-kerr:
The store-locator sitemaps (e.g. www.katespade.com/stores/sitemap.xml)
now return 401/404, so recent runs produced 0 items.

Rewritten to use the RioSEO storefinder, which returns all locations
(US, CA, GB, JP, and EU) from its API.

- branch taken from the per-store name (KSNY prefix stripped)
- email collected from the per-store address
- apply_category(Categories.SHOP_CLOTHES)

### DIFF
--- a/locations/spiders/kate_spade.py
+++ b/locations/spiders/kate_spade.py
@@ -1,29 +1,17 @@
-from scrapy.http import Response
-from scrapy.spiders import SitemapSpider
+from typing import Iterable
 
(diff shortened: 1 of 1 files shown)


Cj-Malone:
website is wrong, and might have been fixed by https://github.com/alltheplaces/alltheplaces/pull/16386

PR #16532, opened 2026-06-02
[McDonald's EE GR LT LV MT] Add common scraper

hfs:
McDonald's in EE, GR, LT, LV, MT and RO is operated by the same company. All their websites except RO have a similar structure and can be scraped using common code.

Turn the existing spider for MT into a shared class in storefinders.

Remove the mcdonalds_baltics spider and replace it with the new variant for EE, LT, LV, because they provide more details.

Add GR.

RO can't be added in the same way, because the website is different.


davidhicks:
I like the brand specific storefinder class and split of spiders. Good approach and quality here :)

PR #16571, opened 2026-06-03
[Farmacie/Parafarmacie IT] Add spiders from salute.gov.it open data (~28k Italian pharmacies)

McCio:
Two spiders sourcing from the Italian Ministry of Health open data portal (IODL 2.0):

- `gov_salute_farmacie_it` — 20,753 active farmacie (~18,950 with coordinates).
  Tipologie: Ordinaria, Succursale, Dispensario, Dispensario Stagionale.
  Dispensari tagged with `pharmacy=dispensario`; seasonal ones additionally `seasonal=yes`.
- `gov_salute_parafarmacie_it` — 7,158 active parafarmacie (~7,130 with coordinates).
  Tagged `amenity=pharmacy` + `dispensing=no` per OSM convention for Italian parafarmacie.

Both include `start_date`, `ref:vatin`, `addr:suburb` (farmacie only) where available.
Data is updated daily by the ministry; the spider scrapes the dataset page to resolve the current download URL.

Also adds `Licenses.IT_IODL2` (Q26805818) to `licenses.py`.

---
_In scope of the [Project of the Month](https://osmit-podoma.wmcloud.org/projects/2026-06_ithealthcare)_

### DIFF
--- a/locations/spiders/government/gov_salute_farmacie_it.py
+++ b/locations/spiders/government/gov_salute_farmacie_it.py
@@ -21,7 +21,6 @@ class GovSaluteFarmacieITSpider(Spider):
         "attribution:website": "https://www.dati.salute.gov.it/it/dataset/farmacie/",
         "use:openstreetmap": "yes",
     }
-    custom_settings = {"DOWNLOAD_TIMEOUT": 120}
 
     def parse(self, response: Response) -> Iterable:
         url = response.xpath('//a[contains(@href, "FRM_FARMA") and contains(@href, ".json")]/@href').get()
--- a/locations/spiders/government/gov_salute_farmacie_it.py
+++ b/locations/spiders/government/gov_salute_farmacie_it.py
@@ -9,14 +9,20 @@ from locations.items import Feature, set_lat_lon
 from locations.licenses import Licenses
 
 
-class GovSaluteParafarmacieITSpider(Spider):
-    name = "gov_salute_parafarmacie_it"
-    # Source: https://www.dati.salute.gov.it/it/dataset/parafarmacie/
-    start_urls = ["https://www.dati.salute.gov.it/it/dataset/parafarmacie/"]
-    dataset_attributes = Licenses.IT_IODL2.value | {"attribution:name": "Ministero della Salute"}
+class GovSaluteFarmacieITSpider(Spider):
+    name = "gov_salute_farmacie_it"
+    # Source: https://www.dati.salute.gov.it/it/dataset/farmacie/
+    # Names are stored in uppercase by the Ministry of Health.
+    start_urls = ["https://www.dati.salute.gov.it/it/dataset/farmacie/"]
+    dataset_attributes = Licenses.IT_IODL2.value | {
+        "attribution:name": "Ministero della Salute",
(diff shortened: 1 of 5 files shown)


locations/spiders/government/gov_salute_farmacie_it.py
```diff
@@ -0,0 +1,74 @@
...
+                    set_lat_lon(item, float(lat_raw.replace(",", ".")), float(lon_raw.replace(",", ".")))
+                except ValueError:
+                    pass
+
+            apply_category(Categories.PHARMACY, item)
+            item["extras"]["dispensing"] = "yes"
+
+            tipologia = record.get("descrizione_tipologia", "").lower()
+            if "dispensario" in tipologia:
+                item["extras"]["pharmacy"] = "dispensario"
```
Cj-Malone, line 70: Should this be `dispensing=yes`? I'd rather OSM tagging if possible, but will make exceptions if required.

locations/spiders/government/gov_salute_farmacie_it.py
```diff
@@ -0,0 +1,74 @@
...
+                if "stagionale" in tipologia:
+                    item["extras"]["seasonal"] = "yes"
```
Cj-Malone, line 72: Should this be indented?

locations/spiders/government/gov_salute_farmacie_it.py
McCio: dispensario was just to not lose the information, but can be ignored as probably in OSM too won't be added.
dispensing=yes is required for pharmacies in Italy

locations/spiders/government/gov_salute_farmacie_it.py
McCio: Yes, they are seasonal based on town-hall decisions (based on tourism they might be open during winter or summer only for example)

locations/spiders/government/gov_salute_farmacie_it.py
Cj-Malone: It can stay for now, but sometime in the future someone will complain that it's not OSM and it'll probably get removed then.

I don't think the output of any of ATP should be blindly added to OSM, people can add local tags (`dispensing`) and remove unknown tags (`pharmacy=dispensario`) when doing OSM editing.

locations/spiders/government/gov_salute_farmacie_it.py
Cj-Malone: But the `if` only runs if the `if` on 69 is positive. IDK if it's intentional, or accidental.

Can `stagionale` be true, even if `dispensario` is false?

McCio: Ah yes that is intentional

Cj-Malone (approved):
Thank you

locations/spiders/government/gov_salute_farmacie_it.py
McCio: Only dispensari can be seasonal

Cj-Malone: Thank you

PR #16572, opened 2026-06-03
[Lafarmacia. IT] Add spider from Algolia store locator

McCio:
Spider for [Lafarmacia.](https://www.lafarmacia.it/) (Q131009869), an Italian pharmacy chain with 647 locations.

Uses the Algolia store locator (`prod_LAF_STORES` index, app `WKANYZFMAK`) discovered in the Next.js JS bundle.

Fields: `ref` (internal store code), `name`, `street_address`, `city`, `postcode`, `state`, `phone`, `email`, `website`, `opening_hours`, `contact:whatsapp` — plus `amenity=pharmacy`, `dispensing=yes`.

---
_In scope of the [Project of the Month](https://osmit-podoma.wmcloud.org/projects/2026-06_ithealthcare)_


McCio:
@Cj-Malone I see in the edit you removed the dispensing=yes, yet that is the standard for actual pharmacies in Italy.

Cj-Malone:
I don't really understand a `amenity=pharmacy` that isn't dispensing, but, if it's accurate in IT and applies universally to Lafarmacia, NSI is probably the better place to store it.

McCio:
In italy there are para-pharmacies, that are only allowed to sell over the counter drugs, so the distinction is explicitly needed in OSM tagging

Cj-Malone:
In GB, that sounds like [shop=chemist](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dchemist). But I don't want to get deep into tagging discussions.

My opinion is still: if it applies to all Lafarmacia, put it in [NSI](https://nsi.guide/index.html?t=*&k=amenity&v=pharmacy&tt=lafarmacia), that's what will actually drive OSM edits. Putting it here won't have much affect, unless you're using the output yourself, in which case, probably just bulk add it to OSM.

PR #16584, opened 2026-06-04
Add Bpet Energy Turkey spider

jhutton1:


Cj-Malone:
Is that street address or a full address?

jhutton1:
> Is that street address or a full address?

You're right.  It's full.  Fixed it.

PR #16590, opened 2026-06-05
[postnet_za] Simplify URL

confusedbuffalo:
The URL without the long cide also works and is neater


Cj-Malone:
I'm not sold on this one, the website links to the long [URLs](https://www.postnet.co.za/stores), not the short ones. The long ones are more likely to be in OSM and other datasets, so trimming them may make it hard to match.

confusedbuffalo:
It's actually 6-3 in favour of shorter URLs in OSM data, but some of them are quite old so maybe it's changed. I'll do some post processing or something then.

PR #16621, opened 2026-06-08
[Burlington] Fix spider

karen-kerr:
The store locator moved off `stores.burlington.com` (now dead) to `www.burlington.com`, which serves store data as React Server Components instead of ld+json, so the old `StructuredDataSpider` produced zero features.

Rewrite as a `SitemapSpider` over the `www.burlington.com` sitemap, parsing the RSC `locations` array (via the shared `parse_rsc` helper) with `DictParser` for address/geo/phone. Branch from `projectName`, website from `storeDetailsHref`, opening hours from `standardHours`, `start_date` from `grandOpeningDate`, and an explicit `SHOP_DEPARTMENT_STORE` category.

Recovers from 0 to 1267 stores.

### DIFF
--- a/locations/spiders/burlington_us.py
+++ b/locations/spiders/burlington_us.py
@@ -1,14 +1,46 @@
+from typing import Any, Iterable
+
+import chompjs
+from scrapy.http import Response
 from scrapy.spiders import SitemapSpider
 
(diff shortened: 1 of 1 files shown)


Cj-Malone:
#16624

PR #16635, opened 2026-06-09
[Toys R Us CA] Add spider

karen-kerr:
Toys R Us Canada (Q85847837) runs on a separate platform from the other
Toys R Us regions, so it needs its own spider. Scrapes the 16 Canadian
stores from the store sitemap: branch, address, phone, opening hours and
website. Category set to SHOP_TOYS.

Replaces the Toys R Us Canada coverage previously served (incorrectly)
by hmv_ca, which has been removed.

### DIFF
--- a/locations/spiders/toys_r_us_ca.py
+++ b/locations/spiders/toys_r_us_ca.py
@@ -1,0 +1,48 @@
+from datetime import datetime
+from typing import Any
+
+from scrapy.http import Response
+from scrapy.spiders import SitemapSpider
+
+from locations.categories import Categories, apply_category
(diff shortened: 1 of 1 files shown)


karen-kerr:
Note on geometry: this source publishes no coordinates, so all POIs ship `addr_full`-only (downstream geocoding fills geometry). Verified there's nothing to extract — the store-page ld+json is only a `WebSite` stub, the HTML has no `data-lat`/`data-lng`, the sitemap entries carry only `loc`/`lastmod`, and the store-locator index has no map widget backing a coordinate feed. So the absent lat/lon is expected, not a parsing miss.

PR #16658, opened 2026-06-09
[Flugger NO] Add Flügger farve spider

gingermusketeer:
Adds a spider for the Norwegian Flügger farve shops. Possible that this can be extended for the other regions where Flügger farve is present but thought this was a good start.


Cj-Malone:
Thank you

PR #16681, opened 2026-06-10
[Erste Bank Polska] Rename from santander_pl and fix

karen-kerr:
Santander Bank Polska was acquired by Erste Group (completed 9 January
2026) and rebranded to Erste Bank Polska. santander.pl now redirects to
erste.pl, and the site, the store feed and Wikidata (Q806653) all use
the Erste name.

Rename santander_pl -> erste_pl, set brand to "Erste Bank Polska" (same
Q806653), and read the erste.pl store feed — recovering all ATMs and
branches (was 0). NSI still labels Q806653 "Santander", so name and the
ATM operator are set explicitly to stop the stale value back-filling.

Also update euronet_pl: the Euronet ATM list already labels these
machines "Erste Bank Polska", so its brand mapping is repointed — these
ATMs were going unbranded under the old "Santander Bank Polska" key.

### DIFF
--- a/locations/spiders/erste_pl.py
+++ b/locations/spiders/erste_pl.py
@@ -7,10 +7,10 @@ from locations.hours import DAYS, OpeningHours
 from locations.items import Feature
 
 
-class SantanderPLSpider(Spider):
-    name = "santander_pl"
-    item_attributes = {"brand": "Santander", "brand_wikidata": "Q806653"}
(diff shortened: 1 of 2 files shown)


karen-kerr:
https://github.com/osmlab/name-suggestion-index/commit/9ae089f3832460fb640b954f89fbd0faf3785d5d

PR #16684, opened 2026-06-10
Sort out different spiders for Texaco and Valero in UK

TheUKHighStreet:
Original Texaco spider now finding Valero branded petrol stations.


TheUKHighStreet:
I've just copied the Valero spider to the texaco_gb_ie spider and added a header. Is that ok or should I rewrite the valero spider to allow a header to be added?

TheUKHighStreet:
Thanks - that was easier than I thought!

Cj-Malone:
Thank you

PR #16695, opened 2026-06-10
[Rema 1000 NO] Add website and tweak hours

gingermusketeer:
Now the website URL is returned from the spider for each store. This makes it easier to confirm opening hours when there are holidays etc.

A minor change was made to the opening hours which is to explicitly handle days when the shop is closed.


Cj-Malone:
Thank you

PR #16702, opened 2026-06-11
[TymeBank] Remove generic phone

confusedbuffalo:
Essentially all phones were identical

### DIFF
--- a/locations/spiders/tymebank_za.py
+++ b/locations/spiders/tymebank_za.py
@@ -19,23 +19,32 @@ PICK_N_PAY_BRANDS = {
 class TymebankZASpider(LocationBankSpider):
     name = "tymebank_za"
     client_id = "05c109b0-604d-4570-9ecf-fee8f55b18fd"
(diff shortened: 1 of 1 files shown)


Cj-Malone:
What is this? I can't find any "TymeBank" in OSM. It should be added to NSI.

confusedbuffalo:
And actually having checked the website, it looks like they are rebranding to "GoTyme Bank".

I have only come across them as 'kiosks', which is like an ATM, but you can't get any money out. I think you can get cards from them and do basic banking services. I didn't think they had any branches, but the spider lists "customer hubs", which matches references on the website to human interaction. But then their kiosk locator lists those locations like kiosks.

I'll have a look and add to NSI.

I had mapped one, but then the store that it was in closed, so it's gone now.

Cj-Malone:
[This](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dpayment_terminal) may be closer than a vending machine?

PR #16749, opened 2026-06-16
[enseignement_premier_second_degre_fr] Fix licensing metadata

brawer:
Before this change, the licensing metadata was declared within the spider, with `license:wikidata` pointing to a [Stone setting in Södermanland, Sweden](https://www.wikidata.org/wiki/Q29949417).

After this change, the licensing metadata for this dataset is taken from `locations/licenses.py`. Also corrected French accents in the attribution string.


Cj-Malone:
Thank you

PR #16763, opened 2026-06-16
[Commerzbank DE] Emit ATM instead of BANK for self-service-only locations

### LINKED ISSUE #7600: commerzbank_de - some locations are merely ATMs, not a proper banks
https://www.alltheplaces.xyz/map/#22/50.9726723/14.2708347
https://filialsuche.commerzbank.de/en/branch-name/Sebnitz%20-%20Markt

```
Self service terminals
ATM
Cash-In machine
```

something like following is not a bank - it is already mapped as ATM, see https://www.openstreetmap.org/node/339009251

screen02

------------------------------------------------------

likely actual bank - https://filialsuche.commerzbank.de/en/branch-name/Aachen

```
Cashier services are available during opening hours.
Self service terminals
ATM
Service terminal
Account statement printer
Cash-In machine
```

iandees:
Fixes #7600.

Standalone Commerzbank ATM kiosks (SB-Standorte) were being emitted as `amenity=bank`, which is incorrect — they have no teller, no cashier, and are self-service only.

The API data includes `orgTyp` (e.g. `"SB"` for Selbstbedienung) and `kasse` (boolean, whether a cashier is present). Locations where `orgTyp == "SB"` or `kasse == False` are now emitted as `amenity=atm`. Full branches with cashier services continue to emit `amenity=bank` with `atm=yes/no` as before.

Closes #7600

### DIFF
--- a/locations/spiders/commerzbank_de.py
+++ b/locations/spiders/commerzbank_de.py
@@ -105,7 +105,13 @@ class CommerzbankDESpider(CrawlSpider):
 
                 properties["opening_hours"] = self.parse_hours(branch)
 
-                apply_category(Categories.BANK, properties)
-                apply_yes_no(Extras.ATM, properties, bool(branch.get("geldautomat")))
+                # Locations with orgTyp="SB" (Selbstbedienung / self-service) and
+                # no cashier (kasse=False) are standalone ATM kiosks, not branches.
+                is_atm_only = branch.get("orgTyp") == "SB" or not branch.get("kasse")
+                if is_atm_only:
+                    apply_category(Categories.ATM, properties)
+                else:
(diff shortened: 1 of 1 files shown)


sebastianfrankaachen:
Could you please revise this classification?

kasse == false does not automatically mean that a location is an ATM-only or self-service location. These can still be regular bank branches with staff providing in-person consultation; they simply do not have a staffed cash counter.

Only locations where orgTyp == "SB" should be classified as self-service locations without staff and emitted as amenity=atm.

Therefore, the distinction should be:

orgTyp == "SB" → amenity=atm
Other organisation types, including those with kasse == false → amenity=bank

PR #16911, opened 2026-06-18
[apply_category] Don't allow multivalue top level tags

Cj-Malone:
#10217 and others has made the categories excessively complex, reverting to 1 category. More can and should be done to simplify categories.

### DIFF
--- a/locations/categories.py
+++ b/locations/categories.py
@@ -457,15 +457,10 @@ class Categories(Enum):
 def apply_category(category: Mapping | Enum, item: Feature | dict) -> None:
     """
     Apply categories to a Feature, where categories can be supplied as a
-    single Enum, or dictionary of key-value strings. If a value for the
-    category key is already defined, the new value for the category key is
-    appended rather than overwritten. When appending the new value, the list
-    of values is sorted and each value is separated with a semi-colon. Any
-    duplication of values is avoided by ignoring second attempts to add an
-    already existing value.
+    single Enum, or dictionary of key-value strings.
     :param category: Either an Enum member representing a single category to
                      add, or a dictionary of key-value strings representing
-                     multiple categories to add.
+                     a category to add.
     :param item: Feature to which categories should be added to.
     """
     if isinstance(category, Enum):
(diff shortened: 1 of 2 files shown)


iandees:
Agree that we should think of a better way of doing categories. Will merge this to unstick things though.

davidhicks:
I've looked into this a bit more, and come across this NSI entry as an example of a brand with two top level categories:

From amenity/restaurant:
```
    {
      "displayName": "Cooper's Hawk",
      "id": "coopershawk-96af40",
      "locationSet": {"include": ["us"]},
      "matchTags": ["shop/alcohol", "shop/wine"],
      "tags": {
        "amenity": "restaurant",
        "brand": "Cooper's Hawk",
        "brand:wikidata": "Q17511079",
        "cuisine": "american",
        "name": "Cooper's Hawk",
        "official_name": "Cooper's Hawk Winery & Restaurants",
        "shop": "wine"
      }
    },
```
Reference: https://github.com/osmlab/name-suggestion-index/blob/ef5d3834bc1c3a833d935d4b3e9de91cb0f25028/data/brands/amenity/restaurant.json#L1943C1-L1946C12

NSI has decided that a single feature should be tagged as both a restaurant and a wine shop. I believe ATP best practice is to instead extract two features (not one) and give one feature the Categories.RESTAURANT category and the other feature the Categories.SHOP_WINE category. Or just pick one of these categories and ignore the other. Existing spider `coopers_hawk_us` does the later. It chooses Categories.RESTAURANT and ignores Categories.SHOP_WINE. Current NSI matching code in ATP I think is OK with this, but the ATP Feature will end up with two top level tags--`amenity=restaurant` as declared through `apply_category(Categories.RESTAURANT, item)` and `shop=wine` as enriched from NSI matching.

Cj-Malone:
Also some Fedex entries and Chuck E. Cheese. NSI is wrong. But internally to NSI, they have 1 real category, "amenity/restaurant.json".

davidhicks:
New NSI matching code will now check not just brand/operator Wikidata ID and name against NSI, as well as geographic location against NSI, but also top level category tags. When top level category tags are checked, if `get_category_tags` returns just one top level tag such as `shop=wine` then an entry in NSI will match if it also contains `shop=wine`. If that NSI entry also happens to have tag `amenity=restaurant` then ATP NSI matching code doesn't care. If `get_category_tags` were to return `cuisine=chicken` as well as `shop=wine` then NSI matching would fail if the NSI entry didn't have both `cuisine=wine` AND `shop=wine`.

To see whether it's ever necessary to match NSI entries with more than a single top level category tags, I tried finding an example of two brands (or a single brand) operating under the same Wikidata ID and name in the same geographic region, both with the same top level tag, e.g. `amenity=fast_food`, but one has `cuisine=chicken` and the other has `cuisine=fish`. The closest example I discovered was `Ikea Bistro` and `Ikea Restaurant` in NSI which share the same brand Wikidata ID and name, same geographic location and same top level category tag. However, there's no good way for ATP to decide which NSI entry to match against unless the spider were to supply a fourth bit of data to match on, the `name` tag. This is because there is no other distinguishing tag such as `Ikea Bistro` having `cuisine=swedish` and `Ikea Restaurant` having `cuisine=swedish;italian` (or whatever).

I haven't looked at the current situation in detail, but it's possible further improvements are needed to NSI matching in ATP to accommodate weird cases such as same brand, same location, same top level category, two NSI entries.

PR #16915, opened 2026-06-19
[Erste PL] Sync brand with NSI

Cj-Malone:
Waiting on a NSI release


iandees:
The QID used here (Q806653) is wrong for this spider. In NSI, Q806653 is used for **Santander (Polska)** — the Polish Santander entity. The current www.erste.pl site is branded "Erste Bank Polska - przedtem Santander Bank" (formerly Santander Bank), meaning the brand recently reverted from Santander back to Erste.

Q806653 on Wikidata carries aliases "Bank Zachodni WBK" and "Santander Bank Polska", and its parent org (P749) is Banco Santander (Q6496310). It is not the right QID for Erste operating in Poland.

The sibling spider `erste_banka_hr.py` (Croatia) uses `Q696867` (Erste Bank, the Austrian parent group), which is also what NSI uses for all Erste Bank entries in AT/HR/HU/RS. That's the appropriate QID here too.

Additionally, the brand name on the live site is "Erste Bank Polska", not "Erste" — changing it to "Erste" would be incorrect.

Suggested fix:

```python
item_attributes = {"brand": "Erste Bank Polska", "brand_wikidata": "Q696867"}
```

This matches the established pattern in `erste_banka_hr.py` and NSI. The `operator_wikidata` in the ATM branch of the parse function should also stay as `Q696867` (or be updated from Q806653 if that was intended to carry over).

PR #16937, opened 2026-06-19
Restore dominant school tagging

Cj-Malone:
Rel: #16734

While https://wiki.openstreetmap.org/wiki/Proposal:Education_tag may have passed, it's still far from used. Lets stick with the simpler categories for now.

### DIFF
--- a/locations/categories.py
+++ b/locations/categories.py
@@ -24,12 +24,12 @@ class Categories(Enum):
     PARKING = {"amenity": "parking"}
     PARKING_SPACE = {"amenity": "parking_space"}
 
-    KINDERGARTEN = {"amenity": "kindergarten", "education": "kindergarten"}
-    SCHOOL = {"amenity": "school", "education": "school"}
-    COLLEGE = {"amenity": "college", "education": "college"}
-    UNIVERSITY = {"amenity": "university", "education": "university"}
-    LANGUAGE_SCHOOL = {"amenity": "school", "education": "language_school"}
-    NATURE_SCHOOL = {"amenity": "school", "education": "nature_school"}
+    KINDERGARTEN = {"amenity": "kindergarten"}
+    SCHOOL = {"amenity": "school"}
+    COLLEGE = {"amenity": "college"}
+    UNIVERSITY = {"amenity": "university"}
+    LANGUAGE_SCHOOL = {"amenity": "language_school"}
+    NATURE_SCHOOL = {"education": "nature_school"}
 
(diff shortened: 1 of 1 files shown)


davidhicks (changes requested):
I had forgotten to check whether the change would break NSI matching, and yes, it does, because NSI is still using `amenity` tags for education facilities.

I have two sets of recommendations that are almost identical to each other, and as far as I know would mostly just impact output from `CountCategoriesPipeline` whether we call a school `education=school` or `amenity=school` in ATP stats. So not a huge difference.

RECOMMENDATION SET 1

1. Keep ATP as-is with use of the `education` tag as a top level tag, but amend `ApplyNSICategoriesPipeline` so if the result of the `get_category_tags` includes `education=school`, this is replaced with the older `amenity=*` equivalent tag from an OSM:NSI category mapping maintained in `ApplyNSICategoriesPipeline`. Code to update:
```
if get_category_tags(item):
            nsi_matches = list(filter(lambda x: self.nsi_entry_has_all_tags(x, get_category_tags(item)), nsi_matches))
```
Code should be changed so result of `get_category_tags(item)` has certain `education` tags removed and replaced with equivalent `amenity` tags which are still used by NSI.


RECOMMENDATION SET 2

1. Leave `education=*` tags in `Categories.KINDERGARTEN`, `Categories.SCHOOL` and so on for ATP consumers that have or are migrating from `amenity` to `education` for education features.

2. Remove `education` from `top_level_tags` in `categories.py`. This will ensure `get_category_tags`, NSI matching, etc in ATP stops caring about `education=*` tags altogether. Whilst doing this, add a code comment explaining the reason for excluding `education` from the list, which I assume is that NSI hasn't yet migrated to `education`, and top level tags are primarily used by ATP for NSI matching.

3. Completely remove the `if "amenity" in categories.keys() and "education" in categories.keys():` branch from `get_category_tags` function in `categories.py` as the function won't care about `education=*` tags if point (2) above is implemented.

davidhicks:
Please see suggested changes of PR #17327 for an alternative approach to handling OSM's allowance during transition periods for having e.g. `amenity=school` and `education=school` used at the same time. Basically we can internally accept both top level tags in ATP, and when doing any NSI matching, pick the category that NSI prefers for matching purposes. That way features output from ATP would have both old and new top level tags from OSM (during the transition period where OSM is migrating to a new scheme) _and_ would have all the tags from NSI matching applied too.

PR #16946, opened 2026-06-20
Improve category handling for the Feature class

davidhicks:
Currently a 'category' in ATP is primarily used for the purpose of matching a feature to one in NSI, and applying additional tags from the matched NSI entry. There are some auxillary uses too--such as compiling statistics at the end of each spider run, which is particularly useful for spiders which extract multiple types of features, such as bank branches and ATMs.

The primary method expected to be used today by a spider to set a category for a feature is use of the 'apply_category' function. This function has had a messy historical history, having changed in functionality a few times, and been used in differing ways by spiders. This is largely due to the historical lack of strict definition for the 'apply_category' function and the ad-hoc way in which ATP has historically wanted spiders to specify a category.

It is now clear from PR #16911 and other changes in the past year:
- A category is an NSI "category", which is based on https://wiki.openstreetmap.org/wiki/Top-level_tag
- A feature extracted by ATP should only have a single NSI "category" applicable. That is, a spider extracting a shop=supermarket and amenity=parking as a single feature for a supermarket that also has a nearby customer carpark should extract two separate features, each with only one category (either shop=supermarket or amenity=parking).
- Spiders should provide minimum tags necessary to permit a feature to be matched to an NSI entry. For example, a spider shouldn't need to specify cuisine=chicken for spider kfc_us because this tag can be obtained from (and centrally managed) through the NSI database. Extras tags should only be supplied where they differ between different features being extracted. If tags apply to all features extracted by a spider, a change request should be made to NSI to update tags in the NSI database.
- If there is a conflict between NSI categories and OSM top-level keys (https://wiki.openstreetmap.org/wiki/Top-level_tag) as is the case for the example of amenity=school/education=school (OSM is transitioning to new education= top level key) then the NSI category is preferred to be used by ATP, because ATP primarily cares about categories just to allow an NSI match to occur.

This change introduces a new 'category' property to the Feature class which expects a type only of a single member of the Categories Enum (from categories.py). This presents a number of benefits to help prevent spiders from messing up categories, and also to make spiders simpler and easier to read and develop:
- `item_attributes = {"brand": "..", "brand_wikidata": "..", "category": Categories.SHOP_WINE}` can be used as the preferred method for specifying a single common category for all features extracted by a spider. It's not possible to accidentally specify two categories for the same feature. It also places all the information NSI uses to perform matching in a single line (single dictionary) so on a quick glance at spider code it's possible to see how it intends NSI matching to occur.
- `item["category"] = Categories.ATM` can be used within a spider to specify/override the category of an individual feature, if the spider is able to extract different feature types such as bank branches and ATMs.
- There is no longer a need to add single line post_process_item functions into spiders which only call `apply_category(Categories.ATM, item)`. This can significantly reduce the number of imports the spider needs and makes spiders faster to develop and easier to read.
- There is no need to use the `apply_category` function--it is now deprecated (and could be decorated as @deprecated when ATP specifies a more recent minimum Python version).

By setting the 'category' property on a Feature,
ApplyNSICategoriesPipeline will use this 'category' when finding a suitable NSI entry to match the Feature against. The 'extras' dictionary of the Feature will then be populated with any tags contained in the NSI entry. For example, if `item["category"] = Categories.FAST_FOOD` and an NSI match is successful, ApplyNSICategoriesPipeline may set `item["extras"] = {"amenity": "fast_food", "cuisine": "chicken"}`.

If an NSI match is unsuccessful (e.g. NSI doesn't yet contain an entry for the feature) then ApplyNSICategoriesPipeline will fall back to copying tags from the 'category' of a Feature into the 'extras' dictionary. For example, if `item["category"] = Categories.FAST_FOOD` but an NSI match is not possible, ApplyNSICategoriesPipeline would set `item["extras"] = {"amenity": "fast_food"}`.

### DIFF
--- a/locations/categories.py
+++ b/locations/categories.py
@@ -454,21 +454,83 @@ class Categories(Enum):
     NATURAL_TREE = {"natural": "tree"}
 
 
-def apply_category(category: Mapping | Enum, item: Feature | dict) -> None:
+def apply_category(category: Mapping | Categories, item: Feature | dict) -> None:
     """
-    Apply categories to a Feature, where categories can be supplied as a
-    single Enum, or dictionary of key-value strings.
-    :param category: Either an Enum member representing a single category to
-                     add, or a dictionary of key-value strings representing
-                     a category to add.
-    :param item: Feature to which categories should be added to.
+    This 'apply_category' method has been replaced by use of the 'categories'
+    property of the 'Feature' class. It is preferred for a spider to set this
+    category property with either of:
+    1. item_attributes = {"brand": "ACME", "category": Categories.SHOP_WINE}
+    2. item["category"] = Categories.SHOP_WINE
+
+    For older spiders which still rely on this function, current behaviour of
+    this function is documented below.
+
+    A 'category' in ATP corresponds to categories defined in NSI
+    (see https://nsi.guide for list of categories). ATP uses this 'category'
+    to look up an entry in NSI for the Feature extracted to allow additional
+    attributes for the Feature to be automatically populated for the Feature.
+    For example, setting a category of 'amenity=fast_food' in ATP will allow
+    ATP to match a feature with the relevant brand in NSI and then additional
+    attributes such as 'cuisine=chicken' can automatically be added to the ATP
+    feature.
+
+    It's important to note this primary intended use of NSI matching because
+    if there exists a conflict between tagging in OSM and NSI, it is
+    preferred to use the NSI tagging scheme. This issue is seen, for example,
+    with deprecation of amenity=school on OSM and replacement with a new
+    tol level category tag of education=school. If NSI has not yet migrated
+    to use of education=school then it is preferred to follow whatever scheme
+    NSI is currently using.
+
+    This function will apply a 'category' to a Feature. A Feature should only
+    ever have one 'category' declared. Historically, ATP didn't enforce this
+    contraint because ATP stored the category in the 'extras' dictionary which
+    is a property of the Feature class. Spiders were free to add any number of
+    category tags into this 'extras' dictionary, such as:
+      item["extras"] = {"shop": "wine", "amenity": "parking"}
+    For this example, a spider may have set these tags for a single POI that
+    corresponds to a wine cellar shop at a vineyard, as well as customer
+    parking at the vineyard. The correct expectation for ATP spiders is the
+    shop and the parking area should be extracted as two separate features,
+    each with one top level category tag only.
+
+    A 'category' property has more recently been added to the Feature class in
+    ATP to ensure that only one top level category is applied at a time. By
+    passing a member of the Categories Enum member to this function, the
+    'category' property will be replaced with the supplied Categories Enum
+    member. No change is made the 'extras' property of the Feature.
+
+    If a Mapping (dict) is instead supplied to this function, a top level
+    category supplied in this dictionary, such as shop=wine, will be added
+    to the supplied Feature within the 'extras' property (a dictionary). This
+    function will do nothing if the supplied Mapping does not contain a top
+    level category. Do not add two top level categories to the Mapping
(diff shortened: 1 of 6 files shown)


davidhicks:
Uploading as an early/incomplete preview. I've still got to fix and improve tests, and also try and refactor ApplyNSICategoriesPipeline.process_item so it's not 1000 lines long.

iandees (commented):
🤖 The new `apply_category` Mapping path has a breaking change that will silently drop tags from ~30 existing call sites.

**The regression**

Before this PR, when a spider called `apply_category({"cuisine": "american"}, item)`, the dict was written directly into `extras` — the tag was preserved.

After this PR, the Mapping branch now calls `get_category_tags(category)` instead of using the dict as-is. `get_category_tags` filters the input down to only the keys in `top_level_tags` (aeroway, amenity, shop, leisure, etc.), so any tag that is not a top-level OSM key is silently discarded and never written to `extras`.

**Affected call sites (sample)**

| Spider | Dict passed | Tags dropped |
|---|---|---|
| `cracker_barrel.py:48` | `{"cuisine": "american"}` | `cuisine` — entire dict dropped, no top-level key present |
| `pudo_tr.py:25/30/36` | `{"addr:neighborhood": …}`, `{"addr:desc": …}`, `{"access": "private"}` | all three entirely dropped |
| `hyundai_nz.py:55-56` | `{"boat:repair": "yes"}`, `{"boat:parts": "yes"}` | both entirely dropped |
| `politix_au.py:40` | `{"clothes": "men"}` | `clothes` dropped (not in `top_level_tags`) |
| `pizzalonga_away_it.py:42` | `{"cuisine": "pizza", "takeaway": "only"}` | both dropped |
| `seattle_parks` spiders | `{"sport": "baseball;softball"}` etc. | `sport` dropped |
| `kaiser_permanente_us.py:198` | `{"brand": "…", "brand:wikidata": "…"}` | both dropped |

There are roughly 30 call sites in total that pass a plain dict to `apply_category`. Some of these are using `apply_category` as a quick way to set non-top-level tags directly into `extras` (e.g. `cuisine`, `sport`, `clothes`, `access`, `addr:*`), which is a pattern the old implementation supported.

**Suggested fix**

Revert the Mapping branch to its original behaviour:

```python
elif isinstance(category, Mapping):
    tags = category  # was: tags = get_category_tags(category)
```

If the intent is to enforce the "one top-level category only" constraint, raise an explicit `ValueError` when the dict contains a non-top-level key instead of silently dropping it. That makes the breakage visible and gives spider authors a clear migration path.

---

The rest of the PR looks sound: adding `category` as a typed `Feature` field, the `item["category"] = Categories.X` shortcut in `item_attributes`, and the `ApplyNSICategoriesPipeline` fallback-copy logic all look correct and non-breaking.

davidhicks:
Thanks @iandees I'll update this PR to remove/replace those instances where `apply_category` is being misused to add tags that aren't top level, such as `cuisine=*`. This ensures `apply_category` will only accept top level tags if a Mapping (dict) is supplied, or will otherwise raise an error advising the spider author that they're misusing `apply_category`.

iandees:
(My comment on this PR and all the PR noise in the last ~30 min was from me trying to use Claude to review PRs locally. It turns out there is a built-in `/review` command that it picked up and started using. That command instructs it to leave comments and open pull requests to fix stuff, which I explicitly did _not_ want. Apologies for the noise here.)

davidhicks:
PR #17029 (merged) resolves misuse of `apply_category` for applying tags that aren't top-level. PR #17327 provides some further enhancements so hopefully now `apply_category` is only used now with `Categories` enum members, and not with dictionaries of random/little used tags.

Once PR #17327 has been considered/merged I'll update this PR #16946, and may consider whether `apply_category` should still accept dictionaries, or whether we enforce use of `Categories` enum members only.

PR #16968, opened 2026-06-21
[Backstube NO] Add spider

iandees:
- Follows `maps.app.goo.gl` short links in each location row to resolve coordinates via redirect, extracting lat/lon from the `/@lat,lon` path in the final Google Maps URL
- Adds `country: NO` to `item_attributes`; 37 Norwegian bakery locations

seen in #16957

### DIFF
--- a/locations/spiders/backstube_no.py
+++ b/locations/spiders/backstube_no.py
@@ -1,0 +1,61 @@
+import re
+from urllib.parse import unquote
+
+from scrapy import Spider
+from scrapy.http import Request, Response
+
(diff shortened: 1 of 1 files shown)


iandees:
🤖 Closing in favour of merging the original #16957 without coordinates. The only coordinate source on backstube.no is maps.app.goo.gl short links — following those redirects and extracting /@lat,lon from the Google Maps URL constitutes using Google as a data source, which we avoid. The address + opening hours data is valuable enough to merge without coords.

PR #16971, opened 2026-06-21
[First Bank of Toyama JP] Add spider

iandees:
- Adds new spider for First Bank of Toyama (富山第一銀行), 91 locations in Japan
- Guards `categories[0]` index access against empty list to avoid IndexError (seen in #16966)


iandees:

PR #16972, opened 2026-06-21
[Seibu Shinkin Bank JP] Add spider

iandees:
- Guards `categories[0]` index access against empty list
- Guards `name` field against `None` before calling `removesuffix`

seen in #16969


iandees:

PR #16973, opened 2026-06-21
[Daito Bank JP] Add spider

iandees:
Closes #16970

Guards `categories[0]` index access against empty or missing list before the match block. 104 locations.


iandees:

PR #17013, opened 2026-06-22
[GBFS] Fix NSI matching

karen-kerr:


Cj-Malone:
This also fixed Burger Kings and a bunch more, thank you.

PR #17071, opened 2026-06-24
[Core] Shift spiders to outputting OpeningHours

Cj-Malone:
For a while now spiders could output the string or the OpeningHours class. Outputting the class lets us skip validation on the raw string. This PR should give a small performance boost by skipping, but really this is to give the LLM overlords more examples of this so they stop using it in new spiders.


iandees:
❤️ I will add this as a preference in the rules I use for LLM spider generation.

iandees (approved):

PR #17101, opened 2026-06-27
[Walmart-US] Fix spider

GA-Kevin-Codes:
Sitemap pointed to a list of 4 different urls based on store type. I also added the Wikidata for Neighborhood markets


Cj-Malone:
Does the sitemap change do anything? The original [url](https://www.walmart.com/sitemap_store_main.xml) points to the 4, in theory, it'll be more stable over time than the individual ones.

GA-Kevin-Codes:
Good point @Cj-Malone im not sure. When I'm back at my computer we can test it! I was just happy it worked eventually!

PR #17104, opened 2026-06-27
[yuppiechef_za] Add requires proxy to fix spider

arrival-spring:
Works locally, but not in the weekly runs for some time, so looks like a proxy issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved store-location data collection on Yuppiechef South Africa by strengthening Cloudflare challenge handling.
  * Added support for recognizing successful verification and handling temporary access-denied responses.

### DIFF
--- a/locations/spiders/yuppiechef_za.py
+++ b/locations/spiders/yuppiechef_za.py
@@ -15,6 +15,7 @@ class YuppiechefZASpider(PlaywrightSpider):
     start_urls = ["https://www.yuppiechef.com/store-directory.htm"]
     custom_settings = DEFAULT_PLAYWRIGHT_SETTINGS_WITH_EXT_JS
     no_refs = True
+    requires_proxy = True
 
     async def start(self) -> AsyncIterator[Request]:
         yield Request(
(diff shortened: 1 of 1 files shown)


arrival-spring:
Still getting 403 here though, despite proxy

iandees:
## Investigation Summary

**Root Cause**: Yuppiechef is protected by Cloudflare with a managed challenge (not Turnstile). All requests return HTTP 403 with  header and a "Just a moment..." challenge page. 

**Why `requires_proxy` fails**: The Zyte proxy IP pool is on Cloudflare's blocklist. This was encountered before with `eighty_five_degrees_c_au` (commit f3d309665) which had to remove proxy entirely.

**Direct curl test** (with real Chrome UA):
```
HTTP/2 403
Server: cloudflare
cf-mitigated: challenge
Content: Just a moment... (Cloudflare challenge)
```

## Possible Fixes

**Option A: Use Camoufox instead of Playwright** (Most likely to work)
- Switch from `PlaywrightSpider` to `CamoufoxSpider`
- Use `DEFAULT_CAMOUFOX_SETTINGS_FOR_CLOUDFLARE_INTERSTITIAL` 
- Set `captcha_type = "cloudflare_interstitial"`
- Set `handle_httpstatus_list = [403]`
- Reference: `sopharmacy_bg.py` (commit d872085fa) implements this pattern

**Option B: Find alternative API endpoint**
- Check if Yuppiechef has a store locator JSON API (similar to other retailers)
- If found, convert to `JSONBlobSpider`

**Option C: Remove spider**
- If Camoufox approach fails and no API exists, the spider cannot be recovered due to Cloudflare's bot protection

I'd recommend trying **Option A** (Camoufox) first, as it's proven to work for Cloudflare interstitial challenges. Happy to implement if you'd like to proceed.

PR #17105, opened 2026-06-27
[van_schaik_bw_na_za] Fix spider

arrival-spring:
They now have a store locator, but still no geometry


arrival-spring:
I have [updated NSI](https://github.com/osmlab/name-suggestion-index/commit/736a338e6499004f224ea7dcf38285ff7f9ec032) to add BW and NA

PR #17132, opened 2026-06-30
[Cashbuild] Fix Spider

mohit-sharma2194:
Spider works fine locally and from US also, so generalise proxy flag .


Cj-Malone:
Proxy and playwright still needed?

arrival-spring:
Can you just check the opening hours parsing too, it looks like Sunday's hours are being mapped to Saturday, Monday's to Sunday, and so on.

PR #17296, opened 2026-07-09
Add new Spider for Changing Places toilets

rjw62:
"Changing Places" branded toilets in the UK, using the list returned by https://www.changing-places.org/api/getToilets

### DIFF
--- a/locations/spiders/changing_places_gb.py
+++ b/locations/spiders/changing_places_gb.py
@@ -2,21 +2,20 @@ from typing import Iterable
 
 from scrapy.http import Response
 
-from locations.hours import OpeningHours,DAYS  
-from locations.items import Feature
 from locations.categories import Categories
+from locations.hours import DAYS, OpeningHours
+from locations.items import Feature
 from locations.json_blob_spider import JSONBlobSpider
+
 
 class ChangingPlacesGBSpiderr(JSONBlobSpider):
     name = "changing_places_gb"
     item_attributes = {
-        "brand": "Changing Places", 
+        "brand": "Changing Places",
         "brand_wikidata": "Q104870811",
         "extras": Categories.TOILETS.value,
     }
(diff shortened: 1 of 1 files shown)


Cj-Malone:
I don't think my local Tesco toilets are branded "Changing Places"?

rjw62:
> I don't think my local Tesco toilets are branded "Changing Places"?

I don't know for sure, but if they're listed on the Changing Places website, I think they may well be - but it might only be a small ["Changing Places" logo](https://www.changing-places.org/news/view/the-changing-places-symbol-1). For OSM, it would be good to know if all the registered sites typically do show the logo/branding.

Cj-Malone:
> For OSM, it would be good to know if all the registered sites typically do show the logo/branding.

Also for ATP. I don't want `brand=x` here if it's a lie.

It sounds like a "network", like Refil. (Not necessary `network`)

rjw62:
I guess it's a bit like refill, but there's a stronger connection since it's applied to a physical facility rather than a service. It's really more of an accreditation / seal of approval.

But certainly the ones I've seen in the wild (all stand-alone toilets in public places) have the appearance of being branded as "Changing Places", to the point where an observer might thing it's run by the organisation. The ones in supermarkets do seem to use some branding as well with the name and recognisable logo: https://corporate.asda.com/article/changes-at-our-abbey-park-store-include-a-new-changing-places-toilet and https://www.purplegoatagency.com/case-studies/tesco-marketing-campaign/ .

I'm not that bothered about the ATP attributes or OSM tagging, but OSM should have common well-defined way to tag any that are marked as such on the ground. The interesting question is whether all the facilities in the Changing Places database are labelled as such on the ground. I would suspect that almost all are, as there's advantages to the owner and the users for making use of the common brand/logo on the facility.

PR #17327, opened 2026-07-10
Improve top level categories tagging throughout brand spiders

davidhicks:
* Add missing members to Categories and use throughout all spiders to
  ensure that apply_category() is now only used with Categories
  enumeration members, not with dictionaries.

* Enhance healthcare spiders which have complex tagging, including with
  use of HealthcareSpecialities.

* Enhance automotive sales/service/parts spiders with improved use of
  service:vehicle:* tags. For some spiders that weren't correctly
  splitting sales/service/parts POIs, these are now split and tagged
  separately.


Cj-Malone (changes requested):
What's your obsession with the education tag?

davidhicks:
I've mostly used `education` here, alongside `healthcare`, as examples where OSM uses two different top level keys for the same feature types during a transition period. Per https://wiki.openstreetmap.org/wiki/Proposal:Education_tag this `education` key was approved (not without complaint though) to replace `amenity` tags for education/training features.

This change essentially removes ATP from any debate OSM, NSI or other communities might have about top level tagging schemes. ATP can just output for all competing schemes and support everyone (well, mostly) and also at the same time still integrate with NSI that _is_ opinionated (necessarily so) and has to pick one scheme to use over all others.

Cj-Malone:
Adding `healthcare`+`amenity` to the enum was a mistake, I'm sorry.

That `education` wiki page has 24 votes for it, it's not a decided topic. I know it's confusing at first, but the wiki isn't authoritative over OSM data. OSM won't become that overnight, in fact it will never match the wiki, by the time the data changes, the wiki will have changed again. ATP can't and wont jump on every fad in the wiki, we should be using the stable, well used tagging. Not niche, not new, not edgy.

`healthcare` is slightly different, it at least has some usage within OSM. Although, that usage comes from iD heavily pushing it, and backlash continues to this day. Notably, iD is not doing that same level of pushing for `education`, it defiantly is still steering, but less so.

Even after all these years of iD strongly pushing `healthcare`, it's still quite clear that the original tagging methods are more common, and most used by OSM consumers. I think the most likely outcome is that `education` is used far less than that, and end up being forgotten.

If ATP jumps on this, we aren't becoming neutral, we aren't removing ourselves for any debate, we are adding to it, encouraging it.

Cj-Malone:
Sorry, I should have used more words in the first comment.

davidhicks:
Thanks, now that #17342 has been merged I've updated this branch to not output `education` tags, except for `education=nature_school` which has no equivalent in `amenity=*`.

_If_ a spider did set `healthcare` or `education` duplicate top level tags where an equivalent tag already exists under `amenity`, this branch will still allow NSI matching to occur by ignoring the duplicate top level tags that NSI doesn't care for.

locations/categories.py
```diff
@@ -527,13 +546,67 @@ def add_list(key: str, value: str, item: Feature) -> None:
     "waterway",
 ]
 
+nsi_preferred_top_level_tag_mapping = [
```
Cj-Malone: I'm not sure this is necessary now we aren't duplicating tags internally? #17342

Cj-Malone: Oh, for use after pipelines?

davidhicks: It's kind of dead code at the moment without a use case, as after this PR is merged, I don't think there are any spiders left setting top level categories manually (beyond just use of `Categories` enum members), therefore no real risk of spiders stuffing up categories. I can remove `nsi_preferred_top_level_tag_mapping` (and corresponding code in `get_category_tags` and associated test cases) or keep it in place in case a use case develops in the future for it, such as use by a pipeline executing after the healthcare tag duplication pipeline.

Cj-Malone: I think it's overcomplicating things at the moment, maybe we'll need it at some point, but if it's dead code now, lets drop it.

I've not looked at the rest of the PR in detail, but I like the gist of it. Thank you.

davidhicks: Done. Also if PR #16946 goes ahead, `get_category_tags` likely can be deleted too because `Feature` gains `category` as an attribute, and only `Categories` enum members would be accepted, and each of these enum members would only have a single top level tag to help guide NSI matching (particularly for operators more than brands). That way `extras` in the future should be limited to just feature-specific tags such as number of parking spaces, presence of a drive-through at some locations but not others, material used for a specific park bench, etc. And all other tags applying brand-wide should come from NSI matching.

Cj-Malone: Thank you, I have briefly discussed a category field with Ian before. It's a big change, and a lot to think about. Not something I'd rush. Also not something I'd personally do in the summer, save the big things for the dark and wet months.

Cj-Malone (approved):
I've not reviewed every file here, but the gist of the PR is good. Thank you

PR #17342, opened 2026-07-13
Apply duplicate healthcare tags at export time

Cj-Malone:
This PR isn't removing `healthcare` from the output, just moving where it's applied. This PR isn't removing all `healthcare` top level tags as categories, categories like `healthcare=dialysis` remain in the enum and used directly. This PR means internally to ATP we have 1 top level tag in the categorises enum values, moving towards simplifying categories internally. But leaving duplicate tagging on the output.

If `education` gets traction in OSM, it could be added her too. See https://github.com/alltheplaces/alltheplaces/pull/17327#issuecomment-4956584750

Or ultimately, if either new tagging methods overtake the originals in terms of numbers in OSM, and usage in OSM consumers, we could switch these around or even drop the original tagging methods. But right now, the traditional, well used, well supported tags are king.


davidhicks:
Is there a need to output `healthcare=clinic` + `amenity=clinic` instead of just the NSI preferred `amenity=clinic`? `id-tagging-schema` is more open to recommending use of new tagging schemes, and will suggest `healthcare=clinic` + `amenity=clinic`. After reading the OSM situation in more detail, I just can't find a good reason why 'amenity' exists, and why `healthcare` exists, and what the long term intention is for each of the top level tagging schemes. OSM top level tagging schemes are largely just historical in nature and have VERY loose criteria for what is meant to be addressed in each scheme. I can't see that `healthcare` and `education` are really any better defined than `amenity` -- which is essentially no clear definition.

OSM is full of random top level tagging scheme use such as:

- Buy a service (hairdressing) from `shop/hairdresser`
- Buy a service (car repair) from `shop/car_repair`
- Buy a service (taxation advice) from `office/tax_advisor`
- Buy a service (legal advice) from `office/lawyer`
- Buy a service (medical advice) from `amenity/doctors` or `amenity/clinic`
- Buy a service (Internet use) from `amenity/internet_cafe`

- Have fun (theme park rides) from `tourism/theme_park`
- Have fun (trampoline use) from `leisure/trampoline_park`
- Have fun (watch movies) from `amenity/cinema`

I assume `healthcare` and `education` were attempts to try and make sense of the random top level tagging schemes, but this hasn't actually happened, and instead they've just made tagging schemes even more random?

Cj-Malone:
> Is there a need to output `healthcare=clinic` + `amenity=clinic` instead of just the NSI preferred `amenity=clinic`?

I don't think there is a need to, and I think I did it because of inertia, not for a specific reason. Our output is so close to OSM for many reasons, one of them being political, another is technical. Lots of people are already reading OSM data, being the same allows those people to ingest ATP without any real effort.

In the OSM world, most people reading OSM data are still reading `amenity=clinic` over `healthcare=clinic` (see osm.org not rendering `healthcare=clinic`), I can't comment on the commercial world, but I would assume it's the same. Maybe even more behind.

> I just can't find a good reason why 'amenity' exists, and why `healthcare` exists, and what the long term intention is for each of the top level tagging schemes.

There is no single intention on anything, just years of building on top ourselves. I do believe `healthcare` had good intentions, I do believe the data would have been better if we had switched. But OSM didn't, and isn't going to. I have less positive beliefs about `education` because we can see what happened before and intentionally walked into it again, learning nothing.

Defining new tags is great, even with new "top level tags" it's fine. Adding `healthcare=physiotherapist` was a clear win for OSM, duplicating `healthcare=hospital` was not.

> OSM top level tagging schemes are largely just historical in nature and have VERY loose criteria for what is meant to be addressed in each scheme. I can't see that `healthcare` and `education` are really any better defined than `amenity` -- which is essentially no clear definition.
> 
> OSM is full of random top level tagging scheme use such as:
> 
>     * Buy a service (hairdressing) from `shop/hairdresser`
> 
>     * Buy a service (car repair) from `shop/car_repair`
> 
>     * Buy a service (taxation advice) from `office/tax_advisor`
> 
>     * Buy a service (legal advice) from `office/lawyer`
> 
>     * Buy a service (medical advice) from `amenity/doctors` or `amenity/clinic`
> 
>     * Buy a service (Internet use) from `amenity/internet_cafe`
> 
>     * Have fun (theme park rides) from `tourism/theme_park`
> 
>     * Have fun (trampoline use) from `leisure/trampoline_park`
> 
>     * Have fun (watch movies) from `amenity/cinema`

Once upon a time some people also wanted `amenity=shop` + `shop=supermarket`. (`shop` not to be a "top level tag") :exploding_head:

Historical, is a great word to use here. I also wouldn't read too much into the literal words used in the key value pairs, most of the world doesn't read English. There is an argument that Wikidata with it's "P numbers", with descriptions in several languages is better than English words that may or may not mean what the dictionary says.

Another fun rabbit hole if you are bored is "OSM uses British English as interpreted by Germans."

> I assume `healthcare` and `education` were attempts to try and make sense of the random top level tagging schemes, but this hasn't actually happened, and instead they've just made tagging schemes even more random?

`education` is still too soon to call, but I have low expectations. `healthcare` seems to have confused more things that it's helped unfortunately.

Cj-Malone:
Going back to "Is there a need to output", the real argument for would be compatibility without ourselves. We have no real concept of breaking changes, but I see no real reason to intentionally break for the norm of last year, or the year before. Because it's there, lets leave it there. IDK how many consumers of ATP we have, but we don't need to make their lives harder.

Another argument for would be equality between NSI matched POIs (branded POIs) and none matched POIs (just apply_category).

The argument against is simple, duplicating tags is confusing, parsing multiple tags is confusing. Encouraging OSM complexity, and so drift (When a POI goes from pharmacy to a hairdresser, `amenity` is updated and too often `healthcare` is left.) is damaging OSM.

The arguments for `education` are similar, lets not overcomplicate the output, lets stick with what we were doing for years. We aren't adding `education` to branded locations via NSI, so we don't need to worry about branded POIs equality. Lets not encourage drift in OSM.

PR #17356, opened 2026-07-14
[Ridley's Family Markets US] Fix spider and rename

karen-kerr:
The hardcoded WordPress admin-ajax nonce went stale, so the store request returned nothing. Scrape a fresh nonce from the locations page first. Also set the category, use DisplayName for the branch, and rename to _us (US-only chain).

35 locations, previously 0.


karen-kerr:
Closing: the fetch fails because Cloudflare returns 403 on the admin-ajax POST from the CI egress IP. The nonce refresh in this PR is the correct code fix (the endpoint returns all stores from a clean IP), but the store-data endpoint is now behind Cloudflare bot-mitigation that blocks datacenter egress, so it can't be crawled without a residential/clean IP. Parking until that's viable.

PR #17360, opened 2026-07-14
[Barko Loans ZA] Fix spider

karen-kerr:
The `Barkomap.js.txt` endpoint now returns 404 after a site rebuild. Discover branches from the new `/Home/GetAllBranches` JSON API instead.

- Repoint to the new endpoint; drop the old province-keyed parsing and Google-URL coordinate extraction.
- New payload provides direct per-branch coordinates, email, town, and province.
- Apply `shop=money_lender` explicitly.

Local crawl now yields 224 branches with coordinates, phone, and email.

### DIFF
--- a/locations/spiders/barko_loans_za.py
+++ b/locations/spiders/barko_loans_za.py
@@ -1,23 +1,10 @@
-from typing import Iterable
+from typing import Any, Iterable
 
-from chompjs import parse_js_object
 from scrapy.http import Response
 
-from locations.dict_parser import DictParser
-from locations.google_url import url_to_coords
+from locations.categories import Categories, apply_category
 from locations.items import Feature
 from locations.json_blob_spider import JSONBlobSpider
-
-PROVINCES = {
-    "MPUMALANGA": "Mpumalanga",
-    "GAUTENG": "Gauteng",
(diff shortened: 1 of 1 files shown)


locations/spiders/barko_loans_za.py
```diff
@@ -1,53 +1,31 @@
...
 
 class BarkoLoansZASpider(JSONBlobSpider):
     name = "barko_loans_za"
     item_attributes = {
         "brand": "Barko Loans",
         "brand_wikidata": "Q118185897",
     }
-    start_urls = ["https://www.barko.co.za/Barkomap.js.txt"]
+    start_urls = ["https://www.barko.co.za/Home/GetAllBranches"]
     no_refs = True
```
Cj-Malone, line 17: looks like `branchOfficeNumber` can be used

karen-kerr: Good catch to check, but `branchOfficeNumber` is the branch phone number (e.g. `879805242`), not a unique identifier — it is already used for `item["phone"]`. The only `id` field in the payload is `0` for every branch, so there is no stable per-branch ref, hence `no_refs`. `branchName` is unique and goes into `branch`.

PR #17385, opened 2026-07-15
Gscaltex kr patch

chiho-kr:
- Set `branch` from `siteNm` (previously unused).
- Tag `self_service` from `selfYn`.
- Tag `fuel:lpg` from `lpgYn` using the existing `Fuel.LPG` enum.
- Tag Korean mobile payment methods (`kakaoPayYn`, `naverPayYn`, `paycoYn`, `zeroPayYn`) using new `PaymentMethods` enum values, since these payment methods are common across Korean fuel stations and other Korean POIs.

The upstream site exposes these fields but the spider was dropping them. Adding a `branch` value also lets multiple stations under the same brand be distinguished.

New `PaymentMethods` entries (`KAKAO_PAY`, `NAVER_PAY`, `PAYCO`, `ZERO_PAY`) were added rather than writing raw tags into `extras`, since these are likely to recur in other Korean spiders.

### DIFF
--- a/locations/categories.py
+++ b/locations/categories.py
@@ -746,6 +746,7 @@ class PaymentMethods(Enum):
     HUAWEI_PAY = "payment:huawei_pay"
     ID = "payment:id"
     JCB = "payment:jcb"
+    KAKAO_PAY = "payment:kakaopay"
     KUSTERS = "payment:kusters"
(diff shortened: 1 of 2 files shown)


Cj-Malone:
Thank you

Issue #16216, opened 2026-05-11
Pipeline step for Punycode normalization


brawer:
What would you think about a pipeline step for URL normalization? Specifically, expanding Punycode hostnames.

Currently, the field checker complains if the hostname of `website` or `image` is [Punycode](https://en.wikipedia.org/wiki/Punycode)-encoded. Personally I think it’s correct to complain, since Punycode is not human-readable (essentially, something between web browsers and the DNS system). However, I also wonder if it really should be up to each spider to know about and fix up Punycode.  Instead of just flagging Punycode hostnames as an error, how about pipeline stage that rewrites `https://www.xn--nni-hoa.ch/` to `https://www.nüni.ch/`?

I wanted to bring it up for discussion before implementing it.

Cj-Malone:
This is a topic I know next to nothing about. What's the OSM way, a bit of both?

iandees:
🤖 This is a feature request/discussion for a pipeline step to auto-decode Punycode hostnames in `website`/`image` fields (e.g. `xn--nni-hoa.ch` → `nüni.ch`). The idea has merit — it would remove the per-spider burden of knowing about Punycode and centralise the normalisation. An implementation would call `encodings.idna.decode()` (stdlib) on the hostname portion of the URL. Leaving open for community discussion and a potential PR.

PR #17398, opened 2026-07-16
[Körperformen] Fix spider

karen-kerr:
The `website` field carried the store URL's IDN host in punycode form (`xn--krperformen-rfb.com`), which failed URL validation and logged an error for every item. Decode the host back to Unicode (`körperformen.com`).

- Take `branch` from the studio name (strip the brand prefix).
- Apply `leisure=fitness_centre` explicitly.

Local crawl yields 275 studios (deduplicated from 550 listing rows) with coordinates, and no errors.


Cj-Malone:
We need to discus punycode #16216 but I'll take the rest of the fixes

PR #17442, opened 2026-07-21
[Suzuki DE] Fix spider

karen-kerr:
The dealer-search endpoint is behind CloudFront, which rejects the POST request the spider sent (403 error page, no dealers returned). Send the search as a GET with the parameters in the query string, which returns the dealer JSON. Add type hints.

### DIFF
--- a/locations/spiders/suzuki_de.py
+++ b/locations/spiders/suzuki_de.py
@@ -1,7 +1,8 @@
-from typing import AsyncIterator
+from typing import Any, AsyncIterator
+from urllib.parse import urlencode
 
 from scrapy import Spider
-from scrapy.http import FormRequest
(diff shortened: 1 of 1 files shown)


karen-kerr:
**Re: the dupe-filter warning** — expected, not a regression. The dealer-search API only returns the **20 nearest dealers to a point**: `radius` is ignored, `count` is capped at 20, and there's no pagination (`start`/`offset`/`page` all return the same 20). Full coverage therefore needs a per-city seed with downstream dedup, and since metros hold >20 dealers, overlapping queries are unavoidable. Trimming seeds to cut duplicates just loses dealers (237 seeds → 497 dealers; 102 → 455; 62 → 407), so the seeding is left as-is to keep all ~497.

PR #17448, opened 2026-07-21
[U.S. Bank] Fix spider

karen-kerr:
The location sitemap's URLs moved to a locations-*.us.bank-dns.com host that no longer resolves, and allowed_domains dropped every one, so the spider produced nothing. The store pages still serve on www.usbank.com, so rewrite the sitemap host back to it via sitemap_filter. Also return the OpeningHours object directly instead of pre-stringifying it.

### DIFF
--- a/locations/spiders/us_bank.py
+++ b/locations/spiders/us_bank.py
@@ -1,4 +1,5 @@
 import json
+from urllib.parse import urlsplit, urlunsplit
 
 from scrapy.spiders import SitemapSpider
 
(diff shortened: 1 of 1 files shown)


karen-kerr:
On the scraper-bot image-uniqueness warning: the branch-images turn out to be a small set of generic stock photos (bank with car park, drive-thru, interior with ATM) keyed to branch ids via unique URLs — not photos of the specific branch — so I've dropped the image field.

PR #17465, opened 2026-07-22
[One NZ] Fix spider

karen-kerr:
One store has no latlng key, so the unconditional feature["latlng"]["latitude"] raised KeyError and aborted the whole crawl (only 19 of 57 stores reached output). Guard the coordinate access with .get() so that store still yields with its address.


karen-kerr:
Two data notes:
- **Phone dropped for all stores** — nearly all list only the shared national `0800` number, which isn't per-store useful; rather than hardcode-filtering that one number, the field is dropped entirely.
- **The one store without coordinates (Westfield Albany) is a real store** — the source has no lat/lon for it, so it's kept with just its address.

PR #17487, opened 2026-07-22
[Volksbank AT] Fix spider

karen-kerr:
The finder endpoint sits under a robots-disallowed path and redirects to a subdomain whose robots.txt disallows everything, so the spider was blocked before its first request (0 items). Disable ROBOTSTXT_OBEY. Also drop the email, which is a per-region shared address rather than per-branch, and move the per-store branch name into branch so NSI fills the brand name.


karen-kerr:
Removed 14 items that produced invalid geometry: they have empty coordinates in the source because they're back-office / non-branch entities — business-customer teams, real-estate & leasing subsidiaries, the training academy, private banking, and self-service ATM lobbies — not retail branches. Skipping stores without coordinates leaves 163 branches, all with valid coordinates.

PR #17513, opened 2026-07-23
Add category to several spiders

ckpiper:
After the refactoring of apply_nsi_categories these spiders no longer match with NSI. Thus, I am adding categories to them. Some of them still do not match with NSI as NSI has none country sized locations. Like New York Public Library for example. But I thought they should still have a category.


locations/spiders/davita_us.py
```diff
@@ -31,4 +32,7 @@ def parse(self, response: Response, **kwargs: Any) -> Any:
             ] = f'https://davita.com/locations/{item["state"]}/{item["city"]}/{location["address1"]}--{location["facilityid"]}'.lower().replace(
                 " ", "-"
             )
+            apply_category(Categories.CLINIC, item)
+            apply_category(Categories.DIALYSIS, item)
```
davidhicks: I think only one of `Categories.CLINIC` or `Categories.DIALYSIS` should apply here?

davidhicks (changes requested):

locations/spiders/davita_us.py
ckpiper: I was curious about that. As NSI has both Dialysis and Clinic for a tag. But left as just clinic with the specialty.

davidhicks: Actually good catch here, this will be of interest to @Cj-Malone (NSI maintainer too) as he recently implemented the `TagDuplicator` pipeline in ATP which adds duplicate `healthcare` tags only at a late stage of output from ATP. The general intent I believe is for ATP to internally use a single top level tag for each type of feature.

In this case we have NSI tagging features with both `amenity=clinic` AND `healthcare=dialysis`. There is no use of `amenity=dialysis` in OSM. The question I have is whether ATP should internally stick with `amenity=clinic` or use the more specific `healthcare=dialysis` instead? It'd be worth us testing to make sure whatever is chosen, this spider correctly looks up and matches NSI on a single top level tag (despite NSI using top level tags for this brand).

ckpiper: I tested and whether you have the DIALYSIS or CLINIC it chooses the correct NSI Entry.

Cj-Malone: I do have things to say here, but I've been procrastinating.

Cj-Malone: I think NSI is wrong. I will raise it there, but I know there will be some turmoil. Dialysis isn't rendered by Carto (the default map on OSM.org) so sneakily double tagging is NSI caused a bunch of OSM to be rendered (incorrectly but still rendered). Removing those tags in NSI will end up with OSM data being inconsistent (it already is) and more confusion in the rendering.

Cj-Malone: Assuming Colton is correct and we currently match, I would merge this. But I would expect this to got moved to dialysis at some point in the future.

ckpiper: Okay. So should I change it to Dialysis? If not can someone give it a greenlight so I can merge?

Cj-Malone (approved):

KarenKerr-TomTom (approved):

iandees:
It looks like all the changes from this PR ended up in #17632.

PR #17518, opened 2026-07-24
[PacSun US] Fix spider

karen-kerr:
- The Demandware `Stores-FindStores` response was restructured: the `stores` array is now skeletal, with details moved into a rendered HTML fragment and coordinates into a separate `locations` JSON string, so the spider crashed with a KeyError and returned zero features.
- Merge the `locations` entries with each store's card in `storesResultsHtml`, and use a generic `Mozilla/5.0` user agent, which the store search endpoint serves without the PerimeterX challenge, removing the need for a proxy.

### DIFF
--- a/locations/spiders/pacsun_us.py
+++ b/locations/spiders/pacsun_us.py
@@ -1,40 +1,49 @@
-import re
-from typing import Iterable
+import json
+from typing import Any, Iterable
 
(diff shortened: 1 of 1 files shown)


karen-kerr:
Missing geometry is valid poi with full address and phone number.

PR #17525, opened 2026-07-24
[Lombard AU] Add spider for party goods stores (13 stores)

davidhicks:


Cj-Malone:
(how much effort is 13 stores worth)

iandees:
🤖 In `locations/spiders/lombard_au.py`, the `else` branch that processes regular store rows (building up `properties` with branch name, address, phone, and opening hours) never ends with a `yield Feature(**properties)` — so even once the site's 403 is resolved, no items would be produced for the non-header rows. Needs a final `yield Feature(**properties)` at the end of that branch.

Separately, the site currently returns HTTP 403 for `/site/pages/party-shop-finder` in CI, which may need `requires_proxy` or investigation independent of the code fix.

Note "Allow edits from maintainers" is off on this PR so I can't push the fix directly.

iandees:
Heads up: this PR's content already landed in `master` — it appears to have been accidentally included in an unrelated merge commit (`b22fb23d2`, "[5 Şarj TR] Add spider #17426"), rather than through this PR. `locations/spiders/lombard_au.py` is currently live in master with identical content to this branch.

Separately: confirmed the spider itself doesn't currently work — `www.lombard.com.au` returns a Cloudflare Turnstile challenge (403, `cf-mitigated: challenge`) on the store-finder page, and CI on this PR shows 0 items / "No output" for the same reason. Tried rewriting as a CamoufoxSpider with Turnstile click-solving; it couldn't find a solvable checkbox after several retries either, which looks like an IP-reputation block rather than a JS-only challenge.

Given the code's already in master (dead-on-arrival), this PR may be worth closing as superseded — but leaving that call to you since it's your PR.

PR #17601, opened 2026-07-29
Add categories to spiders

ckpiper:
After the refactoring of apply_nsi_categories these spiders no longer match with NSI. Thus, I am adding categories to them. Most of them still do not match with NSI for a couple different reasons. 

- dia_br - This does not match with NSI as it has location ambiguity of global (brand=Dia Market) and ar_br_es (brand=Dia). New code does not take into account the brand at all. Maybe as a last ditch resort it should. 
- animalis_fr, biogroup_fr, cigusto_fr, euromaster_fr, linvosges_fr all do not match even with category because fx is the included location set in NSI.
- ismash_gb, tops_pizza_gb, vancouver_coastal_health_ca, juice_press_us, doughnut_time_gb, naf_naf_grill_us, winemark_gb, cerise_et_potiron_fr, leyland_sdm_gb, foodland_grocery_us, cornish_bakery_gb, rush_hair_gb, longdan_gb all have NSI that have includes be parts of countries like state geojsons or very unique includes. Thus, matches do not happen in NSI so we include categories here.


davidhicks:
For `juice_press_us` the last run had `atp/nsi/location_unknown=18` in statistics, which is due to `nsi_entry_includes_location` not being able to extract the US state "NY" from NSI's `{"include":["us-ma.geojson","us-nv.geojson","us-ny-new_york_city.geojson"]}` for this brand.

Could this problem be solved in `nsi_entry_includes_location` by splitting the include/exclude geojson filenames by hyphens, and ignoring anything after the 1st hypen? If there is no hypen--a country code is specified, if there is one hypen--a state code is specified, if there are two hypens--a 3rd level subdivision (such as New York city in New York state) is present and this 3rd level subdivision should be ignored.

davidhicks:
For `rush_hair_gb` the problem is there is no second level subdivision of `ENG` defined for features extracted by ATP. NSI restricts to 2nd level subdivision per `{"include":["gb-eng"]}`.

Spiders might not set an ISO 3166-2 code for `state`, or might not set `state` at all. This is true for `rush_hair_gb` that does not extract a `state` value and just outputs `state=None`. ATP has `StateCodeCleanUpPipeline` which fires first before NSI matching is done, and is meant to reverse geolocate in the case of `state=None`, and is also meant to convert `England` to `ENG` if this is applicable (for other spiders, not applicable to `rush_hair_gb`). Currently this pipeline is only setup to work for Canada and the USA. It should be made generic to cover all countries and 2nd level ISO 3166-2 subdivision codes and names.

davidhicks:
For `dia_br` I agree, a new disambiguation at the bottom of `ApplyNSICategoriesPipeline.process_item` could check the `brand`/`operator` name of the ATP feature against the name of each of the multiple NSI items that have matched.

davidhicks:
NSI's `fx` location is France minus overseas territories. ISO 3166 is a messy situation for France and related overseas territories per https://en.wikipedia.org/wiki/ISO_3166-2:FR

For example, "French Guinea" (GF) is defined by ISO 3166-1 but Clipperton Island is included in "France" (FR).

We should check this situation for other states with overseas territories: Australia, New Zealand, Netherlands, UK, US, Finland, Denmark, China, Norway.

I think `ApplyNSICategoriesPipeline.process_item` is just going to have to implement a lookup table of odd cases where ATP is trying to implement ISO 3166-1 and ISO 3166-2, but ISO standards have some deficiencies that NSI has had to workaround with their own thing such as `fx`.

Cj-Malone:
Some of the excessive stuff can be sorted in [NSI](https://github.com/osmlab/name-suggestion-index/commit/06bd98475e64bb6643b369a62a96974d053f7b3c), some of the unnecessary stuff (fx, gb-eng) could be changed in vendor nsi. Some stuff may need code changes.

PR #17680, opened 2026-08-05
[Ministop JP] Fix spider

karen-kerr:
- Store data moved off the Next.js page props to a Canly locator API.
- Query it across two geohashes covering Japan (1,668 stores); handle the 24-hour stores.


Cj-Malone:
#17690

PR #17776, opened 2026-08-12
[Erciyes Anadolu Holding TR] Fix spider

karen-kerr:
The `GetProvBrFirms` API dropped the `EMailAddress` field, so the unconditional `pop("EMailAddress")` raised `KeyError` on every record and the spider produced no items.

Removed the email field, and set a `ref` from brand + coordinates (the API exposes no id field, so items were being collapsed by the duplicate filter). Now yields dealers across all three brands (İstikbal, Bellona, Mondi) with coordinates.

## Summary by CodeRabbit

* **Improvements**
  * Store listings now include more consistent branch names.
  * Brand information is applied more reliably to listings and categorization.
  * Store references are generated more consistently using brand and location details.
  * Store listing retrieval is faster, with reduced waiting between requests.
  * Store data processing is more reliable for supported locations.

### DIFF
--- a/locations/spiders/erciyes_anadolu_holding_tr.py
+++ b/locations/spiders/erciyes_anadolu_holding_tr.py
@@ -15,7 +15,7 @@ BRANDS = {
 
 class ErciyesAnadoluHoldingTRSpider(scrapy.Spider):
     name = "erciyes_anadolu_holding_tr"
-    custom_settings = {"DOWNLOAD_DELAY": 2}
+    requires_proxy = "TR"
     start_urls = ["https://brandapi.erciyes.com/api/ContactApi/GetCities"]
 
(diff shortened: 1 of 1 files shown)


karen-kerr:
Setting `requires_proxy = "TR"` here: the store API `brandapi.erciyes.com` is behind Cloudflare that rate-limits datacenter egress with HTTP 429s. From a residential IP the spider crawls all ~1,580 dealers (İstikbal / Bellona / Mondi) cleanly with no rate-limiting; from a datacenter IP the fetch hits 429s. I tested it through a residential proxy — requests succeed, items are returned, and the error ratio is 0 — so residential-proxy routing is required for reliable crawling. The fetch check will be empty until proxying is available.

PR #17780, opened 2026-08-12
[Tractor Supply US] Fix spider

karen-kerr:
The store-details endpoint (`/wcs/resources/store/...`) is Akamai-protected and now times out on every request, so the spider produced no items.

Rewrote against the store sitemap (`sitemap_stores.xml`), parsing each store page's embedded Next.js data. This still requires proxy routing (`requires_proxy = "US"`) to get past Akamai — but it uses the exact store list, ~2,474 requests, down from the previous ~5,868-request 25-mile centroid grid. Now yields ~2,474 US stores with coordinates and hours.

## Summary by CodeRabbit

* **New Features**
  * Store locations are now discovered through store sitemap pages.
  * Store listings include branch details, website information, and opening hours.
* **Bug Fixes**
  * Improved store data extraction from page metadata.
  * Opening hours now support split time ranges and correctly identify closed days.
  * Incomplete or unavailable opening-hours entries are handled more reliably.

### DIFF
--- a/locations/spiders/tractor_supply.py
+++ b/locations/spiders/tractor_supply.py
@@ -1,70 +1,50 @@
-from typing import AsyncIterator
+import json
+from typing import Any, Iterable
 
-from scrapy import Spider
-from scrapy.http import Request
+from scrapy.http import Response
+from scrapy.spiders import SitemapSpider
 
-from locations.geo import point_locations
+from locations.dict_parser import DictParser
 from locations.hours import OpeningHours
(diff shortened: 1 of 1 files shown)


karen-kerr:
Applied the closed-day handling defensively — `set_closed` is now called for a `Closed` value. Note: no Tractor Supply store currently reports a closed day (verified across 200+ stores, all open seven days with `H - H` hours), so this only takes effect if that ever changes.

PR #17807, opened 2026-08-13
[585 Gold RU] Fix spider

karen-kerr:
The sitemap no longer lists per-store pages (only the address index, which server-renders a single city), so the spider produced 0 features in recent runs.

Rewrote as a `JSONBlobSpider` against the JSON API the site's own frontend uses, which returns all stores in one request. Pawnshop-only points are skipped and counted; phones are dropped (only 6 unique numbers across the network — central lines), while per-store email, image, opening hours and full address are collected. `ROBOTSTXT_OBEY` is disabled for this spider: the API lives on a backend subdomain with a blanket `Disallow: /`, while being the data source for the site's public store locator.

## Summary by CodeRabbit

* **New Features**
  * Store listings are now retrieved directly from the provider’s data service.
  * Listings include normalized store details, the first available photo, opening hours, and the jewelry-shop category.
  * Lombard-only locations are excluded from results.
* **Bug Fixes**
  * Improved handling of Russian opening hours and store information.
  * Removed outdated sitemap-based extraction that could produce incomplete or inconsistent listings.

### DIFF
--- a/locations/spiders/zoloto_585_ru.py
+++ b/locations/spiders/zoloto_585_ru.py
@@ -1,27 +1,40 @@
-from typing import Any
+from typing import Iterable
 
-from scrapy.http import Response
-from scrapy.spiders import SitemapSpider
+from scrapy.http import TextResponse
 
+from locations.categories import Categories, apply_category
+from locations.hours import DAYS_RU, NAMED_DAY_RANGES_RU, OpeningHours
 from locations.items import Feature
-from locations.user_agents import BROWSER_DEFAULT
+from locations.json_blob_spider import JSONBlobSpider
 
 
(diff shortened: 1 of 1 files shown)


karen-kerr:
Regarding the review comments: the `addr_full` assignment was indeed dead — `addr_full` is not a key `DictParser` recognises, so the value never reached the item. Removed the line rather than remapping it: `addressFull` is just "г. {city}, {street}", a concatenation of the structured fields the spider already collects (street_address, city, postcode), so a redundant `addr_full` isn't needed.

The other two suggestions are intentionally not applied: `country` is derived by the pipeline from the spider-name suffix, and all 249 image URLs in the source are already unique, so an occurrence-count guard would be dead code here.

PR #17935, opened 2026-08-21
[NCP GB] Fix spider

karen-kerr:
Car park pages regularly take longer than the default 30 second navigation timeout, so 404 of 481 requests failed with a Playwright TimeoutError and the spider returned 39 car parks against 161 in the previous run.

The navigation timeout is now 60 seconds, as `rewe_de` already sets for the same reason. A live crawl makes 262 requests with 253 responses and no timeouts.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved page-loading reliability by allowing up to 60 seconds for navigation to complete.

### DIFF
--- a/locations/spiders/ncp_gb.py
+++ b/locations/spiders/ncp_gb.py
@@ -31,6 +31,7 @@ class NcpGBSpider(SitemapSpider):
         "ROBOTSTXT_OBEY": False,
         "USER_AGENT": BROWSER_DEFAULT,
         "DOWNLOAD_DELAY": 2,
+        "PLAYWRIGHT_DEFAULT_NAVIGATION_TIMEOUT": 60 * 1000,
     }
 
     def parse_pois(self, response):


karen-kerr:
Closing this for now, as the change cannot be verified here.

The spider combines Playwright with `DOWNLOAD_DELAY: 2`, so shutdown takes longer than the CI grace period. After the two minute cap fires, the run is sent SIGTERM before the feed is written, which is why the result shows no items. That is not caused by this change: measured locally, shutdown takes 36 seconds with the 60 second navigation timeout proposed here and 40 seconds with the 30 second default already on master, so the behaviour predates it.

The change itself does what it sets out to do. In the CI run above the spider crawled 46 pages and scraped 34 car parks with no navigation timeouts at all, against 404 of 481 requests failing with a Playwright TimeoutError in recent runs at the default timeout.

I will reopen once there is a way to demonstrate the improvement rather than assert it.

Issue #17952, opened 2026-08-22
check_spider_naming_consistency: respect skip_auto_cc_spider_name = True


davidhicks:
`ci/check_spider_naming_consistency.py` should respect the spider class attribute `check_spider_naming_consistency = True` and not try to capitalise country code suffixes if this attribute is set to True. Otherwise, spiders such as that proposed in PR #17950 will incorrectly have the "Co" part of their brand name treated as the country code "CO".

iandees:
So far the only case where I've seen this is when the spider is named "XAndCo" and the problem can be solved by expanding "Co" to "Company". I'd rather do that than add a workaround for the naming rule. If we run into another case that doesn't work this way, let's rethink.

PR #17953, opened 2026-08-22
Add skip_auto_cc_spider_name_check flag to spider naming checker

iandees:
`ci/check_spider_naming_consistency.py` auto-capitalizes a spider name's trailing components when they match a known country code (e.g. `teboil_fi` -> `TeboilFISpider`). This produces a false positive for spider names where a trailing component collides with a country code but isn't one, e.g. `basilic_and_co` -> `BasilicAndCOSpider` (treating "Co" as Colombia's ISO code), as seen in #17950.

This adds an opt-out class attribute, `skip_auto_cc_spider_name_check = True`, that a spider can set to skip the country-code capitalization step for its own name. This is a new, dedicated flag rather than reusing the existing `skip_auto_cc_spider_name` (which already controls a different, unrelated behavior — whether `CountryCodeCleanUpPipeline` auto-derives an item's `country` field from the spider name at runtime). Reusing that flag would have forced renaming several already-correctly-named spiders (`CrocsEUSpider`, `AsicsEUSpider`, `OrangeFRSpider`, etc.) that set it for the unrelated pipeline behavior.

Verified `ci/check_spider_naming_consistency.py` still passes with no errors against the current spider directory.

seen in #17952, #17950

### DIFF
--- a/ci/check_spider_naming_consistency.py
+++ b/ci/check_spider_naming_consistency.py
@@ -143,7 +143,7 @@ COUNTRYCODE_COMPONENTS = {
 }
 
 
-def snake_to_camel(snake_str: str) -> str:
+def snake_to_camel(snake_str: str, skip_auto_cc_spider_name_check: bool = False) -> str:
     components = snake_str.split("_")
 
     # If the first component is a number, spell it out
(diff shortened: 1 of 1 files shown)


iandees:
Closing per my comment on #17952: https://github.com/alltheplaces/alltheplaces/issues/17952#issuecomment-5469230333 — the only case seen so far is the "X And Co" pattern, which is better solved by renaming to "X And Company" than by adding a naming-rule workaround. If another case comes up that doesn't fit this fix, we'll revisit.

PR #17972, opened 2026-08-23
[Baby-Walz DE] Add proxy for Vercel security checkpoint

iandees:
- Every path on www.baby-walz.de, including /robots.txt, returns 429 with `x-vercel-mitigated: challenge` and a Vercel Security Checkpoint page — datacentre IPs are blocked outright.
- Added `requires_proxy = "DE"` and a browser UA, following the same fix used for `evgo_us` (also behind Vercel's checkpoint).
- Seen in #17471

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved access to location data by using a browser-like user agent and a German proxy.
  * Updated crawling behavior to prevent robots.txt restrictions from blocking data retrieval.

### DIFF
--- a/locations/spiders/baby_walz_de.py
+++ b/locations/spiders/baby_walz_de.py
@@ -8,6 +8,7 @@ from scrapy.http import Response
 from locations.categories import Categories, apply_category
 from locations.google_url import url_to_coords
 from locations.items import Feature
+from locations.user_agents import BROWSER_DEFAULT
 
 
 class BabyWalzDESpider(Spider):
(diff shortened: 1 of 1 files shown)


iandees:
Closing without merging — CI shows the proxy connected successfully (success_ratio 1.0) but still scraped 0 items. Likely still hitting the Vercel checkpoint's interstitial page (which needs JS execution to clear via a redirect/cookie, not just a different source IP) — Zyte's plain httpResponseBody won't render that. requires_proxy alone isn't sufficient here; would need Zyte's browserHtml feature or a Camoufox rewrite instead.

PR #17988, opened 2026-08-23
[Howard Hanna] Fix spider

iandees:
- The office map site is fully behind a Cloudflare managed challenge (`cf-mitigated: challenge`), including static HTML pages, so the previous plain `PlaywrightSpider` POST approach was blocked with 403s (seen in #17891).
- Rewrote as a `CamoufoxSpider` with `DEFAULT_CAMOUFOX_SETTINGS_FOR_CLOUDFLARE_TURNSTILE`, same pattern as `odeon_gb` and `northside_hospital_us`. The original `FormRequest` POST to `Office/MapOffices` is unchanged; Camoufox's fingerprint passes the challenge without needing a proxy.
- Verified locally: single request returns 200 with 475 offices.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Howard Hanna listing retrieval when Cloudflare protection or CAPTCHA challenges are encountered.
  * Added handling for HTTP 403 responses to improve scraping reliability.
  * Updated browser and request behavior to better support protected pages.

PR #17992, opened 2026-08-23
[Playwright/Camoufox] Wire proxy support into browser launch options

iandees:
## The bug

`DEFAULT_PLAYWRIGHT_SETTINGS` and `DEFAULT_CAMOUFOX_SETTINGS` in `locations/settings.py` fully replace `DOWNLOAD_HANDLERS`/`DOWNLOADER_MIDDLEWARES`, which drops `ZyteApiByCountryMiddleware` — the only thing that reads a spider's `requires_proxy` and applies proxy routing. Result: any `PlaywrightSpider`/`CamoufoxSpider` with `requires_proxy` set gets **zero** actual proxying — the browser launches straight from the CI host's real IP.

Most Playwright/Camoufox spiders with `requires_proxy` (hilton, tesco_gb, carrefour_pl, rei_us, etc.) have been unaffected in practice because their blocks are fingerprint-based and Camoufox's realistic fingerprint alone gets through — they don't actually need a different source IP. But it's the exact blocker for two spiders with genuine IP-reputation blocks:

- `grainger_us` — seen in #17772. PR #17982 added `requires_proxy = True`; CI still got 403 on all 77 requests, confirming `requires_proxy` is a no-op for Playwright spiders.
- `howard_hanna` — seen in #17891. PR #17988 rewrote it as a `CamoufoxSpider`, which worked perfectly from a residential IP (475 items) but CI's datacenter IP got served a harder, non-interactive Cloudflare Managed Challenge that the click-solver can't beat.

## What this does

Wires Zyte API's [proxy-mode endpoint](https://docs.zyte.com/zyte-api/usage/proxy-mode.html) (`api.zyte.com:8011`, HTTP CONNECT proxy, API key as Basic-auth username) into the actual browser launch, gated per-spider by `requires_proxy` so spiders that don't need it aren't silently routed through the proxy:

- `locations/playwright_spider.py`: `PlaywrightSpider` now has a `requires_proxy` attribute (mirroring `ZyteApiByCountryMiddleware`) and overrides `update_settings()` (a classmethod Scrapy calls once per crawl, before settings freeze and before the browser launches) to inject `{"server": ..., "username": <ZYTE_API_KEY>, "password": ""}` into `PLAYWRIGHT_LAUNCH_OPTIONS["proxy"]` when `requires_proxy` is set and `ZYTE_API_KEY` is present. `CamoufoxSpider` overrides a small `_proxy_launch_options_setting` class attribute to point the same logic at `CAMOUFOX_LAUNCH_OPTIONS` instead (Camoufox's `proxy` option is also launch-time, confirmed by reading `camoufox/utils.py`).
- `locations/middlewares/playwright_middleware.py`: `PlaywrightMiddleware.process_request` now also sets a `Zyte-Geolocation: <CC>` header via the browser context's `extra_http_headers` (`playwright_context_kwargs`/`camoufox_context_kwargs` request meta) when a country code resolves, reusing `get_proxy_location()` from `zyte_api_by_country.py` rather than reimplementing the country-suffix logic.
- Cost: both `DEFAULT_PLAYWRIGHT_SETTINGS`/`DEFAULT_CAMOUFOX_SETTINGS` already abort every non-"document" request client-side before it hits the network, so proxying is bounded to page-navigation requests only, not subresources — same rough cost profile as a plain `requires_proxy` REST spider.
- Test spiders: added `requires_proxy = True` to `grainger_us` and reapplied the closed-PR `CamoufoxSpider` rewrite (`#17988`) plus `requires_proxy = "US"` to `howard_hanna`.

Verified locally (no real Zyte account in this sandbox) that `update_settings()` correctly injects `proxy` into `PLAYWRIGHT_LAUNCH_OPTIONS`/`CAMOUFOX_LAUNCH_OPTIONS` for both spiders, is a no-op without `requires_proxy` or without `ZYTE_API_KEY`, and doesn't affect unrelated spiders — plus the full `pytest` suite and `pre-commit`.

## What's unverified — needs human sign-off before merge

**This cannot be verified locally — this sandbox has no Zyte API credentials.** Two things specifically need CI (which does have real Zyte credentials) to confirm:

1. That Zyte's proxy-mode endpoint actually works as a plain browser-launch proxy for Playwright/Camoufox at all (docs explicitly say "not optimized for use in combination with browser automation tools").
2. That the `Zyte-Geolocation` header sent via `extra_http_headers` is actually honored for HTTPS traffic tunneled through the CONNECT proxy — Zyte's docs don't document this for proxy-mode, and a plain CONNECT proxy is normally blind to encrypted request headers, so this may silently not select the intended country (falling back to Zyte's default IP selection) even if the proxy itself works.

If CI shows `grainger_us`/`howard_hanna` still 403ing or producing no output, that means the whole approach doesn't work as assumed here — please don't guess further on my end, just report the CI result back so we can decide next steps (e.g., dropping the geolocation header, or concluding proxy-mode genuinely isn't viable for browser automation).

Given the cost/architecture implications across the ~19 spiders currently setting `requires_proxy` on Playwright/Camoufox subclasses, **I'm not merging this myself even if CI is green** — flagging for explicit review.

### DIFF
--- a/locations/camoufox_spider.py
+++ b/locations/camoufox_spider.py
@@ -9,6 +9,9 @@ from locations.playwright_spider import PlaywrightSpider
 
 
 class CamoufoxSpider(PlaywrightSpider):
+    # See PlaywrightSpider._apply_zyte_proxy_launch_options().
+    _proxy_launch_options_setting = "CAMOUFOX_LAUNCH_OPTIONS"
+
     @staticmethod
     async def click_solver(page: Page, request: Request, spider: Spider) -> None:
         """
--- a/locations/middlewares/playwright_middleware.py
+++ b/locations/middlewares/playwright_middleware.py
@@ -106,10 +106,12 @@ class PlaywrightMiddleware:
             return
         if not os.environ.get("ZYTE_API_KEY"):
             return
-        requires_proxy = getattr(self.crawler.spider, "requires_proxy", False)
+        if not (spider := self.crawler.spider):
+            return
+        requires_proxy = getattr(spider, "requires_proxy", False)
         if not requires_proxy:
             return
-        if not (country_code := get_proxy_location(requires_proxy, self.crawler.spider.name)):
+        if not (country_code := get_proxy_location(requires_proxy, spider.name)):
             return
 
         context_kwargs = request.meta.setdefault(context_kwargs_meta_key, {})
--- a/locations/middlewares/playwright_middleware.py
+++ b/locations/middlewares/playwright_middleware.py
@@ -1,3 +1,5 @@
+import os
+
 from scrapy import Selector
 from scrapy.crawler import Crawler
 from scrapy.http import Request, Response, TextResponse
@@ -6,6 +8,7 @@ from scrapy_camoufox.page import PageMethod
 
 from locations.camoufox_spider import CamoufoxSpider
 from locations.captcha_solvers import click_solver
+from locations.middlewares.zyte_api_by_country import get_proxy_location
 from locations.playwright_spider import PlaywrightSpider
 
 
@@ -20,7 +23,9 @@ class PlaywrightMiddleware:
         return cls(crawler)
 
     def process_request(self, request: Request) -> None:
+        context_kwargs_meta_key = None
         if issubclass(type(self.crawler.spider), CamoufoxSpider):
+            context_kwargs_meta_key = "camoufox_context_kwargs"
             if "camoufox" not in request.meta:
                 request.meta["camoufox"] = True
             if "camoufox_page_event_handlers" not in request.meta.keys():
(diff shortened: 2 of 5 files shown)


iandees:
Closing without merging — CI confirms this doesn't work as implemented. Both test spiders (grainger_us, howard_hanna) still return 0 items, but for a new, more specific reason than the original bug:

- `NS_ERROR_PROXY_CONNECTION_REFUSED` on some requests — Firefox/Camoufox can't establish a connection to api.zyte.com:8011 at all.
- `SEC_ERROR_UNKNOWN_ISSUER` on others — Zyte's proxy mode does TLS interception with its own certificate, and the browser doesn't trust Zyte's CA, so the TLS handshake fails.

This matches Zyte's own documentation warning that proxy mode 'is not optimized for use in combination with browser automation tools.' A working fix would need the browser context configured to trust Zyte's MITM certificate (e.g. Playwright's `ignoreHTTPSErrors` context option, or explicit CA trust) plus resolving the connection-refused issue (possibly needs the HTTPS proxy interface at :8014 instead of :8011, or different auth handling) — genuinely uncertain without deeper Zyte-specific troubleshooting or contacting their support. The settings.py architecture (gating on requires_proxy, launch-time config via update_settings) is sound and reusable if/when the underlying TLS trust issue gets solved — just the actual proxy connection doesn't work yet.

PR #18097, opened 2026-08-24
[Ionna US] Fix spider

iandees:
- Ionna's store locator changed the pre-launch status text from \"Coming Soon\" to \"Opening Soon\". The spider's filter only matched the old text, so all 79 not-yet-open locations started flowing through.
- One of those locations has only \"4 NACS\" in its connectors spec (no CCS), which crashed \`specs.split(\" | \")\` with an unhandled \`ValueError\`. Since \`parse\` is a generator, that exception silently truncated the crawl (128 of 236 locations yielded, no error surfaced beyond one log line).
- Updated the skip filter to also match \"Opening Soon\", and rewrote the connector-count parsing with regex so it can't crash on a single-connector-type or reordered spec string.
- Verified locally: 236 total locations in the JSON blob, 79 filtered as not-yet-open, 157 yielded cleanly with no exceptions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Locations marked “Opening Soon” are no longer listed as available.
  * Connector counts are now parsed more reliably across different specifications, including entries with missing connector types.
  * Capacity and connector details are reported consistently for NACS and CCS connectors.

### DIFF
--- a/locations/spiders/ionna_us.py
+++ b/locations/spiders/ionna_us.py
@@ -1,4 +1,5 @@
 import json
+import re
 
 import scrapy
 
@@ -39,8 +40,9 @@ class IonnaUSSpider(scrapy.Spider):
         json_data = json.loads(locations)
 
         for location_id, location in json_data.items():
-            # Skip locations that are coming soon. They indicate this with "Coming Soon" in the note field.
-            if "Coming Soon" in location["note"]:
+            # Skip locations that are not yet open. They indicate this with "Coming Soon" or
+            # "Opening Soon" in the note field.
+            if "Coming Soon" in location["note"] or "Opening Soon" in location["note"]:
                 continue
 
             item = Feature(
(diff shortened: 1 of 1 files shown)


GA-Kevin-Codes:
Hey @iandees I don't have a solution off the top of my head but this weeks IONNA run returned 0 results.

iandees:
Root cause of the 0-result run: the site now indirects through `window.allLocations`. Previously the page had `var locations = {...huge JSON...};` followed by `for(var key in locations) {`; now it's just `var locations = window.allLocations;`, with the actual JSON assigned earlier in the page as `window.allLocations = {...};`. The old `extract_text_between(text, "var locations = ", "for(var key in locations) {")" now captures the literal string "window.allLocations;" instead of JSON, so `json.loads()` raises immediately on the first line of `parse()` — 0 items instead of a partial crawl.

Fix in #18770: locate `window.allLocations = ` directly and use `json.JSONDecoder().raw_decode()` to parse the object without relying on an end-of-string marker. Verified locally: 177 items from the live site.

---

PR #18099, opened 2026-08-24
[Marvins Building Materials US] Fix blank item name

iandees:
- Every item yielded by this spider shipped with a blank `name`, since `parse()` never set `item["name"]`.
- The brand (Q109395525) has no NSI entry, so there's no downstream fallback to backfill the name. Now sets `item["name"]` from the spider's `brand` attribute.


iandees:

PR #18100, opened 2026-08-24
[Hyundai UY] Fix blank item name for parts-only branches

iandees:
- \`item["branch"] = item.pop("name", None)\` never reassigns \`item["name"]\` afterward.
- NSI happens to backfill the missing name for \`shop=car\` and \`shop=car_repair\` from its global Hyundai entries, but not for \`shop=car_parts\`, so parts-only dealer branches shipped with a blank name. Now sets \`item["name"]\` explicitly from the brand so it's correct for all three category branches regardless of NSI coverage.


iandees:

PR #18102, opened 2026-08-24
[Nextra AU] Fix blank name in unmatched-brand fallback

iandees:
- The brand-prefix matcher in `post_process_item` only set `item["name"]` in the matched branch of the `for...else`.
- The `for...else` fallback (unmatched store name — not currently triggered by real data, but a latent bug) left `item["name"]` blank; it now backfills it from the raw unprefixed store name.


iandees:

PR #18106, opened 2026-08-24
[Kinleigh Folkard & Hayward GB] Fix missing name field

iandees:
- Every item shipped with a blank \`name\` because \`post_process_item\` explicitly set it to \`None\` after \`branch\` was already populated from the page's \`<h1>\`. There's no \`brand_wikidata\` for this spider so NSI can't backfill the name.
- Set \`item["name"]\` to the brand string instead.


iandees:

PR #18109, opened 2026-08-24
[Řeznictví RABBIT] Fix missing name field

iandees:
- Every item shipped with a blank \`name\` because it was never set in \`parse_store\` (only \`branch\` was populated). The brand's \`brand_wikidata\` (Q140309856) has no NSI matches, so NSI can't backfill the name either.
- Set \`item["name"]\` to the brand string.


iandees:

PR #18212, opened 2026-08-27
[Columbus Café] Drop "& co" from brand name

hagatopaxi:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved extraction of phone numbers from location details.
  - Improved handling of addresses with varying phone-label formats.
  - Preserved the complete cleaned address when separating address and phone information.
  - Improved handling of location records where address details cannot be fully separated.
  - Simplified address presentation for more consistent location information.

### DIFF
--- a/locations/spiders/columbus_cafe_fr.py
+++ b/locations/spiders/columbus_cafe_fr.py
@@ -11,7 +11,7 @@ from locations.pipelines.address_clean_up import clean_address
 
 class ColumbusCafeFRSpider(JSONBlobSpider):
     name = "columbus_cafe_fr"
-    item_attributes = {"brand": "Columbus Café & Co", "brand_wikidata": "Q2984582"}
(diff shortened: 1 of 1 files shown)


Cj-Malone:
`brand` needs to match [NSI](https://nsi.guide/index.html?t=*&k=amenity&v=cafe&tt=q2984582#columbuscafeandco-d3efe9) https://github.com/osmlab/name-suggestion-index/blob/26a1224837cee1aa3bd1b818e5c9381a5da9b212/data/brands/amenity/cafe.json#L1838

I don't typically recommend parsing addressees like this, it's too fragile. But I will merge it if you want and the `brand` is synced with NSI (I've no opinion on which way).

hagatopaxi:
The brand name has changed, sourced by [wikipedia](https://fr.wikipedia.org/wiki/Columbus_Caf%C3%A9_%26_Co). I remove the address parsing if you recommend it.

PR #18358, opened 2026-08-31
[99 Speedmart] Add spider

### LINKED ISSUE #7666: 99 Speedmart (MY)
### Brand name

99 Speedmart

Malaysia mini market chain

### Wikidata ID

Q62075061
https://www.wikidata.org/wiki/Q62075061
https://www.wikidata.org/wiki/Special:EntityData/Q62075061.json

### Store finder url(s)
https://www.99speedmart.com.my/Store
Official Url(s): https://www.99speedmart.com.my/
- [x] I have run the below and if anything in generated, added it.
- [ ] I have manually checked for a storefinder
- [ ] This is not a manufacturer listing outlets operating under their own brands / this will not create duplicates
`pipenv run scrapy sf --brand-wikidata=Q62075061 https://www.99speedmart.com.my/`
```
```

iandees:
Adds a spider for 99 Speedmart, a Malaysian minimart chain (closes #7666). The site's `wp-json/wp/v2/stores` REST API looks like a clean JSON source, but its `acf` address/maps fields are empty for roughly 80% of records even though the same stores render with full address and Google Maps data on the public `/store-locations/` listing page, so the spider scrapes that paginated listing (`?e-page-a09cddc=N`, 147 pages) directly instead.

Coordinates come from each store's "Maps" button link. Most are Google Maps "share this pin" links where the place name is itself a raw DMS coordinate rather than a geocoded address/business name, so extracting them follows the site publishing a coordinate it already has, not a live geocode; a `maps.app.goo.gl` short link is left unresolved rather than followed. A large batch of older stores use a legacy `maps.google.com.my/maps?q=N02.93,E101.31`-style link with inconsistent separators that the shared `url_to_coords()` helper can't parse, so that format is parsed directly in the spider. Verified locally: 3193 items, all unique refs, coordinates within Malaysia's bounding box, `brand` matches NSI exactly, and `addr:state` populated with the correct ISO 3166-2:MY code for all but one record. 22 records (short links plus a couple of corrupted coordinate strings in the source) are published without coordinates.

## Summary by CodeRabbit

- **New Features**
  - Added store listings for 99 Speedmart locations across Malaysia.
  - Listings include branch names, addresses, geographic coordinates, and supermarket categorization.
  - Malaysian state information is included where available to make locations easier to identify and filter.

### DIFF
--- a/locations/spiders/99_speedmart_my.py
+++ b/locations/spiders/99_speedmart_my.py
@@ -1,0 +1,128 @@
+import html
+import re
+from typing import Any, AsyncIterator
+from urllib.parse import unquote
+
+from scrapy import Request, Spider
+from scrapy.http import Response
+
+from locations.categories import Categories, apply_category
+from locations.google_url import url_to_coords
+from locations.items import Feature
+
+# A large number of older stores use a legacy "maps.google.com.my/maps?q=..."
+# link with a raw "N02.93163, E101.31520"-style decimal-degree pair (with an
+# inconsistent mix of separators/whitespace) that locations.google_url's
+# url_to_coords() cannot parse. It is still the site publishing a coordinate
+# pin it already has (not a geocoded address), so it's extracted directly.
+LEGACY_COORD_PATTERN = re.compile(r"q=\s*N?\s*(\d{1,3}\.\d+)\s*,?\s*E?\s*(\d{1,3}\.\d+)")
+
+MY_STATE_ISO_CODES = {
+    "johor": "MY-01",
+    "kedah": "MY-02",
+    "kelantan": "MY-03",
+    "melaka": "MY-04",
+    "negeri_sembilan": "MY-05",
+    "pahang": "MY-06",
+    "penang": "MY-07",
+    "perak": "MY-08",
+    "perlis": "MY-09",
+    "selangor": "MY-10",
+    "terengganu": "MY-11",
+    "sabah": "MY-12",
+    "sarawak": "MY-13",
+    "kuala_lumpur": "MY-14",
+    "labuan": "MY-15",
+    "putrajaya": "MY-16",
+}
+
+
+class NinetynineSpeedmartMYSpider(Spider):
+    name = "99_speedmart_my"
+    item_attributes = {"brand": "99 Speedmart", "brand_wikidata": "Q62075061", "name": "99 Speedmart"}
+    allowed_domains = ["99speedmart.com.my"]
+
+    def make_request(self, page: int) -> Request:
+        return Request(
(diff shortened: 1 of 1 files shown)


davidhicks (changes requested):
Looks like the main API of the storefinder page has been missed. An example for requesting page 1 is:

```
curl 'https://99speedmart.com.my/wp-admin/admin-ajax.php' \
  -X POST \
  -H 'Accept: application/json, text/javascript, */*; q=0.01' \
  -H 'Content-Type: application/x-www-form-urlencoded; charset=UTF-8' \
  -H 'X-Requested-With: XMLHttpRequest' \
  --data-raw 'action=jet_smart_filters&provider=epro-loop-builder%2Fstorelocation&query%5B__s_query%7Csearch%5D=&defaults%5Bhas_custom_pagination%5D=true&defaults%5Bpost_status%5D=publish&defaults%5Bpost_type%5D=stores&defaults%5Borderby%5D=post_date&defaults%5Border%5D=desc&settings%5Bwidget_id%5D=a09cddc&settings%5Bfiltered_post_id%5D=3404&props%5Bfound_posts%5D=1&props%5Bmax_num_pages%5D=1&props%5Bpage%5D=1'
```

This will return a HTML blob of store location details that gets appended to the list already on the page. Each time the user clicks "Load More" the next page gets requested from the API above, and the returned HTML appended.

There are 147 pages stated to exist at present per the API response.

I can't see any need to keep track of already-seen references within this spider. I think the LLM is getting confused with the 'New stores' list on the storefinder page, which is seemingly irrelevant if this raw underlying API is used instead per example above. ATP will automatically deduplicate features based on the `ref` attribute of each `Item` returned by the spider and spiders don't need to deduplicate by themselves.

iandees:
Pushed a fix switching to the store-locations page's Jet Smart Filters AJAX endpoint (`wp-admin/admin-ajax.php`, `action=jet_smart_filters`) instead of scraping the paginated HTML listing, per @davidhicks's review.

One correction to the example curl: `props[page]` turned out to be inert (verified live — sending it with different values returned identical content each time); the field that actually drives which page is returned is `defaults[paged]`. With that, pagination reads `pagination.max_num_pages` from the JSON response directly and each page returns a disjoint set of stores, so the manual `seen_refs` tracking is gone too — confirmed locally that ATP's own `DuplicatesPipeline` still catches the occasional cross-page duplicate on its own (4 dropped in a 524-item test run), same as it would for any spider.

Verified locally: 88+ items parse correctly across multiple pages with address, state, and coordinates intact, using only 1 request per page (down from a full HTML page load per page).

---

PR #17970, opened 2026-08-23
[Fust CH] Re-enable CH proxy for DataDome block

iandees:
- The store-finder page is fronted by DataDome (see CSP references to `datadome.co` / `captcha-delivery.com`). Requests from CI datacentre IPs get 403, while a residential-IP curl with any UA (including our own) returns 200 with the embedded `pointOfService` JSON intact.
- Adds `requires_proxy = "CH"` so requests route through a Zyte CH proxy, following the same pattern used for `livique_ch` and other DataDome/Akamai/Cloudflare-blocked sites.

seen in #17276

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved access to Fust Switzerland listings by routing requests through an appropriate Swiss connection.
  * Reduced request blocks caused by data-centre IP restrictions.

PR #18409, opened 2026-09-01
[Fust CH] Retest CH proxy for DataDome block

iandees:
- Retest of #17970: local curl with a real Chrome UA against the store-finder page succeeds (200, pointOfService JSON present) while a python-requests UA gets 403 — this looks like the CI data-centre IP hitting DataDome, not a hard site-side block.
- #17970 previously failed with 8/8 requests through the CH proxy returning HTTP 520 (Zyte temporary-error), which looked like a Zyte-side infra issue rather than the site rejecting the proxy. Re-enabling `requires_proxy = "CH"` here to see if that has since cleared. Cross-ref seen in #17581 (same 520 signature via US proxy, same session).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved access to Swiss location data by routing requests through an appropriate connection when needed.
  * Reduced the likelihood of blocked or incomplete location results.


iandees:
CI failed, but not with the previous #17970/#17581 signature. This run's log shows `finish_reason: zyte_api_suspended_account` — Zyte API returned 403 `auth/account-suspended` on both requests ("Your account has been suspended... check your spending limit"), i.e. the org's Zyte credits are currently exhausted, not the earlier `status_codes/520` temporary-error pattern. This retest is inconclusive as a result. Leaving open to rerun once Zyte billing/credits are restored.

Issue #17415, opened 2026-07-17
Modernise stealth browser integration


davidhicks:
Some recent developments in stealth browser tech has finally arrived in recent days to replace the ancient and now ineffective Camoufox integration in ATP.

The modern approach now available is:

* Drop dependency 'scrapy-camoufox' that is no longer maintained.
* Use new 'scrapy-playwright' stealth/alternative browser support examples at https://github.com/scrapy-plugins/scrapy-playwright/blob/main/docs/pluggable-browser-providers.md to integrate ATP with scrapy-playwright and allow use of stealth browsers such as Camoufox.
* Add 'camoufox' package as dependency for managing Camoufox version upgrades.
* Use latest Camoufox engine and test various spiders which depend on Camoufox.
* Test 'playwright-captcha' for compatibility/effectiveness and research whether there are any newer "click solvers" available.
* Potentially integrate/test other stealth browser available for use with 'scrapy-playwright'.
* Consider whether a rotating proxy provider could be integrated in place of Zyte (unless Zyte offers such?) for use of residential and country-specific proxies.

PR #18416, opened 2026-09-01
Wire Zyte proxy into Playwright/Camoufox browser launch via a local relay

iandees:
## The bug

`DEFAULT_PLAYWRIGHT_SETTINGS`/`DEFAULT_CAMOUFOX_SETTINGS` in `locations/settings.py` fully replace `DOWNLOAD_HANDLERS`/`DOWNLOADER_MIDDLEWARES`, which drops `ZyteApiByCountryMiddleware` — the only thing that reads a spider's `requires_proxy` and applies proxy routing. Setting `requires_proxy` on a `PlaywrightSpider`/`CamoufoxSpider` subclass today does nothing; the browser launches straight from the crawling host's real IP. That's the reason spiders blocked by IP-reputation-style checks (Cloudflare/DataDome/Akamai) can't currently be fixed just by routing them through Zyte's proxy the way a plain HTTP spider can — seen in #17772, #17147, #17886, #17884.

A previous attempt at this (#17992, closed unmerged) wired Zyte's proxy-mode endpoint directly into `PLAYWRIGHT_LAUNCH_OPTIONS`/`CAMOUFOX_LAUNCH_OPTIONS["proxy"]` with `username`/`password` credentials, but every request then failed with `NS_ERROR_PROXY_CONNECTION_REFUSED`.

## Root-caused, not just retried

I reproduced `NS_ERROR_PROXY_CONNECTION_REFUSED` locally with a throwaway authenticated CONNECT proxy — no Zyte account involved. It's a genuine Firefox bug: an *authenticated* proxy (`username`/`password`, or credentials embedded in the proxy URL) combined with `route.continue_(headers=...)` breaks every navigation. This isn't an edge case for us — `scrapy_playwright`/`scrapy_camoufox` call `route.continue_(headers=...)` **unconditionally on every single request** (to keep the browser request in sync with the original Scrapy `Request`), so any authenticated proxy would hit this 100% of the time for every ATP Playwright/Camoufox spider. Isolated test matrix (Firefox via Playwright, local proxy):

| proxy | route.continue_() | result |
|---|---|---|
| unauthenticated | no override | success |
| unauthenticated | `headers=` override | success |
| authenticated (user/pass fields) | `headers=` override | `NS_ERROR_PROXY_CONNECTION_REFUSED` |
| authenticated (creds in URL) | `headers=` override | `NS_ERROR_PROXY_CONNECTION_REFUSED` |
| authenticated (user/pass fields) | no override | success |

So the fix isn't about credential formatting or the `:8011` vs `:8014` endpoint — it's that the browser can never be given proxy credentials directly.

## What this PR does

**`locations/middlewares/zyte_proxy_relay.py` (new)**: a small local HTTP CONNECT proxy, run on a background thread with its own asyncio loop. The browser is pointed at `127.0.0.1:<ephemeral port>` with *no* credentials (the row that works above); the relay does the real authenticated CONNECT handshake to `api.zyte.com:8011` on the browser's behalf and pipes bytes through unmodified after that. It's started idempotently, once per `ZYTE_API_KEY`, from `PlaywrightSpider.update_settings()`.

**`locations/playwright_spider.py`**: adds `requires_proxy` (mirrors `ZyteApiByCountryMiddleware`, `bool | str`) and a `update_settings()` classmethod hook that starts the relay and injects `{"server": "http://127.0.0.1:<port>"}` into `PLAYWRIGHT_LAUNCH_OPTIONS["proxy"]` — gated strictly on `requires_proxy` and `ZYTE_API_KEY` being set, so spiders that don't opt in are completely unaffected (no cost, no behavior change).

**`locations/camoufox_spider.py`**: three-line override pointing the same logic at `CAMOUFOX_LAUNCH_OPTIONS` instead.

**`locations/middlewares/playwright_middleware.py`**: when `requires_proxy` is set, also sets `ignore_https_errors=True` and a `Zyte-Geolocation` header on the browser context. Zyte's proxy-mode terminates HTTPS itself (needed to inject its own headers/controls), presenting a cert signed by its own CA rather than the destination site's — Firefox doesn't trust that CA, so without this every HTTPS request through the proxy fails with `SEC_ERROR_UNKNOWN_ISSUER`/`SSL_ERROR_UNKNOWN`. Zyte does publish a CA cert that could be installed instead (https://docs.zyte.com/misc/ca.html), which would be the more rigorous fix, but Camoufox launches a fresh ephemeral Firefox profile per run, so there's no persistent trust store to install into — it'd need re-provisioning into every profile at launch. `ignore_https_errors` is the pragmatic tradeoff: scoped to just this context, and only applied when a spider has opted into `requires_proxy`, so it never weakens cert checking for anything else. Flagging this tradeoff explicitly for review.

## What I verified locally

- Reproduced the Firefox authenticated-proxy bug and the relay workaround with a throwaway local CONNECT proxy (table above).
- Verified `ignore_https_errors=True` fixes untrusted-cert failures against a local self-signed HTTPS server (Firefox: `SSL_ERROR_UNKNOWN` without it, success with it).
- Ran full `CrawlerProcess` crawls of a `PlaywrightSpider` and a `CamoufoxSpider`, each with `requires_proxy = True`, `ZYTE_API_KEY` set, and the relay's upstream target monkeypatched to my local test proxy instead of `api.zyte.com` — both completed a real navigation and yielded an item through the whole `update_settings()` → relay-start → launch-options → `PlaywrightMiddleware` chain.
- `uv run ty check ./locations/ --exclude 'locations/spiders/'`, `uv run pytest` (314 passed), and `uv run pre-commit` all pass.
- No spider files are touched — this PR is scoped to the proxy-wiring infrastructure only.

## What's unverified — needs a human with real Zyte credits

The org's Zyte account (185266) was out of credits while building this, so I couldn't test against the actual `api.zyte.com:8011` endpoint or its real TLS-interception certificate. To confirm this once credits are back, add `requires_proxy = True` to `grainger_us` (currently 403ing on every request — #17772) and run it in CI:

- **Success**: real store items come back instead of 403s.
- **Still broken, connection-level**: `NS_ERROR_PROXY_CONNECTION_REFUSED` — would mean Zyte's real endpoint behaves differently from my local proxy stand-in in some way this didn't anticipate (e.g. it drops the connection for another reason at scale).
- **Still broken, cert-level**: `SEC_ERROR_UNKNOWN_ISSUER`/`SSL_ERROR_UNKNOWN` despite `ignore_https_errors=True` — would mean the context option isn't reaching the right place, worth re-checking `playwright_context_kwargs`/`camoufox_context_kwargs` are actually applied at context-creation time for that spider's requests.
- **`zyte_api_suspended_account` / `403 account-suspended`** — not a signal on this code, just means credits are still out.

Given the cost/architecture implications across the spiders that already set `requires_proxy` on Playwright/Camoufox subclasses, I'm not merging this myself even if CI is green — flagging for explicit review.

## Summary by CodeRabbit

- **New Features**
  - Added optional Zyte proxy support for Playwright and Camoufox browser spiders.
  - Proxy settings can be enabled per spider and applied automatically when configured.
  - Added country-specific proxy geolocation and secure HTTPS handling.
  - Improved proxy error reporting, including upstream account and connection failures.
  - Updated supported browser-based crawlers to use shared proxy configuration automatically.


davidhicks:
Also see #17415 which is a need/proposal to modernise stealth browser integration with ATP, including upgrade to latest public Camoufox (or possibly other stealth browser of user choice).

It'd be great to have this proxy support too.

One question for this PR: can Zyte accept a hint to use a residential and/or data centre IP, or is Zyte smart enough to try cheap data centre IPs first, then try residential IPs (expensive) only if necessary?

PR #18484, opened 2026-09-04
[docomo_shop_jp] use the shop list API instead of the sitemap

### LINKED ISSUE #18463: docomo_shop_jp spider does not get full set of stores.
### Spider name

docomo_shop_jp

### Log output

https://alltheplaces-data.openaddresses.io/runs/2026-08-29-13-32-18/logs/docomo_shop_jp.txt

### Backtrace (if applicable)

_No response_

There are only 456 stores grabbed.

There should be (at the time of writing) 2064 stores.

### Possible causes?

The sitemap used does not list all stores. So additional sitemaps with the rest of the stores should be found and added to the spider, if they exist.

Alternatively, a different approach like starting here: https://shop.smt.docomo.ne.jp/result.html?page=1 and then going through each page of results may work better.

chiruu12:
Closes #18463.

The spider crawled `sitemap.xml`, which lists 517 shop pages and yields 456 items. The site's own shop list API reports 2064.

```
before   456 features
after   2064 features
```

Every one of the 456 refs the sitemap version produced is still present. `shop_code + add_shop_code` is the id the detail page is addressed by, so refs are unchanged and nothing churns.

## What changed beyond the count

Comparing the two runs across the 456 shared shops:

| field | differences |
|---|---|
| name, branch, phone, website | 0 |
| coordinates | 0 (largest delta 4.5e-12 degrees, float representation only) |
| opening_hours | 24 |

All 24 are corrections.

22 are shops whose weekly closing day the HTML parser missed entirely. The site states it as `毎週火曜`, and those shops were being published as open seven days:

```
0134108700900   before  Mo-Su 09:00-18:00
                after   Mo 09:00-18:00; Tu closed; We-Su 09:00-18:00
```

The other 2 are the weekend-hours heuristic getting it wrong rather than missing it. The old code applied alternate hours only when the text named Saturday and Sunday together:

```
0400501214800   site says 毎週金曜・毎週土曜・毎週日曜 10:00-21:00
                before  Mo-Fr 10:00-20:00; Sa-Su 10:00-21:00     (Friday dropped)
                after   Mo-Th 10:00-20:00; Fr-Su 10:00-21:00

0601600348100   site says 毎週日曜 11:00-20:30
                before  Mo-Su 11:00-21:00                        (Sunday lost)
                after   Mo-Sa 11:00-21:00; Su 11:00-20:30
```

The API separates the fields the HTML page ran together, which is what removes the guessing:

- `shophour_start1`/`end1` are the shop's hours; `shophour_start2`/`end2` with `shophour_modify2` are genuine alternate hours for named days (65 shops).
- `shophour_name4` is a `受付時間` reception cut-off rather than the hours the shop is open (306 shops). The old spider ignored it by heuristic; it is now ignored by name.
- `shop_holiday_code` is three four-character groups: last character the weekday, the leading three the ordinals it applies to. `6003` is every Wednesday, `2303` the second and third.

I decoded that field and checked it back against the site's own `定休日` text for all 2064 records, and it reconstructs exactly, so the reading is not a guess from a handful of samples.

Only the every-week form is emitted. 1159 shops close on something like the second Wednesday, which `OpeningHours` cannot express, and encoding it would overstate it as every Wednesday. Those keep full hours, as before.

`祝日` (public holidays) appears in the alternate-hours field and is also left unencoded, since there is no day for it.

## Coordinates

Latitude and longitude arrive as degrees, minutes, and thousandths of a second. The conversion agrees with the `data-lat`/`data-lng` the detail page carries:

```
page  43.15565 / 141.39756111111
api   43.15565 / 141.397561
```

## Notes

`DictParser` is not used here. On this payload it returns three fields and one is wrong: it maps `address` onto `addr_full`, but that field holds only the street number (`5-1-10`), with the prefecture and municipality in separate keys.

`phone` is `customer_tel` only. `free_tel` is a regional 0120 number shared between shops, so it is not a per-location value. Four shops publish no `customer_tel` and now carry no phone; all four are among the newly reachable shops, so nothing that exists today loses a number.

Three of the 2064 are not ドコモショップ: two iPhone repair corners and one ハーティプラザ, marked by `shop_category` 3 and 4. They are docomo-operated and the sitemap crawl included them, so I have left them in. Happy to filter on `shop_category == 1` if you would rather they were out.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved collection of Docomo shop information using structured shop-search data.
  * Improved shop location accuracy by converting coordinates to decimal format.
  * Improved opening-hours and regular-closing-day information.
  * Added support for retrieving shop listings across multiple result pages.
  * Corrected handling of invalid late-night time values to prevent inaccurate opening-hour displays.

### DIFF
--- a/locations/spiders/docomo_shop_jp.py
+++ b/locations/spiders/docomo_shop_jp.py
@@ -113,7 +113,9 @@ class DocomoShopJPSpider(Spider):
         if len(hhmm) != 4 or not hhmm.isascii() or not hhmm.isdigit():
             return None
         hours, minutes = int(hhmm[:2]), int(hhmm[2:])
-        return f"{hours:02d}:{minutes:02d}" if hours <= 24 and minutes < 60 else None
+        if minutes > 59 or hours > 24 or (hours == 24 and minutes):
+            return None
+        return f"{hours:02d}:{minutes:02d}"
 
     @staticmethod
     def weekly_closures(holiday_code: Any) -> Iterable[str]:
--- a/locations/spiders/docomo_shop_jp.py
+++ b/locations/spiders/docomo_shop_jp.py
@@ -1,137 +1,135 @@
 import re
-from typing import Any, Iterable
+from typing import Any, AsyncIterator, Iterable
 
-from scrapy.http import Response
-from scrapy.spiders import SitemapSpider
+from scrapy import Spider
+from scrapy.http import JsonRequest, Response
 
 from locations.categories import Categories, apply_category
-from locations.hours import DAYS, OpeningHours
+from locations.hours import DAYS, DAYS_JP, OpeningHours
 from locations.items import Feature
 
-DAY_CHARS = {"月": "Mo", "火": "Tu", "水": "We", "木": "Th", "金": "Fr", "土": "Sa", "日": "Su"}
-FULLWIDTH_DIGITS = str.maketrans("０１２３４５６７８９", "0123456789")
+API_URL = "https://shop.smt.docomo.ne.jp/api/shopsearch/shoplist"
+PAGE_SIZE = 30
 
-HOURS_RE = re.compile(r"(午前|午後)(\d{1,2})時(?:(\d{1,2})分)?[~〜～-](午前|午後)(\d{1,2})時(?:(\d{1,2})分)?")
+# Weekday names inside shophour_modify2. "祝日" (public holidays) also appears there and is
+# skipped, since OpeningHours has no day for it.
+ALTERNATE_DAY_RE = re.compile(r"毎週([月火水木金土日])曜")
+
+WEEKLY_ORDINAL = 6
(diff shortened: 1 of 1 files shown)


chiruu12:
Good catch, fixed in 26f591e.

`clock_time` accepted any minute value at hour 24, so `"2430"` became `"24:30"` and raised out of `add_range`:

```
ValueError: time data '24:30' does not match format '%H:%M'
```

24 is now allowed only at `24:00`, which `add_range` rewrites to `23:59` as a closing time. `"2430"`, `"2459"` and `"2500"` all return `None`.

No shop in the current data hits it: all 2064 records carry a valid four-digit `HHMM`. It matters because the guard exists for the case where one day they do not, and as written it would have taken the rest of that page of 30 shops with it.

Issue #16413, opened 2026-05-26
Spiders whose locations need to be converted to WGS84


j9d3it:
Related: #15640 
Some spiders have their coordinates in a different coordinate system which makes them appear offset from their actual location. The ones below in Japan have a different coordinate system (most likely Tokyo Datum, EPSG:4301 or similar). These need to be converted to WGS84.

- [ ] All Area Marker spiders (japan_post_jp, kuroneko_jp, lawson_jp, lawson_bank_jp, etc.)

The existing spiders have their coordinates in Tokyo Datum (EPSG:4301) so they can possibly be made into a Area Marker storefinder that converts the coordinates from that into WGS84 (EPSG:4326) using pyproj transformer.

- [ ] kirei_jp

This is also EPSG:4301.

iandees:
🤖 Investigated the coordinate systems used by the named spiders:

- `lawson_jp`, `lawson_bank_jp`, `japan_post_jp`, `kuroneko_jp`: These all use the Zenrin/e-map.ne.jp API (`e-map.ne.jp` or `map.japanpost.jp` etc.), **not** the AreaMarker API. Sample coordinates from the latest CI output for `lawson_jp` were reverse-geocoded and correctly placed in Japan (e.g. a Nikaho, Akita store at `39.214°N, 139.905°E` matched the expected address). The e-map.ne.jp API appears to return coordinates in a datum close to WGS84 (JGD2011 difference from WGS84 is <1m).

- `kirei_jp`: Uses the `dnpphoto.jp` API. Sample coordinates (`36.063°N, 138.055°E`) placed correctly in the Okaya/Nagano area, consistent with WGS84.

- `seria_jp`: No spider exists yet (see #16156). When implemented, the AreaMarker API reportedly offers a `datum=WGS84` parameter, so requesting WGS84 coordinates should be straightforward.

If there is a specific spider where misplaced points are observed in practice (with examples), please add a comment with the spider name and affected ref IDs so we can investigate further. Leaving open pending more specific evidence.

j9d3it:
The ones in Tokyo Datum need to be converted using pyproj. I think that's the coordinate system being used. Edit: I tested it using an online converter https://epsg.io/transform#s_srs=4301&t_srs=4326&x=138.3981661&y=35.8316661 and it seemed to work well. I'll have to test others, maybe those will be closer.

PR #18518, opened 2026-09-06
[japan_post_jp] Update spider to include new postboxes (201034 locations)

### LINKED ISSUE #17076: [Japan Post JP] add postboxes, possibly fix offset issue.
https://map.japanpost.jp/smt/search/nmap.htm?tempo=1&post=1&pltype=1&cond152=COL_09%3At&lat=124272024&lon=488205427

The new site for searching for post offices and ATMs on mobile also includes postboxes and the background is Google Maps which hopefully means that we can get a useful lat/lon out of it.

This is an example URL for the request: https://map.japanpost.jp/smt/search/zdcemaphttp.cgi?target=http://127.0.0.1/cgi/nkyoten.cgi?&key=41nQhzBGnwsvBynAXvBlnAof9MmgYPBlidC5SaXurmA7fB6zTbrxmzTZrxDzTzngjzFtmwIXFnmgrzFhmAbPTo&cid=search&postcid=searchPO&search_tempo=1&search_post=1&opt=search&pos=1&cnt=100&enc=UTF-8&lat=124272024&lon=488205427&latlon=124248729,488133193,124295318,488277661&jkn=(((COL_17:t AND (COL_19!:@@NULL@@ OR COL_27!:@@NULL@@ OR COL_36!:@@NULL@@))) AND (COL_09:t))&postjkn=&rad=50000&knsu=50&postknsu=50&exkid=&cust=&polycol=&exarea=&PARENT_HTTP_HOST=map.japanpost.jp&zdccnt=1&enc=EUC&encodeflg=0&NOREF=1

I would paste the example data here but it's probably Shift-JIS and then converted to UTF-8 by the browser data viewer so it's a garbled mess. It doesn't work to paste the request URL as-is into the browser as you don't get any data back.

I don't know what the encoding is for the lat/lon in the browser URL or the request URL. If anyone can get more insights for anything that would be helpful.

shuuji3:
resolve #17076 

The left 4 small circle is the intended process in `_subdivide()`. Since there's much more postboxes exist in a dense city area, the API still truncates additional locations when the limit in the current 5.5km city circle exceeds. So when truncates happens, this calls additonal 4 API calls with small circle areas which covers the original 5.5km circle.

two diagrams. right is green pixel green single circle checking if the city-5.5 circle is covered by 4 sub circles. left is five circles. large city circle and four small circles, union of them covering entire city circle

I'm still checking logs but this is the visualization of the API call results, each circle is correnponding a single API call.

circles_map, japan map with many small circles with three types

## Summary by CodeRabbit

- **Improvements**
  - Improved Japan Post location accuracy by converting source coordinates to WGS 84.
  - Expanded coverage in areas reaching search result limits through automatic subdivision into smaller search regions.
  - Improved handling of paginated search results for more complete location coverage.
  - Added collection schedules and postbox design details to applicable location records.
  - Improved operator information with linked Wikidata identifiers where available.
  - Added clearer query labels and search statistics.

### DIFF
--- a/locations/spiders/japan_post_jp.py
+++ b/locations/spiders/japan_post_jp.py
@@ -1,4 +1,5 @@
 import csv
+import math
 from io import StringIO
 from urllib.parse import urlencode
 
@@ -18,7 +19,7 @@ MAP_ID = "search"
 class JapanPostJPSpider(Spider):
     name = "japan_post_jp"
 
-    def make_request(self, lat, lon, radius, offset=1, count=900, tempo_count=0, post_count=0):
+    def make_request(self, lat, lon, radius, offset=1, count=900, tempo_count=0, post_count=0, source=""):
         params = {
             "cid": MAP_ID,
             "postcid": "searchPO",
@@ -54,17 +55,18 @@ class JapanPostJPSpider(Spider):
                 "count": count,
                 "tempo_count": tempo_count,
                 "post_count": post_count,
+                "source": source,
             },
         )
 
     async def start(self):
         radius_m = RADIUS_KM * 1000
         for lat, lon in country_iseadgg_centroids("JP", RADIUS_KM):
-            yield self.make_request(lat, lon, radius_m)
+            yield self.make_request(lat, lon, radius_m, source="grid-24")
         for city in city_locations("JP", 200000):
-            yield self.make_request(city["latitude"], city["longitude"], 5500)
+            yield self.make_request(city["latitude"], city["longitude"], 5500, source="city-5.5")
 
-    def parse(self, response, lat, lon, radius, offset, count=900, tempo_count=0, post_count=0):
+    def parse(self, response, lat, lon, radius, offset, count=900, tempo_count=0, post_count=0, source=""):
         # response is an EUC-encoded JS file that looks like
         #   ZdcEmapHttpResult[1] = '...';
         # where the string body is a TSV
@@ -83,14 +85,19 @@ class JapanPostJPSpider(Spider):
 
         page = (offset - 1) // count + 1
         self.logger.info(
-            f"Query (lat={lat}, lon={lon}, page={page}, offset={offset}, rec={rec_count}, hit={hit_count}, tempo={tempo_total}, post={post_total})"
+            f"Query (source={source}, lat={lat}, lon={lon}, radius={radius}, page={page}, offset={offset}, rec={rec_count}, hit={hit_count}, tempo={tempo_total}, post={post_total})"
         )
         if tempo_total >= MAX_ITEMS or post_total >= MAX_ITEMS:
-            self.logger.warning("Maximum number of items returned in one query, consider lowering the radius")
+            self.logger.warning(
+                f"Maximum number of items returned in one query, consider lowering the radius  (source={source})"
+            )
+            if source == "city-5.5":
+                yield from self._subdivide(lat, lon, radius, source)
+            return
 
         if offset + rec_count < hit_count:
             yield self.make_request(
-                lat, lon, radius, offset + rec_count, tempo_count=tempo_total, post_count=post_total
+                lat, lon, radius, offset + rec_count, tempo_count=tempo_total, post_count=post_total, source=source
             )
 
         for row in rows:
--- a/locations/spiders/japan_post_jp.py
+++ b/locations/spiders/japan_post_jp.py
@@ -9,7 +9,8 @@ from locations.categories import Categories, apply_category
 from locations.geo import city_locations, country_iseadgg_centroids
 from locations.items import Feature
 
-MAX_ITEMS = 1640  # determined experimentally
+# determined experimentally. per-type cap (TEMPO/POST). a type reaching this is truncated
+MAX_ITEMS = 1640
 RADIUS_KM = 24
 MAP_ID = "search"
 
@@ -21,16 +22,23 @@ class JapanPostJPSpider(Spider):
         params = {
             "cid": MAP_ID,
             "postcid": "searchPO",
+            # include TEMPO (post offices + ATMs + kanpo insurance)
             "search_tempo": "1",
+            # include POST (postboxes)
             "search_post": "1",
             "opt": "search",
+            # starting row (1-based). increased by rec_count to paginate
             "pos": offset,
+            # page size (rows per response). does not limit the total
             "cnt": count,
             "enc": "EUC",
             "lat": lat,
             "lon": lon,
+            # cap on TEMPO rows for this whole query
             "knsu": MAX_ITEMS,
+            # cap on POST rows for this whole query
             "postknsu": MAX_ITEMS,
+            # search radius in metres
             "rad": radius,
             "hour": 1,
         }
@@ -42,7 +50,13 @@ class JapanPostJPSpider(Spider):
                 "X-Requested-With": "XMLHttpRequest",
                 "Referer": "https://map.japanpost.jp/p/search/nmap.htm",
             },
-            cb_kwargs={"lat": lat, "lon": lon, "offset": offset},
+            cb_kwargs={
+                "lat": lat,
+                "lon": lon,
+                "offset": offset,
+                "tempo_count": tempo_count,
+                "post_count": post_count,
+            },
         )
 
     async def start(self):
--- a/locations/spiders/japan_post_jp.py
+++ b/locations/spiders/japan_post_jp.py
@@ -81,13 +81,12 @@ class JapanPostJPSpider(Spider):
         tempo_total = tempo_count + sum(1 for r in rows if r[0] == "TEMPO")
         post_total = post_count + sum(1 for r in rows if r[0] == "POST")
 
+        page = (offset - 1) // count + 1
+        self.logger.info(
+            f"Query (lat={lat}, lon={lon}, page={page}, offset={offset}, rec={rec_count}, hit={hit_count}, tempo={tempo_total}, post={post_total})"
+        )
         if tempo_total >= MAX_ITEMS or post_total >= MAX_ITEMS:
-            f"Maximum number of items returned in one query, consider lowering the radius to avoid locations (lat={lat}, lon={lon}, tempo={tempo_total}, post={post_total})"
-        else:
-            page = (offset - 1) // count + 1
-            self.logger.info(
-                f"Query OK (lat={lat}, lon={lon}, page={page}, offset={offset}, rec={rec_count}, hit={hit_count}, tempo={tempo_total}, post={post_total})"
-            )
+            self.logger.warning("Maximum number of items returned in one query, consider lowering the radius")
 
         if offset + rec_count < hit_count:
             yield self.make_request(
--- a/locations/spiders/japan_post_jp.py
+++ b/locations/spiders/japan_post_jp.py
@@ -1,8 +1,9 @@
 import csv
 from io import StringIO
+from urllib.parse import urlencode
 
 from chompjs import parse_js_object
-from scrapy import Request, Spider
+from scrapy import FormRequest, Spider
 
 from locations.categories import Categories, apply_category
 from locations.geo import city_locations, country_iseadgg_centroids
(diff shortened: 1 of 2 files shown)


shuuji3:
I found that the current spider still doesn't cover all available points 🥲  While I confirmed that all 5.5km city circle areas (blue circles) are fully covered by the subdivision circles (which are not rendered in this map), the city_locations dataset itself does not cover the entire Japan area. As a result, some 24km grid areas (primarily in suburban areas) are still being truncated. (The red dot represents the center of a 24km grid, while the green dot represents an area covered by a 5.5km circle, so no truncation occurs there.)

4 coverage map. Left: Covers city_locations for Japan.
Right: Full coverage visualization.Dots: Green = covered (within 5.5km city circle); Red = uncovered (center of a 24km grid).
(Left: Covers city_locations >=200,000 populations for Japan. Right: All city_locations. Blue = 5.5km city circle / Green = covered (center of a 24km grid within 5.5km city circle) / Red = uncovered (center of a 24km grid).)

j9d3it:
The locations are encoded in Tokyo Datum, please use pyproj.Transformer.from_crs(4301,4326) to reproject them into WGS84. The other Area Marker spiders (japan_post_jp, kuroneko_jp, lawson_jp, lawson_bank_jp) need to have this fix as well, so they should be converted to use a new Area Marker storefinder that implements this fix, rather than duplicating code across all of them. #16413 is the issue I opened about this.

shuuji3:
Thanks for the info. I happened to learn a few years ago that Kuroneko still uses Tokyo Datum. 

I'm wondering if Japan Post still uses it as well, or if they are converting it on the client side, since Google Maps only accepts WGS84. I'll double-check the current coordinates and make sure I reproject them correctly.

locations/spiders/japan_post_jp.py
```diff
@@ -42,25 +116,80 @@ def parse(self, response, lat, lon, offset):
...
+                item = Feature()
+                item["ref"] = ref
+                # post detail page is not accessible without `?post=1`
+                item["website"] = f"https://map.japanpost.jp/p/{MAP_ID}/dtl/{ref}/?post=1"
+                item["lat"] = lat
+                item["lon"] = lon
+                item["postcode"] = postcode
+                item["addr_full"] = addr_full
+                apply_category(Categories.POST_BOX, item)
+                item["name"] = "ポスト"
```
j9d3it: Remove this: postboxes should not be named at all. edit: only line 158 should be removed.

locations/spiders/japan_post_jp.py
```diff
@@ -42,25 +116,80 @@ def parse(self, response, lat, lon, offset):
...
+
+        if offset + rec_count < hit_count:
+            yield self.make_request(
+                lat, lon, radius, offset + rec_count, tempo_count=tempo_total, post_count=post_total, source=source
+            )
+
+        for row in rows:
+            row_type = row[0]
+            ref = row[1]
+            lat = row[2]
```
j9d3it, line 147: Needs to be reprojected from Tokyo Datum (EPSG:4301) to WGS84

locations/spiders/japan_post_jp.py
```diff
@@ -42,25 +116,80 @@ def parse(self, response, lat, lon, offset):
...
+            lon = row[3]
```
j9d3it, line 148: Needs to be reprojected from Tokyo Datum (EPSG:4301) to WGS84

j9d3it:
> I'm wondering if Japan Post still uses it as well, or if they are converting it on the client side, since Google Maps only accepts WGS84. I'll double-check the current coordinates and make sure I reproject them correctly.

I was just browsing around on the preview map, and then opening the website and visually confirming that the locations do not match. I thought they would have switched to WGS84 as well since they are now using a Google Maps embed, but it doesn't look like that's the case.

shuuji3:
Oh, that's surprising. In that case, we might actually be able to provide a more accurate dataset than the official Japan Post website itself.

j9d3it:
> Oh, that's surprising. In that case, we might actually be able to provide a more accurate dataset than the official Japan Post website itself.

No, the locations on the website are accurate, but the ones in our preview are not.

j9d3it:
> I ran the spiders in this pull request for a **2 minute** test and got these results:
> Spider 	Results 	Log
> [`locations/spiders/japan_post_jp.py`](https://github.com/alltheplaces/alltheplaces/blob//locations/spiders/japan_post_jp.py) 	[44273 items](https://alltheplaces-data.openaddresses.io/ci/alltheplaces-fetch:213983ea-898f-40d3-a37d-648d24704bf2/japan_post_jp/output.geojson) ([Map](https://alltheplaces.xyz/preview.html?show=https://alltheplaces-data.openaddresses.io/ci/alltheplaces-fetch:213983ea-898f-40d3-a37d-648d24704bf2/japan_post_jp/output.geojson)) [↑11469](https://alltheplaces-data.openaddresses.io/runs/2026-08-29-13-32-18/stats/japan_post_jp.json) 	Resulted in a `timeout` ([Log](https://alltheplaces-data.openaddresses.io/ci/alltheplaces-fetch:213983ea-898f-40d3-a37d-648d24704bf2/japan_post_jp/log.txt)) ✅
> ℹ️ What does `timeout` mean?

@shuuji3 I was meaning the map preview linked here is not accurate, but when you check the individual locations and go to their website, the Google Maps embed has a different location than the one in the ATP preview.

shuuji3:
Ah, I understand. They might be performing the conversion somewhere else. I'll look into it carefully.

j9d3it:
If the design of the postbox (10号 for example) is in the data, you could encode that in [post_box:design](https://wiki.openstreetmap.org/wiki/Key:post_box:design). It's visible on the Japan Post website so it might be a fun addition.

shuuji3:
Yes, there are many fields in the response and I found `13号` string.

```
POST | 168541 | 35.6574728 | 139.7026094 | 20 | 104.92308378575 | 0 | 東京都渋谷区宇田川町１７−１ | 04 | 010410 | 010410 | 150112 | 渋谷支店 | 130 | 13号 | 縦4cm\u3000横29cm |  | 04 | その他 |  | 東京都渋谷区宇田川町１７−１ | 150-0042 | 13 | 13113 | しぶやく##渋谷区 | 01 |  |  |  |  |  | 厚さ3cm以内のレターパックを投函できます。 | 取集時刻は目安であり、実際の表示と異なることや、「−」表示でも取集している場合があり ます。 | （ポストデータ更新：2026年09月時点） |  |  |  |  |  |  |  |  | 09:10 | 13:10 | 15:00 | 18:30 |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  | 09:10 | 14:30 |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  | 08:50 | 13:50 |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  | 128378560 | 502917826 |  | 
```
https://map.japanpost.jp/p/search/dtl/168541/?&cond200=1&&his=ar,nm&post=1

We can also add [collection_times](https://wiki.openstreetmap.org/wiki/JA:Key:collection_times) as well.

shuuji3:
Regarding the coverage issue, I came up with two ideas:

1. Generate more granular ISEADGG for Japan (12km, 6km, etc.) and use them whenever truncation occurs.
2. Replace the uneven 5.5km circle approach by recursively applying the subdivision function to the 24km grid, fetching with smaller circles only as needed until no truncation remains.

I thought option 1 would be a simpler solution (once grid file is generated), but since it seems that multiple 24km circles cannot cleanly cover a 48km area, I'll go with option 2.

shuuji3:
> They might be performing the conversion somewhere else.

This assumption was correct. While rendered on Google Maps, the conversion is actually run via Zenrin Map SDK's helper function `ZDC.tkyTowgs()` in [emapview_g.js](https://map.japanpost.jp/p/search/emapview_g.js?&cond200=1&), which appears to be Zenrin's custom implementation equivalent to proj reprojection.

To validate the `pyproj` results, I sampled 1/8 of the locations and calculated the distance differences between `from_crs()` and `tkyTowgs()`:
```
count 2853.000000
mean 1.486011
std 0.809664
min 1.236000
25% 1.292000
50% 1.319000
75% 1.387000
max 5.813000
Name: site_vs_pyproj_m, dtype: float64
```

Most locations show errors within 1–2m, but there are anomalies, particularly concentrated in the Kyushu area.

image

The root cause is that when specifying `from_crs(4301, 4326)` (https://epsg.io/4301), `pyproj`'s automatic CRS selection incorrectly selects the Korean CRS (https://epsg.io/1232) when the latitude is below 33.14, instead of the expected Japanese on-shore one (https://epsg.io/15484).

I found three possible approaches to reliably select the correct transformation:

I found three possible ways to select the correct transformer. There may be a better way, but I think option 2 is the most deterministic solution.

1. `TransformerGroup` + `area_of_interest`

```python
from pyproj.aoi import AreaOfInterest
from pyproj.transformer import TransformerGroup

group = TransformerGroup(
    "EPSG:4301", "EPSG:4326",
    always_xy=True,
    area_of_interest=AreaOfInterest(122.83, 20.37, 154.05, 45.54),
)
transformer = next(
    t for t in group.transformers
    if "Tokyo to WGS 84 (108)" in [op.name for op in t.operations]
)
wlon, wlat = transformer.transform(lon0, lat0)
```

3. `from_pipeline` with operation code

```python
from pyproj import Transformer

transformer = Transformer.from_pipeline("EPSG:15484")
wlon, wlat = transformer.transform(lat0, lon0)
```

Note that coords order is different: (lat, lon).

4. `from_crs` + `area_of_interest`

```python
from pyproj import Transformer
from pyproj.aoi import AreaOfInterest

transformer = Transformer.from_crs(
    "EPSG:4301", "EPSG:4326", always_xy=True,
    area_of_interest=AreaOfInterest(122.83, 20.37, 154.05, 45.54),
)
wlon, wlat = transformer.transform(lon0, lat0)
```


image

```py
count    2853.000000
mean        1.486011
std         0.809664
min         1.236000
25%         1.292000
50%         1.319000
75%         1.387000
max         5.813000
Name: site_vs_pyproj_m, dtype: float64
```

---

As a side note, I created [a wrapper of the tky2jgd Python port](https://github.com/shuuji3/tky2jgd) five years ago when pyproj didn't officially supported Tokyo Datum -> WGS 84, and I happened to refactor it recenly. The `tky2jgd` is a conversion program based on GSI Japan's logic. It doesn't just perform linear reprojection but applies region-specific transformation parameters to adjust for the varying distortions within the Tokyo Datum. I compared the `pyproj` results with this tool as well. While they show similar results in mainland Japan, there are large differences in island regions. This suggests that Japan Post's original Tokyo Datum data may not be a strict Tokyo Datum or might have been internally normalized in the past to make it easy to handle.

image

shuuji3:
It looks correct now 🙂 

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>Screenshot from 2026-09-07 01-56-11
 <td>image
</table>

The location matches to the official website: https://map.japanpost.jp/p/search/dtl/400213228600/

Screenshot from 2026-09-07 01-55-49

Another example in a small island in Fukuoka (before/after):

Screenshot from 2026-09-07 02-02-50

https://map.japanpost.jp/p/search/dtl/202267/?post=1

shuuji3:
References about postbox types:
- 郵便ポスト - Wikipedia - https://ja.wikipedia.org/wiki/%E9%83%B5%E4%BE%BF%E3%83%9D%E3%82%B9%E3%83%88
- post_box:design | Keys | Taginfo Japan - https://taginfo.openstreetmap.jp/keys/post_box:design#values (137 nodes tagged in Japan)

j9d3it:
Another point: the postboxes should be tagged with item["operator_wikidata"] = "Q11509260". That will fetch the latest information from Wikidata.

shuuji3:
Recursive subdivision seems working as expected. It took 3156 calls and 52 minutes on local run.

japan map with many circles. each circle represent single api call. nested level represented by defferent rainbow color

tokyo zoom-in

locations/spiders/japan_post_jp.py
```diff
@@ -215,3 +218,18 @@ def parse(
...
 
             yield item
+
+    def get_collection_times(self, row: List[str]) -> str:
+        # POST rows have collection times in three fixed 20-slot groups:
+        #   cols 42-61 weekday (平日), 62-81 Saturday (土曜), 82-101 Sunday/holiday (日曜・休日)
+        groups = {
+            "Mo-Fr": row[42:62],
+            "Sa": row[62:82],
+            "Su": row[82:102],
```
j9d3it: Su, PH

j9d3it (changes requested):
Collection time is Sunday and public holidays so it should be "Su, PH"

shuuji3:
Thanks, good catch 👍🏻

locations/spiders/japan_post_jp.py
```diff
@@ -42,25 +116,80 @@ def parse(self, response, lat, lon, offset):
...
+        self.logger.info(
+            f"Query (source={source}, lat={lat}, lon={lon}, radius={radius}, page={page}, offset={offset}, rec={rec_count}, hit={hit_count}, tempo={tempo_total}, post={post_total})"
+        )
+        if tempo_total >= MAX_ITEMS or post_total >= MAX_ITEMS:
+            self.logger.warning(
+                f"Maximum number of items returned in one query, consider lowering the radius  (source={source})"
+            )
+            if source == "city-5.5":
+                yield from self._subdivide(lat, lon, radius, source)
+            return
```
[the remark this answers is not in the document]
shuuji3: We did a major refactoring. Check the latest states.

PR #18526, opened 2026-09-06
Handle invalid hour-24 close times in OpeningHours.add_range

iandees:
add_range already special-cases the literal string "24:00" (rewriting it to "23:59"), but anything else with hour 24 and nonzero minutes/seconds (e.g. "24:30") fell through to time.strptime() and raised ValueError, crashing the whole spider run. This surfaced in #18484 (docomo_shop_jp), which worked around it with a spider-local hour==24-check — this generalizes that fix into the shared library so every spider benefits, not just ones that happen to validate it themselves.

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid same-day closing times such as `24:30` are now safely ignored instead of causing an error.
  - Valid `24:00` closing times continue to be handled correctly across supported time formats.

- **Tests**
  - Added coverage for invalid hour-24 closing times in standard and seconds-inclusive formats.

### DIFF
--- a/locations/hours.py
+++ b/locations/hours.py
@@ -1152,6 +1152,14 @@ class OpeningHours:
                 close_time = "23:59"
             if close_time in ("24:00:00", "00:00:00"):
                 close_time = "23:59:00"
+            if re.match(r"^24:(?!00(:00)?$)\d{2}(:\d{2})?$", close_time):
+                # Hour 24 is only a valid representation of midnight at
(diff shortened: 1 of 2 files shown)


iandees:
Closing — superseded by #18509, which handles this more completely (treats hour>=24 as a legitimate next-day rollover per the existing over-midnight convention, rather than discarding the range as invalid data).

PR #18603, opened 2026-09-09
fix(dict_parser): separate Japanese prefecture keys for "addr:province"

shuuji3:
I noticed this mistagging in https://github.com/alltheplaces/alltheplaces/pull/18592 and some CodeRabbit reviews.

For addresses in Japan, "addr:province" is used for the prefecture and "state" is unused. So we need a special handling for those Japanese specific keys.

ref.
- Key:addr:province - OpenStreetMap Wiki - https://wiki.openstreetmap.org/wiki/Key:addr:province
- Japan tagging - OpenStreetMap Wiki - https://wiki.openstreetmap.org/wiki/Japan_tagging#Places

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved address parsing for Japanese prefectures.
  - Japanese prefecture values are now stored consistently as province information in parsed address details.
  - Preserved handling of other region fields.

### DIFF
--- a/locations/dict_parser.py
+++ b/locations/dict_parser.py
@@ -153,11 +153,18 @@ class DictParser:
         "store-state",
         "store-province",
         "storeProvince",
-        "prefecture",
-        # JP
-        "都道府県",  # "prefecture"
         # IT
         "regione",  # "region"
+    ]
+
+    # For addresses in Japan, "addr:province" is used for the prefecture and "state" is unused
+    # ref.
+    # - Key:addr:province - OpenStreetMap Wiki - https://wiki.openstreetmap.org/wiki/Key:addr:province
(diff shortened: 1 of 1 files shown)


Cj-Malone:
We can't and wont support every countries address differences. Maybe OSM in Japan should have used `addr:state`, or maybe OSM in US should have used `addr:zip`. Either way, we are where we are, it's the responsibility of whoever uses the output of ATP to convert it to whatever they want to use it.

shuuji3:
@Cj-Malone Sorry, I'm new to this ecosystem and I was simply assuming it's better to follow OSM tagging rules of the targeting location so that it could make the future data process eisier.

I personally don't have any favor as far as it's consistent format, and actually don't know why OSM Japan rule dones't make use of the standard `state` tag at all, which could avoid complexity like this. I'll follow the preferable ways in this repository next time.

PR #18735, opened 2026-09-15
[Graze Craze US] Add spider (133 locations)

valeribarbakadze:
Adds a spider for Graze Craze, a US charcuterie board and catering franchise.

The store locator is a JavaScript application backed by the Gorilla Dash
GraphQL API. Two things about that API shape the spider:

- it only accepts persisted queries, so the query hash is hardcoded rather than
  the query text being sent (sending query text returns 403)
- it requires a bearer token, which the locator page publishes as
  window.graphqlToken. The token is short lived, so the spider reads a fresh
  one from the page on each run rather than carrying a fixed value

The feed returns 155 records with a status field: 133 Active, 21 Opening Soon
and 1 Testing. Only Active locations are emitted. All records are in the US —
the Australian and Canadian sites are separate — so the _us suffix applies.

`ref` is the location slug. Email and social links come from the same records.

No brand:wikidata is set as the chain has no Wikidata item.

## Summary by CodeRabbit

* **New Features**
  * Added Graze Craze location data, including active locations, addresses, contact details, social media links, and deli and charcuterie classifications.


Cj-Malone:
Have you been to one of these? Are they restaurants? OSM has them more as `shop=deli` or `craft=caterer`

PR #18789, opened 2026-09-17
[Japan Post JP] fix post box ref, name of postbox design

### LINKED ISSUE #18780: Japan Post gets wrong ref number for postbox
### Spider name

japan_post_jp

### Log output

https://alltheplaces-data.openaddresses.io/runs/2026-09-12-13-32-21/logs/japan_post_jp.txt

### Backtrace (if applicable)

_No response_

Ref for all postboxes should not be what is in the URL, rather it should be the postbox number. 

For example: this one https://map.japanpost.jp/p/search/dtl/238693/?post=1

ポスト番号：210012 but the ref in the spider output is 238693.

Should collect the postbox number if it's in the data.

### Possible causes?

_No response_

j9d3it:
closes #18780 

in https://taginfo.openstreetmap.jp/keys/post_box:design#values most values start with "差出箱." To keep with that convention that is what I have changed it to. This is also the name on https://www.postmap.org/models

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected postbox reference data for Japan Post locations.
  * Updated the postbox design label format.
  * Preserved the existing postbox website links.


Cj-Malone:
I've merged, but OSMs `ref` and ATPs `ref` are not the same, ATPs refs should very rarely be used in OSM.

j9d3it:
@Cj-Malone understood, thank you

PR #18804, opened 2026-09-17
Add zips_car_wash_us spider (171 locations)

valeribarbakadze:
Adds a spider for ZIPS Car Wash, a US car wash chain across 16 states.

The locations page is a Livewire component whose wire:snapshot attribute holds
every site, nested by state and then by city, with address and coordinates, so
one request covers the whole chain.

Opening hours are only published on the per site pages, in a JSON-LD block that
is not valid JSON — the days are written as object entries inside an array,
{"dayOfWeek": ["Monday": "08:00-19:00", ...]} — so they are read with a regular
expression rather than a JSON parse. All 171 locations produce valid
opening_hours, coordinates and a full address.

`ref` is the site_id from the Livewire payload.

## Summary by CodeRabbit

* **New Features**
  * Added location coverage for ZIPS Car Wash in the United States.
  * Captures branch details, addresses, geographic coordinates, site links, and categorized business information.
  * Includes operating hours when available.


iandees:
Reviewed this — the item count (111 vs 171 expected) is just the 2-minute CI timeout, not a bug; a full run should collect all 171. Spot-checked coordinates, addresses, category, and opening hours and they look correct.

One thing worth fixing before merge: line 70 calls `.as_opening_hours()` inside a truthiness check (`if oh.as_opening_hours():`). Per the repo's review rules, spider code shouldn't call `.as_opening_hours()` directly — the pipeline accepts the `OpeningHours` object as-is. Suggest:

```python
if oh:
```

I don't have push access to this fork's branch, so flagging here rather than pushing directly.

---

valeribarbakadze:
The 172 requests can't finish inside the 120s CI timeout — one for the store
snapshot and one per site for hours, which are only published on the individual
pages. A local run returns all 171 with opening_hours on every feature. Happy to
drop the hours crawl to a single request if you'd prefer CI to complete, but
that loses the hours.

PR #18837, opened 2026-09-18
Add rogers_and_hollands_us spider (72 locations)

valeribarbakadze:
Adds a spider for Rogers & Hollands Jewelers and its sister brand Ashcroft &
Oak Jewelers, which share one store finder.

The find-a-store page carries every store in the map's data-locations
attribute, including coordinates and a chunk of HTML per store holding the
address, phone and store link. Opening hours are only on the store pages, as a
"Monday: 10:00AM - 8:00PM" list, so each store page is also requested.

The feed mixes both brands — 45 Rogers & Hollands and 27 Ashcroft & Oak — so
brand is taken from each store's own name rather than being set for the whole
spider. Only Rogers & Hollands has a Wikidata item.

The address is a run of paragraphs whose count varies by store, so the parse
looks for the line matching "city, ST postcode" and treats what precedes it as
the street.

All 72 locations have coordinates, a full address, phone and opening_hours.
`ref` is the data-store-id.

## Summary by CodeRabbit

- **New Features**
  - Added coverage for Rogers & Hollands and Ashcroft & Oak jewelry store locations in the United States.
  - Captures store details including addresses, phone numbers, websites, coordinates, brands, and jewelry-store categorization.
  - Includes store opening hours when available.


iandees:
🤖 CI is otherwise green (72 items, brand wikidata Q130223859 verified, category correct, address/phone/website/hours all per-location). One thing to fix before merge: CodeRabbit's finding is correct — every location sets `item["brand"]` but never `item["name"]`, so the feature name is missing on all 72 items.

Fix (one line, in `parse`):

```python
if brand := location.get("name"):
    item["name"] = brand
    item["brand"] = brand
    item["brand_wikidata"] = BRANDS.get(brand)
```

Maintainer-edit is enabled on this PR but this environment's git proxy can't push to your fork directly, so leaving this as a comment rather than a fixup commit. Happy to merge as soon as this lands.

---

PR #18938, opened 2026-09-20
[Los Angeles Public Library US] Add spider (73 locations)

watmildon:
Adds Los Angeles Public Library, the second user of the CommunicoSpider storefinder added in #18931 and much the largest system on the Communico Attend API.

* `api.communico.co/v1/lapl/locations` lists all 73 branches with names, addresses, coordinates, phone numbers and photos, and `api.communico.co/v1/lapl/opening-hours/<date>/<days>` supplies hours. Two requests cover the whole system, so the spider is just the client name plus this tenant's own data fixes.
* Every entry the `lapl` client publishes is a real branch: Central Library is a single location with no departments broken out, and there are no bookmobiles, lockers, administrative offices or virtual "locations" to filter, so nothing is skipped.
* Nine branches have a broken `about_url`. Seven are empty (Cypress Park, Echo Park, Encino-Tarzana, Lincoln Heights, Vermont Square, Washington Irving, Will & Ariel Durant), Jefferson points at the branch index rather than its own page, and Cahuenga points at Canoga Park's page. Their real lapl.org slugs are listed in the spider and checked against the site, so all 73 get their own branch page as `website` instead of 64.
* Arroyo Seco Regional is published with Angeles Mesa's exterior photo, so images appearing on more than one branch are dropped; 67 of 73 keep a genuine per-branch photo.
* Each branch's `description` is a "Resources & Services" list, which gives `internet_access=wlan` on the 71 branches that list Wi-Fi.
* Palisades Branch Library is recorded as closed every day of the week. It has been shut since the January 2025 Palisades Fire, with only a mobile service in the branch car park, so the all-week closure is accurate rather than a parsing gap.
* Names are left as the library writes them ("Angeles Mesa Branch Library", "Alma Reaves Woods - Watts Branch Library"), since they are already complete library names, as in king_county_library_system_us and timberland_regional_library_us. OSM has no consistent pattern for these branches to follow — the same system appears as "Angeles Mesa Branch Los Angeles Public Library", "Canoga Park Library", "Edendale Branch L A Public Library" and "Los Angeles Public Library, El Sereno Branch".

73 locations, matching the 73 outlets in the IMLS FY2023 survey, 72 of which are already in OSM — around half of those without addr:street/addr:postcode and around two thirds without phone or opening_hours. All 73 items carry coordinates, a full address, a unique phone number, per-branch hours and a branch page URL.

The broken branch URLs and the duplicated photo have been reported to the library; the workarounds can come out once their feed is corrected.

## Summary by CodeRabbit

- **New Features**
  - Added coverage for Los Angeles Public Library branches.
  - Branch listings now include corrected branch links, categorized library details, Wi‑Fi availability when identified, and cleaned-up image information.
  - Processed branch records include relevant location and opening-hours details.


watmildon:
I am going to mail the library about getting the images fixed on their side so we don't need to hard code the image file name and they'll both come back when fixed on their side.

PR #19013, opened 2026-09-22
[Lids] Fix spider

karen-kerr:
The store API's opening-hours values mix 12-hour and 24-hour clock readings that both carry a meridiem, e.g. `9:00 PM` but also `21:00 PM`. Parsing with `%I:%M %p` raised `ValueError` on the first 24-hour token and killed the parse generator, so recent runs yielded nothing (an earlier run yielded 1025 — the index of the first bad record at the time).

Times are now normalised before parsing: the meridiem is applied only when the hour is 12 or less, and an unparseable token skips that day rather than aborting the crawl. Explicit CLOSED tokens are recorded with `set_closed`.

Also:

- The API has started transposing latitude and longitude on nearly every record. Coordinates are swapped when latitude falls outside (0, 90) — a range no store in this estate can legitimately occupy — so the correction becomes a no-op if the API is repaired.
- `-999` and empty-string sentinels are nulled across scalar fields.
- `website` now uses the API's absolute `url` rather than concatenating `taggedUrl` onto the domain.
- Explicit `shop=clothes` category; the per-store venue label goes to `branch` and `name` is left for NSI.

1027 locations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Lids store data accuracy by handling inconsistent hours, missing values, and coordinate formats.
  * Preserved valid daily opening hours when another day contains invalid or unavailable time data.
  * Corrected store names, website links, and location coordinates in listings.
* **Enhancements**
  * Lids locations are now categorized as clothing stores.


karen-kerr:
The `no output` is the CI egress being blocked by Akamai (a single 403 in the log). Production reaches this API fine — the bug fixed here was a parse crash on a successfully downloaded payload. A local crawl of this branch returns 1027 items.

PR #19017, opened 2026-09-22
[AllSaints] Fix spider

karen-kerr:
The store index at `/stores/` now returns only the handful of stores nearest the requesting IP rather than the full directory, so the crawl had dropped to 5 GB locations.

Switched the start URL to `/storesresult/`, which lists every store grouped by country, and broadened the link pattern to also match stores whose URL is a bare store code (`/stores/TR5631`) rather than `country/city/slug`.

Also in this change:

- Removed `ROBOTSTXT_OBEY: False`; robots.txt allows both `/stores/` and `/storesresult/`.
- Trimmed `Host`, `Connection` and `Referer` from the request headers, keeping the ones needed for a 200.
- Replaced `drop_attributes` with `search_for_email/facebook/twitter = False`. The old keys never matched the item fields, so brand-level Twitter and Facebook values were being emitted on every POI.
- Brand-prefix stripping for `branch` now handles the `Allsaints` and `AlllSaints` spellings present in the source.
- `ref` now uses the store code from `data-store-id`, unique across all locations.
- Empty-string `image` values are nulled.

286 locations across 21 countries.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AllSaints store discovery and link handling.
  * Store records now capture branch references more consistently.
  * Store names are cleaned up for clearer branch identification.
  * Missing store images are handled consistently.
  * Unneeded automated metadata lookups are no longer performed.


karen-kerr:
The `no output` is the CI egress being blocked by Cloudflare (403 on robots.txt and the first page). Production passes this check — the recent 5-item runs successfully fetched the geo-pinned index this PR fixes. Local crawls of this branch return 286 items, all HTTP 200.

